PR #1, opened 2022-12-02
Alterationx10/GitHub publish

alterationx10:
👋 Excited to play around with kyo!

I was greeted with an error loading the project. It seems to be because the `sbt-github-packages` requires a token source to exists, or else it will fail. There doesn't seem to be an option to tell it "Hey, I'm not the CI, I'd just like the run the project", which I've run into before, and perhaps not everyone likes to load up a system with dummy variables, so I tend to just not use it.

If interested, this PR sets up what you'd need to publish to GitHub without a plugin. There's also some code about setting the version published based off a tag/release, but could be removed if you just want to set it manually, or use something like sbt-dynver. ( If you want to keep it, I also tend to shove it in a seperate `.scala` file in  `project/` so it doesn't clutter up the main `build.sbt`)

I didn't add it, but here is a github action I've used with this code before:
```
name: Publish Artifact on Release
on:
  release:
    types: [ created ]
env:
  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
jobs:
  publish:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Set up JDK
        uses: actions/setup-java@v3
        with:
          distribution: 'temurin'
          java-version: '17'
          cache: 'sbt'
      - name: Publish
        run: sbt publish
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ project/project
 target
 *.log
 graal_dumps
+.idea
+metals.sbt
--- a/build.sbt
+++ b/build.sbt
@@ -12,10 +12,6 @@ scalacOptions ++= Seq(
   "-language:implicitConversions",
 )
 
-ThisBuild / githubOwner := "fwbrasil"
-ThisBuild / githubRepository := "kyo"
-ThisBuild / githubTokenSource := TokenSource.Environment("GITHUB_TOKEN")
-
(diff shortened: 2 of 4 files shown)


fwbrasil:
awesome, thank you @alterationx10!!

PR #4, opened 2023-01-27
chore: Remove non-tech comment

nightscape:
Hi @fwbrasil,

I don't have any stakes in either the ZIO nor the TypeLevel community, knowing none of the people personally.
I like(d) working with the TypeLevel stack, and I'm enjoying ZIO quite a bit as well.

I think we need to be able to answer the following questions though:

* How do we make sure we don't scare away the next generation of potential Scala developers?
* How do we make sure that C-Level people don't perceive the Scala community as a major risk when thinking about choosing Scala as a language?
* How do we make sure that there is still cross-pollination of the best ideas?
* How do we make sure that we only drive away the people we actually want to drive away, and not those who chose the other side for non-political reasons, or because other options were not available at the time?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,6 @@ Algebraic effects for Scala 3
 Sorry, documentation coming soon.
 
 
-No Typelevel Policy
--------------------
-
-Given [Typelevel's](https://typelevel.org/) history of online attacks targeting its competitors and members of minority groups, this project adopts a "No Typelevel Policy". Integrations with Typelevel projects are strongly discouraged.
-
 License
 -------
 


fwbrasil:
Hi @nightscape! Thank you for the thoughtful message, these are very important points indeed. I don't believe they supersede my safety concerns, though. The last time I allowed a Typelevel integration with one of my projects, it ended up with full-on cancellation-by-proxy campaign, intimidation, cyberstalking, and ex-coworkers calling me a nazi. 

Typelevel made no efforts to repair the damage, did not investigate clear Scala CoC violations, and took zero steps to prevent similar situations from happening again.

nightscape:
Fair enough. You of course have to protect yourself against being treated in such a way.
I hope these questions get answered someday in a way that allows the community to move beyond it's current bifurcation.

Btw, if you're interested in a model of the development of value systems that gives a rather good explanation of what's going on, I can very much recommend Spiral Dynamics.
I assume TypeLevel is mostly stage green and afaict they sometimes classify other people at stage yellow (a value system where other value systems are accepted) as stage red / blue (which is where Nazis would fall).
Leo from the Actualized YouTube channel has a nice series on this which imo has a lot of explanatory power.

I wish you and this endeavor all the best!

Issue #17, opened 2023-04-05
source file for quests in kyo-chatgpt


giuliohome:
I am testing the demo of ChatGPT with the new effect for AI Quests.
The current version of the code is copied from this [gist](https://gist.github.com/fwbrasil/ea56f7f4e97c9966251db2031c224b7e) and my exact git repo is [here](https://github.com/giuliohome/AIQuestKyoEffect/blob/main/src/main/scala/example/Hello.scala).

It `sbt compile` but there is a runtime exception:
```
[giulio@fedora mykyo]$ sbt run
[info] welcome to sbt 1.8.2 (Red Hat, Inc. Java 19.0.2)
[info] loading project definition from /home/giulio/scala/mykyo/project
[info] loading settings for project hello from build.sbt ...
[info] set current project to Hello (in build file:/home/giulio/scala/mykyo/)
[info] compiling 1 Scala source to /home/giulio/scala/mykyo/target/scala-3.2.2/classes ...
[warn] there was 1 feature warning; re-run with -feature for details
[warn] one warning found
[info] running Hello 
[error] java.io.FileNotFoundException: /home/runner/work/kyo/kyo/kyo-chatgpt/src/main/scala/kyo/quests.scala (No such file or directory)
[error]         at java.base/java.io.FileInputStream.open0(Native Method)
[error]         at java.base/java.io.FileInputStream.open(FileInputStream.java:219)
[error]         at java.base/java.io.FileInputStream.<init>(FileInputStream.java:158)
[error]         at scala.io.Source$.fromFile(Source.scala:94)
[error]         at scala.io.Source$.fromFile(Source.scala:79)
[error]         at scala.io.Source$.fromFile(Source.scala:57)
[error]         at kyo.quests$Quests$.observe(quests.scala:169)
[error]         at kyo.quests$Quests$.$anonfun$1$$anonfun$1$$anonfun$1(quests.scala:146)
[error]         at kyo.core$.kyo$core$$$_$transformLoop$1(core.scala:146)
[error]         at kyo.core$$anon$5.apply(core.scala:142)
[error]         at kyo.core$$anon$5.apply(core.scala:138)
[error]         at kyo.core$$anon$8.liftedTree1$1(core.scala:209)
[error]         at kyo.core$$anon$8.apply(core.scala:213)
[error]         at kyo.core$$anon$8.liftedTree1$1(core.scala:209)
[error]         at kyo.core$$anon$8.apply(core.scala:213)
[...]

PR #19, opened 2023-04-08
one-line fix issue #17 

giuliohome:
Is the [line](https://github.com/getkyo/kyo/blob/77c102684891530161efdda3f319f20e057a50c1/kyo-chatgpt/src/main/scala/kyo/quests.scala#L146) `_  <- observe(self)` really needed? Because just dropping it seems to fix issue #17


fwbrasil:
So sorry for not addressing your PR before. I wasn't ready to receive contributions but that's changing now with the planned release at Functional Scala next week. Unfortunately, I removed the quests effect some time ago so I'm closing the PR. You're more than welcome to contribute again! :)

PR #105, opened 2023-09-25
Fix zio-interop typos

hearnadam:
Fix typos, update to ZIO 2 syntax. I didn't update the comment for `UIO` effect. A `Task` type fails with `Throwable`. Maybe I don't have the full context on the kyo `Effect` type, but can a kyo `T > ZIOs` fail? Should we consider renaming the `fromTask` method?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1463,31 +1463,32 @@ trait Service {
 }
 
 // An example ZIO computation
-val a: ZIO[Service, String, Int] = 
-  ZIO.accessM[Service](service => ZIO.effect(service.compute))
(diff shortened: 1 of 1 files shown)


fwbrasil:
Hi @hearnadam, thanks for the pull request!

> can a kyo T > ZIOs fail?

It should use `Aborts` if there's a known failure case but, as with ZIO, any computation can potentially fail even if not expected.

I ended up deciding to remove the ZIO integration for now :( I'm closing the PR because of that. More context: https://github.com/getkyo/kyo/issues/111

Issue #153, opened 2023-12-11
Add layer mechanism for composing dependencies (or effects in general?)


johnhungerford:
Given an effect `T < (S & Envs[A] & Envs[B] & Envs[C])`, it should be possible to construct a "layer" (terminology taken from ZIO) that will provide some or all of the dependencies `A`, `B`, and `C`, optionally requiring additional dependencies to satisfy these. You should be able to provide this layer to an effect to handle any of the `Envs` dependencies at once, returning an effect without the provided dependencies and (optionally) with additional dependencies.

I envision the signature as being:
```scala
type Layer[Sin, Sout]
```
where `Sout` are the effects being handled, and `Sin` are the effects needed to handle them. 

A layer could can be constructed and then provided to an effect as follows:

```scala
trait ElasticClient:
    def execute(request: ElasticRequest): List[ElasticResult] < (IOs & Aborts[ElasticError])

val elasticEffect: List[ElasticResult] < (IOs & Aborts[ElasticError] & Envs[ElasticClient]) =
    for {
        elasticClient <- Envs[ElasticClient].get
        results <- elasticClient.execute(someRequest)
    } yield results

// Call Envs.layer to construct a layer that will generate the dependency described by `Envs[ElasticClient]`
val elasticClientLayer: Layer[IOs & Tries & Envs[HttpClient], Envs[ElasticClient]] =
    Envs.layer {
        for {
            httpClient <- Envs[HttpClient].get
            elasticClient <- IOs {
                Tries.catching(ElasticClient(httpClient, ElasticClient.Config.default))
            }
        } yield elasticClient
    }

// Note that:
// 1) `Envs[ElasticClient]` is replaced by `Envs[HttpClient]`
// 2) The updated effect now has `Tries` in addition to `Aborts[ElasticError]`, due to the client constructor's effects
val updatedElasticEffect: List[ElasticResult] < (IOs & Tries & Aborts[ElasticError] & Envs[HttpClient]) =
    Layers.run(elasticClientLayer)(elasticEffect)
```

Layers should be composable in two ways:
1. Combine inputs and outputs
```scala
val layer1: Layer[A & B, X & Y]
val layer2: Layer[C, Z]

val composedLayer: Layer[A & B & C, X & Y & Z] = layer1 ++ layer2
```

2. Apply outputs of one layer to the inputs of another:
```scala
val layer1: Layer[A & B, C]
val layer2: Layer[C, X & Y]

[...]

PR #154, opened 2023-12-12
Add Layer trait and Layer factory methods on Envs, Aborts, Tries, and Options

johnhungerford:
Introduces new type `Layer[Sin, Sout]`, representing a handler of `Sout` that requires further handling of `Sin` and that cannot transform the output of the handled effect. The latter restriction allows for commutative composition, so that layers can be combined in any order without affecting (as far as I can tell) the resulting effect.

`Layer` provides a mechanism for dependency injection, by allowing you to compose `Envs` handlers. It should be possible to add a function similar to ZIO's `.provide`, which automatically wires layers together to satisfy all dependencies.

However, `Layer` can be used more broadly as a means to simplify certain kinds of effect handling. For instance, it can be used to between different kinds of failures.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/layers.scala
+++ b/kyo-core/shared/src/main/scala/kyo/layers.scala
@@ -3,18 +3,18 @@ package kyo
 object layers {
 
   trait Layer[In, Out] { self =>
-    def run[T, S](effect: T > (S with In))(implicit fl: Flat[T > (S with In)]): T > (S with Out)
+    def run[T, S](effect: T > (In with S))(implicit fl: Flat[T > (In with S)]): T > (S with Out)
 
     final def add[Out1, In1](other: Layer[In1, Out1]): Layer[In with In1, Out with Out1] =
       new Layer[In with In1, Out with Out1] {
         override def run[T, S](
-            effect: T > (S with In with In1)
+            effect: T > (In with In1 with S)
         )(
-            implicit fl: Flat[T > (S with In with In1)]
+            implicit fl: Flat[T > (In with In1 with S)]
         ): T > (S with Out with Out1) = {
-          val selfRun: T > (S with In1 with Out) =
-            self.run[T, S with In1](effect: T > (S with In1 with In))
-          val otherRun: T > (S with Out with Out1) =
+          val selfRun: T > (In1 with Out with S) =
+            self.run[T, S with In1](effect: T > (In1 with In with S))
+          val otherRun: T > (Out with Out1 with S) =
             other.run[T, S with Out](selfRun)(Flat.unsafe.unchecked)
           otherRun
         }
@@ -46,8 +46,8 @@ object layers {
             layer2: Layer[In2 with Shared, Out2]
         ): Layer[In1, Out1 with Out2] =
           new Layer[In1, Out1 with Out2] {
-            override def run[T, S](effect: T > (S with In1))(implicit
-                fl: Flat[T > (S with In1)]
+            override def run[T, S](effect: T > (In1 with S))(implicit
+                fl: Flat[T > (In1 with S)]
             ): T > (S with Out2 with Out1) = {
               val handled1: T > (S with Out1 with Shared) = layer1.run[T, S](effect)
               val handled2: T > (S with Out2 with Out1) =
@@ -69,8 +69,8 @@ object layers {
             layer2: Layer[In2 with Shared, Out2]
         ): Layer[In1, Out2] =
           new Layer[In1, Out2] {
-            override def run[T, S](effect: T > (S with In1))(implicit
-                fl: Flat[T > (S with In1)]
+            override def run[T, S](effect: T > (In1 with S))(implicit
+                fl: Flat[T > (In1 with S)]
             ): T > (S with Out2) = {
               val handled1: T > (S with Shared) = layer1.run[T, S](effect)
               val handled2: T > (S with Out2) =
@@ -90,8 +90,8 @@ object layers {
             layer2: Layer[Shared, Out2]
         ): Layer[In1, Out1 with Out2] =
           new Layer[In1, Out1 with Out2] {
-            override def run[T, S](effect: T > (S with In1))(implicit
-                fl: Flat[T > (S with In1)]
+            override def run[T, S](effect: T > (In1 with S))(implicit
+                fl: Flat[T > (In1 with S)]
             ): T > (S with Out1 with Out2) = {
               val handled1: T > (S with Out1 with Shared) = layer1.run[T, S](effect)
               val handled2: T > (S with Out1 with Out2) =
@@ -115,10 +115,10 @@ object layers {
             layer2: Layer[Out, Out2]
         ): Layer[In1, Any with Out2] =
           new Layer[In1, Any with Out2] {
-            override def run[T, S](effect: T > (S with In1))(implicit
-                fl: Flat[T > (S with In1)]
+            override def run[T, S](effect: T > (In1 with S))(implicit
+                fl: Flat[T > (In1 with S)]
             ): T > (S with Out2) = {
-              val handled1: T > (S with Out) = layer1.run[T, S](effect)
+              val handled1: T > (Out with S) = layer1.run[T, S](effect)
               val handled2: T > (S with Out2) =
                 layer2.run[T, S](handled1)(Flat.unsafe.unchecked)
               handled2
--- a/kyo-core/shared/src/main/scala/kyo/layers.scala
+++ b/kyo-core/shared/src/main/scala/kyo/layers.scala
@@ -1,0 +1,144 @@
+package kyo
+
+object layers {
+
+  trait Layer[Sin, Sout] { self =>
+    def run[T, S](effect: T > (S with Sout))(implicit fl: Flat[T]): T > (S with Sin)
+
+    final def ++[Sin1, Sout1](other: Layer[Sin1, Sout1]): Layer[Sin with Sin1, Sout with Sout1] =
+      new Layer[Sin with Sin1, Sout with Sout1] {
+        override def run[T, S](
+            effect: T > (S with Sout with Sout1)
+        )(
+            implicit fl: Flat[T]
+        ): T > (S with Sin with Sin1) = {
+          val selfRun: T > (S with Sout1 with Sin) =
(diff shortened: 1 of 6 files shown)


johnhungerford:
Addresses issue #153

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,144 @@
+package kyo
+
+object layers {
+
+  trait Layer[Sin, Sout] { self =>
```
fwbrasil, line 5: It was a bit counterintuitive to me that the effects that the layer can handle is "out" and effects it can produce is "in". I image that's thinking of the operations in the pending effect set. What do you think about flipping the meaning of the type parameters? I think it's more intuitive to think of a `Layer` like an effect-level arrow that takes effects to handle and produces new effects if needed.

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,144 @@
...
+    def run[T, S](effect: T > (S with Sout))(implicit fl: Flat[T]): T > (S with Sin)
+
+    final def ++[Sin1, Sout1](other: Layer[Sin1, Sout1]): Layer[Sin with Sin1, Sout with Sout1] =
```
fwbrasil, line 8: I'd like to avoid any symbolic operators in Kyo. Maybe we could use `add` for `++` and `andThen` for `>>>`?

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,144 @@
...
+        ): T > (S with Sin with Sin1) = {
+          val selfRun: T > (S with Sout1 with Sin) =
+            self.run[T, S with Sout1](effect: T > (S with Sout1 with Sout))
+          val otherRun: T > (S with Sin with Sin1) = other.run[T, S with Sin](selfRun)
+          otherRun
+        }
+      }
+
+    final def >>>[Sin1, Sout1](other: Layer[Sin1, Sout1])(implicit
+        ap: ApplyLayer[Sout, Sin1]
```
fwbrasil, line 23: Would it be possible to express the type-level transformations in the method signature itself instead of relying on the implicit mechanism? I imagine it's not the case, just to double check.

fwbrasil:
I had forgotten how joyful OSS can be! It's great how well the layers compose 🤯

Regarding effect handling ordering, the result of the computations [can change depending on the order](https://github.com/getkyo/kyo#using-effects) but it seems reasonable to assume that ordering is a decision users can make when building layers. We might need better names for the operators, though. `++` gives an impression that ordering isn't important. 

I've been working on a similar mechanism but for another purpose. Users currently need to handle other pending effects before forking a fiber. Since Kyo supports arbitrary effects, it's not easy to "propagate" the effect to the forked computation. For this use case, I need more features than only handling the effect so I created a [new `Joins` mechanism](https://github.com/getkyo/kyo/blob/joins/kyo-core/shared/src/main/scala/kyo/joins.scala) that provides three operations:

1. `save`: Saves any state necessary to handle the effect in the forked computation. For example, `Envs` would save the environment value at caller to provide it while executing the fiber.
2. `handle`: Handles the effect given the state.
3. `resume`: After the fiber returns, this method can suspend the effect again.

The approach works well for methods that don't return a `Fiber` instance like `Fibers.parallel`:

```scala
    def parallel[T, S](j: Joins[S])(l: Seq[T > (S with Fibers)])(
        implicit f: Flat[T > Fibers]
    ): Seq[T] > (S with Fibers) =
      j.save.map { st =>
        j.handle(st, l).map(parallel(_)).map(j.resume)
      }
```

But it doesn't work well for `Fibers.init`. It's not possible to use `resume` in this case because the `M` is inside a `Fiber` value. For now, I'm just returning the `Fiber[M[T]]` but I don't think that provides good usability.

```scala
    def init[T, S](j: Joins[S])( /*inline*/ v: => T > (S with Fibers))(implicit
        f: Flat[T > (S with Fibers)]
    ): Fiber[j.M[T]] > (S with IOs) =
      j.save.map { st =>
        init(j.handle(st, v))
      }
```

I wonder if we could try to unify the joins and layers. The joins impl is in this branch: https://github.com/getkyo/kyo/blob/joins

fwbrasil:
Regarding the build failure, the issue is compilation in Scala 2. The macro for `Flat` is in the same module as the core code so it can't be invoked given that Scala 2 can't use macros from the same compilation unit. I've been working this around by importing `Flat.unsafe`. I'll create an issue to move the macros to a new module, it's not trivial but seems possible.

kyo-core/shared/src/main/scala/kyo/layers.scala
johnhungerford: Yeah I agree. How about `Layer[Handled, S]` or something along those lines, where `Handled` are the effects that are handled and `S` are the new unhandled effects?

kyo-core/shared/src/main/scala/kyo/layers.scala
johnhungerford: I'm not sure. Basically some mechanism is needed to extract the intersection between the effects that one layer handles and the effects that another layer needs to be handled, and keep track of the remaining effects of each. I'm pretty sure I can least simplify and clarify `ApplyLayer`, but I can't think of a way to produce the same functionality with just type bounds. I'll give it some more thought though.

johnhungerford:
Re: the build failure, when I run `sbt +kyo-core/test` or `sbt +kyo-core/compile` locally it works. Is there another command I should be running to reproduce the failure?

kyo-core/shared/src/main/scala/kyo/layers.scala
fwbrasil: Maybe `Input`/`Output` to make it more explicit?

kyo-core/shared/src/main/scala/kyo/layers.scala
fwbrasil: sounds good! the current version is ok with me btw

fwbrasil:
> Re: the build failure, when I run `sbt +kyo-core/test` or `sbt +kyo-core/compile` locally it works. Is there another command I should be running to reproduce the failure?

I get the same error as the CI when I run `sbt +test`. I think it could be the state of your local compilation, maybe the macro class files are already generated so the compiler can expand it. Try `sbt +clean` first.

kyo-core/shared/src/main/scala/kyo/layers.scala
johnhungerford: I wasn't able to figure out an alternative, but I did improve the type class used to chain layers together. Now it's called `ChainLayer`, and it's easier to follow because it reverses the ZIO approach. We can now think of it in this way: the output effects of the first layer are the input effects of the second layer. The main function of `ChainLayer`, then, is to derive which outputs effects from the first layer remain unhandled.

kyo-core/shared/src/main/scala/kyo/layers.scala
johnhungerford: It's now `Layer[In, Out]`, where `In` is the intersection of all effects the layer will handle, and `Out` is the intersection of the effects that will remain to be handled.

kyo-core/shared/src/main/scala/kyo/layers.scala
johnhungerford: Changed `++` to `add` and `>>>` to `chain`.  I'm avoiding `andThen` for now, since that could easily mean what is currently `add` (since `add` just handles applies the first layer *and then* the second).

Whether to keep `add` as `add` and not change it to `andThen` will depend on whether it is effectively commutative so long as it does not change the type of `T` when it handles an effect.

johnhungerford:
I think the build error is resolved: I had not defined the implicit `Flat` parameter for `Layer.run` correctly.

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,131 @@
+package kyo
+
+object layers {
+
+  trait Layer[In, Out] { self =>
+    def run[T, S](effect: T > (S with In))(implicit fl: Flat[T > (S with In)]): T > (S with Out)
+
+    final def add[Out1, In1](other: Layer[In1, Out1]): Layer[In with In1, Out with Out1] =
```
fwbrasil, line 8: Could you rename to `andThen`? I think it'd be better to communicate that order is important.

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,131 @@
+package kyo
+
+object layers {
+
+  trait Layer[In, Out] { self =>
+    def run[T, S](effect: T > (S with In))(implicit fl: Flat[T > (S with In)]): T > (S with Out)
```
fwbrasil, line 6: In Scala 2, methods that remove items from the pending effects need to have the effect to be removed as the first in the type intersection otherwise the compiler isn't able to infer types correctly. You might be able to avoid type annotations if you use `effect: T > (In with S)`.

fwbrasil:
Other the two last minor comments, the change LGTM! Reasoning about Kyo's effect handling and how the layers are composed, I can't see potential issues. I normally try to design effects in Kyo by breaking up problems into smaller reusable pieces that handle specific behaviors. The use case for `Layers` seems different enough from `Joins` for them to be separate implementations. We can revisit this later as well.

Thanks for the contribution! This will be the first meta-effect in Kyo, it's nice to see how it generalized from ZIO's version :)

johnhungerford:
I have renamed `add` to `andThen` so that layers are not too constrained. If/when functionality is added to automatically compose layers, we can add restrictions there to accept only layers that can be combined in any order without changing behavior.

kyo-core/shared/src/main/scala/kyo/layers.scala
fwbrasil: I see. It could be that the compiler isn't able to infer the removal based on the generic type parameter.

johnhungerford:
Ok I think it's good to go. This was fun! (This is my first significant OSS contribution, by the way. Prior to this, I've only made one trivial change to zio-http).

fwbrasil:
Nice!! You are definitely off to a good start 🙏

PR #170, opened 2024-01-17
Remove v1 from base url openai endpoint

fabioyamate:
Follow the same convention in openai-python library where the `/v1` prefix is part of the baseUrl.

https://github.com/openai/openai-python?tab=readme-ov-file#configuring-the-http-client

This is mostly due to Azure endpoints do not have the `v1` path; it is straight `/chat/completions.`

Optionally, we could change the apiUrl to baseUrl to have the same nomenclature if it makes sense.


fwbrasil:
sorry @fabioyamate. I didn't notice your PR and made the change directly on main 🤦

PR #200, opened 2024-03-09
Update `>` to `<` in `Flat` implicit error

kitlangton:
I've been exploring the codebase to try to get a grasp on some of the magic (simultaneously scouring the _Do Be Do Be Do_ paper and attempting to build a toy replica) 🤩 It's such a cool library, @fwbrasil! 

Back in the day, I was a huge fan of Haskell's [Polysemy](https://github.com/polysemy-research/polysemy#readme)—an extensible effects library with some really great ergonomics. When I switched to Scala, I sought out something similar. And after flailing about the Cats ecosystem for a time (trying and failing to enjoy `Eff`), I finally gave ZIO a try. Of course, ZIO has a completely different implementation, but it provided an even better experience than Polysemy, whereas I'd found the Cats ecosystem to be significantly more verbose than Haskell (all those required type annotations 😭).

It's extremely cool to see that original extensible-effects-ish model return with a vengeance 🤘 (that is, with ZIO-like ergonomics + crazy performance).

---

Anyhow, while exploring the code, I did notice that perhaps this error message didn't get updated along with the change from `>` to `<` as the Pending symbol. So here's a one-line—nay, one-character—MR to go along with my accolades 😜.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/internal/FlatImplicits.scala
+++ b/kyo-core/shared/src/main/scala/kyo/internal/FlatImplicits.scala
@@ -37,7 +37,7 @@ object FlatImplicits:
         def print(t: TypeRepr): String =
             t match
                 case Kyo(t, s) =>
-                    s"${print(t)} > ${print(s)}"
+                    s"${print(t)} < ${print(s)}"
                 case _ => t.show
 
         def fail(msg: String) =


nightscape:
Hey @kitlangton, nice to see you back in the Scala + effects world 😃 
I enjoyed your Scala videos very much!!
Hope you get to do some more when circumstances allow ❤️

fwbrasil:
Hey Kit!! Nice to see you here :) Yeah, I'm excited about Kyo! Something nice about the project as well is how the effect mechanism allows for very isolated implementations throughout the codebase. Most effects don't know about each other and It's possible to make changes to major features in isolation. I think this will be an important differentiator as the codebase grows.

Issue #195, opened 2024-03-02
State and Writer effects


ghostdogpr:
Hi!

I am evaluting Kyo as a potential replacement for `ZPure`. I am using `ZPure` in a project with high traffic (gaming) and complex business logic and the amount of allocations is staggering (cats `IndexedReaderWriterStateT` is worse), so Kyo's low overhead approach is definitely interesting.

While I see clearly how to replace the reader and the error effects, I don't see clear equivalents for state and writer. `Locals` are only modified in their inner scope so I don't think it can be used. I can use `Envs[AtomicRef[A]]` but that is overkill and even slower because we don't need concurrent modifications.

Is there a way to do it that I am not seeing? If not, is that something that could be added?

ghostdogpr:
@fwbrasil in case you're interested, I have a small repo with a benchmark comparing various solutions for a reader/writer/state: https://github.com/ghostdogpr/mtl-benchmarks

Each benchmark op is running 1000x a set of 3 operations: read from an env, write a log and modify a state (which is a simplified version of the business logic I run in production).
Here's the commit adding kyo: https://github.com/ghostdogpr/mtl-benchmarks/commit/34304738b58ac40867e2839cbafbffb05658d7b8 (I used `Vars` for the writer, will replace after you change `Sums` like you mentioned).

Kyo fares pretty well overall but is still way slower (and allocates more) than ZPure.
```
[info] Benchmark                               Mode  Cnt     Score     Error  Units
[info] Benchmarks.MTLReaderWriterStateEither  thrpt    5  1522.572 ± 117.366  ops/s
[info] Benchmarks.MTLReaderWriterStateIO      thrpt    5   552.119 ±  55.575  ops/s
[info] Benchmarks.MTLZIO                      thrpt    5   581.533 ± 575.613  ops/s
[info] Benchmarks.MTLZPure                    thrpt    5  1983.791 ±  85.645  ops/s
[info] Benchmarks.ZPure                       thrpt    5  6142.444 ± 120.744  ops/s
[info] Benchmarks.kyo                         thrpt    5  2058.202 ± 264.282  ops/s
[info] Benchmarks.readerWriterStateIO         thrpt    5   919.416 ±   9.923  ops/s
```
Hope this is interesting. Let me know if I did something wrong or unidiomatic.

fwbrasil:
[...]

PR #202, opened 2024-03-15
Add ZPure and RWST to mtl benchmarks

ghostdogpr:
See https://github.com/getkyo/kyo/issues/195 (originally from https://github.com/ghostdogpr/mtl-benchmarks)

```
[info] Benchmark            Mode  Cnt     Score     Error  Units
[info] MtlBench.syncKyo    thrpt    5  2822.207 ±  66.487  ops/s
[info] MtlBench.syncRWST   thrpt    5  2574.217 ± 155.879  ops/s
[info] MtlBench.syncZPure  thrpt    5  9352.017 ± 303.491  ops/s
```


fwbrasil:
awesome! thank you 🙏

PR #206, opened 2024-03-20
Use sys.env and sys.props instead of Java equivalents

nightscape:
Just a minor improvement: Scala has nice built-in ways to handle environment variables and system properties 😃


fwbrasil:
cool! I didn't know about them 🙏

PR #230, opened 2024-03-26
improve Fibers.race to wait for the first successful result

fwbrasil:
This seems a common behavior for effect systems. The current design was based on Twitter's Future, which also doesn't wait for a successful result by default but I think the new behavior is more intuitive given the similarity with ZIO and other similar solutions. Background from discord: https://discord.com/channels/1087005439859904574/1087005440623259681/1222265340319371334

cc/ @jamesward

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/fibers.scala
+++ b/kyo-core/shared/src/main/scala/kyo/fibers.scala
@@ -11,6 +11,7 @@ import scala.collection.immutable.ArraySeq
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
+import scala.runtime.AbstractFunction1
 import scala.util.*
 import scala.util.control.NonFatal
 import scala.util.control.NoStackTrace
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/fibers.scala
```diff
@@ -171,24 +172,33 @@ object Fibers extends Joins[Fibers]:
...
     def raceFiber[T](l: Seq[T < Fibers])(using f: Flat[T < Fibers]): Fiber[T] < IOs =
         l.size match
             case 0 => IOs.fail("Can't race an empty list.")
             case 1 => Fibers.run(l(0))
-            case _ =>
+            case size =>
                 Locals.save.map { st =>
                     IOs {
-                        val p = new IOPromise[T]
+                        class State extends AbstractFunction1[T < IOs, Unit]:
```
fwbrasil, line 184: I've added this class to reduce the memory footprint of the method. It avoids the closure allocation in the `foreach` body.

kyo-core/shared/src/main/scala/kyo/fibers.scala
```diff
@@ -171,24 +172,33 @@ object Fibers extends Joins[Fibers]:
...
+                            val p       = new IOPromise[T]
+                            val pending = new AtomicInteger(size)
+                            def apply(v: T < IOs): Unit =
+                                val last = pending.decrementAndGet() == 0
+                                try discard(p.complete(IOs.run(v)))
+                                catch
+                                    case ex if (NonFatal(ex)) =>
```
hearnadam, line 191: Do we need to branch on `NonFatal`? When would we not want to route errors through `IOs.fail`?

fwbrasil: Yes, fatal exceptions, which include interruptions, linkage errors, and OOMs are meant to immediately halt the execution of the current thread. A system can end up in a “zombie” state if that’s not the case.

hearnadam (approved):

Issue #229, opened 2024-03-26
Forking fibers with other effects


fwbrasil:
## Problem

There's an important design challenge with Kyo's flexible effects: forking fibers can't be easily done with arbitrary effects. ZIO for example is able to "hardcode" the effect handling when forking a fiber because both effects it supports, dependency injection and short circuiting, can be unambiguously handled by the runtime when forking a fiber. It only needs to save the environment values from the forking call site and propagate failures correctly. That's not the case with Kyo since the pending effect set is essentially unbounded and can have arbitrary effects. We still don't recommend users creating new effects, but that's likely to become the case for advanced users at some point.

Currently, users have to manually handle other effects before forking a fiber:

```scala
def ko(v: Int < Aborts[String]) = Fibers.init(v) // Compilation failure

def ok(v: Int < (Aborts[String] & IOs)): Fiber[Either[String, Int]] < IOs =
    Fibers.init(Aborts[String].run(v))
```

This approach is confusing and cumbersome. I've been exploring this problem for some time but I couldn't identify a design that seems to have the usability I'm looking for yet. This problem can actually be split into two:

1. **Explicit forking**: Methods that return a `Fiber` instance like `Fibers.init` and `Fibers.parallelFiber`. This is the most difficult case to provide a good UX since there's an expectation that the `Fiber` instance represents a computation that will finalize at some point, not suspend again with some effect.
2. **Implicit forking**: Methods that fork fibers but don't return a `Fiber` instance like `Fibers.parallel` and `Fibers.race`. In these cases where the method immediately joins a fiber are a bit easier because the API doesn't need to provide a `Fiber` representing the end of the computation.

### Forking combinators for larger effects

This has been the main approach for some time now. The `Joins` [trait](https://github.com/getkyo/kyo/blob/e03e9d3317a535b55eea17d33c2e0524bb3d7107/kyo-core/shared/src/main/scala/kyo/Joins.scala#L3) provides forking combinators that can be extended by other effects. The main example of this approach is the `AIs` effect. It extends `Joins` and [implements](https://github.com/getkyo/kyo/blob/main/kyo-llm/shared/src/main/scala/kyo/llm/ais.scala#L213) the correct wiring for its specific underlying effects.

[...]

PR #235, opened 2024-03-29
unison-inspired streams

fwbrasil:
An initial implementation of the `Streams` effect with an encoding inspired by Unison's `Stream` [ability](https://exercism.org/tracks/unison/exercises/stream-ops). With this encoding, computations can emit multiple streams. For example: `Int < (Stream[Int] & Stream[String])` is a computation that will emit two streams and return an `Int`.

Although the encoding is elegant, I'm not sure about usability. Performing more complex operations on elements of the pending effect set is something other effects don't do and users might have trouble to understand the approach:

```scala
// Note how 'emit' returns a computation that
// produces 'Unit' and has the 'Streams[Int]' 
// pending effect indicating that it also produces 
// a stream of 'Int' values.
val a: Unit < Streams[Int] =
    Streams[Int].emit(Seq(1, 2, 3))

// It's possible to transform the computation freely
// and 'Streams[Int]' continues present indicating
// that the computation also produces a stream.
val b: Int < Streams[Int] =
    a.andThen(42)

// A computation can produce multiple streams.
val c: Int < (Streams[Int] & Streams[String]) =
    b.map(v => Streams[String].emit("42").andThen(v))

// It's possible to perform operations on the
// streams a computation produces. For example,
// it's possible to transform 'Stream[String]'
// to 'Stream[Int]'.
val d: Int < Streams[Int] =
    Streams[String].transform(c)(v => v.toInt)

// A great property of this encoding is how
// pure the effect is. Currently, the only exception is
// the 'buffer' method that uses 'Fibers' to provide
// backpressured processing with buffering via
// channels.
val e: Int < (Streams[Int] & Fibers) =
    Streams[Int].buffer(10)(d)

// It's possible to handle the 'Streams' effect,
// which returns a tuple with the stream contents
// and the result of the computation.
val f: (Seq[Int], Int) < Fibers =
    Streams[Int].runSeq(d)

// prints '(List(1, 2, 3, 42), 42)'
println(IOs.run(Fibers.run(f)))
```

I wonder if we could improve usability with a wrapper API. For example, maybe we could add a `type Stream[T, V, S] = T < (Streams[V] & S)` or `type Stream[V, S] = Unit < (Streams[V] & S)` with extension methods. 

Although the implementation isn't micro-optimized and is missing chunking, which is a major optimization, the initial benchmark [results](https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/bef1f4e507f05ee7dcc877f79d2da646/raw/21ddf4c2fb84c1db648ed872e4f7b2cfe5efd386/jmh-result.json) indicate the performance isn't too far from alternatives:

image

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/streams.scala
+++ b/kyo-core/shared/src/main/scala/kyo/streams.scala
@@ -1,0 +1,154 @@
+package kyo
+
+import kyo.*
+import kyo.core.*
+import kyo.core.internal.*
+
+class Streams[V] extends Effect[Streams[V]]:
+    type Command[T] = V
+
+end Streams
+
+object Streams:
+
+    import internal.*
+
+    class Done
+    object Done extends Done
+
+    private case object streams extends Streams[Any]
+    def apply[V]: Streams[V] = streams.asInstanceOf[Streams[V]]
+
+    extension [V](s: Streams[V])(using tag: Tag[Streams[V]], flat: Flat[V])
+
+        def emit(v: V): Unit < Streams[V] =
+            suspend(s)(v)
+
+        def emit(v: V, tail: V*): Unit < Streams[V] =
+            emit(v +: tail)
+
+        def emit(v: Seq[V]): Unit < Streams[V] =
+            if v.isEmpty then ()
+            else emit(v.head).andThen(emit(v.tail))
+
+        def buffer[T: Flat](size: Int)(v: T < Streams[V]): T < (Streams[V] & Fibers) =
+            Channels.init[V | Done](size).map { ch =>
+                Fibers.init(runChannel(ch)(v)).map(f => emit(ch).andThen(f.get))
+            }
+
+        def take[T: Flat, S](n: Int)(v: T < (Streams[V] & S)): T < (Streams[V] & S) =
+            Vars[TakeN].run(n) {
+                reemit(v) { v =>
+                    Vars[TakeN].get.map {
+                        case 0 =>
+                            ()
+                        case n =>
+                            Vars[TakeN].set(n - 1)
+                                .andThen(emit(v))
+                    }
+                }
+            }
+
+        def drop[T: Flat, S](n: Int)(v: T < (Streams[V] & S)): T < (Streams[V] & S) =
+            Vars[DropN].run(n) {
+                reemit(v) { v =>
+                    Vars[DropN].get.map {
+                        case 0 =>
+                            emit(v)
+                        case n =>
+                            Vars[DropN].set(n - 1)
+                    }
+                }
+            }
+
+        def filter[T: Flat, S](v: T < (Streams[V] & S))(
+            f: V => Boolean < S
+        ): T < (Streams[V] & S) =
+            reemit(v) { v =>
+                f(v).map {
+                    case false => ()
+                    case true  => emit(v)
+                }
+            }
+
+        def transform[T: Flat, V2: Flat, S, S2](v: T < (Streams[V] & S))(
+            f: V => V2 < (S & S2)
+        )(using Tag[Streams[V2]]): T < (Streams[V2] & S & S2) =
+            reemit(v) { v =>
+                f(v).map(Streams[V2].emit)
+            }
+
+        def collect[T: Flat, S, S2, V2: Flat](v: T < (Streams[V] & S))(
+            f: PartialFunction[V, Unit < (Streams[V2] & S2)]
+        ): T < (Streams[V2] & S & S2) =
+            reemit(v) { v =>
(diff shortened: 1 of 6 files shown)


kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -0,0 +1,158 @@
...
+
+        def emit(v: Seq[V]): Unit < Streams[V] =
+            if v.isEmpty then ()
+            else emit(v.head).andThen(emit(v.tail))
+
+        def buffer[T: Flat](size: Int)(v: T < Streams[V])(
+            using Flat[V]
+        ): T < (Streams[V] & Fibers) =
+            Channels.init[V | Done](size).map { ch =>
+                Fibers.init(runChannel(ch)(v)).map(f => emit(ch).andThen(f.get))
```
fwbrasil, line 38: Ideally, `v` should allow other effects (`T < (Streams[V] & S)`) but it's not possible because of https://github.com/getkyo/kyo/issues/229.

kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -0,0 +1,154 @@
...
+                    def resume[T, U: Flat, S3](command: V, k: T => U < (Streams[V] & S3)) =
+                        handle(handler(f(acc, command)), k(().asInstanceOf[T]))
+            val a = handle(handler(acc), v)
+            a
+        end runFold
+
+        def runSeq[T: Flat, S](v: T < (Streams[V] & S))(
+            using Flat[V]
+        ): (Seq[V], T) < S =
+            def handler(acc: List[V])
```
hearnadam, line 117: Perhaps worth adding a specific version of `runSeq` that only handles `Stream[V]` to avoid the somewhat awkward return type.

kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -0,0 +1,154 @@
...
+                : ResultHandler[Const[V], [T] =>> (Seq[V], T), Streams[V], Any] =
+                new ResultHandler[Const[V], [T] =>> (Seq[V], T), Streams[V], Any]:
+                    def pure[T](v: T) = (acc.reverse, v)
+                    def resume[T, U: Flat, S](command: V, k: T => U < (Streams[V] & S)) =
+                        handle(handler(command :: acc), k(().asInstanceOf[T]))
+            val a = handle(handler(Nil), v)
+            a
```
hearnadam, line 124: Did you forget to clean this temporary val up? Or does it force specific behavior from the jvm?

kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -0,0 +1,158 @@
...
+
+        private inline def reemit[T: Flat, S, S2, V2: Flat](v: T < (Streams[V] & S))(
+            inline f: V => Unit < (Streams[V2] & S2)
+        ): T < (Streams[V2] & S & S2) =
+            def loop[T2, S](v: T2 < (Streams[V] & S))(using
+                flat: Flat[T2]
+            ): T2 < (Streams[V2] & S & S2) =
+                val handler = new Handler[Const[V], Streams[V], Streams[V2] & S2]:
+                    def resume[T3, U: Flat, S3](command: V, k: T3 => U < (Streams[V] & S3)) =
+                        f(command).andThen(loop(k(().asInstanceOf[T3])))
```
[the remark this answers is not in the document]
hearnadam: Can we test the safety via generating a long running test?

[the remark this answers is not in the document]
fwbrasil: I've added a few tests but couldn't reproduce a stack overflow yet. I'm also exploring alternatives to make handlers more flexible to address this case. I don't see this issue as a blocker but we should create a ticket for it.

kyo-core/shared/src/main/scala/kyo/streams.scala
fwbrasil: I think that's a fundamental characteristic of the encoding since the the computation produces a value and a stream. We could have a `runSeqUnit` that discards the result of the computation but I'd prefer to keep only the canonical APIs for now.

kyo-core/shared/src/main/scala/kyo/streams.scala
fwbrasil: I'm not sure why type inference fails if I don't use the val.

fwbrasil:
@hearnadam I'm merging this to create a follow up PR with usability improvements but please feel free to raise issues or concerns with this change so we can address them. Thank you for the review!

Issue #204, opened 2024-03-19
KyoApp.run should accept `Aborts[Throwable]`


hearnadam:
When developing applications, I expect most Kyo users will want to replicate existing effect systems' behaviors when it comes to error short circuiting. As such, we should provide an opinionated default in KyoApp which sets the effect evaluation order.

While It's possible we could offer multiple functions, eg `runPure` - the wider type can always accept the narrower error free type.

### Simple Reproducer
```
import kyo._

object Main extends KyoApp {
  run {
    for {
      _ <- Consoles.println(s"Main Args: ${args.mkString(",")}")
      _ <- Aborts(new RuntimeException("FAILED"))
    } yield ()
  }
}
```

```
[error] 9 |  }
[error]   |   ^
[error]   |Method doesn't accept nested Kyo computations.
[error]   |Detected: 'scala.Unit > kyo.Aborts[java.lang.RuntimeException] > kyo.fibers$package.Fibers & kyo.Resources'. Consider using 'flatten' to resolve. 
[error]   |Possible pending effects mismatch: Expected 'kyo.fibers$package.Fibers & kyo.Resources', found 'kyo.Aborts[java.lang.RuntimeException]'.
[error] one error found
```

PR #237, opened 2024-04-01
Enable `KyoApp.run` to accept `Aborts[Throwable]`

hearnadam:
Resolves https://github.com/getkyo/kyo/issues/204

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
+++ b/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
@@ -34,12 +34,17 @@ object KyoApp:
             initCode += (() => handle(v))
     end Base
 
-    type Effects = Fibers & Resources
+    type Effects = Fibers & Resources & Aborts[Throwable]
+
+    def trying[T](timeout: Duration)(v: T < Effects)(
+        using f: Flat[T < Effects]
+    ): Try[T] =
+        IOs.run(runFiber(timeout)(v).block(timeout))
 
     def run[T](timeout: Duration)(v: T < Effects)(
         using f: Flat[T < Effects]
     ): T =
-        IOs.run(runFiber(timeout)(v).block(timeout).map(_.get))
+        trying(timeout)(v).get
 
     def run[T](v: T < Effects)(
         using f: Flat[T < Effects]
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/test/scala/kyoTest/KyoAppTest.scala
```diff
@@ -5,18 +5,25 @@ import scala.concurrent.duration.*
 
 class KyoAppTest extends KyoTest:
 
-    "effects" in run {
-        val app = new App:
```
hearnadam, line 9: Was it intentional to create a new `App` instead of a `KyoApp`?

kyo-core/shared/src/main/scala/kyo/aborts.scala
```diff
@@ -46,6 +47,15 @@ object Aborts:
                     fail(ex.asInstanceOf[V])
             }
 
+        def trying[T, S](v: T < (Aborts[V] & S))(
```
hearnadam, line 50: This function is useful, but perhaps unnecessary. We can also just `.map(_.toTry)` at the call site of `Aborts[Throwable].run`.

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -34,12 +34,17 @@ object KyoApp:
             initCode += (() => handle(v))
     end Base
 
-    type Effects = Fibers & Resources
+    type Effects = Fibers & Resources & Aborts[Throwable]
+
+    def trying[T](timeout: Duration)(v: T < Effects)(
```
fwbrasil, line 39: What do you think of naming this method `attempt`? We already have `IOs.attempt`.

kyo-core/shared/src/test/scala/kyoTest/KyoAppTest.scala
fwbrasil: No, I think it's a leftover of refactorings.

kyo-core/shared/src/main/scala/kyo/aborts.scala
fwbrasil: I'd prefer not to add it. It doesn't seem it'd be a common operation and there's `IOs.attempt` that is similar.

fwbrasil:
Thank you for working on this @hearnadam! 🙏

fwbrasil:
The JS build is failing because it's not possible to block fibers: 

```
[info] - main *** FAILED *** (6 milliseconds)
[info]   java.lang.UnsupportedOperationException: fiber.block is not supported in ScalaJS
[info]   at java.util.concurrent.locks.LockSupport$.fail(/home/runner/work/kyo/kyo/kyo-core/js/target/scala-3.4.1/kyo-core-test-fastopt/main.js:6936:51)
[info]   at kyo.scheduler.IOPromise$$anon$1.apply(/home/runner/work/kyo/kyo/kyo-core/js/target/scala-3.4.1/kyo-core-test-fastopt/main.js:60550:7)
[info]   at kyo.scheduler.IOPromise.kyo$scheduler$IOPromise$$_$f$proxy1$1(/home/runner/work/kyo/kyo/kyo-core/js/target/scala-3.4.1/kyo-core-test-fastopt/main.js:60481:14)
[info]   at kyo.scheduler.IOPromise$$anon$2.apply(/home/runner/work/kyo/kyo/kyo-core/js/target/scala-3.4.1/kyo-core-test-fastopt/main.js:87701:62)
```

You can use `Platform.isJVM` to [limit the test](https://github.com/getkyo/kyo/blob/main/kyo-core/shared/src/test/scala/kyoTest/KyoTest.scala#L42) to the JVM. I think we'd need a follow up ticket to see how can we support 'KyoApp' in JS, though.

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
hearnadam: Sure, that seems more idiomatic.

kyo-core/shared/src/test/scala/kyoTest/Tagged.scala
```diff
@@ -0,0 +1,8 @@
+package kyoTest
+
+import org.scalatest.Ignore
+import org.scalatest.Tag
+
+object Tagged:
+    val jvmOnly = Tag(if Platform.isJVM then "" else classOf[Ignore].getName)
```
hearnadam, line 7: @fwbrasil I added this so that the tests will properly be marked as `ignored` in the console. Let me know what you think.

fwbrasil: Cool, I didn't know about `Tag`. LGMT. Is it possible to tag individual tests as well?

hearnadam: Actually, due to the inclusion of `AsyncFreeSpec` that's the only reasonable way to use Tags, unless we define them as a java annotation - which I am not inclined to do.

Issue #145, opened 2023-12-10
Unified API for effects with extension methods


johnhungerford:
The current design of kyo keeps effect types segregated, so that a separate import is required for each one, and each has a separate API.

```scala
import kyo.options.*
import kyo.ios.*
import kyo.fibers.*

val optEffect = Options.get(Some("hello world"))
val ioEffect = IOs(println("hell world"))
val forkedEffect = Fibers.init(ioEffect)

val handledOpt = Options.run(optEffect)
val handledIo = IOs.run(ioEffect)
val handledFiber = IOs.run(Fibers.run(forkedEffect))
```

It would be much easier for the developer to use a single export for core effects, and have extension methods on `T < Fs` to more easily handle effects. I'm obviously thinking of the ZIO API here:

```scala
import kyo.*

val optEffect = KYO.fromOption(Some("hello world"))
val ioEffect = KYO.attempt(println("hello world")
val forkedEffect = ioEffect.fork

val handledOpt = optEffect.runOptions
val handledIo = ioEffect.runIO
val handledFiber = forkedEffect.join
```

fwbrasil:
I think moving effects to extension methods of `>` would hurt usability given that users would have to browse over several methods to encounter the ones related to the effects they need to use. Calling effect methods explicitly doesn't introduce much syntax overhead and makes code more friendly to newcomers IMO.

I agree that'd be nice to avoid having to import each of the effects, though. I used to have a `kyo.all` object that would export all core effects but I removed it once I introduced Scala 2 support. Kyo also has multiple modules now, which complicates things a bit. Maybe we could have a `kyo-all` artifact that depends on all other modules and exports all effects?

johnhungerford:
From what I can see, everything that I would want to be in the unified export is in the `core` module (`Aborts`, `Envs`, `Options`, `IOs`, `Fibers`), so I'm doubtful another module would be necessary, at least for my own needs.

One more plug for using extensions methods: 
1. using static functions on objects like `Options`, `Aborts`, etc, requires "back-tracking." You come up with an expression, and then you have to return to the beginning of the expression to then handle those effects, rather than calling a method at the end of the expression at the end, which is where you were when you thought of it:

```scala
val effect = Consoles
    .readLine
    .map(line => Tries.catching(v.toInt)
    .map(int => Options.get(Try(23 / int).toOption))
    // At this point I realize I want to handle `Tries` and `Options` to just get an `Option[Double]`
[...]

PR #238, opened 2024-04-01
streams: usability improvements

fwbrasil:
The `Streams` effect introduced by https://github.com/getkyo/kyo/pull/235 has an inconvenient API that doesn't allow users to chain stream transformations and operates at a high level of abstraction by transforming `Streams` in the pending effect set. This PR aims to restructure the API to provide better usability. The main idea is providing a `Stream` opaque type that makes it more convenient to use `Streams`:

```scala
// A Stream produces a final value 'T', 
// emits stream elements 'V' and can 
// have other pending effects 'S'.
opaque type Stream[+T, V, -S] = 
    T < (Streams[V] & S)

// The 'initSource' method takes a regular
// computation and transforms it into a
// 'Stream'.
def a(v: Int < (Streams[String] & IOs)): Stream[Int, String, IOs] =
    Streams.initSource[String](v)

// It's possible to freely "rotate" between
// a 'Stream' and its underlying computation
// using 'Streams.initSource' and 'stream.get'.
def b(v: Stream[Int, String, IOs]): Int < (Streams[String] & IOs) =
    v.get

// If a computation produces multiple streams,
// the 'initSource' type parameter allows users to
// select which 'Streams[V]' they want to work with.
def c(v: Int < (Streams[String] & Streams[Int])): Stream[Int, String, Streams[Int]] =
    Streams.initSource[String](v)

// It continues possible to emit elements as before
// without a 'Stream'.
val d: Unit < Streams[Int] = 
    Streams.emitValue(1)

// There are a few 'Streams.init*' methods and
// stream transformations can now only be performed
// on a 'Stream' type. Note how running a stream still
// produces a tuple since it produces the stream + a value.
val e: (Int, Unit) < Any =
    Streams.initSeq(0 until 1000)
            .filter(_ % 2 == 0)
            .transform(_ + 1)
            .runFold(0)(_ + _)
```

I wanted to avoid types with multiple parameters as much as possible but I think this case calls for it given the low usability without it. A simplification we could make is allowing only computations that return unit (`opaque type Stream[V, S] = Unit < (Streams[V] & S)`) but I think the restriction would degrade experience since users wouldn't be able to freely rotate between the `Stream` representation and regular computations.

This PR has some intersection with https://github.com/getkyo/kyo/issues/145. This new approach to encode effects could be a good technique to make Kyo's APIs more fluent.

I've also added a new benchmark that uses buffering. The performance is [still not far](https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/eb2856d3e66430bb115f03494d368ea6/raw/716940c109179009c08d6d772ac577375f5c7fcb/jmh-result.json) from alternatives even without chunking and micro-optimizations:

image

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/StreamBufferBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/StreamBufferBench.scala
@@ -1,6 +1,6 @@
 package kyo.bench
 
-class StreamBench extends Bench.SyncAndFork[Int]:
+class StreamBufferBench extends Bench.ForkOnly[Int]:
 
     val seq = (0 until 10000).toList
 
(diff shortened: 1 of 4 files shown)


fwbrasil:
I'm merging this PR and will continue the development of the effect in main directly to speed things up since the overall design is covered by the two PRs so far. The next changes I'll work on are basically new functionality (takeWhile, concat, etc). Please feel free to provide feedback on the design here, though!

Issue #231, opened 2024-03-27
Envs.Elide cannot be reduced for uninhabitable intersections


johnhungerford:
When I try to compile the following:

```scala
val effect: Unit < Envs[String & Int & Boolean] = ???
val provided: Unit < Envs[Int & Boolean] = Envs[String].run("string")(effect)
```

I get the following error:

```
[error]     |                  Found:    Unit < kyo.Envs.Elide[Int & Boolean]
[error]     |                  Required: Unit < kyo.Envs[Int & Boolean]
[error]     |
[error]     |                  Note: a match type could not be fully reduced:
[error]     |
[error]     |                    trying to reduce  kyo.Envs.Elide[Int & Boolean]
[error]     |                    failed since selector Int & Boolean
[error]     |                    is uninhabited (there are no values of that type).
```

It looks like the issue is that match types won't reduce when it recognizes that type intersection being matched is uninhabitable (there cannot be a value of that type). Since this will frequently be the case for `Envs` intersections, I don't think match types will work here.

I see two alternatives:

1. Make `Envs` contravariant with `E` so that types are tracked by union rather than intersection.
2. Use a type class with an inner (dependent) type to derive the "remainder"

fwbrasil:
thanks for the report! I agree with making `Envs` contravariant. I also noticed this other bug with the new encoding: https://github.com/getkyo/kyo/issues/215

hearnadam:
I personally don't see the value of using union types in the `Env`. Happy to try my hand at a PR this week.

johnhungerford:
My preference is for approach 2. In general, I've found match types don't work super well and I wouldn't recommend making decisions around them. I would only use unions in those cases where it makes semantic sense, like aborts.

Deriving the remainder of a dependency intersection is not too complicated (from discord):

```scala
type HasEnv[E, ES]:
  type Out

trait LowPriorityHasEnvs:
  given hasEnv[E, ER]: HasEnv[E, E & ER] with
    type Out = Envs[ER]

object HasEnv extends LowPriorityHasEnvs:
  given isEnv[E]: HasEnv[E, E] with
    type Out = Any
```

fwbrasil:
For visibility: we had a [discussion](https://discord.com/channels/1087005439859904574/1087005440623259681/1222610350688436305) on this topic yesterday reached a conclusion that using different variances for `Envs` and `Aborts` is necessary for soundness, so we'd need something like approach 2. 

[...]

Issue #215, opened 2024-03-24
Invalid `Envs` type inference


fwbrasil:
I've been doing some testing before cutting a new release and noticed a type inference issue with the new `Envs` implementation. It's possible to specify type parameters to `Envs.run` in a way to produces an invalid pending effect set:

```scala
def t1(v: Int < Envs[Int & String]) =
    Envs[Int].run[Int, Any, Nothing](1)(v)
val v: Int < Any = t1(42) 
```

By specifying `Nothing`, other pending environment values are incorrectly removed from the pending set. I'll move forward with the release but we should find a way to avoid this issue and verify if other APIs could also be misused this way. I tried doing the same with the new `Aborts` but I couldn't reproduce it, I think because of Scala's custom rules for `Nothing` that don't apply to `Any`.

A potential fix would be making `Envs` contravariant like `Aborts`. Although that would preclude using union type environment values as @jiveshungerford has pointed out, I wonder if it could be a reasonable tradeoff to make the APIs more regular for newcomers so they don't need to learn and remember when to use type unions or intersections. I'd prefer to see all elements in the pending effect type parameter as as elements in sets rather than representing type signatures of actual values.

I've added a pending test for this bug: https://github.com/getkyo/kyo/commit/48dd95352e129033b6b562e6bd35870dc9d4e473

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/240

PR #240, opened 2024-04-03
Replace Evns.Elide and Aborts.Elide with dependant types

johnhungerford:
See #231

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/envs.scala
+++ b/kyo-core/shared/src/main/scala/kyo/envs.scala
@@ -17,19 +17,19 @@ object Envs:
         def use[T, S](f: V => T < S)(using Tag[Envs[V]]): T < (Envs[V] & S) =
             get.map(f)
 
-        def run[T, S, VS](env: V)(value: T < (Envs[VS] & S))(
+        def run[T, S, VS, VR](env: V)(value: T < (Envs[VS] & S))(
             using
-            he: HasEnvs[V, VS],
-            t: Tag[Envs[V]],
-            f: Flat[T < (Envs[VS] & S)]
-        ): T < (S & he.Out) =
+            HasEnvs[V, VS] { type Remainder = VR },
+            Tag[Envs[V]],
+            Flat[T < (Envs[VS] & S)]
+        ): T < (S & VR) =
             val handler = new Handler[Const[Unit], Envs[V], Any]:
                 def resume[T2, U: Flat, S1](
                     command: Unit,
                     k: T2 => U < (Envs[V] & S1)
                 ) = handle(k(env.asInstanceOf[T2]))
 
-            handle(handler, value).asInstanceOf[T < (S & he.Out)]
+            handle(handler, value).asInstanceOf[T < (S & VR)]
         end run
     end extension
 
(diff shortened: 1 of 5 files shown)


fwbrasil:
I've played with the change locally and inference seems great now! I think we can consider this a fix for https://github.com/getkyo/kyo/issues/215 as well. With this new approach, an explicit `Nothing` type parameter produces an `Envs[Nothing]`, which I'd say is a reasonable behavior. Previously, `Elide` was completely removing `Envs` from the pending set.

@johnhungerford was the removal of layers intended?

kyo-core/shared/src/test/scala/kyoTest/envsTest.scala
```diff
@@ -40,7 +40,18 @@ class envsTest extends KyoTest:
...
 
-    "invalid inference" in pendingUntilFixed {
+    "reduce large intersection" in {
+        val t1: Int < Envs[Int & String & Boolean & Float & Char & Double] = 18
+        val t2                                                             = Envs[Int].run(42)(t1)
+        val t3 = Envs[String].run("a")(t2)
+        val t4 = Envs[Boolean].run(false)(t3)
+        val t5 = Envs[Float].run(0.23f)(t4)
+        val t6 = Envs[Char].run('a')(t5)
+        val t7 = Envs[Double].run(0.23d)(t6)
```
hearnadam, line 50: If you structure the runs like:

```
Envs[Char & Double].run('a') {
  Envs[Double].run(0.23d)(???)
}
```
Does the type inference work correctly? I think it would be good to have a test for this.

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -22,25 +17,41 @@ object Envs:
...
-            Tag[Envs[V]],
-            Flat[T < (Envs[V & V2] & S)]
-        ): T < (S & Elide[V2]) =
+            he: HasEnvs[V, VS],
+            t: Tag[Envs[V]],
+            f: Flat[T < (Envs[VS] & S)]
+        ): T < (S & he.Out) =
             val handler = new Handler[Const[Unit], Envs[V], Any]:
-                def resume[T2, U: Flat, S](
+                def resume[T2, U: Flat, S1](
```
hearnadam, line 27: Nice catch of the S shadowing.

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -22,25 +17,41 @@ object Envs:
...
+      *
+      * @tparam V
+      *   the dependency included in `VS`
+      * @tparam VS
+      *   all of the `Envs` dependencies represented by type intersection
+      */
+    sealed trait HasEnvs[V, VS]:
+        /** Remaining effect type, once the `V` dependency has been handled
+          */
+        type Out
```
hearnadam, line 47: Minor: instead of `Out` should we call this `Remainder`/`Remaining`?

Not sure if the type name or the type represented will be outputted in editors

hearnadam (approved):

johnhungerford:
@fwbrasil I removed layers for `Aborts` and `Envs` because I'm not sure how to implement it properly. Since the output type of `Envs[A].run` will depend on the type of effect you're running it on, it's not quite right to say the layer "removes `Envs[A]`". Given that `Layer` is up in the air as an approach (at this point I think we should get rid of it), I didn't spend time on it.

hearnadam (approved):

fwbrasil:
@johnhungerford sounds good. I can follow up with a PR to remove layers for now.

hearnadam:
Perhaps it's still possible to support Layers via a more generalized Memoization effect. I will need to think if there are any limitations to generalized effect memoizations.

PR #241, opened 2024-04-03
MOAR Randoms (fixes #234)

### LINKED ISSUE #234: More Randoms
I'd like some more `Randoms` :)

Like:
```scala
  def nextString(length: Int): String < Fibers =
    val lettersFiber = Fibers.parallel:
      Seq.fill(length):
        Randoms.nextValue('a' to 'z')

    lettersFiber.map { chars =>
      chars.mkString
    }

  def nextBytes(length: Int): Array[Byte] < Fibers =
    val random = java.util.Random()
    val array = Array.ofDim[Byte](length)
    random.nextBytes(array)
    IOs(array)
```

fwbrasil:
I've added a few new methods and restructured the implementation to enable unsafe access for performance-sensitive code.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/randoms.scala
+++ b/kyo-core/shared/src/main/scala/kyo/randoms.scala
@@ -1,6 +1,6 @@
 package kyo
 
-abstract class Random:
+trait Random:
     def nextInt: Int < IOs
     def nextInt(n: Int): Int < IOs
     def nextLong: Long < IOs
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/randoms.scala
```diff
@@ -34,8 +115,8 @@ object Randoms:
     val nextInt: Int < IOs =
         local.use(_.nextInt)
 
-    def nextInt[S](n: Int < S): Int < (S & IOs) =
```
fwbrasil, line 37: I've been moving away from providing effectful parameters where not necessary. It introduces overhead and I think being more explicit about where binds happen helps usability.

kyo-core/shared/src/main/scala/kyo/randoms.scala
```diff
@@ -1,27 +1,108 @@
...
     def nextLong: Long < IOs
     def nextDouble: Double < IOs
     def nextBoolean: Boolean < IOs
     def nextFloat: Float < IOs
     def nextGaussian: Double < IOs
+    def nextValue[T](seq: Seq[T]): T < IOs
+    def nextValues[T](length: Int, seq: Seq[T]): Seq[T] < IOs
+    def nextString(length: Int): String < IOs
+    def nextString(length: Int, chars: Seq[Char]): String < IOs
+    def nextBytes(length: Int): Seq[Byte] < IOs
```
fwbrasil, line 15: I've made this method return a `Seq` instead of `Array` since it's immutable and might prevent newcomers from making mistakes.

fwbrasil, line 15: @jamesward I've made this method return a Seq instead of Array since it's immutable and might prevent newcomers from making mistakes.

kyo-core/shared/src/test/scala/kyoTest/randomsTest.scala
```diff
@@ -4,45 +4,168 @@ import kyo.*
...
-        assert(v == 30d)
-    }
-    "nextFloat" in {
-        val v = IOs.run(Randoms.let(testRandom)(Randoms.nextFloat))
-        assert(v == 40f)
-    }
-    "nextGaussian" in {
-        val v = IOs.run(Randoms.let(testRandom)(Randoms.nextGaussian))
-        assert(v == 50d)
+    "mocked" - {
```
hearnadam, line 7: Can we use a java seed instead?

kyo-core/shared/src/main/scala/kyo/randoms.scala
```diff
@@ -1,27 +1,108 @@
...
+                def nextDouble: Double           = random.nextDouble()
+                def nextBoolean: Boolean         = random.nextBoolean()
+                def nextFloat: Float             = random.nextFloat()
+                def nextGaussian: Double         = random.nextGaussian()
+                def nextValue[T](seq: Seq[T]): T = seq(random.nextInt(seq.size))
+                def nextValues[T](length: Int, seq: Seq[T]): Seq[T] =
+                    Seq.fill(length)(nextValue(seq))
+
+                val alphanumeric = (('a' to 'z') ++ ('A' to 'Z') ++ ('0' to '9')).toIndexedSeq
+                def nextString(length: Int): String =
```
hearnadam, line 53: Minor: let's call this `nextStringAlphanumeric`

kyo-core/shared/src/main/scala/kyo/randoms.scala
```diff
@@ -1,27 +1,108 @@
...
-            val nextInt         = IOs(random.nextInt())
-            def nextInt(n: Int) = IOs(random.nextInt(n))
-            val nextLong        = IOs(random.nextLong())
-            val nextDouble      = IOs(random.nextDouble())
-            val nextBoolean     = IOs(random.nextBoolean())
-            val nextFloat       = IOs(random.nextFloat())
-            val nextGaussian    = IOs(random.nextGaussian())
+    trait Unsafe:
+        def nextInt: Int
+        def nextInt(n: Int): Int
```
hearnadam, line 24: Would prefer to use descriptive names `n` -> `inclusive`/`exclusive`

hearnadam (approved):

kyo-core/shared/src/test/scala/kyoTest/randomsTest.scala
fwbrasil: I'd prefer to keep simple mocked tests. It reproduces the scenario where users provide a custom Random.

fwbrasil:
good points about naming! I've updated the code

hearnadam (approved):

Issue #233, opened 2024-03-29
Resources close with Effects


jamesward:
I have an effectful `close` on a `Closeable`:
```scala
case class MyResource(id: String) extends Closeable:
  override def close(): Unit =
    IOs.run(Fibers.run(req(uri"${scenarioUrl(8)}?close=$id")))
```

It'd be great if that could just be:
```scala
case class MyResource(id: String) extends Closeable:
  override def close(): Unit < Fibers =
    req(uri"${scenarioUrl(8)}?close=$id")
```

hearnadam:
A few quick thoughts after reviewing the existing `resources` implementation:

- Java `Closable` can throw - yet we directly wrap it in `IO`. Does this need to be handled?
  - https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html
- It should be simply to create our own trait `Closeables` which we can use for type bounding on the API;
  - `def acquireEffectful[T <: Closeables](resource: => T < IOs): T < Resources =` - needs a better name
  - Then the Java compatible API can be defined in terms of the effectful API
- Supporting multiple effect types in the same queue seems problematic... @fwbrasil any ideas?

johnhungerford:
Instead of requiring users to implement a new type, how about replacing adding a more general `acquireRelease` that supports an `IOs` cleanup:

```scala
def acquireRelease[T, S](resource: => T < S)(release: T => Unit < IOs): T < (S & Resources) =
    resource.flatMap(r => ensure(release(r)).andThen(r))
```

Then we can implement `acquire` as:

```scala
def acquire[T <: Closable, S](closeable: => T & S): T < (S & Resources) =
   acquireRelease(closeable)(c => IOs(closeable.close()))
```

hearnadam:
The reason for needing a new type is to support arbitrary effects via a type parameters. I am inclined to believe that we should not worry about that for now. I will cut a PR this week to implement a basic IOs based `acquireRelease`

fwbrasil:
`Resources` already provides a more low-level API:

```scala
def ensure(v: => Unit < IOs): Unit < Resources
```

The challenge with this change is that `Resources` has to rely on `IOs.ensure` that provides a different flow to ensure execution even in the case of a fiber interruption/cancellation. I can't see a way to support arbitrary effects but maybe we could at least support `Fibers`? We'd need a `Promise` to ensure close finishes before returning. 

```scala
    def run[T, S](v: T < (Resources & S)): T < (Fibers & S) =
[...]

PR #246, opened 2024-04-05
Enable effectful (IOs) `acquireRelease`

hearnadam:
Partially resolves https://github.com/getkyo/kyo/issues/233

Simplified the comment from John to only use `IOs` - we can support other effect types later.

Additionally, we should consider how we can only expose a `Resources.use` function to convert `Resources` to `IOs`... Currently, 'dead' (already closed) Resources can be referenced outside the scope.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/resources.scala
+++ b/kyo-core/shared/src/main/scala/kyo/resources.scala
@@ -20,9 +20,15 @@ object Resources:
                 }
         }
 
-    def acquire[T <: Closeable](resource: => T): T < Resources =
-        lazy val v = resource
-        ensure(v.close()).andThen(v)
+    def acquireRelease[T](acquire: => T < IOs)(release: T => Unit < IOs): T < Resources =
+        IOs {
+            acquire.map { resource =>
+                ensure(release(resource)).andThen(resource)
+            }
+        }
+
+    def acquire[T <: Closeable](resource: => T < IOs): T < Resources =
+        acquireRelease(resource)(r => IOs(r.close()))
 
     def run[T, S](v: T < (Resources & S)): T < (IOs & S) =
         Queues.initUnbounded[Unit < IOs](Access.Mpsc).map { q =>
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/resources.scala
```diff
@@ -20,9 +20,13 @@ object Resources:
                 }
         }
 
+    def acquireRelease[T](acquire: => T < IOs)(release: T => Unit < IOs): T < Resources =
+        acquire.map { resource =>
+            ensure(release(resource)).andThen(resource)
+        }
+
     def acquire[T <: Closeable](resource: => T): T < Resources =
```
fwbrasil, line 28: Maybe we could remove this method? It feels too unsafe since making a resource should typically require `IOs`.

kyo-core/shared/src/main/scala/kyo/resources.scala
```diff
@@ -20,9 +20,13 @@ object Resources:
                 }
         }
 
+    def acquireRelease[T](acquire: => T < IOs)(release: T => Unit < IOs): T < Resources =
+        acquire.map { resource =>
```
fwbrasil, line 24: When I define a method that takes a by-name `T < IOs`, I normally also add an `IOs` suspension since it's easy for users to forget to add it:

```scala
IOs {
   acquire.map { resource =>
      ...
   }  
}
```

kyo-core/shared/src/main/scala/kyo/resources.scala
hearnadam: What if we add an IOs to the type?? Or are you suggesting that it's too unsafe to interop with `Closeable`s?
```
def acquire[T <: Closeable](resource: => T < IOs): T < Resources =
```

fwbrasil: Yeah, I'm not sure relying on Java's unsafe API is ideal but I'm ok with it and adding `IOs`.

fwbrasil:
Thank you!

Issue #244, opened 2024-04-04
Problem with tagged Aborts


johnhungerford:
Minified example:

```scala
                val effect1: Int < Aborts[String | Boolean] =
                    Aborts[String | Boolean].fail("failure")
                val handled1: Either[String, Int] < Aborts[Boolean] =
                    Aborts[String].run(effect1)
                val handled2: Either[Boolean, Either[String, Int]] < Any =
                    Aborts[Boolean].run(handled1)
                handled2.pure
```

This block of code throws the following crazy runtime exception:

```
java.lang.IllegalStateException: Unhandled effect: -520960862|
scala.Boolean
scala.AnyVal
kyo.Abortskyo.core$.Effectkyo.core	
|30
```

The reason, I believe, is that when the original `Aborts` effect is constructed, it summons a `ClassTag` for `String | Boolean`, permitting a handler to match on `Left[String | Boolean, Nothing]`, but when the effect is handled subsequently, the handler matches on `Left[String, Nothing]` and `Left[Boolean, Nothing]`.

Conversely, it will also fail if you compose two `Aborts` effects constructed with different failure types and try to handle the union:

```scala
                val effect: Int < Aborts[String | Boolean] =
                    Aborts[String].fail("failure")
                val handled: Either[String | Boolean, Int] < Any =
                    Aborts[String | Boolean].run(effect)
                handled.pure
```

This is a big problem: it means that, given an effect `A < Aborts[E1 | E2 | E3| ... | EN]`, it is impossible to know how these errors can be handled without causing a runtime error.

I can see two possible ways to solve this problem:

1. Use reflection to check whether an `Aborts` command (i.e., `Left[V, Nothing]`) has an error type that is a _subtype_ of the handler's error type. I think this is possible by comparing `LightTypeTag`s, but I'm not sure.
2. Don't check types at all: when you handle `Envs` you have to handle all of them. In this case, you'd probably want to make `Options` a separate effect.

PR #248, opened 2024-04-05
Improve error messages in case of bugs + reproduce #244

fwbrasil:
Kyo uses a custom tag mechanism based on Izumi that is designed to avoid allocations. The macro generates a string at compile time that becomes a constant in the class pool, so tags can be obtained and compared for equality without any runtime allocations.

That's the reason the error message in #244 is so cryptic. I've fixed the error messages to parse the `Tag` string into a regular Izumi tag to provide a better string representation. With this change, the error message in #244 becomes:

```
[info] - union tags *** FAILED *** (8 milliseconds)
[info]   kyo.package$KyoBugException: Unhandled effect: Aborts[-{String | Boolean}]. Please file a ticket.
[info]   at kyo.package$KyoBugException$.apply(package.scala:61)
[info]   at kyo.package$.bug(package.scala:66)
[info]   at kyo.package$.pure(package.scala:38)
[info]   at kyoTest.abortsTest.$init$$$anonfun$1$$anonfun$6(abortsTest.scala:58)
```

I've also fixed a bug in `tag.parse` when splitting the string payload.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/ios.scala
+++ b/kyo-core/shared/src/main/scala/kyo/ios.scala
@@ -89,7 +89,7 @@ sealed trait IOs extends Effect[IOs]:
         @tailrec def runLoop(v: T < IOs): T =
             v match
                 case kyo: Suspend[IO, Unit, T, IOs] @unchecked =>
-                    require(kyo.tag == tag, "Unhandled effect: " + kyo.tag)
+                    bug(kyo.tag != tag, "Unhandled effect: " + kyo.tag.parse)
                     runLoop(kyo(()))
                 case _ =>
                     v.asInstanceOf[T]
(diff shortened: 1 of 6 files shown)


hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -89,7 +89,7 @@ sealed trait IOs extends Effect[IOs]:
         @tailrec def runLoop(v: T < IOs): T =
             v match
                 case kyo: Suspend[IO, Unit, T, IOs] @unchecked =>
-                    require(kyo.tag == tag, "Unhandled effect: " + kyo.tag)
+                    bug(kyo.tag != tag, "Unhandled effect: " + kyo.tag.parse)
```
hearnadam, line 92: Perhaps it's useful to make a helper function for this specific case:
```
def ensureTagsEq(handlerTag: Tag, effectTag: Tag): Unit =
  if (handlerTag != effectTag) bug(...)
```

And ensure that both tags are printed. I'm guessing when users file a ticket, it would be useful to know which handler threw the exception.

fwbrasil:
oh sorry! I missed your suggestion, I'll follow up with a new PR.

PR #253, opened 2024-04-08
More flexible and lightweight handlers

### LINKED ISSUE #247: More flexible and lightweight handlers
I've been optimizing the `Streams` implementation and I'm facing two major challenges:

**1. Handler allocations**

Stream transformations like `take` and `drop` [use effect handling](https://github.com/getkyo/kyo/blob/04ddd36831b8fd35943754c7ef54b510aade2c34/kyo-core/shared/src/main/scala/kyo/streams.scala#L22). Kyo's current core requires `Handler` allocations in order to update internal state during handling, which is the case for `Stream` transformations. This is generally ok with other effects because they typically handle a single computation but the overhead for streams is more pronounced since they process many computations for emitted values.

**2. Stack safety**

Stream transformations like `take` handle effect suspensions of `Streams[V]` but they emit new elements using the same effect `Streams[V]`. This means that it's not possible to recurse the handling directly. The `reemit` [method](https://github.com/getkyo/kyo/blob/04ddd36831b8fd35943754c7ef54b510aade2c34/kyo-core/shared/src/main/scala/kyo/streams.scala#L132) uses a stack-unsafe loop in order to emit the command then loop to continue handling the rest of the emits with the handler:

```scala
// this would loop forever since the element would be reemited and handled again
def resume[T3, U: Flat, S3](command: V, k: T3 => U < (Streams[V] & S3)) =
    Handle(f(command).andThen(k(().asInstanceOf[T3])))

// as a workaround, I'm using a stack-unsafe loop
def resume[T3, U: Flat, S3](command: V, k: T3 => U < (Streams[V] & S3)) =
    f(command).andThen(loop(k(().asInstanceOf[T3])))
``` 

The core of the issue is that we can't perform effects before returning a `Handle`. 

**Proposed Solution**

It's been some time since our last offering to the gist gods so I've explored a new core encoding 😅   https://gist.github.com/fwbrasil/f8a25450fa2428db207aaea8296003d7
(note: I'm not using tags in the core for simplicity but we'd need tags for performance. Don't miss the fiesta at the end!)

The new encoding has the following `Handler` definition:

```scala
abstract class Handler[E <: Effect[E], State, Result[_], S](val effect: E):

    import effect.Command

    case class Resume[U, S2](st: State, v: U < (E & S2))

    def done[T](st: State, v: T): Result[T] < S

    def resume[T, U, S2](
        st: State,
        command: Command[T],
        k: T => U < (E & S)
    ): (Result[U] | Resume[U, S]) < S

    def handle[T, S2](st: State, value: T < (E & S2)): Result[T] < S = ...
}
```

Main changes:

1. I've renamed `pure`/`apply` to `done`/`resume` and `Handle` to `Resume`. I think it might be easier to understand their purpose.
2. Note how the handler has an extra `S` type parameter, which specifies effects that can be produced during effect handling.
3. Both the `done` and `resume` methods allow effects `S`. This enables `resume` to perform effects `S` before resuming handling, which is what I need to solve the stack safety issue.
4. First-class support for `State` allows us to allocate handlers once and then reuse them by passing the state when calling `handler.handle`.

The key insight of this new encoding is that we have all the information we need in the computation itself to ensure stack safety in `handler.handle`. If `resume` returns a computation with some other effect suspension, we can simply suspend the handling with a `Continue` (via `map` in the impl) since effect suspensions collapse the stack but if it returns a pure `Resume` with no pending effects, we can recurse directly in a stack-safe manner:

```scala
final private def _handle[T, S2](st: State, value: T < (E & S2)): Result[T] < S =
    handle(st, value)
@tailrec final def handle[T, S2](st: State, value: T < (E & S2)): Result[T] < S =
    value match
        case kyo: Suspend[Command, Any, T, S2] @unchecked
            if kyo.effect == effect =>
            resume(st, kyo.command, kyo) match
                // no suspension so we recurse directly!
                case r: Resume[T, S2] @unchecked =>
                    handle(r.st, r.v)
                case v =>
                    v.map {
                        case r: Resume[T, S2] @unchecked =>
                            _handle(r.st, r.v)
                        case r: Result[T] @unchecked =>
                            r
                    }
            end match
        case kyo: Suspend[MX, Any, T, S2] @unchecked =>
            new Continue[MX, Any, Result[T], S](kyo):
                def apply(v: Any) =
                    _handle(st, kyo(v))
        case v: T @unchecked =>
            done(st, v)
end handle
```

With this approach, it's possible to define `stream.take` as:

```scala
def take(n: Int): Stream[T, V, S] =
    val handler =
        new Handler[Streams[V], Int, Id, Streams[V] & S](Streams(Tag[V])):
            def done[T](st: Int, v: T) = v
            def resume[T, U, S2](
                st: Int,
                command: V,
                k: T => U < (Streams[V] & (Streams[V] & S))
            ) =
                st match
                    case 0 =>
                        Resume(0, k(().asInstanceOf[T]))
                    case n =>
                        Streams.emit(command).andThen(Resume(n - 1, k(().asInstanceOf[T])))
            end resume
    Stream(handler.handle(n, s))
end take
```

In the example gist I'm not caching handlers for simplicity but it's possible to do that if the core is encoded in terms of `Tag`. 

@kitlangton in case you want to have some effectful fun 😉

fwbrasil:
Fixes #247. Please see https://github.com/getkyo/kyo/issues/247 for background.


kyo-core/shared/src/main/scala/kyo/aborts.scala
```diff
@@ -47,23 +41,25 @@ object Aborts:
             Tag[Aborts[V]],
             ClassTag[V]
         ): Either[V, T] < (S & VR) =
-            handle(handler, v).asInstanceOf[Either[V, T] < (S & VR)]
+            self.handle(handler)((), v).asInstanceOf[Either[V, T] < (S & VR)]
```
fwbrasil, line 44: I couldn't find a good way to avoid having to pass `()` when a handler doesn't have state.

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -12,26 +12,31 @@ object Envs:
...
-            handle(handler, value).asInstanceOf[T < (S & VR)]
-        end run
+            Envs[V].handle(handler[V])(env, value).asInstanceOf[T < (S & VR)]
+
     end extension
 
+    private def handler[V]: ResultHandler[V, Const[Unit], Envs[V], Id, Any] =
+        cachedHandler.asInstanceOf[ResultHandler[V, Const[Unit], Envs[V], Id, Any]]
+
+    private val cachedHandler =
```
fwbrasil, line 34: This is the one of the main benefits of this change: we can cache handlers and only pipe state through them.

fwbrasil:
I've opened the PR for visibility but I'll merge it to continue the optimization work on `Streams`. Please let me know if this kind of change/merge is becoming too intrusive to anyone!

PR #254, opened 2024-04-08
aborts: improve support for union types (fixes #244)

### LINKED ISSUE #244: Problem with tagged Aborts

fwbrasil:
Kyo currently only checks effect tags for equality, which isn't enough for effects like `Aborts`. See https://github.com/getkyo/kyo/issues/244 for more background.

This PR introduces a new method to `Handler` called `accepts`, which can be overriden with custom tag matching logic. I've added the override to `Aborts` to consider sub-typing for it and it seems to fix the issue.

I noticed another issue with the API not enabling handling of a super class and added a pending test for it. I think we'd need to add variance to `HasAborts` to fix it.

### DIFF
--- a/kyo-core/jvm/src/test/scala/kyoTest/coreBytecodeSizeTest.scala
+++ b/kyo-core/jvm/src/test/scala/kyoTest/coreBytecodeSizeTest.scala
@@ -38,7 +38,7 @@ class coreBytecodeSizeTest extends KyoTest:
         assert(map == Map(
             "test"        -> 28,
             "resultLoop"  -> 112,
-            "handleLoop"  -> 280,
+            "handleLoop"  -> 264,
             "_handleLoop" -> 12
         ))
     }
--- a/kyo-core/shared/src/main/scala/kyo/aborts.scala
+++ b/kyo-core/shared/src/main/scala/kyo/aborts.scala
@@ -13,8 +13,8 @@ object Aborts:
 
     extension [V](self: Aborts[V])
 
-        def fail[T](value: T)(using ev: T => V, t: Tag[Aborts[V]]): Nothing < Aborts[V] =
-            self.suspend[Nothing](Left(ev(value)))
+        def fail[T <: V](value: T)(using t: Tag[Aborts[T]]): Nothing < Aborts[T] =
+            self.suspend[Nothing](Left(value))
 
         def when(b: Boolean)(value: V)(using Tag[Aborts[V]]): Unit < Aborts[V] =
             if b then fail(value)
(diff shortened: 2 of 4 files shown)


kyo-core/jvm/src/test/scala/kyoTest/coreBytecodeSizeTest.scala
```diff
@@ -38,7 +38,7 @@ class coreBytecodeSizeTest extends KyoTest:
         assert(map == Map(
             "test"        -> 28,
             "resultLoop"  -> 112,
-            "handleLoop"  -> 280,
+            "handleLoop"  -> 264,
```
fwbrasil, line 41: There are fewer bytecodes to invoke `handler.accepts` than to perform the equality check.

kyo-core/shared/src/main/scala/kyo/aborts.scala
```diff
@@ -13,8 +13,8 @@ object Aborts:
 
     extension [V](self: Aborts[V])
 
-        def fail[T](value: T)(using ev: T => V, t: Tag[Aborts[V]]): Nothing < Aborts[V] =
-            self.suspend[Nothing](Left(ev(value)))
+        def fail[T <: V](value: T)(using t: Tag[Aborts[T]]): Nothing < Aborts[T] =
```
fwbrasil, line 16: I've changed this method to suspend the specific type of the parameter. For example `Aborts[Throwable].fail(new Exception` will now return a `T < Aborts[Exception]`.

hearnadam (approved):

Issue #148, opened 2023-12-11
Logs review


fwbrasil:
I recently changed the [logging support](https://github.com/getkyo/kyo/blob/main/kyo-core/shared/src/main/scala/kyo/logs.scala) to avoid having to create logger instances. Instead, the sourcecode library is used to identify the source of the log via the file name and line. This simplification should be helpful for newcomers but there are still a few topics to explore:

1. I'd be nice to have a way to switch the underlying logger like `Consoles` does so users can dynamically change log levels and where the log is emitted to. It's something useful in environments where it's not easy to define or edit the logging xml configs like scastie.
2. The sourcecode library doesn't provide an implicit for the combination of the file name and line number so two objects need to be instantiated, which also means the `IOs` thunk uses two pointers.
3. Even though the values inferred by sourcecode are statically known and become strings in the class constant pool, the library uses [wrapper objects](https://github.com/com-lihaoyi/sourcecode/blob/main/sourcecode/src/sourcecode/SourceContext.scala#L4C16-L4C27) that require an allocation. The `Logs` code is optimized to avoid capturing the object so the allocation should be elided by the JIT in most cases but it'd be nice to simplify the code and have stronger guarantees without depending on JIT behavior, which can be surprising at times.
4. The log source file and line number are currently part of the log message and a single logger `kyo.logs` is reused. Ideally, the logger name should contain the source info but we'd need to check if there's a way to do that without having to allocate `Logger` instances on each produced log line.
5. Could we cache the `logger.is*Enabled` flags on class loading? That'd allow us to avoid allocating the `IOs` thunk when the logging is disabled.

hearnadam:
#259 addressed the remaining concerns here, with the exception of 4. There are some mechanisms we could use to cache per-file loggers, but I think the overhead does not fit with Kyo's direction in general.

PR #255, opened 2024-04-09
Logs API improvements

hearnadam:
This change is the first to target some of the issues with `Logs` described in https://github.com/getkyo/kyo/issues/148

First and most importantly, we need a mechanism to allow users to define a logging interface which is not necessarily coupled to slf4j. I created an 'Unsafe' pure API which can enable users to provide their own logging interface. We can also use this to prevent the need for the slf4j interfaces in Kyo JS.

Unresolved issues:

1. Should we expose the default `unsafe` instance?
2. Should we use local or implicit for logger instance injection? For Kyo internals, it seems to make sense that we should use an implicit...
3. Should we expose `*Enabled` functions in the unsafe API?


fwbrasil:
Geazi mentioned https://github.com/outr/scribe in Discord. Maybe we could integrate with it instead?

> Should we expose the default unsafe instance?

Yes, Kyo normally offers unsafe APIs since they might be required for performance-sensitive code.

> Should we use local or implicit for logger instance injection? For Kyo internals, it seems to make sense that we should use an implicit...

Maybe we could use a `local` users can override like in `Clocks` and `Consoles`?

> Should we expose *Enabled functions in the unsafe API?

I think so, they might be needed for performance-sensitive code.

hearnadam:
> Geazi mentioned https://github.com/outr/scribe in Discord. Maybe we could integrate with it instead?

I'm not especially concerned with what logging interface we choose. I expect many companies are still using slf4j/logback. Most importantly, we should expose an API in `KyoApp` trait to provide an unsafe logging interface. This should enable users to provide any logger they want.

fwbrasil:
LGTM! Thanks :)

hearnadam:
@fwbrasil can you provide guidance where we should use `PlatformSpecific` mixins? I think it's reasonable to assume `val unsafe` can be defined per platform.

Do you have an concern with internal components using the default `unsafe` instance instead of the user provided one?

Perhaps we need a mechanism like ZIO's `bootstrap` for default instance values.

PR #258, opened 2024-04-12
Loops: Stack-Safe Recursion

fwbrasil:
Kyo's execution model evaluates pure code strictly as [described in the readme](https://github.com/getkyo/kyo/blob/457d9f5958afb6af8b847245164d85de3362e439/README.md?plain=1#L1741), which makes recursive computations without effect suspensions not stack safe by default. The current recommendation is introducing an effect suspension like `IOs` to achieve stack safety.

Although introducing an effect suspension can be convenient in some cases, it can become problematic in code that doesn't have a reasonable way to introduce suspensions. For example, I've been working on the `Streams` effect and adding `IOs` to its execution just to ensure stack safety doesn't seem reasonable. I could emit an empty chunk but that would introduce unnecessary overhead and doesn't feel like a proper solution.

This PR introduces `Loops`, which provides generic stack-safe recursion using a technique identified as part of https://github.com/getkyo/kyo/pull/253 that is currently [used in Kyo's core](https://github.com/getkyo/kyo/blob/457d9f5958afb6af8b847245164d85de3362e439/kyo-core/shared/src/main/scala/kyo/core.scala#L72). The technique reflects on the result of the computation to determine if it can continue the loop immediately (using @tailrec) or if the loop should suspend (using `map`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -655,6 +655,79 @@ val d: Seq[Int] < Any =
 
 The `Choices` effect becomes exceptionally powerful when combined with other effects. This allows you not just to make decisions or explore options in isolation but also to do so in contexts that may involve factors such as asynchronicity, resource management, or even user interaction.
 
+
+### Loops: Stack-Safe Recursion
+
+`Loops` provides a solution for general stack-safe recursion in Kyo. It offers a set of methods to transform input values through repeated applications of a function until a termination condition is met, allowing for safe and efficient recursive computations without the need for explicit effect suspensions.
+
+```scala
+import kyo._
+
+// Iteratively increment an 'Int' value
+// until it reaches 5
+val a: Int < Any = 
+  Loops.transform(1)(i => 
+    if i < 5 then Loops.continue(i + 1) 
+    else Loops.done(i)
+  )
+
+// Transform with multiple input values
+val b: Int < Any =
+  Loops.transform2(1, 1)((i, j) => 
+    if i + j < 5 then Loops.continue(i + 1, j + 1) 
+    else Loops.done(i + j)
+  )
+
+// Transform with three input values
+val c: Int < Any = 
+  Loops.transform3(1, 1, 1)((i, j, k) => 
+    if i + j + k < 5 then Loops.continue(i + 1, j + 1, k + 1)
+    else Loops.done(i + j + k)
+  )
+
+// Mixing 'IOs' with 'Loops'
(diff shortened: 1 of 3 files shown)


kyo-core/shared/src/main/scala/kyo/loops.scala
```diff
@@ -0,0 +1,104 @@
...
+                            _loop(next.input)
+                        case res =>
+                            res.asInstanceOf[Output]
+                    }
+                case res =>
+                    res.asInstanceOf[Output]
+        loop(input)
+    end transform
+
+    inline def transform2[Input1, Input2, Output: Flat, S](
```
hearnadam, line 56: Personally, I dislike `functionN` naming - is there a reason we can't overload the `transform` name?

kyo-core/shared/src/test/scala/kyoTest/loopsTest.scala
```diff
@@ -0,0 +1,293 @@
+package kyoTest
+
+import kyo.*
+
+class loopsTest extends KyoTest:
```
hearnadam, line 5: Perhaps it's worth adding a test with a `List` accumulator - assuming we want users referencing tests. It seems more inline with how I might use this in practical application.

README.md
```diff
@@ -655,6 +655,79 @@ val d: Seq[Int] < Any =
...
+  Loops.transform3(1, 1, 1)((i, j, k) => 
+    if i + j + k < 5 then Loops.continue(i + 1, j + 1, k + 1)
+    else Loops.done(i + j + k)
+  )
+
+// Mixing 'IOs' with 'Loops'
+val d: Int < IOs = 
+  Loops.transform(1)(i => 
+    if i < 5 then 
+      IOs(println(s"Iteration: $i")).map(_ => Loops.continue(i + 1))
```
hearnadam, line 692: Can we use `Consoles` here (or in a test)? It may be valuable to highlight that effects other than `IOs` can be combined with Loops. This may be non-obvious when not referencing the code...

Though to be honest, most newcomers may not realize Consoles is a different effect type.

hearnadam (approved):

fwbrasil:
Thanks for the review! I've pushed changes addressing the points you raised. I've also added `indexed` methods since it's a common case and it can avoid boxing to `Integer`.

hearnadam (approved):

PR #259, opened 2024-04-13
Logs: use custom implicit macro for filename:line

hearnadam:
In this PR I create a new custom macro to find the filename + line a function is called on. It uses opaque type, avoiding allocations or the need to elide.

This resolves a majority of the remaining concerns in https://github.com/getkyo/kyo/issues/148

- Use locals for `Unsafe` instance injection (last PR) ✅
- Combine filename and line into single type - reducing the `IOs` thunk to one pointer ✅ 
- Use opaque types to avoid allocations ✅ 
- Push logged filename + line into the logger itself ❌ 
   - not sure this is easily possible without allocating new loggers...
- avoid allocating `IOs` thunk using `IOs.unit` ☑️
   - We could require `unsafe` to have (lazy?) `val`s for *Enabled booleans, but I fear this may increase class size...

---

- Combining filename and line number should reduce IOs to only capture one pointer
- Remove dependency on com.lihaoyi.sourcecode
  - note: this also means kyo now should work better in worksheets/consoles https://github.com/com-lihaoyi/sourcecode/issues/125


hearnadam:
This was the result of running Kyo `Logs` in console previously:

```
scala> IOs.run(Logs.error("test"))
-- Error: ----------------------------------------------------------------------
 1 |IOs.run(Logs.error("test"))
   |                          ^
   |Exception occurred while executing macro expansion.
   |java.lang.NullPointerException: Cannot invoke "java.nio.file.Path.getFileName()" because the return value of "scala.quoted.Quotes$reflectModule$SourceFileMethods.jpath(Object)" is null
   |    at sourcecode.Macros$.fileNameImpl(Macros.scala:155)
   |
   |----------------------------------------------------------------------------
   |Inline stack trace
   |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
   |This location contains code that was inlined from Macros.scala:33
    ----------------------------------------------------------------------------
1 error found
```

After this PR, `Logs` works in console as expected.

kyo-core/shared/src/main/scala/kyo/internal/FileImplicits.scala
```diff
@@ -0,0 +1,21 @@
+package kyo.internal
+
+import scala.language.implicitConversions
+import scala.quoted.*
+
+opaque type FileNameWithLine <: String = String
```
fwbrasil, line 6: What do you think about `Position` instead of `FileNameWithLine`?

kyo-core/shared/src/main/scala/kyo/internal/FileImplicits.scala
```diff
@@ -0,0 +1,21 @@
...
+object FileNameWithLine:
+    def apply(value: String): FileNameWithLine   = value
+    implicit inline def derive: FileNameWithLine = ${ Macros.fileNameWithLine }
```
fwbrasil, line 9: The idiomatic way in Scala 3 would be `inline given derive: FileNameWithLine`

kyo-core/shared/src/main/scala/kyo/logs.scala
```diff
@@ -86,110 +66,45 @@ object Logs:
             inline def errorEnabled: Boolean = logger.isErrorEnabled
 
             inline def trace(msg: => String)(
```
fwbrasil, line 68: Can you make `msg` and `file` also `inline` in these methods? With this change, the message won't even be evaluated/allocated if the logging level isn't enabled. Inlining `file` doesn't avoid allocations or evaluation but it saves some byte codes.

fwbrasil:
Nice! This should be a noticeable performance improvement. Could you also verify the performance with the logging benchmark and share the results? https://github.com/getkyo/kyo/blob/main/kyo-bench/src/main/scala/kyo/bench/LoggingBench.scala

kyo-core/shared/src/main/scala/kyo/internal/FileImplicits.scala
hearnadam: Good to know, thanks

kyo-core/shared/src/main/scala/kyo/internal/FileImplicits.scala
hearnadam: My only slight fear is that we are colliding with Scala's `quotes.reflect.Position`... perhaps `FilePosition`?

I do think `internal.Position` is much cleaner and concise in function definitions. We can proceed with it for now.

kyo-core/shared/src/main/scala/kyo/logs.scala
hearnadam: From my initially testing, this is not possible. I can try experimenting with an abstract class, though not sure if that will change anything.

```
Coordinator.scala:68:16 
[error] 68 |                Logs.unsafe.error("Kyo coordinator failure", ex)
[error]    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]    |          Deferred inline method error in trait Unsafe cannot be invoked
```

Inlining method parameters requires the full function to be inlined, and that results in these errors wherever `Unsafe` is used.

kyo-core/shared/src/main/scala/kyo/logs.scala
```diff
@@ -17,43 +17,43 @@ object Logs:
         def errorEnabled: Boolean
 
         def trace(msg: => String)(
-            using file: internal.FileNameWithLine
+            using file: internal.Position
```
hearnadam, line 20: I intentionally left this named `file` - I think it reads well as 'file position'

hearnadam:
Note: slf4j is using noop logger here - not sure if that's also true for cats/zio, but we should probably try to make the tests equivalent.

```[info] Benchmark                                  Mode  Cnt        Score      Error   Units
[info] LoggingBench.forkCats                     thrpt    5     3005.973 ±  160.386   ops/s
[info] LoggingBench.forkCats:gc.alloc.rate       thrpt    5     5510.125 ±  293.883  MB/sec
[info] LoggingBench.forkCats:gc.alloc.rate.norm  thrpt    5  1922325.558 ±    1.961    B/op
[info] LoggingBench.forkCats:gc.count            thrpt    5      106.000             counts
[info] LoggingBench.forkCats:gc.time             thrpt    5       67.000                 ms
[info] LoggingBench.forkKyo                      thrpt    5    13577.762 ±  742.306   ops/s
[info] LoggingBench.forkKyo:gc.alloc.rate        thrpt    5     7254.873 ±  396.099  MB/sec
[info] LoggingBench.forkKyo:gc.alloc.rate.norm   thrpt    5   560369.277 ±    0.083    B/op
[info] LoggingBench.forkKyo:gc.count             thrpt    5      114.000             counts
[info] LoggingBench.forkKyo:gc.time              thrpt    5       67.000                 ms
[info] LoggingBench.forkZio                      thrpt    5     1413.801 ±   88.186   ops/s
[info] LoggingBench.forkZio:gc.alloc.rate        thrpt    5     5095.815 ±  318.348  MB/sec
[info] LoggingBench.forkZio:gc.alloc.rate.norm   thrpt    5  3780138.685 ±  556.738    B/op
[info] LoggingBench.forkZio:gc.count             thrpt    5       94.000             counts
[info] LoggingBench.forkZio:gc.time              thrpt    5       70.000                 ms
[info] LoggingBench.syncCats                     thrpt    5     3545.430 ±  556.141   ops/s
[info] LoggingBench.syncCats:gc.alloc.rate       thrpt    5     6497.747 ± 1020.868  MB/sec
[info] LoggingBench.syncCats:gc.alloc.rate.norm  thrpt    5  1922050.211 ±  129.579    B/op
[info] LoggingBench.syncCats:gc.count            thrpt    5       62.000             counts
[info] LoggingBench.syncCats:gc.time             thrpt    5       47.000                 ms
[info] LoggingBench.syncKyo                      thrpt    5    12753.201 ± 7234.914   ops/s
[info] LoggingBench.syncKyo:gc.alloc.rate        thrpt    5     6810.520 ± 3863.021  MB/sec
[info] LoggingBench.syncKyo:gc.alloc.rate.norm   thrpt    5   560056.552 ±    0.383    B/op
[info] LoggingBench.syncKyo:gc.count             thrpt    5      105.000             counts
[info] LoggingBench.syncKyo:gc.time              thrpt    5       65.000                 ms
[info] LoggingBench.syncZio                      thrpt    5     1482.786 ±   23.023   ops/s
[info] LoggingBench.syncZio:gc.alloc.rate        thrpt    5     5343.574 ±   81.647  MB/sec
[info] LoggingBench.syncZio:gc.alloc.rate.norm   thrpt    5  3779502.492 ±  579.068    B/op
[info] LoggingBench.syncZio:gc.count             thrpt    5      101.000             counts
[info] LoggingBench.syncZio:gc.time              thrpt    5       62.000                 ms
```

kyo-core/shared/src/main/scala/kyo/logs.scala
hearnadam: Even if we switch to a class implementation of SL4J (extending a trait), the inlining will break usage of the local - as the implementation of the trait is not known:

```
[error]  12 |            _ <- Logs.trace("trace")
[error]     |                 ^^^^^^^^^^^^^^^^^^^
[error]     |        Deferred inline method trace in trait Unsafe cannot be invoked
[error]     |---------------------------------------------------------------------------
[error]     |Inline stack trace
[error]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]     |This location contains code that was inlined from logs.scala:123
[error] 123 |        logWhen(_.traceEnabled)(_.trace(msg))
[error]     |                                ^^^^^^^
[error]      ---------------------------------------------------------------------------

```

fwbrasil: I see. Would it work if `Unsafe`'s methods aren't inlined? Inlining the main API to avoid the evaluation of the message in case the logging is disabled seems more valuable.

fwbrasil:
Yes, the benchmarks run without logging disabled since the goal is to measure the overhead of the effect systems. If logging were enabled, the IO overhead would likely make the benchmark results not very meaningful.

hearnadam:
Baseline: `9d562f45cd6460155cc685130455b4cf6a913fee`
```
[info] Benchmark                                  Mode  Cnt        Score      Error   Units
[info] LoggingBench.forkCats                     thrpt    5     1230.646 ± 3790.368   ops/s
[info] LoggingBench.forkCats:gc.alloc.rate       thrpt    5     2249.841 ± 6941.172  MB/sec
[info] LoggingBench.forkCats:gc.alloc.rate.norm  thrpt    5  1922462.257 ±  357.650    B/op
[info] LoggingBench.forkCats:gc.count            thrpt    5       32.000             counts
[info] LoggingBench.forkCats:gc.time             thrpt    5      126.000                 ms
[info] LoggingBench.forkKyo                      thrpt    5    11164.766 ±  628.747   ops/s
[info] LoggingBench.forkKyo:gc.alloc.rate        thrpt    5     9372.402 ±  527.592  MB/sec
[info] LoggingBench.forkKyo:gc.alloc.rate.norm   thrpt    5   880401.478 ±    0.132    B/op
[info] LoggingBench.forkKyo:gc.count             thrpt    5       86.000             counts
[info] LoggingBench.forkKyo:gc.time              thrpt    5       62.000                 ms
[info] LoggingBench.forkZio                      thrpt    5     1341.414 ±  282.118   ops/s
[info] LoggingBench.forkZio:gc.alloc.rate        thrpt    5     4834.787 ± 1016.071  MB/sec
[info] LoggingBench.forkZio:gc.alloc.rate.norm   thrpt    5  3780127.987 ±  538.274    B/op
[info] LoggingBench.forkZio:gc.count             thrpt    5       77.000             counts
[info] LoggingBench.forkZio:gc.time              thrpt    5       73.000                 ms
[info] LoggingBench.syncCats                     thrpt    5     3703.498 ±   40.331   ops/s
[info] LoggingBench.syncCats:gc.alloc.rate       thrpt    5     6787.468 ±   74.101  MB/sec
[info] LoggingBench.syncCats:gc.alloc.rate.norm  thrpt    5  1922025.856 ±    0.103    B/op
[info] LoggingBench.syncCats:gc.count            thrpt    5      101.000             counts
[info] LoggingBench.syncCats:gc.time             thrpt    5       60.000                 ms
[info] LoggingBench.syncKyo                      thrpt    5    10987.902 ±  716.313   ops/s
[info] LoggingBench.syncKyo:gc.alloc.rate        thrpt    5     9220.937 ±  602.575  MB/sec
[info] LoggingBench.syncKyo:gc.alloc.rate.norm   thrpt    5   880088.629 ±    0.042    B/op
[info] LoggingBench.syncKyo:gc.count             thrpt    5       99.000             counts
[info] LoggingBench.syncKyo:gc.time              thrpt    5       64.000                 ms
[info] LoggingBench.syncZio                      thrpt    5     1487.958 ±  132.302   ops/s
[info] LoggingBench.syncZio:gc.alloc.rate        thrpt    5     5362.333 ±  476.576  MB/sec
[info] LoggingBench.syncZio:gc.alloc.rate.norm   thrpt    5  3779458.715 ±  473.367    B/op
[info] LoggingBench.syncZio:gc.count             thrpt    5      101.000             counts
[info] LoggingBench.syncZio:gc.time              thrpt    5       58.000                 ms
```

Appears to be an improvement

kyo-core/shared/src/main/scala/kyo/logs.scala
hearnadam: Currently the methods in `Logs` object are inlined, along with the methods in `Unsafe` on the SL4J instance. We just cannot inline the `Unsafe` function parameters as they are not in the trait.

fwbrasil:
Allocation rate dropped by almost a half!! 👏👏

PR #267, opened 2024-04-15
introduce chunks

fwbrasil:
Kyo provides stronger guarantees regarding the purity of its effects than other effect systems typically do. A common pattern in libraries like ZIO is to [leverage mutable iterators internally](https://github.com/zio/zio/blob/62173a1aad4be0a369a16cb234f43db5c896337b/core/shared/src/main/scala/zio/ZIO.scala#L3333-L3341) to process sequences:

```scala
  def foreach[R, E, A, B, Collection[+Element] <: Iterable[Element]](in: Collection[A])(
    f: A => ZIO[R, E, B]
  )(implicit bf: BuildFrom[Collection[A], B, Collection[B]], trace: Trace): ZIO[R, E, Collection[B]] =
    ZIO.suspendSucceed {
      val iterator = in.iterator
      val builder  = bf.newBuilder(in)

      ZIO.whileLoop(iterator.hasNext)(f(iterator.next()))(builder += _).as(builder.result())
    }
```

This form of local mutability doesn't seem ideal but it's generally ok to do in a monad that has `IO` as its base effect. That's not the case in Kyo. Pure effects that don't include `IOs` suspensions can't leverage mutable state across transformations since they're meant to support multi-shot continuations and the provided continuations can be executed multiple times by a handler, which would produce invalid state if an iterator is used for example. This is currently an issue with `Seqs` in Kyo as well. I'll follow up later to fix it.

This PR introduces `Chunks` as a mechanism to provide efficient processing of data in a purely functional manner without leveraging mutable state across transformations. The data structure is designed to provide `O(1)` operations for the common needs of recursive computations. Please see the readme for more information and the use in [streams I've working on](https://github.com/getkyo/kyo/blob/efe4f4ce371c808179ed13ce70b4baf639b6752e/kyo-core/shared/src/main/scala/kyo/streams.scala).

I've also updated `Sums` to remove `Summer` and accumulate items in a `Chunk` instead. I'm noticing that's the pattern for the effect in other libraries like ZPure, it's a good simplification of the effect, and it provides better performance than working with collections like `List`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -867,12 +867,14 @@ val m: Int < Any = a.foldLeft(0)(_ + _)
 // Copying to arrays
 val n: Array[Int] = a.toArray
 
-// Other operations
-val o: Chunk[Int] = Chunks.init(a, b)
-  .flatten
-val p: Chunk[Int] = Chunks.init(1, 1, 2, 3, 3)
-  .changes
-val q: Chunks.Indexed[Int] = a.toIndexed
+// Flatten a nested chunk
+val o: Chunk[Int] = 
+  Chunks.init(a, b).flatten
+
+// Obtain sequentially distict elements.
+// Outputs: Chunk(1, 2, 3, 1)
+val p: Chunk[Int] = 
+  Chunks.init(1, 1, 2, 3, 3, 1, 1).changes
 ```
 
 ### Aspects: Aspect-Oriented Programming
--- a/README.md
+++ b/README.md
@@ -859,7 +859,8 @@ val j: Chunk[String] < Any = a.map(_.toString)
 val k: Chunk[Int] < Any = a.filter(_ % 2 == 0)
 
 // Effectful side effects
-val l: Unit < IOs = a.foreach(v => IOs(println(v)))
+val l: Unit < Consoles = 
+  a.foreach(v => Consoles.println(v))
 
 // Effectful fold
 val m: Int < Any = a.foldLeft(0)(_ + _)
--- a/README.md
+++ b/README.md
@@ -762,6 +762,119 @@ val c: Int < Any =
   )
 ```
 
+## Chunks: Efficient Sequences
+
+`Chunks` is an efficient mechanism for processing sequences of data in a purely functional manner. It offers a wide range of operations optimized for different scenarios, ensuring high performance without compromising functional programming principles.
+
+`Chunk` is designed as a lightweight wrapper around arrays, allowing for efficient random access and transformation operations. Its internal representation is carefully crafted to minimize memory allocation and ensure stack safety. Many of its operations have an algorithmic complexity of `O(1)`, making them highly performant for a variety of use cases.
+
+```scala
+import kyo._
+
+// Construct chunks
+val a: Chunk[Int] = Chunks.init(1, 2, 3)
+val b: Chunk[Int] = Chunks.initSeq(Seq(4, 5, 6))
+
+// Perform O(1) operations
+val c: Chunk[Int] = a.append(4)
+val d: Chunk[Int] = b.take(2)
+val e: Chunk[Int] = c.dropLeft(1)
+
+// Perform O(n) operations
+val f: Chunk[String] = d.map(_.toString).pure
+val g: Chunk[Int] = e.filter(_ % 2 == 0).pure
+```
+
+`Chunks` provides two main subtypes: `Chunk` for regular chunks and `Chunks.Indexed` for indexed chunks. The table below summarizes the time complexity of various operations for each type:
+
+| Description                | Operations                                           | Regular Chunk | Indexed Chunk |
+|----------------------------|------------------------------------------------------|---------------|---------------|
+| Creation                   | `Chunks.init`, `Chunks.initSeq`                      | O(n)          | O(n)          |
+| Size and emptiness         | `size`, `isEmpty`                                    | O(1)          | O(1)          |
+| Take and drop              | `take`, `dropLeft`, `dropRight`, `slice`             | O(1)          | O(1)          |
+| Append and last            | `append`, `last`                                     | O(1)          | O(1)          |
+| Element access             | `apply`, `head`, `tail`                              | N/A           | O(1)          |
+| Concatenation              | `concat`                                             | O(n)          | O(n)          |
+| Effectful map and filter   | `map`, `filter`, `collect`, `takeWhile`, `dropWhile` | O(n)          | O(n)          |
+| Effectful side effects     | `foreach`, `collectUnit`                             | O(n)          | O(n)          |
+| Effectful fold             | `foldLeft`                                           | O(n)          | O(n)          |
+| Copying to arrays          | `toArray`, `copyTo`                                  | O(n)          | O(n)          |
+| Other operations           | `flatten`, `changes`, `toSeq`, `toIndexed`           | O(n)          | O(n)          |
+
+When deciding between `Chunk` and `Chunks.Indexed`, consider the primary operations you'll be performing on the data. If you mainly need to `append` elements, `take` slices, or `drop` elements from the beginning or end of the sequence, `Chunk` is a good choice. Its `O(1)` complexity for these operations makes it efficient for such tasks.
+
+```scala
+import kyo._
+
+val a: Chunk[Int] = Chunks.init(1, 2, 3, 4, 5)
+
+// Efficient O(1) operations with Chunk
+val b: Chunk[Int] = a.append(6)
+val c: Chunk[Int] = a.take(3)
+val d: Chunk[Int] = a.dropLeft(2)
+```
+
+On the other hand, if you frequently need to access elements by index, `Chunks.Indexed` is the better option. It provides `O(1)` element access and supports `head` and `tail` operations, which are not available in `Chunk`.
+
+```scala
+import kyo._
+
+val a: Chunks.Indexed[Int] = 
+  Chunks.init(1, 2, 3, 4, 5).toIndexed
+
+// Efficient O(1) operations with Chunks.Indexed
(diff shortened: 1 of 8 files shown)


kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -0,0 +1,464 @@
...
+                            Loops.continue(acc.append(u))
+                        }
+                    else
+                        Loops.continue(acc)
+                    end if
+                else
+                    Loops.done(acc)
+            }
+
+    final def collectUnit[U, S](pf: PartialFunction[T, Unit < S]): Unit < S =
```
hearnadam, line 186: this should be a very useful operator. Consider also naming it `collectDiscard`

kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -0,0 +1,464 @@
...
+            val indexed = this.toIndexed
+            Loops.indexed(()) { (idx, _) =>
+                if idx < size then
+                    val v = indexed(idx)
+                    if pf.isDefinedAt(v) then
+                        pf(v).andThen {
+                            Loops.continue(())
+                        }
+                    else
+                        Loops.continue(())
```
hearnadam, line 199: I think we should add a `Loops.continueUnit` member to avoid allocations. Possibly `Loops.doneUnit` as well.

kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -0,0 +1,464 @@
...
+                        loop(chunk, left, right - 1)
+                    case _ =>
+                        Drop(c, left, right, size)
+            end if
+        end loop
+        loop(this, left, right)
+    end dropLeftAndRight
+
+    final def append(v: T): Chunk[T] =
+        Append(this, v, size + 1)
```
hearnadam, line 49: aside from adding complexity to run loops, is there a reason we cannot support `Append` with `Chunk[T]` member? It seems like a linked list of `Arrays` should be fairly efficient...

I understand there is some benefit in forcing amortization of re-allocating with the `concat` operator, but perhaps there is another way to do so?

kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -0,0 +1,464 @@
...
+        initSeq(values)
+
+    def initSeq[T](values: Seq[T]): Chunk[T] =
+        if values.isEmpty then init[T]
+        else
+            values match
+                case seq: IndexedSeq[T] => FromSeq(seq)
+                case _                  => Compact(values.toArray(using ClassTag(classOf[Any])))
+
+    def fill[T, S](n: Int)(v: T): Chunk[T] =
```
hearnadam, line 393: Where is `S` used?

kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -0,0 +1,464 @@
...
+        else
+            val array = (new Array[Any](n)).asInstanceOf[Array[T]]
+            @tailrec def loop(idx: Int = 0): Unit =
+                if idx < n then
+                    array(idx) = v
+                    loop(idx + 1)
+            loop()
+            Compact(array)
+    end fill
+
```
hearnadam, line 404: Perhaps worth adding a `tabulate` method? ```def tabulate[T](n: Int)(f: Int => T): Chunk[T] =```

kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -0,0 +1,464 @@
...
+                        Tail(c, 1, size - 1)
+    end Indexed
+
+    def init[T]: Chunk[T] =
+        cachedEmpty.asInstanceOf[Chunk[T]]
+
+    def init[T](values: T*): Chunk[T] =
+        initSeq(values)
+
+    def initSeq[T](values: Seq[T]): Chunk[T] =
```
hearnadam, line 386: Will there be a collision if we overload with `init`?

kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -0,0 +1,464 @@
...
+            if size <= 1 then cachedEmpty.asInstanceOf[Indexed[T]]
+            else
+                this match
+                    case Tail(chunk, offset, size) =>
+                        Tail(chunk, offset + 1, size - 1)
+                    case c =>
+                        Tail(c, 1, size - 1)
+    end Indexed
+
+    def init[T]: Chunk[T] =
```
hearnadam, line 380: Let's also alias to `def empty` for consistency with Scala's collections

README.md
```diff
@@ -762,6 +762,119 @@ val c: Int < Any =
...
+val m: Int < Any = a.foldLeft(0)(_ + _)
+
+// Copying to arrays
+val n: Array[Int] = a.toArray
+
+// Other operations
+val o: Chunk[Int] = Chunks.init(a, b)
+  .flatten
+val p: Chunk[Int] = Chunks.init(1, 1, 2, 3, 3)
+  .changes
```
hearnadam, line 874: I think it would be helpful to comment the resulting value here (or any other non-trivial operator). That or we should find a mechanism to have runnable docs...

kyo-core/shared/src/main/scala/kyo/chunks.scala
fwbrasil: good idea! I'll add it

kyo-core/shared/src/main/scala/kyo/chunks.scala
fwbrasil: It is possible but if we add a `Concat` sub-class, all methods that do traversal need to use allocations to walk the structure since they'd have to handle a tail that is also a collection. With the current structure, all traversals are allocation-free even when the chunk isn't indexed. 

We could still attempt to include a `Concat` sub-class and use temporary buffers via thread locals to avoid allocations but there'd be some overhead as well. I think `concat` flattening the chunk to a contiguous array is a good property to avoid the data structure accumulating too many nested operations but I imagine we might revisit this decision if we see good evidence that `Concat` would be beneficial.

kyo-core/shared/src/main/scala/kyo/chunks.scala
fwbrasil: I'm removing it

kyo-core/shared/src/main/scala/kyo/chunks.scala
fwbrasil: I think there are many methods we'll need to add yet but I'd prefer to keep the scope of this PR to the current methods. I need this change to open the PR with the new chunked streams.

kyo-core/shared/src/main/scala/kyo/chunks.scala
fwbrasil: I think so. It becomes ambiguous in case you want to make a chunk of `Seq`s:
```scala
Chunk.init(Seq(1)) // should it be a Chunk[Int] or a Chunk[Seq[Int]]?
```

kyo-core/shared/src/main/scala/kyo/chunks.scala
fwbrasil: I've added a `Chunk.empty`.

kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -0,0 +1,467 @@
...
+            Compact(array)
+        end if
+    end concat
+
+    final def map[U: Flat, S](f: T => U < S): Chunk[U] < S =
+        if isEmpty then Chunks.init
+        else
+            val size    = this.size
+            val indexed = this.toIndexed
+            Loops.indexed(Chunks.init[U]) { (idx, acc) =>
```
hearnadam, line 89: Shouldn't it be cheaper to pre-allocate a backing Array instead of growing the `init`? We can still use the `Loops` to fill it...

kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -0,0 +1,467 @@
...
+                    Loops.done(acc)
+            }
+    end map
+
+    final def filter[S](f: T => Boolean < S): Chunk[T] < S =
+        if isEmpty then Chunks.init
+        else
+            val size    = this.size
+            val indexed = this.toIndexed
+            Loops.indexed(Chunks.init[T]) { (idx, acc) =>
```
hearnadam, line 102: Is it worth building a private (inlined) helper method to abstract this pattern?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -0,0 +1,467 @@
...
+    //////////////////
+    // O(1) methods //
+    //////////////////
+
+    def size: Int
+
+    final def isEmpty: Boolean = size == 0
+
+    final def take(n: Int): Chunk[T] =
+        dropLeftAndRight(0, size - Math.min(Math.max(0, n), size))
```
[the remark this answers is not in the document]
fwbrasil: hmm I didn't that but I took a look and `Math` here is the Java one. Maybe it's different in Scala 3.

kyo-core/shared/src/main/scala/kyo/chunks.scala
fwbrasil: It's not ideal because it'd introduce mutable state in continuations, similarly to ZIO's approach. If an effect handler calls a continuation multiple times, the continuation could have a different behavior on each call.

PR #268, opened 2024-04-15
attempt to fix missed interrupt signals (#260)

### LINKED ISSUE #260: Flaky (slow) FIbers test
'multiple fibers' test is reaching 5 second timeout (intermittently). Is this a bug in the test? Or are we facing non-determinism with fibers?

```
[info] - multiple fibers *** FAILED *** (5 seconds, 6 milliseconds)
[info]   kyo.fibers$package$Fibers$Interrupted$:
[info]   ...
```

fwbrasil:
I'm not sure if this will fix #260 since I'm not able to reproduce it on my machine (M2). I noticed that the interrupt signal can get lost in the `IOTask` run loop when `state` is updated. I've updated the code to maintain the interrupt signal via a `VarHandle` and `compareAndSwap` to avoid an `AtomicInteger` allocation.

This code can use some refactoring but I think it's best to attempt to fix the issue before improving it. @hearnadam could you check if it fixes the issue for you?

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/ios.scala
+++ b/kyo-core/shared/src/main/scala/kyo/ios.scala
@@ -139,9 +139,9 @@ sealed trait IOs extends Effect[IOs]:
                             ensureLoop(v2, np)
                         end apply
                 case _ =>
-                    p.remove(ensure)
                     IOs {
                         ensure()
+                        p.remove(ensure)
                         v
                     }
         ensureLoop(v, Preempt.never)
--- a/kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
+++ b/kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
@@ -1,5 +1,6 @@
 package kyo.scheduler
 
+import java.lang.invoke.MethodHandles
 import kyo.*
 import kyo.Locals.State
 import kyo.core.*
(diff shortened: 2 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -139,9 +139,9 @@ sealed trait IOs extends Effect[IOs]:
                             ensureLoop(v2, np)
                         end apply
                 case _ =>
-                    p.remove(ensure)
                     IOs {
                         ensure()
+                        p.remove(ensure)
```
fwbrasil, line 144: It seems the ordering here could make the `ensure` get lost. The ensure would be removed but then the `IOs` instance might never get executed if the execution is interrupted.

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -82,22 +87,29 @@ private[kyo] class IOTask[T](
...
         if !isNull(curr) then
             Task.Preempted
         else
+            state = -1
             Task.Done
         end if
     end run
 
     def ensure(f: () => Unit): Unit =
-        if !isNull(curr) then
```
fwbrasil, line 99: not needed anymore after the ordering change in `IOs.ensure`

hearnadam:
Still seeing the same issue with your commit: `3dfd09ee7aebb0d76768e3528a545dc0efb92bac`

PR #270, opened 2024-04-16
streams chunking

fwbrasil:
Introduces stream chunking based on https://github.com/getkyo/kyo/pull/267. The implementation isn't very sophisticated yet since, for example, it doesn't enforce chunk sizes and any effect in a transformation can hold the processing of the entire chunk. The feature set is enough for my needs in the LLM module for now, though.

I'm actually not sure how to handle effects vs chunking. It seems ZIO's approach breaking chunks can become problematic so we might need to figure out a new and more ergonomic solution.

The performance is closer to the alternatives in the non-buffered benchmark but it's better in the buffered version since Kyo's channels and fibers are much more lightweight:

image
https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/83f4cd0ccd4d3779561dc774bec2f294/raw/485789ff92724d438362ebf5ebd39dd8f9aa0449/jmh-result.json

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/streams.scala
+++ b/kyo-core/shared/src/main/scala/kyo/streams.scala
@@ -4,6 +4,7 @@ import kyo.*
 import kyo.core.*
 import kyo.core.internal.*
 import scala.annotation.implicitNotFound
+import scala.annotation.tailrec
 
 opaque type Stream[+T, V, -S] = T < (Streams[V] & S)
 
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -182,88 +175,60 @@ object Stream:
         ): Stream[(T, T2), V, S & S2] =
             s.map(t => s2.map((t, _)))
 
-        def accumulate[S2, V2: Flat](init: V2)(f: (V2, V) => V2 < S2)(
```
fwbrasil, line 185: I've removed this method because I wasn't sure about its semantics. We can add it back later.

kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -182,88 +175,60 @@ object Stream:
...
         ): Stream[T, V, Fibers & S] =
-            Channels.init[V | Stream.Done](size).map { ch =>
+            Channels.init[Chunk[V] | Stream.Done](size).map { ch =>
                 val s2 = s.asInstanceOf[Stream[T, V, Fibers & S]]
                 Fibers.init(s2.runChannel(ch)).map { f =>
                     Streams.emitChannel[V](ch).andThen(f.get)
                 }
             }
 
-        def throttle(meter: Meter): Stream[T, V, S & Fibers] =
```
fwbrasil, line 227: Also removed this method since the throttle would hold the entire chunk.

kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -182,88 +175,60 @@ object Stream:
...
-                            Streams.emitValueAndThen(newAcc)(
-                                Resume(newAcc, k(().asInstanceOf[T]))
-                            )(using tag2)
-                        }
-            Streams.emitValueAndThen(init)(
-                Streams[V].handle(handler)(init, s)
-            )(using tag2)
-        end accumulate
-
-        def reemit[S2, V2: Flat](
```
fwbrasil, line 206: This was more of an internal method that isn't necessary anymore.

hearnadam:
Getting chunking right will be a big part of Kyo's success story. One of the issues I have seen with ZStreams is a 'hidden' re-chunking behavior which can be both hard to debug and can cause performance issues.

I will need to spend some time understanding the code changes, but I believe it's key that we make it extremely obvious where changes to chunking can occur.

kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -273,60 +238,61 @@ object Stream:
...
-            discard.asInstanceOf[Handler[Const[V], Streams[V], Any]]
+        def discardHandler[V]: Handler[Const[Chunk[V]], Streams[V], Any] =
+            discard.asInstanceOf[Handler[Const[Chunk[V]], Streams[V], Any]]
     end internal
 
 end Stream
 
 class Streams[V] extends Effect[Streams[V]]:
-    type Command[T] = V
+    type Command[T] = Chunk[V]
```
hearnadam, line 252: pre-existing: is `T` used?

kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -15,65 +16,68 @@ object Stream:
...
+            else
+                val handler =
+                    new ResultHandler[Int, Const[Chunk[V]], Streams[V], Id, Streams[V]]:
+                        def done[T](st: Int, v: T) = v
+                        def resume[T, U: Flat, S2](
+                            st: Int,
+                            command: Chunk[V],
+                            k: T => U < (Streams[V] & S2)
+                        ) =
+                            if st == 0 then
```
hearnadam, line 56: is it possible for the state to dip below 0?

kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -15,65 +16,68 @@ object Stream:
...
+                                Streams.emitChunkAndThen(command) {
+                                    k(().asInstanceOf[T])
+                                }
+                            else
+                                Streams.emitChunkAndThen(command.dropLeft(st)) {
+                                    Resume(st - Math.min(command.size, st), k(().asInstanceOf[T]))
```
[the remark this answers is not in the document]
fwbrasil: That's actually passing the result of the suspension to resume the computation. `Streams` resume with `Unit` but other effects resume with other types. Take a look at the methods that do the suspension, they return `Unit`.

kyo-core/shared/src/main/scala/kyo/streams.scala
fwbrasil: More or less, this effect suspends with a constant value so the the suspended type isn't directly related to the command. This allows returning a `Unit` in `Streams.emit*`, which then gets resumed in the handlers like in the previous comment. This mechanism isn't very type safe but I consider the core API still internal. We'll need to find ways to make it safer and more ergonomic to allow users to create new effects.

kyo-core/shared/src/main/scala/kyo/streams.scala
fwbrasil: I don't think so. The method checks `n` at the start and then `dropLeft` shouldn't return more elements than `st`.

hearnadam (approved):

PR #272, opened 2024-04-18
add: multi-platform console backed logger

hearnadam:
- move SL4J logger to JVM specific file

This should be an improved experience as users can explicitly rely on console logger if they so choose.


kyo-core/shared/src/main/scala/kyo/logs.scala
```diff
@@ -58,53 +55,57 @@ object Logs:
...
 
     object Unsafe:
-        def apply(logger: org.slf4j.Logger): Unsafe = new Unsafe:
-            inline def traceEnabled: Boolean = logger.isTraceEnabled
-            inline def debugEnabled: Boolean = logger.isDebugEnabled
-            inline def infoEnabled: Boolean  = logger.isInfoEnabled
-            inline def warnEnabled: Boolean  = logger.isWarnEnabled
-            inline def errorEnabled: Boolean = logger.isErrorEnabled
+        class ConsoleLogger(name: String) extends Logs.Unsafe:
+            inline def traceEnabled: Boolean = true
```
hearnadam, line 59: I figured if this is a class, users can extend and override these values.

kyo-core/shared/src/main/scala/kyo/logs.scala
```diff
@@ -58,53 +55,57 @@ object Logs:
...
+            inline def infoEnabled: Boolean = true
+
+            inline def warnEnabled: Boolean = true
+
+            inline def errorEnabled: Boolean = true
 
             inline def trace(msg: => String)(
                 using file: internal.Position
-            ): Unit = if traceEnabled then logger.trace(s"[$file] $msg")
+            ): Unit = if traceEnabled then println(s"TRACE $name -- [$file] $msg")
```
hearnadam, line 71: I tried to follow the SLF4J format I saw in the tests (without the cost of clock or thread id)

`21:02:54.003 [scala-execution-context-global-30] WARN kyo.logs -- [logsTest.scala:35] warn`

fwbrasil:
Nice cleanup! I haven't been very disciplined creating the JS stubs :)

kyo-core/shared/src/main/scala/kyo/logs.scala
fwbrasil: Nice!!

Issue #275, opened 2024-04-19
Aborts effect is very slow with its use of Izumi


lukestephenson:
When running Kyo applications, performance is slower than expected due to the Aborts effect handling. Specifically the type comparisons [here](https://github.com/getkyo/kyo/blob/e748b7ca4645b5de9b1b208f5d2e6bbad7f9bf2e/kyo-core/shared/src/main/scala/kyo/aborts.scala#L51) are generating a huge amount of memory allocations and are the suspected cause of the problem.

Introduced in https://github.com/getkyo/kyo/releases/tag/v0.9.2 as part of resolving #244 in PR #254. Running the above code with version `0.9.1` is substantially faster.

This was initially discovered while experimenting with kyo and writing a Kafka producer. That code is available if it helps with resolving https://github.com/lukestephenson/kyo-playground/blob/main/src/main/scala/playground/KyoKafkaProducer.scala.

Originally raised on discord https://discord.com/channels/1087005439859904574/1087005440623259681/1230729800751251567

fwbrasil:
I was taking a look at how ZIO handles type failures and it doesn't use tags like Kyo does. The reflection happens via user-provided partial functions that will match the correct types without a need for tags:

```scala
  final def catchSome[R1 <: R, E1 >: E, A1 >: A](
    pf: PartialFunction[E, ZIO[R1, E1, A1]]
  )(implicit ev: CanFail[E], trace: Trace): ZIO[R1, E1, A1] 
```
https://github.com/zio/zio/blob/series/2.x/core/shared/src/main/scala/zio/ZIO.scala#L358

Optimizing the use of Izumi tags seems difficult, we'd probably need a new tag implementation. Maybe we could adopt ZIO's approach instead? The API would have to change to something like:

```scala
val a: Int < Aborts[Ex1 | Ex2] = ...
val b: Int < Aborts[Ex2] =
    Aborts.runCatching(a) {
        case ex: Ex1 => 42
    }
val c: Int < Aborts[Ex1] =
    Aborts.runCatching(a) {
        case ex: Ex2 => 42
    }
val d: Int < Any =
    Aborts.runCatching(a) {
        case ex: Ex1 => 42
        case ex: Ex2 => 42
    }
val e: Either[Ex1 | Ex2, Int] < Any
    Aborts.runEither(computation)
```

hearnadam:
The new API seems preferable to me.

lukestephenson:
[...]

PR #277, opened 2024-04-21
introduce failure benchmark

fwbrasil:
This new benchmark is designed to reproduce https://github.com/getkyo/kyo/issues/275. Current results on my M2:

```
[info] Benchmark               Mode  Cnt        Score       Error  Units
[info] FailureBench.forkCats  thrpt    5   132505.248 ± 12212.895  ops/s
[info] FailureBench.forkKyo   thrpt    5    28640.191 ±  1648.773  ops/s
[info] FailureBench.forkZio   thrpt    5    92255.834 ± 11252.393  ops/s
[info] FailureBench.syncCats  thrpt    5   133276.418 ±  4492.660  ops/s
[info] FailureBench.syncKyo   thrpt    5    43961.788 ±  2972.598  ops/s
[info] FailureBench.syncZio   thrpt    5  1453020.306 ± 54500.737  ops/s
```

Kyo's profiling sessions are dominated by Izumi's parsing and subtype checking:

Allocations
alloc

CPU
cpu

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/FailureBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/FailureBench.scala
@@ -1,10 +1,10 @@
 package kyo.bench
 
-class Ex1  extends Exception
+class Ex1  extends Exception derives CanEqual
 object Ex1 extends Ex1
-class Ex2  extends Exception
+class Ex2  extends Exception derives CanEqual
 object Ex2 extends Ex2
-class Ex3  extends Exception
+class Ex3  extends Exception derives CanEqual
 object Ex3 extends Ex3
 
 class FailureBench extends Bench.SyncAndFork[Either[Ex1 | Ex2, Int]]:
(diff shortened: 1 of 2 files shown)


kyo-bench/src/main/scala/kyo/bench/FailureBench.scala
```diff
@@ -0,0 +1,89 @@
...
+        import cats.effect.*
+
+        def loop(i: Int): IO[Either[Ex1, Either[Ex2, Int]]] =
+            if i > depth then IO.pure(Right(Right(i)))
+            else
+                (i % 4) match
+                    case 0 => IO.pure(Left(Ex1))
+                    case 1 => IO.pure(Right(Left(Ex2)))
+                    case 2 =>
+                        loop(i - 1).map {
```
hearnadam, line 22: I must be missing something. If `i` starts at `0` and is only decremented, will this ever reach the `i > depth` condition? Is this meant to run forever?

fwbrasil: My bad, I hadn't noticed my last push had failed. I've just pushed the correct version, this one was actually running a single iteration since it'd start from zero and get a failure on the first loop.

hearnadam (approved):

PR #278, opened 2024-04-22
New Scheduler

fwbrasil:
Kyo's current scheduler is functional and has been serving the projects need's well but the code is focused on simplicity using global objects and with only indirect coverage via tests that use fibers. This PR introduces a new version of the scheduler with several major improvements:

- The scheduler is now in a separate module with no dependencies to other kyo modules. This allows users to adopt the scheduler in isolation, for example with another effect system like ZIO.
- The code was refactored to become more testable via function composition and constructor injection. Its test coverage is quite high now.
- A new `Regulator` abstraction was introduced to replace the concurrency management by `Coordinator`, which has been removed.
- In addition to the concurrency regulator, a new `Admission` regulator probes the scheduler's queuing delay in order to provide a back pressure signal. The signal isn't integrated in other modules yet but it should be used to reject requests in `kyo-tapir` for example.
- The `Task` interface is now a first-class public API that provides built-in runtime tracking, which simplified `IOTask`.

I realize this PR is too large to review. I apologize for that but isolating changes would slow things down too much, which doesn't seem necessary in the current phase of the project. Please let me know if this kind of change is getting too intrusive, though. I'm also planning to eventually write a separate readme for the scheduler but I want to have some experience with it first.


kyo-core/shared/src/main/scala/kyo/core.scala
```diff
@@ -145,12 +145,12 @@ object core:
     end Handler
 
     trait Safepoint[-E]:
-        def check(): Boolean
+        def preempt(): Boolean
```
fwbrasil, line 148: Minor refactoring for clarity.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -129,7 +129,7 @@ private[kyo] class IOPromise[T](state: State[T])
             promise.get() match
                 case _: Pending[T] @unchecked =>
                     IOs {
-                        Scheduler.flush()
+                        Scheduler.get.flush()
```
fwbrasil, line 132: `Scheduler` isn't an object anymore. The global instance returned by `get` isn't lazily initialized because it'd require a memory barrier to read the field. It uses a regular `val` loaded when the companion object class is loaded.

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -12,28 +12,24 @@ import scala.util.control.NonFatal
 private[kyo] class IOTask[T](
     private var curr: T < Fibers,
     private var ensures: Ensures,
-    @volatile private var state: Int // Math.abs(state) => runtime; state < 0 => preempting
-) extends IOPromise[T] with Task
+    initialRuntime: Int
+) extends IOPromise[T] with Task(initialRuntime)
```
fwbrasil, line 16: Runtime and preemption management are now done by `Task`.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/InternalClock.scala
```diff
@@ -0,0 +1,33 @@
+package kyo.scheduler
+
+import java.util.concurrent.Executor
+import java.util.concurrent.locks.LockSupport
+
+final private class InternalClock(executor: Executor):
```
fwbrasil, line 6: `Task`'s runtime tracking has to measure execution times and obtaining the time from the system clock is prohibitively expensive. This internal clock provides approximate time measurements by updating a volatile field every ~1ms. Readers only pay the price of the read barrier and eventual cache misses when the field is updated.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/InternalTimer.scala
```diff
@@ -0,0 +1,31 @@
+package kyo.scheduler
+
+import InternalTimer.*
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration.Duration
+
+abstract private class InternalTimer:
```
fwbrasil, line 8: Introduced for testability.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Queue.scala
```diff
@@ -1,27 +1,28 @@
...
 import java.util.concurrent.atomic.AtomicBoolean
 import kyo.*
 import scala.collection.mutable.PriorityQueue
 
-final private[kyo] class Queue[T](using ord: Ordering[T]) extends AtomicBoolean:
+final private class Queue[T](using ord: Ordering[T]) extends AtomicBoolean:
 
     private val queue = PriorityQueue[T]()
 
-    @volatile private var items = 0
```
fwbrasil, line 11: This is a micro optimization. If `items` is marked as `volatile`, methods like `add` have to use memory barriers to update it within a `modify`/`tryModify` block, which is unnecessary since there's already a write barrier at the end. to release the lock.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Queue.scala
```diff
@@ -1,27 +1,28 @@
...
+    private var items = 0
 
     def isEmpty() =
-        items == 0
+        size() == 0
 
     def size(): Int =
+        VarHandle.acquireFence()
         items
```
fwbrasil, line 19: Force a read barrier since the field isn't volatile anymore so all threads see the latest value.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Queue.scala
```diff
@@ -61,7 +62,7 @@ final private[kyo] class Queue[T](using ord: Ordering[T]) extends AtomicBoolean:
             !isEmpty() && to.isEmpty() && to.tryModify {
                 t = queue.dequeue()
                 val s = size() - 1
-                var i = s - (s / 2)
+                var i = s - Math.ceil(s.toDouble / 2).intValue()
```
fwbrasil, line 65: When a worker had 3 tasks, this code was stealing 2 for example. This new logic makes it steal only 1. Leaving a worker with only one task increases the likelihood of it'll soon go to sleep again. If a thief is successful, it'll naturally try stealing again when it's out of tasks.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/util/MovingStdDev.scala
```diff
@@ -0,0 +1,39 @@
+package kyo.scheduler.util
+
+final private[kyo] class MovingStdDev(window: Int):
```
fwbrasil, line 3: Since this class isn't used in a very hot path anymore and is called only by regulators that execute periodically, it was updated to favor precision over performance.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -0,0 +1,230 @@
...
+    private val clock   = InternalClock(clockExecutor)
+    private val workers = new Array[Worker](maxWorkers)
+    private val flushes = new LongAdder
+
+    @volatile private var allocatedWorkers = 0
+    @volatile private var maxConcurrency   = coreWorkers
+
+    val a1, a2, a3, a4, a5, a6, a7 = 0L // padding
+
+    @volatile private var cycles = 0L
```
fwbrasil, line 41: This field used to be in `Coordinator`.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -0,0 +1,230 @@
...
+        admissionRegulator.reject(key)
+
+    def asExecutor: Executor =
+        (r: Runnable) => schedule(Task(r.run()))
+
+    def asExecutionContext: ExecutionContext =
+        ExecutionContext.fromExecutor(asExecutor)
+
+    @tailrec
+    private def schedule(t: Task, submitter: Worker): Unit =
```
fwbrasil, line 74: These scheduling methods have not changed.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -0,0 +1,230 @@
...
+    private def updateWorkers(delta: Int) =
+        maxConcurrency = Math.max(minWorkers, Math.min(maxWorkers, maxConcurrency + delta))
+        ensureWorkers()
+
+    private def ensureWorkers() =
+        for idx <- allocatedWorkers until maxConcurrency do
+            workers(idx) = new Worker(idx, pool, schedule, steal, () => cycles, clock)
+            allocatedWorkers += 1
+
+    private val cycleTask =
```
fwbrasil, line 160: Moved from `Coordinator`

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -0,0 +1,230 @@
...
+    private def cycleWorkers(): Unit =
+        try
+            val curr = cycles + 1
+            cycles = curr
+            var i = 0
+            while i < allocatedWorkers do
+                val w = workers(i)
+                if w != null then
+                    if i >= maxConcurrency then
+                        w.drain()
```
fwbrasil, line 177: Improved logic to continue cycling all workers but also draining workers that have been stopped (are above the concurrency limit)

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Worker.scala
```diff
@@ -0,0 +1,149 @@
...
+    exec: Executor,
+    scheduleTask: (Task, Worker) => Unit,
+    stealTask: Worker => Task,
+    getCurrentCycle: () => Long,
+    clock: InternalClock
+) extends Runnable:
+
+    val a1, a2, a3, a4, a5, a6, a7 = 0L // padding
+
+    @volatile private var running       = false
```
fwbrasil, line 22: This class hasn't changed much but I've made `running` a `volatile` that is updated using `VarHandle` as a micro optimization to avoid the pointer chasing with `AtomicBoolean` since the flag used in the hot path of task execution.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Worker.scala
```diff
@@ -0,0 +1,149 @@
...
+            c.doPreempt()
+    end cycle
+
+    def handleBlocking() =
+        val m = mount
+        val r = m != null && running && {
+            val state = m.getState().ordinal()
+            state == Thread.State.BLOCKED.ordinal() ||
+            state == Thread.State.WAITING.ordinal() ||
+            state == Thread.State.TIMED_WAITING.ordinal()
```
fwbrasil, line 78: Previous implementation was missing some of these states.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/InternalTimer.scala
```diff
@@ -0,0 +1,31 @@
...
+        new InternalTimer:
+            def schedule(interval: Duration)(f: => Unit): TimerTask =
+                val future = executor.scheduleWithFixedDelay(() => f, interval.toNanos, interval.toNanos, TimeUnit.NANOSECONDS)
+                new TimerTask:
+                    def cancel() = future.cancel(true)
+            end schedule
+
+            def scheduleOnce(delay: Duration)(f: => Unit): TimerTask =
+                val future = executor.schedule((() => f): Runnable, delay.toNanos, TimeUnit.NANOSECONDS)
+                new TimerTask:
```
hearnadam, line 27: Is it possible to avoid this allocation?

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/regulator/Regulator.scala
```diff
@@ -0,0 +1,96 @@
...
+        try
+            if loadAvg() < loadAvgTarget then
+                measure(0)
+            else
+                stats.probes.inc()
+                probe()
+            end if
+        catch
+            case ex if NonFatal(ex) =>
+                log.error(s"🙈 !!Kyo Scheduler Bug!! ${getClass.getSimpleName()} regulator's probe collection has failed.", ex)
```
hearnadam, line 43: now that we are using a logger instance for this class, we can remove `getClass.getSimpleName()`

hearnadam (commented):
Looks like a huge improvement! Nice work!

fwbrasil:
[Benchmark results](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/dadc149d9fadfcbd299763d95b4ab7fb/raw/221ab9cfa48d4cf6bf875e582c9ab2f15943cf51/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/dadc149d9fadfcbd299763d95b4ab7fb/raw/221ab9cfa48d4cf6bf875e582c9ab2f15943cf51/jmh-result-candidate.json) with the new scheduler: 

image

The results are mixed and the `RandomBench` improvement seems just noise. I think the results are reasonable, though. The scheduler now does much more in terms of handling blocking, adjusting concurrency, and providing a backpressure signal. Not having a significant regression is already a good win and we can keep optimizing it over time.

@hearnadam would you be able to check out this branch and run [this main](https://github.com/getkyo/kyo/blob/37467bfc13fab6ed262e2aecb22607fa31aff0bd/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/util/SelfCheck.scala#L90)? It checks that the regulators are working properly, it'd be nice to get data from other machines as well. I'm doing some more testing using containers and CPU quotas to validate the behavior with CPU throttling and I'm planning to merge this PR if everything looks ok.

kyo-scheduler/js/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -3,13 +3,19 @@ package kyo.scheduler
 import scala.scalajs.concurrent.JSExecutionContext
 
 object Scheduler:
+    lazy val get = new Scheduler
```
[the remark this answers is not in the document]
fwbrasil: fixed!

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/InternalTimer.scala
fwbrasil: You're getting used to Kyo's focus on performance :) In this case, I don't think it's necessary because only two tasks are submitted when the scheduler starts. It'd also be more difficult to mock this class in tests if we used an opaque type to alias `ScheduledFuture`.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/regulator/Regulator.scala
fwbrasil: good catch!

fwbrasil: ah actually I'm not using Kyo's logging because the idea is for the scheduler module to be isolated without a dependency to kyo-core. This is the regular slf4j logger.

hearnadam: Right, but you are using SLF4J instance for each class which should include the classname. See line 92:

```
private[Regulator] val log = LoggerFactory.getLogger(getClass)
```

This goes throughout

hearnadam:
@fwbrasil will take a look at running this tonight.

fwbrasil:
Self-check is ok with CPU quotas as well:

```
fwbrasil@flavios-mbp kyo % docker run --rm --cpus=1 -v /Users/fwbrasil/workspace/kyo/kyo-scheduler/jvm/target/scala-3.4.1:/app adoptopenjdk/openjdk11 java -cp /app/kyo-scheduler-assembly-0.9.2+71-82298105+20240423-1138-SNAPSHOT.jar kyo.scheduler.util.SelfCheckMain
Map(clients -> 0, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 1, rejectionPercent -> 0.02040816326530612, rejectionThreshold -> 0.2)
Map(clients -> 2, rejectionPercent -> 0.8846153846153846, rejectionThreshold -> 0.2)
Success
fwbrasil@flavios-mbp kyo % docker run --rm --cpus=2 -v /Users/fwbrasil/workspace/kyo/kyo-scheduler/jvm/target/scala-3.4.1:/app adoptopenjdk/openjdk11 java -cp /app/kyo-scheduler-assembly-0.9.2+71-82298105+20240423-1138-SNAPSHOT.jar kyo.scheduler.util.SelfCheckMain
Map(clients -> 0, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 1, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 2, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 3, rejectionPercent -> 1.2727272727272727, rejectionThreshold -> 0.2)
Success
fwbrasil@flavios-mbp kyo % docker run --rm --cpus=3 -v /Users/fwbrasil/workspace/kyo/kyo-scheduler/jvm/target/scala-3.4.1:/app adoptopenjdk/openjdk11 java -cp /app/kyo-scheduler-assembly-0.9.2+71-82298105+20240423-1138-SNAPSHOT.jar kyo.scheduler.util.SelfCheckMain
Map(clients -> 0, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 1, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 2, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 3, rejectionPercent -> 0.06521739130434782, rejectionThreshold -> 0.2)
Map(clients -> 4, rejectionPercent -> 0.21951219512195122, rejectionThreshold -> 0.2)
Success
fwbrasil@flavios-mbp kyo % docker run --rm --cpus=4 -v /Users/fwbrasil/workspace/kyo/kyo-scheduler/jvm/target/scala-3.4.1:/app adoptopenjdk/openjdk11 java -cp /app/kyo-scheduler-assembly-0.9.2+71-82298105+20240423-1138-SNAPSHOT.jar kyo.scheduler.util.SelfCheckMain
Map(clients -> 0, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 1, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 2, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 3, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 4, rejectionPercent -> 0.06521739130434782, rejectionThreshold -> 0.2)
Map(clients -> 5, rejectionPercent -> 0.21951219512195122, rejectionThreshold -> 0.2)
Success
fwbrasil@flavios-mbp kyo % docker run --rm --cpus=5 -v /Users/fwbrasil/workspace/kyo/kyo-scheduler/jvm/target/scala-3.4.1:/app adoptopenjdk/openjdk11 java -cp /app/kyo-scheduler-assembly-0.9.2+71-82298105+20240423-1138-SNAPSHOT.jar kyo.scheduler.util.SelfCheckMain
Map(clients -> 0, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 1, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 2, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 3, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 4, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 5, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 6, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 7, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 8, rejectionPercent -> 0.020833333333333332, rejectionThreshold -> 0.2)
Map(clients -> 9, rejectionPercent -> 1.5, rejectionThreshold -> 0.2)
Success
fwbrasil@flavios-mbp kyo % docker run --rm --cpus=6 -v /Users/fwbrasil/workspace/kyo/kyo-scheduler/jvm/target/scala-3.4.1:/app adoptopenjdk/openjdk11 java -cp /app/kyo-scheduler-assembly-0.9.2+71-82298105+20240423-1138-SNAPSHOT.jar kyo.scheduler.util.SelfCheckMain
Map(clients -> 0, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 1, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 2, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 3, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 4, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 5, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 6, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 7, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 8, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 9, rejectionPercent -> 1.380952380952381, rejectionThreshold -> 0.2)
Success
fwbrasil@flavios-mbp kyo % docker run --rm --cpus=7 -v /Users/fwbrasil/workspace/kyo/kyo-scheduler/jvm/target/scala-3.4.1:/app adoptopenjdk/openjdk11 java -cp /app/kyo-scheduler-assembly-0.9.2+71-82298105+20240423-1138-SNAPSHOT.jar kyo.scheduler.util.SelfCheckMain
Map(clients -> 0, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 1, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 2, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 3, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 4, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 5, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 6, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 7, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 8, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 9, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 10, rejectionPercent -> 0.1951219512195122, rejectionThreshold -> 0.2)
Map(clients -> 11, rejectionPercent -> 0.7241379310344828, rejectionThreshold -> 0.2)
Success
fwbrasil@flavios-mbp kyo % docker run --rm --cpus=8 -v /Users/fwbrasil/workspace/kyo/kyo-scheduler/jvm/target/scala-3.4.1:/app adoptopenjdk/openjdk11 java -cp /app/kyo-scheduler-assembly-0.9.2+71-82298105+20240423-1138-SNAPSHOT.jar kyo.scheduler.util.SelfCheckMain
Map(clients -> 0, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 1, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 2, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 3, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 4, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 5, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 6, rejectionPercent -> 0.041666666666666664, rejectionThreshold -> 0.2)
Map(clients -> 7, rejectionPercent -> 0.53125, rejectionThreshold -> 0.2)
Success
fwbrasil@flavios-mbp kyo % docker run --rm --cpus=9 -v /Users/fwbrasil/workspace/kyo/kyo-scheduler/jvm/target/scala-3.4.1:/app adoptopenjdk/openjdk11 java -cp /app/kyo-scheduler-assembly-0.9.2+71-82298105+20240423-1138-SNAPSHOT.jar kyo.scheduler.util.SelfCheckMain
Map(clients -> 0, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 1, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 2, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 3, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 4, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 5, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 6, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 7, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 8, rejectionPercent -> 0.0425531914893617, rejectionThreshold -> 0.2)
Map(clients -> 9, rejectionPercent -> 1.0, rejectionThreshold -> 0.2)
Success
fwbrasil@flavios-mbp kyo % docker run --rm --cpus=10 -v /Users/fwbrasil/workspace/kyo/kyo-scheduler/jvm/target/scala-3.4.1:/app adoptopenjdk/openjdk11 java -cp /app/kyo-scheduler-assembly-0.9.2+71-82298105+20240423-1138-SNAPSHOT.jar kyo.scheduler.util.SelfCheckMain
Map(clients -> 0, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 1, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 2, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 3, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 4, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 5, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 6, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 7, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
Map(clients -> 8, rejectionPercent -> 0.041666666666666664, rejectionThreshold -> 0.2)
Map(clients -> 9, rejectionPercent -> 0.16666666666666666, rejectionThreshold -> 0.2)
Map(clients -> 10, rejectionPercent -> 1.2727272727272727, rejectionThreshold -> 0.2)
Success
```

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/regulator/Regulator.scala
fwbrasil: The `log` in this case would be `Regulator`, not the subclass. But I've decided to remove the `slf4j` dependency since logging is used only for bugs. There's the `LoomSupport` warning but I think it's ok to log that with a regular `println`. The module now has zero dependencies, which should help avoid issues in case it's used in isolation without other Kyo modules.

fwbrasil:
I've made a few new optimizations and the [benchmark results](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/35925aba0bb39e80c15e9b6cb7a88a7a/raw/0ced03630b2263355d04ca878e7c4b26e6eff44b/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/35925aba0bb39e80c15e9b6cb7a88a7a/raw/0ced03630b2263355d04ca878e7c4b26e6eff44b/jmh-result-candidate.json) are reporting more consistent good results:

image

The main cause for the regression in `ForkSpawnBench` is the path for the fiber to check for the preemption signal involving more code now that `Task` manages the preemption instead of `IOTask`. Since `Task` is a `trait`,  Scala encodes methods in the companion object and accessing the `state` field has to go through a getter.

hearnadam:
@fwbrasil 
```
sbt:kyo-scheduler> runMain kyo.scheduler.util.SelfCheckMain
[info] running (fork) kyo.scheduler.util.SelfCheckMain 
[info] Map(clients -> 0, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
[info] Map(clients -> 1, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
[info] Map(clients -> 2, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
[info] Map(clients -> 3, rejectionPercent -> 0.0, rejectionThreshold -> 0.2)
[info] Map(clients -> 4, rejectionPercent -> 0.02127659574468085, rejectionThreshold -> 0.2)
[info] Map(clients -> 5, rejectionPercent -> 0.75, rejectionThreshold -> 0.2)
[info] Failure: Expected between 6.4 and 14.0 clients for 8 cores but found 5.
```

On: `6e838b14047e9eb0f57caab112c34ba49196ef2b`

fwbrasil:
@hearnadam can you check if you have other processes that could be using some of the CPU?

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/regulator/Regulator.scala
He-Pin: Cool emoji!

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/package.scala
```diff
@@ -0,0 +1,10 @@
+package kyo
+
+package object scheduler:
+
+    private def statsScope(path: String*) =
+        "kyo" :: "scheduler" :: path.toList
+
+    private def bug(msg: String, ex: Throwable) =
+        (new Exception("🙈 !!Kyo Scheduler Bug!! " + msg, ex)).printStackTrace(System.err)
```
sideeffffect, line 9:
```suggestion
    import java.util.logging.*
    private lazy val logger = Logger.getLogger(getClass.getName)
    
    private def bug(msg: String, ex: Throwable) =
        logger.log(Level.SEVERE, "🙈 !!Kyo Scheduler Bug!!", new Exception(msg, ex))
```
Since you want 0 dependencies, we can still use java.util.logging, since that's part of the standard library. This will give people better interoperability with other logging frameworks or other tools hooked on logging, like Sentry.

fwbrasil: good idea! 🙏

kyo-scheduler/js/src/main/scala/kyo/scheduler/InternalClock.scala
```diff
@@ -0,0 +1,13 @@
+package kyo.scheduler
+
+import java.util.concurrent.Executor
+import scala.annotation.nowarn
+
+@nowarn
+final class InternalClock(executor: Executor = null):
+
+    def currentMillis(): Long = System.currentTimeMillis()
```
He-Pin, line 9: Should it be nano time?

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/InternalClock.scala
```diff
@@ -0,0 +1,44 @@
...
+
+    @volatile private var _stop = false
+
+    val a1, a2, a3, a4, a5, a6, a7 = 0L // padding
+
+    @volatile private var millis = System.currentTimeMillis()
+
+    val b1, b2, b3, b4, b5, b6, b7 = 0L // padding
```
He-Pin, line 15: Cool padding

fwbrasil:
I've changed the code use java logging as @sideeffffect suggested. I've also tried [another optimization](https://github.com/getkyo/kyo/pull/278/commits/e762687d112f48c27a9b214ededa656ce73d8e40) moving the preemption handling from `Task` to `Worker` but the results weren't good because obtaining the current worker is more expensive.

I'm merging this PR since it keeps getting larger but please feel free to provide feedback on the change!

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -0,0 +1,232 @@
...
+            if w != null && !w.handleBlocking() then
+                val l = w.load()
+                if l > maxLoad && (w ne thief) then
+                    maxLoad = l
+                    worker = w
+            end if
+            i += 1
+        end while
+        if worker != null then
+            worker.steal(thief)
```
He-Pin, line 119: Worker stealing thief not thief stealing worker?

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/util/LoomSupport.scala
```diff
@@ -0,0 +1,40 @@
...
+                schedulerField.setAccessible(true)
+                schedulerField.set(ofVirtual, exec)
+
+                val factoryMethod = ofVirtual.getClass.getDeclaredMethod("factory")
+                factoryMethod.setAccessible(true)
+                val factory = factoryMethod.invoke(ofVirtual)
+
+                classOf[Executors]
+                    .getMethod("newThreadPerTaskExecutor", classOf[ThreadFactory])
+                    .invoke(null, factory).asInstanceOf[Executor]
```
He-Pin, line 29: Cool, I just did the same too
 toohttps://github.com/apache/rocketmq/pull/8063

He-Pin: So you don't plan to pooling the virtual thread right?

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/package.scala
He-Pin: You can define your own logging and then there will be a adapter. 

Netty 5 is using slf4j again

kyo-scheduler/js/src/main/scala/kyo/scheduler/InternalClock.scala
fwbrasil: `System.nanoTime` isn't monotonic across threads. It also helps to hide the imprecision of the clock.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
fwbrasil: That's `thief` taking tasks from `worker` but yeah, the API isn't very clear.

He-Pin: Maybe stealing or divisionBy

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/util/LoomSupport.scala
fwbrasil: Cool! I tried polling but putting virtual threads to sleep is too expensive and it's cheaper to allocate new ones. Virtual threads are allocated only when workers start, though. A worker mounted on a virtual thread can several tasks until there's no more pending work, which alleviates the cost of the virtual thread creation on each task execution.

He-Pin: Yeah,  If totally move to none pooling will need to change a lot of the current work base at work, where thread pool is using like some kind of concurrency limiter too

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/package.scala
```diff
@@ -0,0 +1,25 @@
...
+
+package object scheduler:
+
+    private val log = Logger.getLogger("kyo.scheduler")
+
+    private def statsScope(path: String*) =
+        "kyo" :: "scheduler" :: path.toList
+
+    private def bug(msg: String, ex: Throwable) =
+        log.severe(s"🙈 !!Kyo Scheduler Bug!! $msg \n Caused by: ${stackTrace(ex)}")
```
sideeffffect, line 15: Just curious, why not use the overload of this method where you can pass the `Throwable` as an argument?

I was hoping that the logging backend will do the printing of the Throwable's stack trace for us, instead of us having to do it manually with out own ad hoc method (`private def stackTrace(ex: Throwable)`).

Supplying an intact Throwable instance to the logger will also have better interop with 3rd parties. (You can always turn a Throwable to String, but you can't go the other way around.)

Issue #266, opened 2024-04-15
KyoApp review


hearnadam:
KyoApp should be the default place users run Kyo applications. Currently, a few minor issues and we should aim to address prior to 1.0 release.

1. Provide configuration mechanism for default set of services, Consoles, Logs, Clocks, etc
2. Provide configuration for default app level timeout (useful for 'task' related scripts)
3. Configurable error handling? Can we offer a mechanism to support arbitrary `Aborts` types?

Feel free to append more wants/needs as you come across them

hearnadam:
Closing this issue.
1 is resolved
2 seems less useful when I think about it more
3 I think may deserve attention in its own issue.

PR #280, opened 2024-04-23
Enable providing custom services in KyoApp

hearnadam:
Begin addressing https://github.com/getkyo/kyo/issues/266

Later, these services will enable us to build a test integration on `KyoApp` instances. Using KyoApp should enable users to run tests as `Main`s.

This will also simplify providing a custom logger.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
+++ b/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
@@ -5,11 +5,22 @@ import scala.concurrent.duration.Duration
 import scala.util.Try
 
 abstract class KyoApp extends KyoApp.Base[KyoApp.Effects]:
+    def log: Logs.Unsafe = Logs.unsafe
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -5,11 +5,22 @@ import scala.concurrent.duration.Duration
 import scala.util.Try
 
 abstract class KyoApp extends KyoApp.Base[KyoApp.Effects]:
+    val log: Logs.Unsafe = Logs.unsafe
+    val random: Random   = Random.default
+    val clock: Clock     = Clock.default
```
fwbrasil, line 10: There are sometimes issues with class initialization in app classes. I think it's safer to keep these as `def`.

hearnadam: Done.

PR #282, opened 2024-04-24
ZIO Test Integration

hearnadam:
Working, albeit messy zio-test integration. I think this is generally preferable to Scalatest (for many reasons), but open to feedback.

Much of this is closely tied to the implementation of `ZIOSpecDefault` or more accurately `ZIOSpec`.

I currently cannot get sbt to detect the tests to run, but that may be a fault in my configuration.

---

One of the perks of Kyo's model is that we can simply assume all assertions may have pending effects. As such, we can always invoke the `run`. In ZIO's case, there is logic to determine which `TestConstructor[_, _]` to use via implicit resolution. If a pure `TestResult` is passed as an assertion, it's necessary to wrap it in `ZIO.succeed`

### DIFF
--- a/kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
+++ b/kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
@@ -3,7 +3,7 @@ package kyoTest
 import kyo.*
 import kyo.KyoApp.Effects
 import scala.concurrent.Future
-import scala.concurrent.duration.Duration
+import scala.concurrent.duration.*
 import scala.language.postfixOps
 import zio.EnvironmentTag
 import zio.Trace
@@ -49,9 +49,9 @@ end KyoSpecAbstract
 
 abstract class KyoSpecDefault extends KyoSpecAbstract[KyoApp.Effects]:
     final override def run[In](v: => In < KyoApp.Effects)(using Flat[In < KyoApp.Effects]): ZIO[Environment, Throwable, In] =
-        ZIO.fromFuture { implicit ec => IOs.run(KyoApp.runFiber(timeout)(IOs(v)).toFuture).map(_.get) }
+        ZIO.fromFuture { implicit ec => IOs.run(KyoApp.runFiber(timeout)(v).toFuture).map(_.get) }
 
-    def timeout: Duration = Duration.Inf
+    def timeout: Duration = if Platform.isDebugEnabled then Duration.Inf else 5.seconds
 
     def spec: Spec[Any, Any]
 
@@ -61,25 +61,25 @@ object ExampleSpec extends KyoSpecDefault:
     def spec: Spec[Any, Throwable] =
         suite("suite!")(
             test("pure") {
-                assertTrue(true)
+                assertCompletes
             },
             test("IOs Succeed") {
-                IOs(assertTrue(true))
+                IOs(assertCompletes)
             },
             test("IOs Fail") {
                 for
                     _ <- IOs.fail("ERROR!")
-                yield assertTrue(true)
+                yield assertCompletes
             } @@ TestAspect.failing,
             test("Aborts!") {
                 for
                     _ <- Aborts[Throwable].fail(new RuntimeException("Aborts!"))
-                yield assertTrue(true)
+                yield assertCompletes
             } @@ TestAspect.failing,
             test("Fibers.sleep") {
                 for
                     _ <- Fibers.sleep(Duration.Inf)
-                yield assertTrue(true)
+                yield assertCompletes
             } @@ TestAspect.timeout(zio.Duration.Zero) @@ TestAspect.ignore
         )
 end ExampleSpec
--- a/kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
+++ b/kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
@@ -2,55 +2,62 @@ package kyoTest
 
 import kyo.*
 import kyo.KyoApp.Effects
-import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
 import scala.language.postfixOps
+import zio.EnvironmentTag
+import zio.Trace
 import zio.ZIO
 import zio.ZLayer
 import zio.internal.stacktracer.SourceLocation
 import zio.test.*
 
-abstract class KyoTestZIO[S] extends ZIOSpecAbstract:
-    final val environmentTag: zio.EnvironmentTag[Any] = zio.EnvironmentTag[Any]
+abstract class KyoSpecAbstract[S] extends ZIOSpecAbstract:
+    override type Environment = Any
+    override val environmentTag: EnvironmentTag[Any] = EnvironmentTag[Environment]
+    override def bootstrap: ZLayer[Any, Any, Any]    = ZLayer.empty
 
+    // copied from `TestAnnotation` as it's `private[zio]`
+    // https://github.com/zio/zio/blob/series/2.x/test/shared/src/main/scala/zio/test/TestAnnotation.scala#L91-L92
     private val trace: TestAnnotation[List[SourceLocation]] =
         TestAnnotation("trace", List.empty, _ ++ _)
 
-    override type Environment = Any
-    override def bootstrap: ZLayer[Any, Any, Any] = ZLayer.empty
+    def spec: Spec[Environment, Any]
 
-    def spec: Spec[Any, Any]
-    def timeout: Duration = Duration.Inf
-    def run[In](v: In < S)(using ExecutionContext, Flat[In < S]): Future[In]
+    def run[In](v: => In < S)(using Flat[In < S]): ZIO[Environment, Throwable, In]
 
-    def test[In <: TestResult](label: String)(assertion: In < S)(
-        using
+    def test[In <: TestResult](label: String)(assertion: => In < S)(using
         fl: Flat[In < S],
         sl: SourceLocation,
-        tr: zio.Trace
+        tr: Trace
     ): Spec[Any, Throwable] =
         Spec.labeled(
             label,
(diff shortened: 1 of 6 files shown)


kyo-core/shared/src/test/scala/kyoTest/KyoTestZIO.scala
```diff
@@ -0,0 +1,78 @@
...
+                for
+                    _ <- IOs.fail("ERROR!")
+                yield assertTrue(true)
+            } @@ TestAspect.failing,
+            test("Aborts!") {
+                for
+                    _ <- Aborts[Throwable].fail(new RuntimeException("Aborts!"))
+                yield assertTrue(true)
+            } @@ TestAspect.failing,
+            test("Fibers.sleep") {
```
hearnadam, line 72:
```
[info]   - Fibers.sleep
[info] Timeout of 0 ns exceeded.
[info] 4 tests passed. 1 tests failed. 0 tests ignored.
[info]   - suite! - Fibers.sleep
[info] Timeout of 0 ns exceeded.
[info] Executed in 288 ms
[info] timestamp=2024-04-24T06:22:45.194897Z level=ERROR thread=#zio-fiber-1 message="" cause="Exception in thread "zio-fiber-4" java.lang.RuntimeException: null
[info]  at zio.test.ZIOSpecAbstract.runSpec$$anonfun$1$$anonfun$2$$anonfun$1$$anonfun$3$$anonfun$2$$anonfun$1(ZIOSpecAbstract.scala:79)
[info]  at zio.ZIO$.fail$$anonfun$1(ZIO.scala:3151)
[info]  at zio.ZIO$.failCause$$anonfun$1$$anonfun$1$$anonfun$1(ZIO.scala:3160)"
```

Can we improve this error message? ie should the exception have a message?

hearnadam: It looks like there are two causes for the bad error messages. When timing out fibers directly (from Kyo code), the `NoStackTrace` exception is throwing with a message of `null`.  This can be fixed via adding a message here:

```
    case object Interrupted
        extends RuntimeException("Fiber Interrupted")
        with NoStackTrace

    private[kyo] val interrupted = IOs.fail(Interrupted)
```

However, this still does not solve the `cause="Exception in thread "zio-fiber-4" java.lang.RuntimeException: null`. Will need to dig in further.

hearnadam:
@fwbrasil please review when you get a chance. We can split this out into `kyo-zio-test` prior to publishing. I will also split out these into multiple files for each class.

kyo-core/shared/src/test/scala/kyoTest/KyoTestZIO.scala
hearnadam: Fixed with latest commit:

```
[info]   - Fibers.sleep
[info] Timeout of 0 ns exceeded.
[info] 4 tests passed. 1 tests failed. 0 tests ignored.
[info]   - suite! - Fibers.sleep
[info] Timeout of 0 ns exceeded.
[info] Executed in 248 ms
[info] timestamp=2024-04-24T19:33:53.526752Z level=ERROR thread=#zio-fiber-1761416989 message="" cause="Exception in thread "zio-fiber-1308002466" java.lang.RuntimeException: Tests failed.
[info]  at zio.test.ZIOSpecAbstract.runSpec$$anonfun$1$$anonfun$2$$anonfun$1$$anonfun$3$$anonfun$2$$anonfun$1(ZIOSpecAbstract.scala:80)
[info]  at zio.ZIO$.fail$$anonfun$1(ZIO.scala:3154)
[info]  at zio.ZIO$.failCause$$anonfun$1$$anonfun$1$$anonfun$1(ZIO.scala:3163)"

```

I think this is fixed by switching to ZIO 2.1 which has improved ZIO.fromFuture

kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
```diff
@@ -0,0 +1,85 @@
...
+import scala.concurrent.duration.Duration
+import scala.language.postfixOps
+import zio.EnvironmentTag
+import zio.Trace
+import zio.ZIO
+import zio.ZLayer
+import zio.internal.stacktracer.SourceLocation
+import zio.test.*
+
+abstract class KyoSpecAbstract[S] extends ZIOSpecAbstract:
```
hearnadam, line 15: `ZIOSpecAbstract` has a timeout warning after 60 seconds. This works correctly with `Fiber.sleep`:

```
timestamp=2024-04-24T19:41:29.306520Z level=WARN thread=#zio-fiber-160508830 message="Test suite! - Fibers.sleep has taken more than 1 m to execute. If this is not expected, consider using TestAspect.timeout to timeout runaway tests for faster diagnostics."
```

kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
```diff
@@ -0,0 +1,85 @@
...
+            test("Aborts!") {
+                for
+                    _ <- Aborts[Throwable].fail(new RuntimeException("Aborts!"))
+                yield assertTrue(true)
+            } @@ TestAspect.failing,
+            test("Fibers.sleep") {
+                for
+                    _ <- Fibers.sleep(Duration.Inf)
+                yield assertTrue(true)
+            } @@ TestAspect.timeout(zio.Duration.Zero) @@ TestAspect.ignore
```
hearnadam, line 83: Timeouts work great with the interop, which is slightly surprising - but glad it works.

kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
```diff
@@ -0,0 +1,85 @@
...
+        sl: SourceLocation,
+        tr: Trace
+    ): Spec[sc.OutEnvironment, sc.OutError] =
+        zio.test.suite(label)(specs*)
+
+end KyoSpecAbstract
+
+abstract class KyoSpecDefault extends KyoSpecAbstract[KyoApp.Effects]:
+    final override def run[In](v: => In < KyoApp.Effects)(using Flat[In < KyoApp.Effects]): ZIO[Environment, Throwable, In] =
+        ZIO.fromFuture { implicit ec => IOs.run(KyoApp.runFiber(timeout)(IOs(v)).toFuture).map(_.get) }
```
hearnadam, line 52: Maybe we can add a `private[kyo] def KyoApp.runToFuture` method.

Note, I am wrapping in `IOs` here to ensure suspension - not sure that's necessary.

kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
```diff
@@ -0,0 +1,85 @@
...
+
+    def timeout: Duration = Duration.Inf
```
hearnadam, line 54: Not sure we need this. Users can use the ZIO timeout aspects.

kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
fwbrasil: I think we need to propagate ZIO's cancelation to Kyo's fibers otherwise the test could continue running in the background.

kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
fwbrasil: Maybe `KyoApp.runFuture`? I think it could be public as well, it seems useful. There's no need for the `IOs` suspension since the code handles `IOs` and is also suspended in ZIO.

kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
fwbrasil: I think a default timeout is important since people might not add the aspect. `KyoTest` has a behavior to disable the timeout in case there's an active debugging session. It's quite useful when working on tests so we'd need to port it as well.

fwbrasil:
I haven't used `zio-test` yet, could you share more on its advantages? Something to consider is if it'd impact the execution time of the tests. I'd be much more inclined to migrate in case it's faster for example. A fast test suite is essential in my development flow.

kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
hearnadam: Sure - I was just hesitant to suggest a public function as it seems like it could be misused.

hearnadam:
zio-test will likely be equivalent in performance to scalatest - though there are ways to speed it up. You can use the `TestAspect.parallel` - I don't remember scalatest having this feature. The best benefits of zio-test in my experience are:
- test generators for property testing: https://zio.dev/reference/test/property-testing/how-generators-work/
- easy to use assertion mechanism, with clear error rendering


In scalatest:

```
val x = 1
val y = 2

x should be 
y
```

kyo-core/shared/src/test/scala/kyoTest/KyoSpecAbstract.scala
hearnadam: Can we can update `Fibers.timeout` to ignore `Duration.Inf`? This should lower the overhead for each test

fwbrasil:
LGTM. Is this ready to merge?

hearnadam:
@fwbrasil it can be merged in the current state, but the remaining work:

- split to new project/module
- add KyoApp.runFuture / equivalent 
- add unit tests

hearnadam:
@fwbrasil I removed the `if debug` for the fiber timeout, as I don't want to migrate the existing `Platform` implementation. I think we can use test aspects or another ZIO specific mechanism for this...

I split the code to new files + project, so it should be consumable from all kyo projects. We can resolve the remaining issues later.

PR #284, opened 2024-04-25
Cross compile scheduler to Scala 2.12 and 2.13

fwbrasil:


build.sbt
```diff
@@ -81,6 +83,8 @@ lazy val `kyo-scheduler` =
         .in(file("kyo-scheduler"))
         .settings(
             `kyo-settings`,
+            scalacOptions -= "-language:strictEquality",
```
hearnadam, line 86: It should be possible to configure scalacOptions per compiler version.

Maybe it's worth adding `-Xsource:3` for Scala 2.13.

https://github.com/scala/scala/releases/tag/v2.13.13

build.sbt
```diff
@@ -1,6 +1,8 @@
 import sys.process.*
 
-val scala3Version = "3.4.1"
+val scala3Version   = "3.4.1"
+val scala212Version = "2.12.19"
```
hearnadam, line 4: I see no value in supporting 2.12 - there are some unexpected differences in collection behavior which may be undesirable.

hearnadam (approved):

build.sbt
fwbrasil: The use of collections is very limited in the scheduler. There are companies still on 2.12, it'd be nice to have the scheduler available for them as well.

build.sbt
fwbrasil: I've tried a few ways but sbt is always a big mystery for me. I've added the flag in a naive way and it's working but sbt is emitting a warning. I think it's ok for now.

PR #286, opened 2024-04-25
Log the exception in logger

sideeffffect:
continuation of https://github.com/getkyo/kyo/pull/278#discussion_r1579100968

I think logging the exception itself should be more helpful, with better interop, etc.

But if there's a good reason to print it manually then no worries, feel free to close the PR.


fwbrasil:
Sorry, I had missed your feedback. Thank you!

Issue #208, opened 2024-03-21
Make Choices handler tail-recursive


fwbrasil:
The implementation of the `Choices` handler isn't tail-recursive since it invokes a nested `Choices.run`: https://github.com/getkyo/kyo/blob/457d9f5958afb6af8b847245164d85de3362e439/kyo-core/shared/src/main/scala/kyo/choices.scala#L36

More background: https://github.com/getkyo/kyo/issues/201#issuecomment-2007622533

fwbrasil:
The effect is now stack safe in more scenarios and the remaining one is likely not possible to make tail-recursive. Closing this.

PR #287, opened 2024-04-26
Loops-based Seqs

fwbrasil:
Reimplements `Seqs` based on `Loops`. With this change, all methods are stack safe, have good performance with low allocation rate, and are compatible with multi-shot continuations since they don't leverage mutable state across transformations like other effect systems typically do by leveraging iterators and mutable buffers internally.

This change partially solves https://github.com/getkyo/kyo/issues/208 for the `collect` call but the handler is still not fully stack safe when there's nested effect handling. I've added a new test to track the issue.

[Benchmark results](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/8d3122c20a513b6de0b403d10df7a27d/raw/e36a985242db929c78c293cbdf31a5d2ad1ca622/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/8d3122c20a513b6de0b403d10df7a27d/raw/e36a985242db929c78c293cbdf31a5d2ad1ca622/jmh-result-candidate.json) show a regression for the two benchmarks that use these methods but I think it's acceptable given the benefits of the change. We might find ways to optimize these methods further and the performance in comparison to alternatives looks ok. On a side note: I hadn't noticed Kyo passed ZPure in the MtlBench! This benchmarks pulls it a little back, though 🙃

image


kyo-bench/src/main/scala/kyo/bench/MtlBench.scala
```diff
@@ -14,7 +14,7 @@ class MtlBench extends Bench:
     def syncKyo() =
         import kyo.*
         def testKyo: Unit < (Aborts[Throwable] & Envs[Env] & Vars[State] & Sums[Event]) =
-            Seqs.traverseUnit(loops)(_ =>
+            Seqs.foreach(loops)(_ =>
```
fwbrasil, line 17: Renamed for clarity.

kyo-bench/src/main/scala/kyo/bench/SchedulingBench.scala
```diff
@@ -41,10 +41,10 @@ class SchedulingBench extends Bench.ForkOnly[Int]:
                 }
             }
 
-        Seqs.traverse(range) { i =>
+        Seqs.map(range) { i =>
```
fwbrasil, line 44: ditto

kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -81,7 +81,7 @@ sealed abstract class Chunk[T] derives CanEqual:
         end if
     end concat
 
-    final def map[U: Flat, S](f: T => U < S): Chunk[U] < S =
```
fwbrasil, line 84: `Flat` wasn't necessary here. The `Loops` call needs a proof that `Chunk[U]` is flat, which is a given, not that `U` is flat.

kyo-core/shared/src/main/scala/kyo/loops.scala
```diff
@@ -72,7 +72,7 @@ object Loops:
                     loop(next.input1, next.input2)
                 case kyo: Kyo[Output | Continue2[Input1, Input2], S] @unchecked =>
                     kyo.map {
-                        case next: Continue2[Input1, Input2] =>
+                        case next: Continue2[Input1, Input2] @unchecked =>
```
fwbrasil, line 75: I'm noticing some non-determinism with the compiler. Sometimes unchanged code starts reporting a warning it. It seems related to inlining.

kyo-core/shared/src/main/scala/kyo/loops.scala
```diff
@@ -107,6 +107,27 @@ object Loops:
         loop(input1, input2, input3)
     end transform
 
+    inline def indexed[Output: Flat, S](
```
fwbrasil, line 110: Added an indexed loop without an input to be used in `Seq`.

kyo-core/shared/src/main/scala/kyo/seqs.scala
```diff
@@ -4,70 +4,118 @@ import kyo.core.*
...
+                            else f(seq(idx)).map(u => Loops.continue(acc.append(u)))
+                        }
+                    case seq: List[T] =>
+                        Loops.transform(seq, Chunks.init[U]) { (seq, acc) =>
+                            seq match
+                                case Nil          => Loops.done(acc.toSeq)
+                                case head :: tail => f(head).map(u => Loops.continue(tail, acc.append(u)))
+                        }
+                    case seq =>
+                        Chunks.initSeq(seq).map(f).map(_.toSeq)
```
fwbrasil, line 25: All methods follow a similar pattern specializing for small collections, indexed seqs, and lists. The "slow path" uses `Chunks`, which we can probably optimize further.

kyo-core/shared/src/main/scala/kyo/loops.scala
```diff
@@ -107,6 +107,27 @@ object Loops:
...
+        inline run: Int => Result[Unit, Output] < S
+    ): Output < S =
+        def _loop(idx: Int): Output < S =
+            loop(idx)
```
hearnadam, line 114: Why is this necessary?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/loops.scala
fwbrasil: On each iteration of a loop execution, it can either return a pure value, which triggers immediate recursion with `@tailrec`, or a suspended computation, which then suspends the loop via `map`. When the loop is suspended, it needs to start again once the computation resumes but it isn't a tail-recursive call, and it doesn't need to be since the effect suspension will unfold the stack. This method alias allows recursion in such cases while still keeping the `@tailrec` annotation for the fast path when there's no suspension.

hearnadam: Ah, thanks for the concise explanation. Very clever.

PR #289, opened 2024-04-26
add type inferred Aborts API

hearnadam:
- pass failure value for `Aborts.when` by name, potentially avoiding allocations
- add unit test for `Aborts.when`

I intentionally did not create an API for `run`/`catching` as it's unergonomic with type unions. I think followup work can be done there depending on how https://github.com/getkyo/kyo/issues/275 is resolved.

This massively improves the ux of Aborts in my opinion.


fwbrasil:
Makes sense. I think we can try to move to this encoding as part of #275

PR #290, opened 2024-04-26
loops review

fwbrasil:
I did a review of the codebase to locate loops that can leverage the new `Loops` effect, are missing `@tailrec`, or are stack unsafe.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/channels.scala
+++ b/kyo-core/shared/src/main/scala/kyo/channels.scala
@@ -120,14 +120,14 @@ object Channels:
                                     def dropTakes(): Unit < IOs =
                                         Loops.foreach {
                                             takes.poll() match
-                                                case null => Loops.doneUnit
-                                                case p    => p.interrupt.map(_ => Loops.continueUnit)
+                                                case null => Loops.done
+                                                case p    => p.interrupt.map(_ => Loops.continue)
                                         }
                                     def dropPuts(): Unit < IOs =
                                         Loops.foreach {
                                             puts.poll() match
-                                                case null   => Loops.doneUnit
-                                                case (_, p) => p.interrupt.map(_ => Loops.continueUnit)
+                                                case null   => Loops.done
+                                                case (_, p) => p.interrupt.map(_ => Loops.continue)
                                         }
                                     dropTakes()
                                         .andThen(dropPuts())
--- a/kyo-core/shared/src/main/scala/kyo/channels.scala
+++ b/kyo-core/shared/src/main/scala/kyo/channels.scala
@@ -118,15 +118,17 @@ object Channels:
                                     None
                                 case r: Some[Seq[T]] =>
                                     def dropTakes(): Unit < IOs =
-                                        takes.poll() match
-                                            case null => ()
-                                            case p =>
-                                                p.interrupt.map(_ => dropTakes())
+                                        Loops.foreach {
+                                            takes.poll() match
+                                                case null => Loops.doneUnit
+                                                case p    => p.interrupt.map(_ => Loops.continueUnit)
+                                        }
                                     def dropPuts(): Unit < IOs =
-                                        puts.poll() match
-                                            case null => ()
-                                            case (_, p) =>
-                                                p.interrupt.map(_ => dropPuts())
+                                        Loops.foreach {
+                                            puts.poll() match
+                                                case null   => Loops.doneUnit
+                                                case (_, p) => p.interrupt.map(_ => Loops.continueUnit)
+                                        }
                                     dropTakes()
                                         .andThen(dropPuts())
                                         .andThen(r)
--- a/kyo-core/shared/src/main/scala/kyo/hubs.scala
+++ b/kyo-core/shared/src/main/scala/kyo/hubs.scala
@@ -18,7 +18,7 @@ object Hubs:
                                     listeners.toArray
                                         .toList.asInstanceOf[List[Channel[T]]]
                                         .map(child => IOs.attempt(child.put(v)))
-                                Fibers.parallel(puts).map(_ => Loops.continueUnit)
+                                Fibers.parallel(puts).map(_ => Loops.continue)
                             }
                         }
                     }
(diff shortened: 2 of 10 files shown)


kyo-core/shared/src/main/scala/kyo/channels.scala
```diff
@@ -118,15 +118,17 @@ object Channels:
...
                                     def dropPuts(): Unit < IOs =
-                                        puts.poll() match
-                                            case null => ()
-                                            case (_, p) =>
-                                                p.interrupt.map(_ => dropPuts())
+                                        Loops.foreach {
+                                            puts.poll() match
+                                                case null   => Loops.doneUnit
+                                                case (_, p) => p.interrupt.map(_ => Loops.continueUnit)
+                                        }
```
fwbrasil, line 131: These cases were already stack safe since `interrupt` always suspends with `IOs` but it's best to be on the safe side since the code can change over time and `Loops` introduces little overhead.

kyo-core/shared/src/main/scala/kyo/hubs.scala
```diff
@@ -3,23 +3,25 @@ package kyo
...
+                                val puts =
                                     listeners.toArray
                                         .toList.asInstanceOf[List[Channel[T]]]
                                         .map(child => IOs.attempt(child.put(v)))
-                                Fibers.parallel(l).map(_ => loop())
+                                Fibers.parallel(puts).map(_ => Loops.continueUnit)
                             }
                         }
-                    loop()
+                    }
```
fwbrasil, line 24: ditto since `Fibers.parallel` suspends

kyo-core/shared/src/main/scala/kyo/hubs.scala
```diff
@@ -74,17 +77,16 @@ class Hub[T: Flat] private[kyo] (
     def close: Option[Seq[T]] < IOs =
         fiber.interrupt.map { _ =>
             ch.close.map { r =>
-                val it = listeners.iterator()
-                def loop(): Unit < IOs =
-                    IOs {
-                        if it.hasNext() then
-                            IOs.attempt(it.next().close)
-                                .map(_ => loop())
+                IOs {
```
fwbrasil, line 80: This `IOs` suspension was missing here since it deals with a mutable collection. I've not used `iterator` to avoid mutable state between the transformations.

kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -97,7 +97,7 @@ sealed trait IOs extends Effect[IOs]:
     end run
 
     def runLazy[T: Flat, S](v: T < (IOs & S)): T < S =
-        def runLazyLoop(v: T < (IOs & S)): T < S =
+        @tailrec def runLazyLoop(v: T < (IOs & S)): T < S =
```
fwbrasil, line 100: only missing the annotation

kyo-core/shared/src/main/scala/kyo/meters.scala
```diff
@@ -86,47 +86,45 @@ object Meters:
...
-                            case h +: t =>
-                                h.available.map(v => loop(t, acc + v))
-                    loop(meters, 0)
-                end available
+                    Loops.indexed(0) { (idx, acc) =>
+                        if idx == meters.length then Loops.done(acc)
+                        else meters(idx).available.map(v => Loops.continue(acc + v))
+                    }
 
                 def run[T, S](v: => T < S) =
```
fwbrasil, line 100: `run` and `tryRun` aren't stack safe because it's necessary to produce a thunk representing the execution of all other nested meters to handle permits correctly. That's not problematic unless for some reason there's a pipeline with hundreds of meters, which doesn't seem to make sense.

kyo-core/shared/src/main/scala/kyo/queues.scala
```diff
@@ -43,15 +43,15 @@ object Queues:
         def peek(): T
 
         def drain(): Seq[T] =
-            def loop(acc: List[T]): List[T] =
+            val b = Seq.newBuilder[T]
+            @tailrec def loop(): Unit =
```
fwbrasil, line 47: No effects here so a regular loop + builder is fine.

kyo-core/shared/src/main/scala/kyo/meters.scala
```diff
@@ -86,47 +86,45 @@ object Meters:
...
-                        l match
-                            case Seq() => v.map(Some(_))
-                            case h +: t =>
-                                h.tryRun(loop(t)).map {
-                                    case None => None
-                                    case r    => r.flatten: Option[T]
-                                }
-                    loop(meters)
+                    def loop(idx: Int = 0): Option[T] < (S & IOs) =
+                        if idx == meters.length then v.map(Some(_))
```
hearnadam, line 109: I wonder if there is an allocation free strategy for `Option` that we can use within or publish for Kyo. I don't really love the semantics of `A | Null`...

kyo-core/shared/src/main/scala/kyo/retries.scala
```diff
@@ -28,18 +28,18 @@ object Retries:
...
                         }
                     else
                         IOs.fail(ex)
-                case Success(value) =>
-                    value
+                case Success(value: T) =>
+                    Loops.done(value)
             }
-        loop(0)
+        }
```
hearnadam, line 43: Much improved!

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/meters.scala
fwbrasil: In a sense, T < Options is already that, especially now that Aborts doesn’t allocate a handler. I don’t think it’s a good API, though. It feels like in a case like this one, short circuiting is a first-class concern the caller of the API needs to typically immediately handle. In these cases, I think effects should be avoided to make the code simpler. In terms of performance, JIT compilers can normally avoid allocations unless the code is highly dynamic, which isn’t typically the case for Kyo computations.

kyo-core/shared/src/main/scala/kyo/retries.scala
fwbrasil: I’ve been a bit surprised by how well the code reads with a more explicit API like Loops. I’m almost preferring it to regular recursive methods. This might be a good direction to evolve the library: more explicit APIs that don’t incur into runtime cost like in Result, which seems helpful especially to newcomers.

PR #291, opened 2024-04-26
Interop Kyo Fiber & ZIO

hearnadam:
- Ensure interruptions work correctly
- migrate KyoSepcDefault

### DIFF
--- a/kyo-test/shared/src/main/scala/kyo/test/interop/zio.scala
+++ b/kyo-test/shared/src/main/scala/kyo/test/interop/zio.scala
@@ -1,0 +1,15 @@
+package kyo.test.interop
+
+import kyo.*
+import zio.Trace
+import zio.ZIO
+
+extension (capture: ZIO.type)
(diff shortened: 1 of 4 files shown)


kyo-test/shared/src/main/scala/kyo/test/interop/zio.scala
```diff
@@ -0,0 +1,15 @@
+package kyo.test.interop
+
+import kyo.*
+import zio.Trace
+import zio.ZIO
+
+extension (capture: ZIO.type)
+    private[kyo] def fromKyoFiber[A](fiber: Fiber[A])(using Flat[A], Trace): ZIO[Any, Throwable, A] =
+        ZIO.asyncInterrupt[Any, Throwable, A] { cb =>
+            val async: ZIO[Any, Throwable, A] = ZIO.fromFuture(_ => IOs.run(fiber.toFuture))
```
hearnadam, line 10: We could probably replace `fromFuture`, but I'm not too worried about the performance or cost given this should only be used in tests.

PR #292, opened 2024-04-26
Avoid Tags in Aborts (fixes #275)

### LINKED ISSUE #275: Aborts effect is very slow with its use of Izumi

fwbrasil:
As discussed in #275, this PR changes `Aborts` to rely solely on regular reflection of aborted values instead of type tags. It's an approach similar to ZIO. The [benchmark results](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/3964a7ff775d855268b809cfee7fab84/raw/31e8b5b5b1dab3f70ec6aee60ef12873ba68dd1b/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/3964a7ff775d855268b809cfee7fab84/raw/31e8b5b5b1dab3f70ec6aee60ef12873ba68dd1b/jmh-result-candidate.json) confirm the fix of the bottleneck very well 🤣

image

### DIFF
--- a/kyo-core/jvm/src/test/scala/kyoTest/coreBytecodeSizeTest.scala
+++ b/kyo-core/jvm/src/test/scala/kyoTest/coreBytecodeSizeTest.scala
@@ -38,7 +38,7 @@ class coreBytecodeSizeTest extends KyoTest:
         assert(map == Map(
             "test"        -> 28,
             "resultLoop"  -> 112,
-            "handleLoop"  -> 264,
+            "handleLoop"  -> 294,
             "_handleLoop" -> 12
         ))
     }
--- a/kyo-core/shared/src/main/scala/kyo/aborts.scala
+++ b/kyo-core/shared/src/main/scala/kyo/aborts.scala
@@ -4,95 +4,97 @@ import kyo.core.*
 import scala.reflect.ClassTag
 import scala.util.control.NonFatal
 
-class Aborts[-V] extends Effect[Aborts[V]]:
-    opaque type Command[T] >: Left[V, Nothing] = Left[V, Nothing]
+// Can't use an opaque type because
+// kyo-direct tests crash the compiler.
+type Aborts[-V] //= DoAbort
 
 object Aborts:
-    private case object aborts extends Aborts[Any]
-    def apply[V]: Aborts[V] = aborts.asInstanceOf[Aborts[V]]
 
-    def fail[V](value: V)(using t: Tag[Aborts[V]]): Nothing < Aborts[V] =
-        Aborts[V].fail(value)
+    import internal.*
 
-    def when[V](b: Boolean)(value: => V)(using Tag[Aborts[V]]): Unit < Aborts[V] =
+    def fail[V](v: V): Nothing < Aborts[V] =
+        DoAbort.suspend(v).asInstanceOf[Nothing < Aborts[V]]
+
+    def when[V](b: Boolean)(value: => V): Unit < Aborts[V] =
         if b then fail(value)
         else ()
 
-    def get[V, T](e: Either[V, T])(using Tag[Aborts[V]]): T < Aborts[V] =
-        Aborts[V].get(e)
+    def get[V, T](e: Either[V, T]): T < Aborts[V] =
+        e match
+            case Right(v) => v
+            case Left(v)  => DoAbort.suspend(v).asInstanceOf[Nothing < Aborts[V]]
 
-    extension [V](self: Aborts[V])
-
-        def fail[T <: V](value: T)(using t: Tag[Aborts[T]]): Nothing < Aborts[T] =
-            self.suspend[Nothing](Left(value))
-
-        def when(b: Boolean)(value: => V)(using Tag[Aborts[V]]): Unit < Aborts[V] =
-            if b then fail(value)
-            else ()
-
-        def get[T](e: Either[V, T])(using Tag[Aborts[V]]): T < Aborts[V] =
-            e match
-                case Right(value) => value
-                case Left(e)      => fail(e)
-
-        def catching[T, S](v: => T < S)(
+    class RunDsl[V]():
+        def apply[T: Flat, S, VS, VR](v: T < (Aborts[VS] & S))(
             using
-            ClassTag[V],
-            Tag[Aborts[V]]
+            h: HasAborts[V, VS] { type Remainder = VR },
+            ct: ClassTag[V]
+        ): Either[V, T] < (VR & S) =
+            DoAbort.handle(handler)(ct, v).asInstanceOf[Either[V, T] < (VR & S)]
+    end RunDsl
+
+    def run[V]: RunDsl[V] = RunDsl()
+
+    class CatchingDsl[V]():
+        def apply[T: Flat, S](v: => T < S)(
+            using ct: ClassTag[V]
         ): T < (Aborts[V] & S) =
-            IOs.handle(v) {
-                case ex: V =>
-                    fail(ex.asInstanceOf[V])
+            IOs.catching(v) {
+                case ex: V => Aborts.fail(ex)
             }
+    end CatchingDsl
 
-        def run[T: Flat, S, VS, VR](v: T < (Aborts[VS] & S))(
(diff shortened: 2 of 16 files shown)


kyo-core/jvm/src/test/scala/kyoTest/coreBytecodeSizeTest.scala
```diff
@@ -38,7 +38,7 @@ class coreBytecodeSizeTest extends KyoTest:
         assert(map == Map(
             "test"        -> 28,
             "resultLoop"  -> 112,
-            "handleLoop"  -> 264,
+            "handleLoop"  -> 294,
```
fwbrasil, line 41: Due to the changes in `handler.accepts`.

kyo-core/shared/src/main/scala/kyo/aborts.scala
```diff
@@ -4,95 +4,90 @@ import kyo.core.*
 import scala.reflect.ClassTag
 import scala.util.control.NonFatal
 
-class Aborts[-V] extends Effect[Aborts[V]]:
-    opaque type Command[T] >: Left[V, Nothing] = Left[V, Nothing]
+// Can't use an opaque type because
+// kyo-direct tests crash the compiler.
```
fwbrasil, line 8: I'll create a separate issue for this. I'm not sure why the compiler crashes when some effects are marked as `opaque`. It's an important limitation we should address.

kyo-core/shared/src/main/scala/kyo/aborts.scala
```diff
@@ -4,95 +4,90 @@ import kyo.core.*
...
-            self.suspend[Nothing](Left(value))
+    def get[V, T](e: Either[V, T]): T < Aborts[V] =
+        e match
+            case Right(v) => v
+            case Left(v)  => DoAbort.suspend(v).asInstanceOf[Nothing < Aborts[V]]
 
-        def when(b: Boolean)(value: => V)(using Tag[Aborts[V]]): Unit < Aborts[V] =
-            if b then fail(value)
-            else ()
+    class RunDsl[V]():
```
fwbrasil, line 27: I'm not a big fan of this approach with DSL classes but I couldn't find a better way. There's a new feature for clause interleaving in Scala 3 that might enable defining a single method but it's still experimental. Suggestions are appreciated!

kyo-core/shared/src/main/scala/kyo/aborts.scala
```diff
@@ -4,95 +4,90 @@ import kyo.core.*
...
+    def catching[V]: CatchingDsl[V] = CatchingDsl()
 
-        private def handler(using ClassTag[V], Tag[Aborts[V]]) =
-            new ResultHandler[Unit, self.Command, Aborts[V], [T] =>> Either[V, T], Any]:
-                def done[T](st: Unit, v: T) = Right(v)
+    private object internal:
 
-                override inline def accepts[T, U](self: Tag[T], other: Tag[U]): Boolean =
-                    self == other || self.parse <:< other.parse
+        val handler =
```
fwbrasil, line 51: This change also ended up saving the handler allocation! The `ClassTag` is passed as the state of the handler.

kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -59,7 +59,7 @@ sealed trait IOs extends Effect[IOs]:
         end try
     end attempt
 
-    def handle[T, S, U >: T, S2](v: => T < S)(
+    def catching[T, S, U >: T, S2](v: => T < S)(
```
fwbrasil, line 62: Renamed to be more consistent with `Aborts.catching`. I think `handle` can also give an impression that the effect itself is getting handled.

kyo-core/shared/src/main/scala/kyo/aborts.scala
```diff
@@ -4,95 +4,97 @@ import kyo.core.*
...
-        Aborts[V].get(e)
-
-    extension [V](self: Aborts[V])
-
-        def fail[T <: V](value: T)(using t: Tag[Aborts[T]]): Nothing < Aborts[T] =
-            self.suspend[Nothing](Left(value))
+    def get[V, T](e: Either[V, T]): T < Aborts[V] =
+        e match
+            case Right(v) => v
+            case Left(v)  => DoAbort.suspend(v).asInstanceOf[Nothing < Aborts[V]]
```
hearnadam, line 25: Isn't this equivalent to `fail`?

kyo-core/shared/src/main/scala/kyo/aborts.scala
hearnadam: minor:
- unnecessary parenthesis on the class
- Would prefer the name `RunPartiallyApplied` as it can more clearly communicate why this class is like this


I think in Scala 3 you can supply partial type parameter lists. Can you test that? The `V` should come first so it can be specified without the others. Either way I don't think this class is a big deal.

kyo-core/shared/src/main/scala/kyo/aborts.scala
```diff
@@ -4,95 +4,97 @@ import kyo.core.*
...
+                    command match
+                        case v: V => true
+                        case _    => false
+                end accepts
+
+                def resume[T, U: Flat, S2](st: ClassTag[?], command: Any, k: T => U < (DoAbort & S2)) =
+                    Left(command)
+        end handler
+
+        class DoAbort extends Effect[DoAbort]:
```
hearnadam, line 75: How about just calling this `Abort`

kyo-examples/jvm/src/main/scala/kyo/examples/ledger/api/Handler.scala
```diff
@@ -26,8 +26,8 @@ object Handler:
 
     final class Live(db: DB) extends Handler:
 
-        private val notFound            = Aborts[StatusCode].fail(StatusCode.NotFound)
-        private val unprocessableEntity = Aborts[StatusCode].fail(StatusCode.UnprocessableEntity)
+        private val notFound            = Aborts.fail[StatusCode](StatusCode.NotFound)
```
hearnadam, line 29: Can you remove the type and get proper inference here?

kyo-core/shared/src/main/scala/kyo/aborts.scala
hearnadam: Was there an exception you can share?

kyo-core/shared/src/main/scala/kyo/aborts.scala
hearnadam: After testing locally, I think partial application is best. Do we need to use a dummy+AnyVal to avoid allocations?

```
    class RunPartiallyApplied[V](private val dummy: Boolean = false) extends AnyVal:
        def apply[T: Flat, S, VS, VR](v: T < (Aborts[VS] & S))(
            using
            h: HasAborts[V, VS] { type Remainder = VR },
            ct: ClassTag[V]
        ): Either[V, T] < (VR & S) =
            Abort.handle(handler)(ct, v).asInstanceOf[Either[V, T] < (VR & S)]
    end RunPartiallyApplied
```

hearnadam:
Merging PR so I can start working on some more refinements.

PR #295, opened 2024-04-28
Improve Aborts handling of sub/super types

hearnadam:
Followup to https://github.com/getkyo/kyo/pull/292

### DIFF
--- a/kyo-core/shared/src/test/scala/kyoTest/abortsTest.scala
+++ b/kyo-core/shared/src/test/scala/kyoTest/abortsTest.scala
@@ -181,13 +181,16 @@ class abortsTest extends KyoTest:
                     t3(42)
                 succeed
             }
-            "super" in pendingUntilFixed {
-                assertCompiles("""
-                    val ex                          = new Exception
-                    val a: Int < Aborts[Exception]  = Aborts.fail(ex)
-                    val b: Either[Throwable, Int]   = Aborts.run[Throwable](a).pure
-                    assert(b == Left(ex))
-                """)
+            "super" in {
+                val ex                         = new Exception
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/test/scala/kyoTest/abortsTest.scala
```diff
@@ -255,6 +263,32 @@ class abortsTest extends KyoTest:
...
+                    val a = Aborts.run[Distinct1](distinct)
+                    val b = Aborts.run[Distinct2](a).pure
+                    assert(b == Right(Left(d1)))
+
+                    val c = Aborts.run(distinct).pure
+                    assert(c == Left(d1))
+                }
+                "ClassTag inference" in pendingUntilFixed {
+                    assertCompiles("""
+                        val r = Aborts.run(Aborts.catching(throw new RuntimeException)).pure
```
fwbrasil, line 287: What happens here? Does the `ClassTag` get inferred to `Nothing`/`Object`?

fwbrasil (approved):

kyo-core/shared/src/test/scala/kyoTest/abortsTest.scala
hearnadam:
```
[error] -- [E172] Type Error:
[error] 286 |                    val r = Aborts.run(Aborts.catching(throw new RuntimeException)).pure
[error]     |                                                                                  ^
[error]     |           No ClassTag available for V
[error]     |
[error]     |           where:    V is a type variable with constraint <: Throwable
[error] one error found
[error] (Test / compileIncremental) Compilation failed
```

PR #296, opened 2024-04-29
Initial Duration module

hearnadam:
Needs tests + some minor reworking, but should be usable.

Improves compatibility with ZIO interop and general API interop

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,7 +368,6 @@ While the companion object of `KyoApp` provides utility methods to run isolated 
 
 ```scala
 import kyo._
-import scala.concurrent.duration._
 
 // An example computation
 val a: Int < IOs =
@@ -1368,17 +1367,16 @@ The `sleep` and `timeout` methods pause a computation or time it out after a dur
 
 ```scala
 import kyo._
-import scala.concurrent.duration._
 
 // A computation that sleeps for 1s
 val a: Unit < Fibers =
-  Fibers.sleep(1.second)
+  Fibers.sleep(1.seconds)
 
 // Times out and interrupts the provided 
 // computation in case it doesn't produce 
 // a result within 1s
 val b: Int < Fibers =
-  Fibers.timeout(1.second)(Math.cos(42).toInt)
+  Fibers.timeout(1.seconds)(Math.cos(42).toInt)
 ```
 
 The `fromFuture` methods provide interoperability with Scala's `Future`.
@@ -1451,7 +1449,6 @@ Similarly to `IOs`, users should avoid handling the `Fibers` effect directly and
 
 ```scala
 import kyo._
-import scala.concurrent.duration._
 
 // An example computation with fibers
 val a: Int < Fibers =
@@ -1735,7 +1732,6 @@ The `Meters` effect offers utilities to regulate computational execution, be it 
 
 ```scala
 import kyo._
-import scala.concurrent.duration._
 
 // 'mutex': One computation at a time
 val a: Meter < IOs = 
@@ -1749,7 +1745,7 @@ val b: Meter < IOs =
 val c: Meter < IOs =
   Meters.initRateLimiter(
     rate = 10, 
-    period = 1.second
+    period = 1.seconds
   )
 
 // 'pipeline': Combine multiple 'Meter's
@@ -1791,7 +1787,6 @@ The `Timers` effect is designed for control over the timing of task execution.
 
 ```scala
 import kyo._
-import scala.concurrent.duration._
 
 // An example computation to
 // be scheduled
@@ -1800,34 +1795,34 @@ val a: Unit < IOs =
 
 // Schedule a delayed task
 val b: TimerTask < IOs =
-  Timers.schedule(delay = 1.second)(a)
+  Timers.schedule(delay = 1.seconds)(a)
 
 // Recurring task with
 // intial delay
 val c: TimerTask < IOs =
   Timers.scheduleAtFixedRate(
-    initialDelay = 1.minute,
-    period = 1.minute
+    initialDelay = 1.minutes,
+    period = 1.minutes
   )(a)
 
 // Recurring task without
 // initial delay
 val d: TimerTask < IOs =
   Timers.scheduleAtFixedRate(
-    period = 1.minute
+    period = 1.minutes
   )(a)
 
 // Schedule with fixed delay between tasks
 val e: TimerTask < IOs =
   Timers.scheduleWithFixedDelay(
-    initialDelay = 1.minute,
-    period = 1.minute
+    initialDelay = 1.minutes,
+    period = 1.minutes
   )(a)
 
 // without initial delay
 val f: TimerTask < IOs =
   Timers.scheduleWithFixedDelay(
-    period = 1.minute
+    period = 1.minutes
   )(a)
 
 // Specify the 'Timer' explictly
@@ -1839,11 +1834,10 @@ val i: TimerTask < IOs =
 
 ```scala
 import kyo._
-import scala.concurrent.duration._
 
 // Example TimerTask
 val a: TimerTask < IOs = 
-  Timers.schedule(1.second)(())
+  Timers.schedule(1.seconds)(())
 
 // Try to cancel the task
 val b: Boolean < IOs =
--- a/kyo-core/shared/src/main/scala/kyo/Duration.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Duration.scala
@@ -4,24 +4,25 @@ import java.time.Duration as JavaDuration
 import java.time.temporal.ChronoUnit.*
 import kyo.Duration.Units
 import kyo.Duration.Units.*
+import kyo.duration.*
 import scala.concurrent.duration.Duration as ScalaDuration
 
-opaque type Duration = Long
-given CanEqual[Duration, Duration] = Duration.canEqual
+type Duration = duration.Duration
+given CanEqual[Duration, Duration] = canEqual
 
 object Duration:
-    val Zero: Duration     = 0
-    val Infinity: Duration = Long.MaxValue
+    val Zero: Duration     = _Zero
+    val Infinity: Duration = _Infinity
 
-    inline def fromNanos(value: Long): Duration = value
+    inline def fromNanos(value: Long): Duration = _fromNanos(value)
     def fromJava(value: JavaDuration): Duration =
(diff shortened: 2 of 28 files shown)


kyo-core/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -0,0 +1,74 @@
...
+
+object Duration:
+    val Zero: Duration                   = JDuration.ZERO
+    val Infinity: Duration               = Long.MaxValue.nanos
+    def fromNanos(value: Long): Duration = value.nanos
+    def from(value: JDuration): Duration =
+        (value.nanos: @switch) match
+            case 0                        => Zero
+            case n if n >= Infinity.nanos => Infinity
+            case _                        => value
```
hearnadam, line 20: Should we reallocate here? Are Java Durations mutable?

kyo-core/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -0,0 +1,74 @@
+package kyo
+
+import java.time.Duration as JDuration
+import java.time.temporal.ChronoUnit.*
+import java.time.temporal.TemporalUnit
+import scala.annotation.switch
+import scala.concurrent.duration.Duration as SDuration
+
+opaque type Duration <: JDuration = JDuration
+given CanEqual[Duration, Duration] = CanEqual.derived
```
hearnadam, line 10: Can we define this manually?

kyo-core/shared/src/main/scala/kyo/Duration.scala
fwbrasil: The javadoc mentions it being immutable and thread safe:

```
 * @implSpec
 * This class is immutable and thread-safe.
 *
 * @since 1.8
 */
@jdk.internal.ValueBased
public final class Duration
```

kyo-core/shared/src/main/scala/kyo/Duration.scala
fwbrasil: You'd just to implement an instance of `CanEqual` here I think. Something important about Scala 3's encoding of `given` is that it introduces lazy initialization under the hood, which requires a memory barrier on each read and prevents JIT optimizations in some cases. I haven't migrated the codebase yet but this is the encoding I've been using to avoid the overhead:

```
class SomeType

private val someValue = new SomeType
inline given SomeType = someValue
```

kyo-core/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -0,0 +1,74 @@
+package kyo
+
+import java.time.Duration as JDuration
+import java.time.temporal.ChronoUnit.*
+import java.time.temporal.TemporalUnit
+import scala.annotation.switch
+import scala.concurrent.duration.Duration as SDuration
+
+opaque type Duration <: JDuration = JDuration
```
fwbrasil, line 9: Do you think we could use `opaque type Duration = Long`? It'd be nice to avoid the allocation of Java's wrapper.

kyo-core/shared/src/main/scala/kyo/Duration.scala
hearnadam: Awesome!

kyo-core/shared/src/main/scala/kyo/Duration.scala
hearnadam: It's possible, but my goal was to directly 'replace' Java Duration. Can we make it Java duration compatible - ie, a value class that extends `Duration`?

fwbrasil: I don't think so, value classes can only extend "universal traits": https://docs.scala-lang.org/overviews/core/value-classes.html.

Maybe a `toJava` would be enough to provide interop?

hearnadam (commented):
@fwbrasil please take a look. Needs some minor cleanup and better tests, but I think this is pretty awesome. Should be fully allocation free.

README.md
```diff
@@ -368,7 +368,6 @@ While the companion object of `KyoApp` provides utility methods to run isolated
 
 ```scala
 import kyo._
-import scala.concurrent.duration._
```
fwbrasil, line 371: Not requiring this import is a nice improvement to!

kyo-core/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -1,74 +1,108 @@
...
+        (value.toNanos: @annotation.switch) match
+            case 0                  => Zero
+            case n if n >= Infinity => Infinity
+            case n                  => n.nanos
+
+    def fromScala(value: ScalaDuration): Duration =
+        if value.isFinite then Math.max(value.toNanos.nanos, Zero) else Infinity
+
+    enum Units(val factor: Double):
+        case Nanos   extends Units(NANOS.getDuration.toNanos.toDouble)
```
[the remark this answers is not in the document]
fwbrasil: I think it's better this way than hardcoding values!

kyo-core/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -0,0 +1,136 @@
...
+    inline def toSeconds: Long = self.to(Seconds)
+    inline def toMinutes: Long = self.to(Minutes)
+    inline def toHours: Long   = self.to(Hours)
+    inline def toDays: Long    = self.to(Days)
+    inline def toWeeks: Long   = self.to(Weeks)
+    inline def toMonths: Long  = self.to(Months)
+    inline def toYears: Long   = self.to(Years)
+
+    def toScala: ScalaDuration =
+        (self: @annotation.switch) match
```
fwbrasil, line 81: cool, I didn't know this annotation!

hearnadam: it directly compiles `match` to a java switch. It's a good way to ensure performance, even if the compiler should do the right thing without it. It should fail compilation where the compilation is not possible.

kyo-core/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -0,0 +1,137 @@
...
+    def toJava: JavaDuration =
+        (self: @annotation.switch) match
+            case Duration.Zero => JavaDuration.ZERO
+            case n             => JavaDuration.of(n.toNanos, NANOS)
+
+    // Is this Robust enough?
+    private[kyo] inline def isFinite: Boolean = self < Duration.Infinity
+end extension
+
+private[kyo] object duration:
```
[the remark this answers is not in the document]
fwbrasil: What part of the implementation did you want to avoid exposing here? Does anything fail if you move the opaque type to line 10?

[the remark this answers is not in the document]
hearnadam: The issue is that internal aspects of Kyo's code base become `Duration` (Longs, Chars, etc) meaning the functions like `to` are tied to Duration:

This line will cause a failure to compile
https://github.com/getkyo/kyo/blob/main/kyo-core/shared/src/main/scala/kyo/randoms.scala#L54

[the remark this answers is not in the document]
fwbrasil: That's odd, I think the opaque type should be visible only in the same file.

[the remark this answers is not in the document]
fwbrasil: oh I see, the issue is the extension methods. Wouldn't that be a problem for users as well?

kyo-core/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -0,0 +1,136 @@
...
+
+import java.time.Duration as JavaDuration
+import java.time.temporal.ChronoUnit.*
+import kyo.Duration.Units
+import kyo.Duration.Units.*
+import kyo.duration.*
+import scala.concurrent.duration.Duration as ScalaDuration
+
+type Duration = duration.Duration
+given CanEqual[Duration, Duration] = canEqual
```
fwbrasil, line 11: Can you move this to the companion object?

kyo-core/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -0,0 +1,136 @@
...
+            else if factor <= Long.MaxValue / self.toDouble then Math.round(self.toDouble * factor)
+            else Duration.Infinity
+        end multiply
+
+        inline def gtEq(that: Duration): Boolean = self >= that
+        inline def ltEq(that: Duration): Boolean = self <= that
+        inline def gt(that: Duration): Boolean   = self > that
+        inline def lt(that: Duration): Boolean   = self < that
+        inline def eqEq(that: Duration): Boolean = self == that
+        inline def neEq(that: Duration): Boolean = self == that
```
fwbrasil, line 131: `!=`

hearnadam: Great catch

kyo-core/shared/src/main/scala/kyo/Duration.scala
hearnadam: Will do

hearnadam: I don't think it's possible. The type is necessary to be in the top level namespace for the extension methods.

kyo-test/shared/src/test/scala/kyo/test/DurationSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+                    TestResult.allSuccesses(
+                        assertTrue(i.nanos.toNanos == i),
+                        assertTrue(i.micros.toMicros == i),
+                        assertTrue(i.millis.toMillis == i),
+                        assertTrue(i.seconds.toSeconds == i),
+                        assertTrue(i.minutes.toMinutes == i),
+                        assertTrue(i.hours.toHours == i)
+//                        assertTrue(i.days.toDays == i),
+//                        assertTrue(i.weeks.toWeeks == i),
+//                        assertTrue(i.months.toMonths == i)
```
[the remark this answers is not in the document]
hearnadam: Let's proceed with the current module, we can get some feedback from users on if they expect exceptions for overflow. I'm concerned with the safety of random exceptions - but also not excited about silent overflow.

PR #297, opened 2024-04-30
Aborts-like encoding for Sums

fwbrasil:
Follows the new `Aborts` approach favoring companion object methods.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/sums.scala
+++ b/kyo-core/shared/src/main/scala/kyo/sums.scala
@@ -1,25 +1,35 @@
 package kyo
 
 import kyo.core.*
-
-object Sums:
-    private object sums extends Sums[Any]
-    def apply[V]: Sums[V] = sums.asInstanceOf[Sums[V]]
+import scala.annotation.implicitNotFound
+import scala.util.NotGiven
 
 class Sums[V] extends Effect[Sums[V]]:
-    opaque type Command[T] = V
-
-    def add(v: V)(using Tag[Sums[V]]): Unit < Sums[V] =
-        this.suspend[Unit](v)
-
-    def run[T: Flat, S](v: T < (Sums[V] & S))(
-        using Tag[Sums[V]]
-    ): (Chunk[V], T) < S =
-        this.handle(handler)(Chunks.init, v)
+    type Command[T] = V
 
     private val handler =
         new ResultHandler[Chunk[V], Const[V], Sums[V], [T] =>> (Chunk[V], T), Any]:
             def done[T](st: Chunk[V], v: T) = (st, v)
             def resume[T, U: Flat, S](st: Chunk[V], command: V, k: T => U < (Sums[V] & S)) =
                 Resume(st.append(command), k(().asInstanceOf[T]))
+end Sums
+
+object Sums:
+    private object sums extends Sums[Any]
+    private def sums[V]: Sums[V] = sums.asInstanceOf[Sums[V]]
+
+    def add[V](v: V)(using Tag[Sums[V]]): Unit < Sums[V] =
+        sums[V].suspend[Unit](v)
+
+    class RunDsl[V]:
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/sums.scala
```diff
@@ -1,25 +1,35 @@
...
+
+    class RunDsl[V]:
+        def apply[T: Flat, S](v: T < (Sums[V] & S))(
+            using Tag[Sums[V]]
+        ): (Chunk[V], T) < S =
+            sums[V].handle(sums[V].handler)(Chunks.init, v)
+    end RunDsl
+
+    def run[V](using
+        @implicitNotFound("Please specify the type to be handled. Example: 'Sums.run[Int](...)'") ng: NotGiven[V => Nothing]
```
fwbrasil, line 32: This effect is a bit different since its type parameter is invariant. I had to include this restriction otherwise Scala infers `Nothing` as `V` when it's not specified.

hearnadam: Can you try `V >: Nothing`?

fwbrasil: Nice! I've pushed this change and it made Scala able to infer the type correctly when there's only one sum pending 👏

kyo-core/shared/src/main/scala/kyo/sums.scala
```diff
@@ -1,25 +1,33 @@
 package kyo
 
 import kyo.core.*
-
-object Sums:
-    private object sums extends Sums[Any]
-    def apply[V]: Sums[V] = sums.asInstanceOf[Sums[V]]
+import scala.annotation.implicitNotFound
+import scala.util.NotGiven
```
hearnadam, line 5: unused

kyo-core/shared/src/main/scala/kyo/sums.scala
```diff
@@ -1,25 +1,33 @@
...
             def resume[T, U: Flat, S](st: Chunk[V], command: V, k: T => U < (Sums[V] & S)) =
                 Resume(st.append(command), k(().asInstanceOf[T]))
 end Sums
+
+object Sums:
+    private object sums extends Sums[Any]
+    private def sums[V]: Sums[V] = sums.asInstanceOf[Sums[V]]
+
+    def add[V](v: V)(using Tag[Sums[V]]): Unit < Sums[V] =
+        sums[V].suspend[Unit](v)
```
hearnadam, line 22: Just to be clear, though this is a polymorphic call site, the handler should only be allocated once, right?

Does this have any performance benefit?
```
private def sumsHandler[V]: Sums[V] = sums.handler.asInstanceOf[..]
```

kyo-core/shared/src/test/scala/kyoTest/sumsTest.scala
```diff
@@ -7,65 +7,66 @@ class sumsTest extends KyoTest:
...
             for
-                _ <- Sums[Int].add(3)
-                _ <- Sums[String].add("1")
-                _ <- Sums[Int].add(2)
-                _ <- Sums[String].add("2")
-                _ <- Sums[Int].add(1)
-                _ <- Sums[String].add("3")
+                _ <- Sums.add(3)
+                _ <- Sums.add("1")
+                _ <- Sums.add(2)
```
hearnadam, line 32: Much cleaner!

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/sums.scala
hearnadam: ah I realize now that you are not invoking the handler, but the method on the effect. It's fine as is.

kyo-core/shared/src/main/scala/kyo/sums.scala
fwbrasil: fixed

PR #298, opened 2024-04-30
Aborts-like encoding for Envs

fwbrasil:
Follows the new Aborts approach favoring companion object methods.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/envs.scala
+++ b/kyo-core/shared/src/main/scala/kyo/envs.scala
@@ -1,32 +1,36 @@
 package kyo
 
 import kyo.core.*
+import scala.util.NotGiven
 
 class Envs[+V] extends Effect[Envs[V]]:
     type Command[T] = Unit
 
 object Envs:
     private case object envs extends Envs[Any]
-    def apply[V]: Envs[V] = envs.asInstanceOf[Envs[V]]
+    private def envs[V]: Envs[V] = envs.asInstanceOf[Envs[V]]
 
-    extension [V](self: Envs[V])
+    def get[V](using Tag[Envs[V]]): V < Envs[V] =
+        envs[V].suspend[V](())
 
-        def get(using Tag[Envs[V]]): V < Envs[V] =
-            self.suspend[V](())
-
-        inline def use[T, S](inline f: V => T < S)(
+    class UseDsl[V]:
+        inline def apply[T, S](inline f: V => T < S)(
             using inline tag: Tag[Envs[V]]
         ): T < (Envs[V] & S) =
-            self.suspend[V, T, S]((), f)
+            envs[V].suspend[V, T, S]((), f)
+    end UseDsl
 
-        def run[T: Flat, S, VS, VR](env: V)(value: T < (Envs[VS] & S))(
-            using
-            HasEnvs[V, VS] { type Remainder = VR },
-            Tag[Envs[V]]
-        ): T < (S & VR) =
-            Envs[V].handle(handler[V])(env, value).asInstanceOf[T < (S & VR)]
+    def use[V](
+        using ng: NotGiven[V => Nothing]
+    ): UseDsl[V] =
+        new UseDsl[V]
 
-    end extension
+    def run[V, T: Flat, S, VS, VR](env: V)(value: T < (Envs[VS] & S))(
+        using
+        HasEnvs[V, VS] { type Remainder = VR },
+        Tag[Envs[V]]
+    ): T < (S & VR) =
+        Envs[V].handle(handler[V])(env, value).asInstanceOf[T < (S & VR)]
 
     private def handler[V]: ResultHandler[V, Const[Unit], Envs[V], Id, Any] =
         cachedHandler.asInstanceOf[ResultHandler[V, Const[Unit], Envs[V], Id, Any]]
--- a/kyo-core/shared/src/main/scala/kyo/envs.scala
+++ b/kyo-core/shared/src/main/scala/kyo/envs.scala
@@ -1,7 +1,6 @@
 package kyo
 
 import kyo.core.*
-import scala.util.NotGiven
 
 class Envs[+V] extends Effect[Envs[V]]:
     type Command[T] = Unit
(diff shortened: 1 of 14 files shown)


kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -1,32 +1,36 @@
...
+    end UseDsl
 
-        def run[T: Flat, S, VS, VR](env: V)(value: T < (Envs[VS] & S))(
-            using
-            HasEnvs[V, VS] { type Remainder = VR },
-            Tag[Envs[V]]
-        ): T < (S & VR) =
-            Envs[V].handle(handler[V])(env, value).asInstanceOf[T < (S & VR)]
+    def use[V](
+        using ng: NotGiven[V => Nothing]
```
hearnadam, line 24: Same comment as `Sums`, you should be able to fix this with a type bound: `def use[V >: Nothing]`

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -1,32 +1,36 @@
...
+    ): UseDsl[V] =
+        new UseDsl[V]
 
-    end extension
+    def run[V, T: Flat, S, VS, VR](env: V)(value: T < (Envs[VS] & S))(
```
hearnadam, line 28: I suspect we may need to improve manually providing type `V` in the case where `env` needs to be a specific type (think type that extends multiple traits) - though don't have any way to confirm that. Perhaps worth proactively adding a DSL.

kyo-core/shared/src/test/scala/kyoTest/abortsTest.scala
```diff
@@ -298,16 +298,16 @@ class abortsTest extends KyoTest:
...
                         ).pure ==
                             Right(5)
                     )
                 }
                 "failure" in {
                     assert(
-                        Envs[Int].run(0)(
-                            Aborts.run[Ex1](Aborts.catching[Ex1](test(Envs[Int].get)))
+                        Envs.run(0)(
+                            Aborts.run[Ex1](Aborts.catching[Ex1](test(Envs.get[Int])))
```
hearnadam, line 310: Given the parameter includes the type: `def test(v: Int < Env[Int]`
Can you try invoking it like: `test(Envs.get)`? not that this is particularly safe...

kyo-core/shared/src/test/scala/kyoTest/envsTest.scala
```diff
@@ -6,61 +6,61 @@ class envsTest extends KyoTest:
 
     "value" in {
         val v1 =
-            Envs[Int].get.map(_ + 1)
+            Envs.get[Int].map(_ + 1)
```
hearnadam, line 9: This becomes much more 'readable' in my mind. 'get int' vs 'int get'

kyo-core/shared/src/test/scala/kyoTest/envsTest.scala
```diff
@@ -6,61 +6,61 @@ class envsTest extends KyoTest:
...
-                    Envs[Float].run(0.23f)(
-                        Envs[Boolean].run(false)(
-                            Envs[String].run("a")(
-                                Envs[Int].run(42)(t)
+            Envs.run(0.23d)(
+                Envs.run('a')(
+                    Envs.run(0.23f)(
+                        Envs.run(false)(
+                            Envs.run("a")(
+                                Envs.run(42)(t)
```
hearnadam, line 63: Awesome!

kyo-core/shared/src/main/scala/kyo/envs.scala
hearnadam: Minor: I would rename `env` -> `value` and `value` -> `effect`

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -1,32 +1,36 @@
...
+        using
+        HasEnvs[V, VS] { type Remainder = VR },
+        Tag[Envs[V]]
+    ): T < (S & VR) =
+        Envs[V].handle(handler[V])(env, value).asInstanceOf[T < (S & VR)]
```
hearnadam, line 33: I see you use the capital `Envs[V]` vs the static `envs[V]` instance. Is this necessary?

hearnadam (approved):

kyo-core/shared/src/test/scala/kyoTest/abortsTest.scala
fwbrasil: yep! that works too

kyo-core/shared/src/test/scala/kyoTest/envsTest.scala
fwbrasil: true that

kyo-core/shared/src/main/scala/kyo/envs.scala
fwbrasil: it was an oversight

kyo-core/shared/src/main/scala/kyo/envs.scala
fwbrasil: yeah, I think we should follow up trying to make parameter naming more explicit and regular in all APIs. I'll keep as is since other APIs typically use `v` or `value` for a computation parameter. I'm not a big fan of `effect`  because in Kyo's context I'd expect an `effect` parameter to be a algebraic effect, not a computation.

kyo-core/shared/src/test/scala/kyoTest/abortsTest.scala
hearnadam: Type unsafe 🙂 - inference could change depending on how the function is defined. Obviously not a problem with Kyo's encoding.

It's likely not a concern for most users, as they should rarely define functions where the result type is generic.

eg

```scala
type Pending = Envs[Float] & Envs[Double]
def myFunc[T](effect: T < Pending) = effect.map(_ / 2)

val result: ??? = myFunc(Envs.get)
```

Determinism in the Scala compiler is generally good, but I generally prefer not to take changes.

kyo-core/shared/src/main/scala/kyo/envs.scala
hearnadam: Makes sense, thanks.

kyo-core/shared/src/test/scala/kyoTest/abortsTest.scala
fwbrasil: I tested this and it's actually unsafe as you suspected. It's not much about the method inference but Scala infers a type intersection as the environment type, which fails when handling the effect. I think we'll need to add a constraint similar to `Flat` to validate that the type is not an intersection. Another option is using `Tag` to check types during handling but that's too expensive.

I added a couple pending tests because I think this doesn't seem a blocker.

hearnadam (approved):

PR #299, opened 2024-04-30
Aborts-like encoding for Vars

fwbrasil:


kyo-bench/src/main/scala/kyo/bench/MtlBench.scala
```diff
@@ -18,14 +18,14 @@ class MtlBench extends Bench:
                 for
                     conf <- Envs.use[Env](_.config)
                     _    <- Sums.add(Event(s"Env = $conf"))
-                    _    <- Vars[State].update(state => state.copy(value = state.value + 1))
+                    _    <- Vars.update[State](state => state.copy(value = state.value + 1))
```
hearnadam, line 21: Can you try adding the type parameter to the lambda instead of the 'update'?

hearnadam (approved):

kyo-bench/src/main/scala/kyo/bench/MtlBench.scala
fwbrasil: Yep, that works too.

PR #301, opened 2024-04-30
fix kyo-test build

fwbrasil:


.github/workflows/build.yml
```diff
@@ -23,8 +23,8 @@ jobs:
 
     - name: Build JVM
       run: |
-        sbt 'kyo-bench/test' 'kyo-cache/test' '+kyo-scheduler/test' 'kyo-core/test' 'kyo-direct/test' 'kyo-examples/test' 'kyo-os-lib/test' 'kyo-stats-otel/test' 'kyo-sttp/test' 'kyo-tapir/test'
+        sbt 'kyo-bench/test' 'kyo-cache/test' '+kyo-scheduler/test' 'kyo-core/test' 'kyo-direct/test' 'kyo-test/test' 'kyo-examples/test' 'kyo-os-lib/test' 'kyo-stats-otel/test' 'kyo-sttp/test' 'kyo-tapir/test'
 
     - name: Build JS
       run: |
-        sbt '+kyo-schedulerJS/test' 'kyo-coreJS/test' 'kyo-directJS/test' 'kyo-sttpJS/test' 
+        sbt '+kyo-schedulerJS/test' 'kyo-coreJS/test' 'kyo-directJS/test' 'kyo-sttpJS/test' 'kyo-testJS/test'
```
fwbrasil, line 30: I tried to find a way to separate JVM and JS modules without having to manually specify them in the past but I couldn't find a good way.

build.sbt
```diff
@@ -185,18 +186,19 @@ lazy val `kyo-tapir` =
         )
 
 lazy val `kyo-test` =
-  crossProject(JVMPlatform, JSPlatform)
-    .withoutSuffixFor(JVMPlatform)
-    .crossType(CrossType.Full)
-    .in(file("kyo-test"))
-    .dependsOn(`kyo-core` % "test->test;compile->compile")
```
fwbrasil, line 192: The test dependency to `kyo-core` was the cause of the issue with `@EnableReflectiveInstantiation`.

hearnadam (approved):
Thanks for the fix!

PR #303, opened 2024-04-30
zio integration

fwbrasil:
I've recovered the ZIO integration and identified a way to handle both Kyo and ZIO computations by manually implementing the evaluation loop handling all effects at once, [similarly to `IOTask`](https://github.com/getkyo/kyo/blob/be752ed1e6d8477da28a42d9847fea8867922790/kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala#L27). This approach doesn't require blocking threads and it seems it can handle async boundaries and propagate interrupts/failures correctly.

### DIFF
--- a/kyo-zio/shared/src/main/scala/kyo/zios.scala
+++ b/kyo-zio/shared/src/main/scala/kyo/zios.scala
@@ -1,20 +1,29 @@
 package kyo
 
+import ZIOs.internal.*
 import core.*
 import core.internal.*
 import kyo.fibersInternal.*
 import scala.util.control.NonFatal
 import zio.Task
-import zio.ZEnvironment
 import zio.ZIO
 
-class ZIOs extends Effect[ZIOs]:
-    type Command[T] = Task[T]
+opaque type ZIOs <: Fibers = Tasks & Fibers
 
-object ZIOs extends ZIOs:
+object ZIOs:
 
-    def run[T: Flat](v: T < (Fibers & ZIOs)): Task[T] =
-        def loop(v: T < (Fibers & ZIOs)): Task[T] =
+    def get[E >: Nothing: Tag, A](v: ZIO[Any, E, A]): A < (Aborts[E] & ZIOs) =
+        val task = v.fold[A < Aborts[E]](e => Aborts.fail(e), a => a)
+        Tasks.suspend[A < Aborts[E], A, Aborts[E]](task, identity)
+
+    def get[A](v: ZIO[Any, Nothing, A]): A < ZIOs =
+        Tasks.suspend[A](v)
+
+    inline def get[R: zio.Tag, A](v: ZIO[R, ?, A])(using tag: Tag[Envs[R]]): A < (Envs[R] & ZIOs) =
+        compiletime.error("ZIO environments are not supported yet. Please handle them before calling this method.")
+
+    def run[T: Flat](v: T < ZIOs): Task[T] =
+        def loop(v: T < ZIOs): Task[T] =
             v match
                 case kyo: Suspend[?, ?, ?, ?] =>
                     try
@@ -32,7 +41,7 @@ object ZIOs extends ZIOs:
                                         Left(ZIO.succeed(p.interrupt()))
                                     }
                             end match
-                        else if kyo.tag == Tag[ZIOs] then
+                        else if kyo.tag == Tag[Tasks] then
                             val k = kyo.asInstanceOf[Suspend[Task, Any, T, ZIOs]]
                             k.command.flatMap(v => loop(k(v)))
                         else
@@ -48,14 +57,9 @@ object ZIOs extends ZIOs:
         loop(v)
     end run
 
-    // TODO: fix Envs with type intersections
-    private[kyo] def get[R: zio.Tag, E: Tag, A](v: ZIO[R, E, A])(using tag: Tag[Envs[R]]): A < (Envs[R] & Aborts[E] & ZIOs) =
-        Envs.get[R](using tag).map(r => get(v.provideEnvironment(ZEnvironment(r))))
-
-    def get[E >: Nothing: Tag, A](v: ZIO[Any, E, A]): A < (Aborts[E] & ZIOs) =
-        val task = v.fold[A < Aborts[E]](e => Aborts.fail(e), a => a)
-        this.suspend(task, identity)
-
-    def get[A](v: ZIO[Any, Nothing, A]): A < ZIOs =
-        this.suspend(v)
+    private[kyo] object internal:
+        class Tasks extends Effect[Tasks]:
+            type Command[T] = Task[T]
+        object Tasks extends Tasks
+    end internal
 end ZIOs
--- a/kyo-zio/shared/src/main/scala/kyo/zios.scala
+++ b/kyo-zio/shared/src/main/scala/kyo/zios.scala
@@ -19,7 +19,7 @@ object ZIOs:
     def get[A](v: ZIO[Any, Nothing, A]): A < ZIOs =
         Tasks.suspend[A](v)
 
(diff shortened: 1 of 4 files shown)


kyo-zio/shared/src/main/scala/kyo/zios.scala
```diff
@@ -0,0 +1,71 @@
...
+
+class ZIOs extends Effect[ZIOs]:
+    type Command[T] = Task[T]
+
+object ZIOs extends ZIOs:
+
+    def run[T: Flat](v: T < (Fibers & ZIOs)): Task[T] =
+        def loop(v: T < (Fibers & ZIOs)): Task[T] =
+            v match
+                case kyo: Suspend[?, ?, ?, ?] =>
```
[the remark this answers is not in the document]
fwbrasil: There's no need to handle other effects here. This method handles these effects as the "last ones". `Aborts` for example needs to be handled before calling it. We need this multi handler because `IOs`, `Fibers`, and `Zios` require deep handlers, which is essentially a regular [monad interface](https://github.com/getkyo/kyo/blob/d89f10fd0b7ab6182d43d038918648da844749b7/kyo-core/shared/src/main/scala/kyo/core.scala#L181).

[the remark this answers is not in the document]
hearnadam: Right... my suggestion was to enable handling `Aborts` via a mapping to `ZIO.fail`, though I think the problem may be with multiple aborts.

kyo-zio/shared/src/main/scala/kyo/zios.scala
```diff
@@ -0,0 +1,71 @@
...
+        end loop
+
+        loop(v)
+    end run
+
+    def get[R: zio.Tag, E: Tag, A](v: ZIO[R, E, A])(using tag: Tag[Envs[R]]): A < (Envs[R] & Aborts[E] & ZIOs) =
+        Envs.get[R](using tag).map(r => get(v.provideEnvironment(ZEnvironment(r))))
+
+    def get[R: zio.Tag, A](v: URIO[R, A])(using tag: Tag[Envs[R]]): A < (Envs[R] & ZIOs) =
+        Envs.get[R](using tag).map(r => get(v.provideEnvironment(ZEnvironment(r))))
```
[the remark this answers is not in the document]
fwbrasil: Makes sense. I've added a method overload to provide a better compile-time error message.

kyo-zio/shared/src/main/scala/kyo/zios.scala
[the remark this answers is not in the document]
fwbrasil: I see! It makes sense to add `Aborts[Throwable]` since `Task` supports failures. I've added it by handling `Aborts` first and using ZIO's `flatMap`.

fwbrasil:
This is ready for review. It seems we could eventually add integration for `Resources`/`Scope` as well but we need to figure out type inference. `Envs` support requires a new `Tag` mechanism.

kyo-zio/shared/src/main/scala/kyo/zios.scala
```diff
@@ -0,0 +1,68 @@
...
+                            ZIO.fail(ex)
+                case v =>
+                    ZIO.succeed(v.asInstanceOf[U])
+            end match
+        end loop
+
+        loop(Aborts.run(v)).flatMap {
+            case Left(ex)     => ZIO.fail(ex)
+            case Right(value) => ZIO.succeed(value)
+        }
```
hearnadam, line 60: I believe this is correct, but I want to ensure that Aborts short circuit in the correct order. I was trying to add a test myself, but was having issues with the scalatest runner.

```
        val a = Aborts.fail(new RuntimeException("Aborts"))
        val b = ZIOs.get(ZIO.fail(new RuntimeException("ZIOs")))

        def run(expected: String, v: Nothing < (Aborts[Throwable] & ZIOs)): Task[Assertion] =
            ZIOs.run(v).fold(
                e => assert(e.getMessage == expected),
                _ => fail("Unexpected success")
            )

        run("Aborts", a.map(_ => b)) *> run("ZIOs", a.map(_ => b))
```

It's important to have this test to ensure future refactorings don't break compatibility

kyo-zio/shared/src/main/scala/kyo/zios.scala
```diff
@@ -0,0 +1,68 @@
...
+
+    inline def get[R: zio.Tag, A](v: ZIO[R, ?, A])(using tag: Tag[Envs[R]]): A < (Envs[R] & ZIOs) =
+        compiletime.error("ZIO environments are not supported yet. Please handle them before calling this method.")
+
+    def run[T: Flat](v: T < (Aborts[Throwable] & ZIOs)): Task[T] =
+        def loop[U](v: U < ZIOs): Task[U] =
+            v match
+                case kyo: Suspend[?, ?, ?, ?] =>
+                    try
+                        if kyo.tag == Tag[IOs] then
```
hearnadam, line 30: is there a reason to avoid pattern match?

```
kyo.tag match {
  case Tag[IOs]
```
etc

kyo-zio/shared/src/main/scala/kyo/zios.scala
```diff
@@ -0,0 +1,68 @@
...
+object ZIOs:
+
+    def get[E >: Nothing: Tag, A](v: ZIO[Any, E, A]): A < (Aborts[E] & ZIOs) =
+        val task = v.fold[A < Aborts[E]](e => Aborts.fail(e), a => a)
+        Tasks.suspend[A < Aborts[E], A, Aborts[E]](task, identity)
+
+    def get[A](v: ZIO[Any, Nothing, A]): A < ZIOs =
+        Tasks.suspend[A](v)
+
+    inline def get[R: zio.Tag, A](v: ZIO[R, ?, A])(using tag: Tag[Envs[R]]): A < (Envs[R] & ZIOs) =
```
hearnadam, line 22: I think this function needs the `E`. Even if we later plan to add overloads for `Nothing` case, it's better for compatibility to keep the `E`

```
inline def get[R: zio.Tag, E, A](v: ZIO[R, E, A])(using tag: Tag[Envs[R]]): A < (Envs[R] & ZIOs) =
```

kyo-zio/shared/src/main/scala/kyo/zios.scala
fwbrasil: I've added a couple tests for this.

kyo-zio/shared/src/main/scala/kyo/zios.scala
fwbrasil: It doesn't compile. I'm trying to design a "multi handler" in Kyo's core to avoid having this kind of code here and in `IOTask`.

kyo-zio/shared/src/main/scala/kyo/zios.scala
fwbrasil: done

kyo-zio/shared/src/main/scala/kyo/zios.scala
```diff
@@ -0,0 +1,68 @@
...
+object ZIOs:
+
+    def get[E >: Nothing: Tag, A](v: ZIO[Any, E, A]): A < (Aborts[E] & ZIOs) =
+        val task = v.fold[A < Aborts[E]](e => Aborts.fail(e), a => a)
+        Tasks.suspend[A < Aborts[E], A, Aborts[E]](task, identity)
+
+    def get[A](v: ZIO[Any, Nothing, A]): A < ZIOs =
+        Tasks.suspend[A](v)
+
+    inline def get[R: zio.Tag, E, A](v: ZIO[R, E, A])(using tag: Tag[Envs[R]]): A < (Envs[R] & ZIOs) =
```
hearnadam, line 22: Note for later: when `E >: Nothing` `Aborts[E]` should be in the result

hearnadam (approved):

PR #304, opened 2024-04-30
introduce fork/join benchmark

fwbrasil:
Port of ZIO's benchmark: https://github.com/zio/zio/blob/b84b12892e8b6c329e5ace56f4c4ecc2d21ce094/benchmarks/src/main/scala/zio/ForkJoinBenchmark.scala#L1

Results with M2 Max:

https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/ea9fcff9a00b6f79a1799b3342d625d7/raw/5d4876a9257176924c8bd19a63c48da4769b3821/jmh-result.json

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/ForkJoinBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/ForkJoinBench.scala
@@ -1,0 +1,47 @@
+package kyo.bench
+
+import org.openjdk.jmh.annotations.*
+
+class ForkJoinBench extends Bench.ForkOnly[Unit]:
+
+    val depth = 10000
+    val range = (0 until depth).toList
+
+    def catsBench() =
+        import cats.*
+        import cats.effect.*
+
+        val forkFiber     = IO.unit.start
+        val forkAllFibers = Traverse[List].traverse(range)(_ => forkFiber)
+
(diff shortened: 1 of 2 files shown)


kyri-petrou:
Just a small recommendation based on my learnings with benchmarks recently:

When benchmarking fork-join performance, I would recommend settings the `@Threads` annotation on the JMH benchmark to ~number of CPUs in the machine(ideally a little bit higher). Since forking usually involves globally defined resources (queues, etc.), having a single thread forking tasks doesn't capture contention in those resources which can be a surprisingly big bottleneck.

fwbrasil:
@kyri-petrou I'd say it's not as straightforward. The most difficult part of microbencharking is providing a clear signal regarding the performance of specific operations to guide optimization efforts. Setting the number of threads to the number of cores puts the system in a state of overloading where there's typically a significant amount of noise. Threads start getting preempted by the linux scheduler, contention can present itself in unrelated parts of the code being measured, and the overhead to pause and restart JMH's threads in case of async execution can dominate the performance profile.

I typically prefer to design benchmarks specifically for testing contention. You make a good point regarding fork/join since it's a scenario that is valuable to test contention for. I'll prepare a benchmark for it.

PR #305, opened 2024-05-01
fork/join contention bench

fwbrasil:
As discussed in https://github.com/getkyo/kyo/pull/304#issuecomment-2088666500. This benchmark shows some opportunity for optimization in Kyo! \o/

[Current results](https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/d4ad8ddd2d9819e3fb6fac9ea5ff1490/raw/16d4999c04859e26c3177ac41226934758a0dfe1/jmh-result.json) on my machine:

image

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/ForkJoinContentionBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/ForkJoinContentionBench.scala
@@ -1,0 +1,57 @@
+package kyo.bench
+
+import org.openjdk.jmh.annotations.*
+
+class ForkJoinContentionBench extends Bench.ForkOnly[Unit]:
+
+    val depth     = 1000
+    val parallism = Runtime.getRuntime().availableProcessors()
+    val range     = (0 until depth).toList
+
+    def catsBench() =
+        import cats.*
+        import cats.effect.*
+        import cats.implicits.*
+
+        val forkFiber         = IO.unit.start
+        val forkAllFibers     = Traverse[List].traverse(range)(_ => forkFiber)
+        val forkJoinAllFibers = forkAllFibers.flatMap(fibers => Traverse[List].traverse(fibers)(_.join).void)
+
+        Seq.fill(parallism)(forkJoinAllFibers).parSequence.void
+    end catsBench
+
+    override def kyoBenchFiber() =
+        import kyo.*
+
+        val forkFiber         = Fibers.init(())
+        val forkAllFibers     = Seqs.map(range)(_ => forkFiber)
+        val forkJoinAllFibers = forkAllFibers.flatMap(fibers => Seqs.map(fibers)(_.get).unit)
+
+        Fibers.parallel(Seq.fill(parallism)(forkJoinAllFibers)).unit
+    end kyoBenchFiber
+
+    def zioBench() =
+        import zio.*
+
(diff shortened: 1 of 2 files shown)


kyo-bench/src/main/scala/kyo/bench/ForkJoinContentionBench.scala
```diff
@@ -0,0 +1,57 @@
+package kyo.bench
+
+import org.openjdk.jmh.annotations.*
+
+class ForkJoinContentionBench extends Bench.ForkOnly[Unit]:
+
+    val depth     = 1000
+    val parallism = Runtime.getRuntime().availableProcessors()
```
hearnadam, line 8: Do you know what value this evaluated to on your machine? Do we need a minimum number here?

kyo-bench/src/main/scala/kyo/bench/ForkJoinContentionBench.scala
```diff
@@ -0,0 +1,57 @@
...
+        val forkAllFibers     = Seqs.map(range)(_ => forkFiber)
+        val forkJoinAllFibers = forkAllFibers.flatMap(fibers => Seqs.map(fibers)(_.get).unit)
+
+        Fibers.parallel(Seq.fill(parallism)(forkJoinAllFibers)).unit
+    end kyoBenchFiber
+
+    def zioBench() =
+        import zio.*
+
+        val forkFiber         = ZIO.unit.forkDaemon
```
hearnadam, line 36: Why forkDaemon and not fork? It shouldn't matter in this context, but I presume you want to avoid overhead of monitoring parent fiber interruption?

hearnadam (approved):

kyo-bench/src/main/scala/kyo/bench/ForkJoinContentionBench.scala
fwbrasil: I'm not sure as well, that came from the original in ZIO: https://github.com/zio/zio/blob/9eb127057495aa36d15c9a3b01f4d2188df9a5ff/benchmarks/src/main/scala/zio/ForkJoinBenchmark.scala#L59

kyo-bench/src/main/scala/kyo/bench/ForkJoinContentionBench.scala
fwbrasil: It returns `12` on my M2 Max. Setting to the number of cores helps to keep the benchmark testing similar behavior regardless of how many cores the machine has.

kyri-petrou:
@fwbrasil thanks for writing this benchmark - it actually made me aware of a non-neglible performance issue with `ZIO.collectAllPar` / `ZIO.foreachPar` (and other `ZIO.xPar` variants) 🥲. For long-lived forks, the impact should be almost negligible, but it affects this benchmark quite significantly.

[I opened an issue in the zio repo](https://github.com/zio/zio/issues/8813) to track this, but in the meantime, may I recommend to use the following for running ZIO effects in parallel to test for contention?

```scala
ZIO.collectAll(Seq.fill(parallism)(forkJoinAllFibers.forkDaemon)).flatMap(ZIO.foreach(_)(_.join))
```

PR #306, opened 2024-05-01
introduce option to replace Zio's executor with Kyo's in benchmarks

fwbrasil:
It's functional but there's a performance regression. I'm not sure why yet but even benchmarks that don't use the scheduler present new entry for `itable stubs`, which performs interface dispatches, as one of top methods using CPU. Example for a sync benchmark:

image

A possibility is some call that used to be mono/bimorphic with the default scheduler becomes polymorphic with the scheduler config. I'll follow up on it.

The config is disabled by default and it's necessary to pass the JMH parameter manually when executing benchmarks to activate it.

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/Bench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/Bench.scala
@@ -6,6 +6,7 @@ import kyo.*
 import org.openjdk.jmh.annotations.*
 import scala.concurrent.duration.Duration
 import zio.UIO
+import zio.internal.ExecutionMetrics
 
 @State(Scope.Benchmark)
 @Fork(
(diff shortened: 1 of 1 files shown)


kyo-bench/src/main/scala/kyo/bench/Bench.scala
```diff
@@ -20,7 +21,29 @@ import zio.UIO
...
+            given Unsafe = Unsafe.unsafe(identity)
+            val exec =
+                new Executor:
+                    val scheduler                     = kyo.scheduler.Scheduler.get
+                    def metrics(using unsafe: Unsafe) = None
+                    def submit(runnable: Runnable)(implicit unsafe: Unsafe): Boolean =
+                        scheduler.schedule(kyo.scheduler.Task(runnable.run()))
+                        true
+            val kExecutorLayer = Runtime.setExecutor(exec) ++ Runtime.setBlockingExecutor(exec)
+            Runtime.unsafe.fromLayer(kExecutorLayer).unsafe
```
hearnadam, line 43: We may need to avoid these 'layer' based operators and use the `zio` namespace to manually invoke Runtime creation

fwbrasil: Yeah, I wonder if the fact that we're using layers for the config is what generating the itable stub overhead

PR #308, opened 2024-05-02
introduce kyo-pure

fwbrasil:
I've been considering the long-term maintenance of the library and something that worries me is the lack of a clear separation between 100% pure effects from those that rely on `IOs`. Ideally, the library should evolve to have as much as possible of its code not depending on `IOs`, which should help keep the code quality and maintainability.

Additionally, some users might want to use Kyo in the same kind of scope which `zio-pure` is currently used for. This new module would be Kyo's equivalent to it.

This initial PR only creates the new module and copies all pure effects to it. Some effects like `Streams` has a few `IOs`-based methods to handle `Channel`s. I've removed these methods and am planning to reintroduce them in `kyo-core` via extension methods.

After this change, I'm planning follow up with a new PR changing `kyo-core` to depend on `kyo-pure` and removing the duplicate effects that were moved to `kyo-pure`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,8 +23,8 @@ jobs:
 
     - name: Build JVM
       run: |
-        sbt 'kyo-bench/test' 'kyo-cache/test' '+kyo-scheduler/test' 'kyo-core/test' 'kyo-direct/test' 'kyo-test/test' 'kyo-examples/test' 'kyo-os-lib/test' 'kyo-stats-otel/test' 'kyo-sttp/test' 'kyo-tapir/test'
+        sbt 'kyo-pure/test' '+kyo-scheduler/test' 'kyo-core/test' 'kyo-direct/test' 'kyo-test/test' 'kyo-bench/test' 'kyo-cache/test' 'kyo-examples/test' 'kyo-os-lib/test' 'kyo-stats-otel/test' 'kyo-sttp/test' 'kyo-tapir/test'
 
     - name: Build JS
       run: |
-        sbt '+kyo-schedulerJS/test' 'kyo-coreJS/test' 'kyo-directJS/test' 'kyo-sttpJS/test' 'kyo-testJS/test'
+        sbt 'kyo-pureJS/test' '+kyo-schedulerJS/test' 'kyo-coreJS/test' 'kyo-directJS/test' 'kyo-sttpJS/test' 'kyo-testJS/test'
--- a/kyo-pure/shared/src/main/scala/kyo/aborts.scala
+++ b/kyo-pure/shared/src/main/scala/kyo/aborts.scala
@@ -1,0 +1,80 @@
+package kyo
+
+import kyo.core.*
+import scala.reflect.ClassTag
+
+// Can't use an opaque type because
+// kyo-direct tests crash the compiler.
+type Aborts[-V] //= DoAbort
+
+object Aborts:
+
+    import internal.*
+
+    def fail[V](v: V): Nothing < Aborts[V] =
+        DoAbort.suspend(v).asInstanceOf[Nothing < Aborts[V]]
+
+    def when[V](b: Boolean)(value: => V): Unit < Aborts[V] =
+        if b then fail(value)
+        else ()
+
+    def get[V, T](e: Either[V, T]): T < Aborts[V] =
+        e match
+            case Right(v) => v
+            case Left(v)  => fail(v)
+
+    class RunDsl[V]:
+        def apply[V0 <: V, T: Flat, S, VS, VR](v: T < (Aborts[VS] & S))(
+            using
+            h: HasAborts[V0, VS] { type Remainder = VR },
+            ct: ClassTag[V0]
+        ): Either[V, T] < (VR & S) =
+            DoAbort.handle(handler)(ct, v).asInstanceOf[Either[V, T] < (VR & S)]
+    end RunDsl
+
+    def run[V]: RunDsl[V] = RunDsl[V]
+
+    private object internal:
+
+        val handler =
+            new ResultHandler[ClassTag[?], Const[Any], DoAbort, [T] =>> Either[Any, T], Any]:
+                def done[T](st: ClassTag[?], v: T) = Right(v)
+
+                override def accepts[T](st: ClassTag[?], command: Any) =
+                    type V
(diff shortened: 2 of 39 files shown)


.github/workflows/build.yml
```diff
@@ -23,8 +23,8 @@ jobs:
 
     - name: Build JVM
       run: |
-        sbt 'kyo-bench/test' 'kyo-cache/test' '+kyo-scheduler/test' 'kyo-core/test' 'kyo-direct/test' 'kyo-test/test' 'kyo-examples/test' 'kyo-os-lib/test' 'kyo-stats-otel/test' 'kyo-sttp/test' 'kyo-tapir/test'
+        sbt 'kyo-pure/test' '+kyo-scheduler/test' 'kyo-core/test' 'kyo-direct/test' 'kyo-test/test' 'kyo-bench/test' 'kyo-cache/test' 'kyo-examples/test' 'kyo-os-lib/test' 'kyo-stats-otel/test' 'kyo-sttp/test' 'kyo-tapir/test'
```
fwbrasil, line 26: Changed the order here to build modules with fewer dependencies first.

kyo-pure/shared/src/main/scala/kyo/aspects.scala
```diff
@@ -0,0 +1,65 @@
+package kyo
+
+object Aspects:
```
fwbrasil, line 3: This effect was based on `IOs` only in order to use `Locals`. It's using the new `Defers` effect now, which is similar to `IOs` but designed to provide only effect suspension without side effects.

kyo-pure/shared/src/main/scala/kyo/core.scala
```diff
@@ -0,0 +1,214 @@
...
+                                resultLoop(kyo)
+                            case r =>
+                                r.asInstanceOf[Result[T]]
+                        end match
+                    case kyo: Suspend[MX, Any, T, E & S & S2] @unchecked =>
+                        new Continue[MX, Any, Result[T], S & S2](kyo):
+                            def apply(v: Any, s: Safepoint[S & S2], l: Locals.State) =
+                                _handleLoop(st, kyo(v, s, l))
+                    case v =>
+                        handler.done(st, v.asInstanceOf[T])
```
fwbrasil, line 96: I made a few minor optimizations to reduce bytecode size. Instead of matching the pure type at the end, I changed the code to use a type cast since it's the last type check.

kyo-pure/shared/src/main/scala/kyo/defers.scala
```diff
@@ -0,0 +1,25 @@
+package kyo
+
+import kyo.core.*
+import kyo.core.internal.*
+
+abstract private[kyo] class Defer[T, S] extends Suspend[Const[Unit], Unit, T, S & Defers]:
```
fwbrasil, line 6: This new effect provides suspension similarly to `IOs` but it's designed to handle only pure computations without side effects.

kyo-pure/shared/src/main/scala/kyo/locals.scala
```diff
@@ -0,0 +1,82 @@
+package kyo
+
+import core.*
+import core.internal.*
+
+abstract class Local[T]:
+
+    import Locals.*
```
fwbrasil, line 8: New implemented in terms of `Defers` instead of `IOs`.

kyo-pure/shared/src/main/scala/kyo/streams.scala
```diff
@@ -0,0 +1,265 @@
+package kyo
+
+import kyo.*
+import kyo.core.*
+import kyo.core.internal.*
+import scala.annotation.tailrec
+
+opaque type Stream[+T, V, -S] = T < (Streams[V] & S)
```
fwbrasil, line 8: Removed some methods that depend on `IOs` to handle `Channels`. The idea is to reintroduce them as extension methods in `kyo-core`.

kyo-pure/shared/src/main/scala/kyo/aborts.scala
```diff
@@ -0,0 +1,80 @@
+package kyo
+
+import kyo.core.*
+import scala.reflect.ClassTag
+
+// Can't use an opaque type because
+// kyo-direct tests crash the compiler.
+type Aborts[-V] //= DoAbort
+
+object Aborts:
```
fwbrasil, line 10: I'm not sure if that's a good decision yet but I'm also considering exception handling and throwing as not pure so I've removed `Aborts.catching`, which I'm planning to reintroduce as an extension method in `kyo-core` as well. I'd love to hear thoughts on this aspects!

kyo-pure/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -0,0 +1,467 @@
+package kyo
+
+import Chunks.Indexed
+import scala.annotation.tailrec
+import scala.reflect.ClassTag
+
+sealed abstract class Chunk[T] derives CanEqual:
```
fwbrasil, line 7: This is the only implementation in `kyo-pure` that uses impure code internally.

kyo-pure/shared/src/main/scala/kyo/core.scala
```diff
@@ -0,0 +1,214 @@
...
+            using inline _tag: Tag[E]
+        ): U < (E & S) =
+            new Suspend[e.Command, T, U, S]:
+                def command = cmd
+                def tag     = _tag.asInstanceOf[Tag[Any]]
+                def apply(v: T, s: Safepoint[S], l: State) =
+                    f(v)
+    end extension
+
+    inline def eval[T, S, S2](v: T < S)(inline f: (Safepoint[S & S2], () => T < S) => T < (S & S2)): T < (S & S2) =
```
fwbrasil, line 39: This is a new method I introduced to enable implementing `IOs` methods like `handle` and `ensure` without depending on internal core classes.

hearnadam:
As a general comment, I would prefer you use `git mv` instead of duplication the files - if possible. This will preserve the git history when. It's fine if `kyo-core` is broken until the next PR.

fwbrasil:
> As a general comment, I would prefer you use `git mv` instead of duplication the files - if possible. This will preserve the git history when. It's fine if `kyo-core` is broken until the next PR.

The difficulty is that I needed to introduce `Defers` and `IOs` will need to be reimplemented based on it. Let take another path and introduce these changes before the move.

PR #309, opened 2024-05-02
fork/join contention benchmark improvement

fwbrasil:
ZIO's `collectAllPar` is getting [optimized](https://github.com/zio/zio/issues/8813). This PR changes the benchmark to manually fork and join fibers as a workaround for now. 

@kyri-petrou Thank you for sharing this suggestion! 🙏

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/ForkJoinContentionBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/ForkJoinContentionBench.scala
@@ -37,7 +37,7 @@ class ForkJoinContentionBench extends Bench.ForkOnly[Unit]:
         val forkAllFibers     = ZIO.foreach(range)(_ => forkFiber)
         val forkJoinAllFibers = forkAllFibers.flatMap(fibers => ZIO.foreach(fibers)(_.await).unit)
 
-        ZIO.collectAllPar(Seq.fill(parallism)(forkJoinAllFibers)).unit
+        ZIO.collectAll(Seq.fill(parallism)(forkJoinAllFibers.forkDaemon)).flatMap(ZIO.foreach(_)(_.join)).unit
     end zioBench
 
     @Benchmark


kyo-bench/src/main/scala/kyo/bench/ForkJoinContentionBench.scala
```diff
@@ -37,7 +37,7 @@ class ForkJoinContentionBench extends Bench.ForkOnly[Unit]:
         val forkAllFibers     = ZIO.foreach(range)(_ => forkFiber)
         val forkJoinAllFibers = forkAllFibers.flatMap(fibers => ZIO.foreach(fibers)(_.await).unit)
 
-        ZIO.collectAllPar(Seq.fill(parallism)(forkJoinAllFibers)).unit
+        ZIO.collectAll(Seq.fill(parallism)(forkJoinAllFibers.forkDaemon)).flatMap(ZIO.foreach(_)(_.join)).unit
```
hearnadam, line 40: I am a bit torn on this change. On one hand, it's best to be fair to all libraries and attempt to benchmark the true performance. However, this is a bit misleading to potential users. It's probable that a majority of ZIO code will directly use the `*Par` variants, thus this benchmark is not representative of that. Programmers will use the functions provided by the library.

hearnadam (approved):

kyo-bench/src/main/scala/kyo/bench/ForkJoinContentionBench.scala
fwbrasil: Not sure I agree. The main value of these benchmarks is to guide optimization efforts, hopefully even beyond Kyo. The semantics of the code is the same with this change and the execution seems also the same as the other targets. Removing the noise due to the current collectAllPar performance issue provides a better signal to measure the fork/join performance.

hearnadam: I agree with that.

PR #312, opened 2024-05-03
introduce core eval method

fwbrasil:
I'm working on cleaning up use of core internal classes in effects. This PR introduces a new `eval` method designed to replace a few usage patterns that appear in the `IOs` effect. The method essentially gives the user control over the evaluation of a computation in order to access the `Safepoint` and provide features like exception handling.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/core.scala
+++ b/kyo-core/shared/src/main/scala/kyo/core.scala
@@ -106,6 +106,23 @@ object core:
         end handle
     end extension
 
+    inline def eval[T, S, S2](v: => T < S)(
+        inline resume: (Safepoint[S & S2], () => T < (S & S2)) => T < (S & S2),
+        inline done: (Safepoint[S & S2], T) => T < (S & S2) = (s: Safepoint[S & S2], v: T) => v,
+        inline suspend: T < (S & S2) => T < (S & S2) = (v: T < (S & S2)) => v
+    ): T < (S & S2) =
+        def evalLoop(s: Safepoint[S & S2], v: T < (S & S2)): T < (S & S2) =
+            v match
+                case kyo: Suspend[MX, Any, T, S & S2] @unchecked =>
+                    new Continue[MX, Any, T, S & S2](kyo):
+                        def apply(v: Any, s: Safepoint[S & S2], l: Locals.State) =
+                            suspend(evalLoop(s, resume(s, () => kyo(v, s, l))))
+                case v =>
+                    suspend(done(s, v.asInstanceOf[T]))
+        val s = Safepoint.noop
+        evalLoop(s, suspend(resume(s, () => v)))
+    end eval
+
     abstract class ResultHandler[State, Command[_], E <: Effect[E], Result[_], S]:
 
         case class Resume[U, S2](st: State, v: U < (E & S & S2))
--- a/kyo-core/shared/src/main/scala/kyo/ios.scala
+++ b/kyo-core/shared/src/main/scala/kyo/ios.scala
@@ -36,54 +36,22 @@ sealed trait IOs extends Effect[IOs]:
         }
 
     def attempt[T, S](v: => T < S): Try[T] < S =
-        def attemptLoop(v: T < S): Try[T] < S =
-            v match
-                case kyo: Suspend[MX, Any, T, S] @unchecked =>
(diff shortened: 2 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -118,34 +86,25 @@ sealed trait IOs extends Effect[IOs]:
...
-                                catch
-                                    case ex if (NonFatal(ex)) =>
-                                        ensure()
-                                        throw ex
-                            ensureLoop(v2, np)
-                        end apply
-                case _ =>
-                    p.remove(ensure)
-                    IOs {
+        eval(v: T < (IOs & S))(
```
fwbrasil, line 89: This is the most complete example of the usage of the API. The `suspend` parameter can be used to suspend the evaluation with an effect. In this case, both the execution of `resume` and `done` will be wrapped into `IOs` since they perform side effects.

kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -178,6 +137,15 @@ private[kyo] object iosInternal:
             IOs(v)
     end Preempt
     object Preempt:
+        def ensure(s: Safepoint[?], e: Ensure): Unit =
```
fwbrasil, line 140: A minor refactoring to simplify `ensure`.

kyo-core/shared/src/main/scala/kyo/core.scala
```diff
@@ -106,6 +106,23 @@ object core:
         end handle
     end extension
 
+    inline def eval[T, S, U >: T, S2](v: => T < S)(
+        inline resume: (Safepoint[S & S2], () => U < (S & S2)) => U < (S & S2),
+        inline done: (Safepoint[S & S2], T) => U < (S & S2) = (s: Safepoint[S & S2], v: T) => v,
+        inline suspend: U < (S & S2) => U < (S & S2) = (v: U < (S & S2)) => v
```
fwbrasil, line 112: I'm sorry for this symbol salad 😂 I tried defining an `EvalHandler` with a more readable API but I couldn't avoid overhead with it.

kyo-core/shared/src/main/scala/kyo/core.scala
```diff
@@ -106,6 +106,23 @@ object core:
         end handle
     end extension
 
+    inline def eval[T, S, S2](v: => T < S)(
+        inline resume: (Safepoint[S & S2], () => T < (S & S2)) => T < (S & S2),
+        inline done: (Safepoint[S & S2], T) => T < (S & S2) = (s: Safepoint[S & S2], v: T) => v,
+        inline suspend: T < (S & S2) => T < (S & S2) = (v: T < (S & S2)) => v
```
fwbrasil, line 112: I'm sorry for this symbol salad 😂 I tried defining an EvalHandler with a more readable API but I couldn't avoid overhead with it.

kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -36,54 +36,22 @@ sealed trait IOs extends Effect[IOs]:
...
-            case ex: Throwable if (NonFatal(ex)) =>
-                Failure(ex)
-        end try
-    end attempt
+        eval(v.map(Success(_): Try[T]))((s, k) =>
+            try k()
+            catch
+                case ex if NonFatal(ex) =>
+                    Failure(ex)
+        )
```
fwbrasil, line 44: This method only specifies `resume`, both `done` and `suspend` use no-op default parameters.

kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -118,34 +86,25 @@ sealed trait IOs extends Effect[IOs]:
...
-                            val v2 =
-                                try kyo(v, s, l)
-                                catch
-                                    case ex if (NonFatal(ex)) =>
-                                        ensure()
-                                        throw ex
-                            ensureLoop(v2, np)
-                        end apply
-                case _ =>
-                    p.remove(ensure)
```
fwbrasil, line 142: This was incorrectly placed outside of `IOs`.

fwbrasil:
I'm merging this PR to follow up with a new `IOs` impl based on `Defers`. Please feel free to provide feedback here!

PR #313, opened 2024-05-03
loops: repeat and forever

fwbrasil:
Moves `repeat` from `Seqs` to `Loops` and introduces a new `forever` method to loop indefinitely.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/loops.scala
+++ b/kyo-core/shared/src/main/scala/kyo/loops.scala
@@ -220,4 +220,30 @@ object Loops:
                     ()
         loop()
     end foreach
+
+    inline def repeat[S](n: Int)(inline run: => Unit < S): Unit < S =
+        def _loop(i: Int): Unit < S =
+            loop(i)
+        @tailrec def loop(i: Int = 0): Unit < S =
+            if i == n then ()
+            else
+                run match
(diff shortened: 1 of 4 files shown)


fwbrasil:
I'm merging this PR to follow up with a new `IOs` impl based on `Defers`. Please feel free to provide feedback here!

kyo-core/shared/src/main/scala/kyo/loops.scala
```diff
@@ -220,4 +220,30 @@ object Loops:
...
+                        loop(i + 1)
+        loop()
+    end repeat
+
+    inline def forever[S](inline run: Unit < S): Unit < S =
+        def _loop(): Unit < S =
+            loop()
+        @tailrec def loop(): Unit < S =
+            run match
+                case kyo: Kyo[Unit, S] @unchecked =>
```
hearnadam, line 243: Can this be `Kyo[?, S]` or `Kyo[Any, S]`? Does that impact cost of typecheck?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/loops.scala
fwbrasil: We'd need to do a type cast or use `map(_ => ...` in the next line. The type itself has no cost, generic type parameters aren't checked at runtime.

PR #314, opened 2024-05-03
fix io task runtime tracking

fwbrasil:
This is a bug introduced by https://github.com/getkyo/kyo/pull/278. I noticed it while testing the new scheduler under stress. I'm not sure how to introduce an automated test for this given that it's a performance issue and this code isn't very testable in isolation. I've tried designing a benchmark to measure preemption performance a few times unsuccessfully.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
+++ b/kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
@@ -49,7 +49,7 @@ private[kyo] class IOTask[T](
                         k.command match
                             case Promise(p) =>
                                 this.interrupts(p)
-                                val runtime = (clock.currentMillis() - startMillis).toInt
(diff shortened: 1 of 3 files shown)


kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -49,7 +49,7 @@ private[kyo] class IOTask[T](
                         k.command match
                             case Promise(p) =>
                                 this.interrupts(p)
-                                val runtime = (clock.currentMillis() - startMillis).toInt
+                                val runtime = (clock.currentMillis() - startMillis + this.runtime()).toInt
```
fwbrasil, line 52: This scenario happens when the fiber/io task suspends because of an async boundary. Here, `runtime` should contain the time spent on the current execution plus all of the scheduling time the fiber got before it.

hearnadam (approved):

PR #315, opened 2024-05-03
cache Streams handlers

fwbrasil:
Some handlers don't depend on functions and can be cached. This optimization doesn't show improvement for the current benchmarks likely because the JIT is able to avoid the allocation but that may not be the case in production code.

image
https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/c7689837c75289c5ada664fa0d4427eb/raw/a9d115eebc6ae293d85f00413d8cef78cfafc8d9/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/c7689837c75289c5ada664fa0d4427eb/raw/a9d115eebc6ae293d85f00413d8cef78cfafc8d9/jmh-result-candidate.json

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/streams.scala
+++ b/kyo-core/shared/src/main/scala/kyo/streams.scala
@@ -23,45 +23,14 @@ object Stream:
             if n <= 0 then
                 runDiscard
             else
-                val handler =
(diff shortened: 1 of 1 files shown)


fwbrasil:
ouch, tags giving headaches again. I'll close this for now.

PR #326, opened 2024-05-06
Make HasAborts public again

johnhungerford:


fwbrasil:
That's so odd, I'd expect Kyo's tests to be failing. It seems the compiler somehow leaks the type as public since it's used in `RunDsl`, which I think should fail compilation.

PR #328, opened 2024-05-06
http client benchmarks

fwbrasil:
I'm working on the performance of `Requests`. I couldn't find a way to instantiate `zio-http`'s client once and reuse it so I've disabled the benchmark for ZIO by returning a constant. Please ignore ZIO's results for now.

image

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/HttpClientBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/HttpClientBench.scala
@@ -1,0 +1,53 @@
+package kyo.bench
+
+import org.http4s.ember.client.EmberClientBuilder
+
+class HttpClientBench extends Bench.ForkOnly[String]:
+
+    val port = 9999
+    val url  = TestHttpServer.start(port)
+
+    lazy val catsClient =
+        import cats.effect.*
+        import cats.effect.unsafe.implicits.global
+        EmberClientBuilder.default[IO].build.allocated.unsafeRunSync()._1
+    end catsClient
+
+    val catsUrl =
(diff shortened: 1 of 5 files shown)


hearnadam:
I think for ZIO applications you will need to add a method which enables creating runtimes from layers (I think some of the framework is already there for the scheduler change)

You should be able to do
```
val clientRuntime: Runtime[Client] = unsafeRun(Runtime.default)(Client.default.toRuntime)
```

kyo-bench/src/main/scala/kyo/bench/HttpClientBench.scala
```diff
@@ -0,0 +1,53 @@
...
+
+    override def kyoBenchFiber() =
+        import kyo.*
+
+        Requests.run(Requests[String](_.get(kyoUrl)))
+    end kyoBenchFiber
+
+    val zioUrl =
+        import zio.http.*
+        URL.decode(this.url).toOption.get
```
hearnadam, line 42: do you need the `this` prefix?

fwbrasil: yes, it conflicts with `zio.http.url`

fwbrasil:
I'm merging this to follow up with optimizations. I'll open a ticket for ZIO's benchmark.

> ```
> val clientRuntime: Runtime[Client] = unsafeRun(Runtime.default)(Client.default.toRuntime)
> ```

Wouldn't that still initialize the client on each benchmark execution? I think we need to get a hold of a client instance to reuse across executions.

PR #330, opened 2024-05-06
requests: avoid switching to ForkJoinPool

fwbrasil:
Java's HTTP client switches execution to `ForkJoinPool` by default, which isn't necessary since STTP already shifts the execution to Kyo's scheduler to process the request. This PR makes the client use an "immediate" executor by default, flushing any `CompletionStage` transformations immediately.

image
https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/63f13c78df90063bf403a584b647ccd7/raw/7116437f81dc06cdb6c06fdc8f8a711614375314/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/63f13c78df90063bf403a584b647ccd7/raw/7116437f81dc06cdb6c06fdc8f8a711614375314/jmh-result-candidate.json


hearnadam (approved):
Nice!

Issue #327, opened 2024-05-06
Choices.run incorrectly infers pending Effects


Tomczik76:
Choices.run incorrectly infers pending Effects
  
```
  val app = for {
    x <- Choices.get(1 to 5)
    y <- 
      if x == 2 then Aborts.fail("boom")
      else 42
  } yield y

  run(Choices.run(app))
```
```
[error] kyo.package$bug$KyoBugException: Unexpected effect 'aborts$package::Aborts::internal::DoAbort' found while handling '{IOs & fibersInternal::FiberGets}'. Please file a ticket.
[error]         at kyo.package$bug$KyoBugException$.apply(package.scala:70)
[error]         at kyo.package$bug$.apply(package.scala:94)
[error]         at kyo.scheduler.IOTask$$anon$2.apply(IOTask.scala:21)
[error]         at kyo.scheduler.IOTask$$anon$2.apply(IOTask.scala:72)
[error]         at kyo.scheduler.IOTask.eval(IOTask.scala:46)
[error]         at kyo.scheduler.IOTask.run(IOTask.scala:83)
[error]         at kyo.scheduler.Worker.runTask(Worker.scala:134)
[error]         at kyo.scheduler.Worker.run(Worker.scala:112)
[error]         at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
[error]         at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
[error]         at java.base/java.lang.Thread.run(Thread.java:833)
[error] stack trace is suppressed; run last Compile / run for the full output
[error] (Compile / run) kyo.package$bug$KyoBugException: Unexpected effect 'aborts$package::Aborts::internal::DoAbort' found while handling '{IOs & fibersInternal::FiberGets}'. Please file a ticket.
```

Expectation: This should not compile

PR #334, opened 2024-05-07
reproduce flat check bug (#327)

fwbrasil:
Reproduces #327. I'm pretty sure this used to work, I imagine something broke it along the way and we didn't have a test for it.

### DIFF
--- a/kyo-core/shared/src/test/scala/kyoTest/FlatTest.scala
+++ b/kyo-core/shared/src/test/scala/kyoTest/FlatTest.scala
@@ -61,6 +61,20 @@ class FlatTest extends KyoTest:
             succeed
         }
 
+        "effect mismatch" in {
+            def test[T](v: T < Fibers)(using Flat[T < Fibers]): T < Fibers = v
+            test(1)
+            test(1: Int < Fibers)
+            assertDoesNotCompile("test(1: Int < Options)")
+        }
+
+        "flat flat" in pendingUntilFixed {
+            def test[T](v: T < Fibers)(using Flat[T]): T < Fibers = v
+            test(1)
+            test(1: Int < Fibers)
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/test/scala/kyoTest/FlatTest.scala
```diff
@@ -61,6 +61,20 @@ class FlatTest extends KyoTest:
...
+            test(1)
+            test(1: Int < Fibers)
+            assertDoesNotCompile("test(1: Int < Options)")
+        }
+
+        "flat flat" in pendingUntilFixed {
+            def test[T](v: T < Fibers)(using Flat[T]): T < Fibers = v
+            test(1)
+            test(1: Int < Fibers)
+            assertDoesNotCompile("test(1: Int < Options)")
```
fwbrasil, line 75: This is currently compiling but shouldn't. It seems a simple workaround would be migrating all flat checks to the format with pending effects (`Flat[T < Fibers]` instead of `Flat[T]`) but it'd be ideal to fix this bug to reduce the noise of the flat check in Kyo's codebase.

kyo-core/shared/src/test/scala/kyoTest/KyoAppTest.scala
```diff
@@ -55,6 +55,20 @@ class KyoAppTest extends KyoTest:
...
     }
 
+    "effect mismatch" taggedAs jvmOnly in {
+        assertDoesNotCompile("""
+            new KyoApp:
+                run(1: Int < Options)
+        """)
+    }
+
+    "indirect effect mismatch" taggedAs jvmOnly in pendingUntilFixed {
```
fwbrasil, line 65: Another scenario that shouldn't compile.

hearnadam (approved):

PR #335, opened 2024-05-07
Locals and IOs based on Defers

fwbrasil:
I'm working on isolating Kyo's core in preparation for the new `kyo-pure` module (see https://github.com/getkyo/kyo/pull/308 for more context). This PR integrates `Defers` into both `IOs` and `Locals`. Computations that use `IOs` can also perform `Defer` suspensions since the opaque type declares it contains `Defers`:

```scala
opaque type IOs <: Defers = Defers & SideEffects
```

This change also makes `Locals` a pure effect based on `Defers` instead of `IOs`.

**Important:** This change breaks any use of `IOs` in `kyo-direct`. Any effect declared as `opaque` crashes the compiler for `kyo-direct`. I've changed the majority of tests to use `Defers` instead of `IOs` and commented out some others. The example http server is also commented out. We'll need to fix this issue with opaque types in the direct syntax before the next release.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/core.scala
+++ b/kyo-core/shared/src/main/scala/kyo/core.scala
@@ -27,13 +27,13 @@ object core:
                 def apply(v: T, s: Safepoint[E], l: State) =
                     v
 
-        inline def suspend[T, U, S](inline cmd: e.Command[T], inline f: T => U < S)(
+        inline def suspend[T, U, S](inline cmd: e.Command[T], inline f: T => U < (E & S))(
             using inline _tag: Tag[E]
         ): U < (E & S) =
-            new Suspend[e.Command, T, U, S]:
+            new Suspend[e.Command, T, U, E & S]:
                 def command = cmd
                 def tag     = _tag.asInstanceOf[Tag[Any]]
-                def apply(v: T, s: Safepoint[S], l: State) =
+                def apply(v: T, s: Safepoint[E & S], l: State) =
                     f(v)
     end extension
 
--- a/kyo-core/shared/src/main/scala/kyo/ios.scala
+++ b/kyo-core/shared/src/main/scala/kyo/ios.scala
@@ -2,7 +2,6 @@ package kyo
 
 import IOs.internal.*
 import core.*
-import core.internal.*
 import java.util.concurrent.atomic.AtomicReference
 import scala.annotation.tailrec
 import scala.util.*
--- a/kyo-core/shared/src/main/scala/kyo/ios.scala
+++ b/kyo-core/shared/src/main/scala/kyo/ios.scala
@@ -1,25 +1,29 @@
 package kyo
 
+import IOs.internal.*
 import core.*
 import core.internal.*
-import iosInternal.*
 import java.util.concurrent.atomic.AtomicReference
 import scala.annotation.tailrec
 import scala.util.*
 import scala.util.control.NonFatal
 
-sealed trait IOs extends Effect[IOs]:
+type IOs = Defers & SideEffects
 
-    private val tag = Tag[IOs]
-
-    val unit: Unit < IOs = ()
+object IOs:
 
     inline def apply[T, S](
         inline f: => T < (IOs & S)
     ): T < (IOs & S) =
-        new KyoIO[T, S]:
-            def apply(v: Unit, s: Safepoint[IOs & S], l: Locals.State) =
-                f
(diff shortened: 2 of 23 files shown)


kyo-core/shared/src/main/scala/kyo/core.scala
```diff
@@ -27,13 +27,13 @@ object core:
                 def apply(v: T, s: Safepoint[E], l: State) =
                     v
 
-        inline def suspend[T, U, S](inline cmd: e.Command[T], inline f: T => U < S)(
+        inline def suspend[T, U, S](inline cmd: e.Command[T], inline f: T => U < (E & S))(
```
fwbrasil, line 30: Just makes the API more flexible by allowing the `f` continuation to also perform the effect.

kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -1,25 +1,29 @@
 package kyo
```
fwbrasil, line 1: No internal core classes anymore \o/

kyo-sttp/shared/src/main/scala/kyo/internal/KyoSttpMonad.scala
```diff
@@ -35,7 +35,7 @@ object KyoSttpMonad:
             override def eval[T](t: => T) =
                 IOs[T, Fibers](t)
 
-            override def suspend[T](t: => M[T]) =
+            override def suspend[T](t: => M[T]): M[T] =
```
fwbrasil, line 38: Not sure why the compiler now needs the explicit return type here.

kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -1,25 +1,28 @@
...
 import core.*
-import core.internal.*
-import iosInternal.*
 import java.util.concurrent.atomic.AtomicReference
 import scala.annotation.tailrec
 import scala.util.*
 import scala.util.control.NonFatal
 
-sealed trait IOs extends Effect[IOs]:
+opaque type IOs <: Defers = Defers & SideEffects
```
hearnadam, line 10: Because this is a subtype of `Defers`, I think this should be possible: 

`Defers.run(IOs(sideEffect)).pure`

Will this work?
`opaque type IOs >: Defers = Defers & SideEffects`

kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -1,25 +1,28 @@
...
+
+    inline def unit: Unit < IOs = ()
+
+    def run[T](v: T < IOs)(using f: Flat[T < IOs]): T =
+        runLazy(v).pure
+
+    def runLazy[T: Flat, S](v: T < (IOs & S)): T < S =
+        Defers.run(SideEffects.handle(handler)((), v))
 
     def fail[T](ex: Throwable): T < IOs =
```
hearnadam, line 27: pre-existing - shouldn't this return `Nothing`?

kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -1,25 +1,28 @@
...
-    val unit: Unit < IOs = ()
+object IOs:
 
     inline def apply[T, S](
         inline f: => T < (IOs & S)
     ): T < (IOs & S) =
-        new KyoIO[T, S]:
-            def apply(v: Unit, s: Safepoint[IOs & S], l: Locals.State) =
-                f
+        SideEffects.suspend[Unit, T, Defers & S]((), _ => f: T < (Defers & SideEffects & S))
```
hearnadam, line 17: I don't love the `SideEffects` name, but I'm not sure of anything better.

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/ios.scala
fwbrasil: good catch! I've fixed it

kyo-core/shared/src/main/scala/kyo/ios.scala
fwbrasil: Yeah, Kyo typically doesn't use composite names. Maybe `Unsafes`? It doesn't seem to sound good, though

fwbrasil:
The workaround in https://github.com/getkyo/kyo/pull/361 didn't work for `IOs` because it has to be declared with ` <: Defers`. I'll drop this change for now to avoid having to maintain the branch. Hopefully we'll eventually figure out the issue with `opaque` and the direct syntax to enable this change.

PR #345, opened 2024-05-09
benchs: detect runtime leaks

fwbrasil:
It's easy to make mistakes in benchmarks that end up initializing targets other than the one being tested. For example, if `Bench.zioRuntime` were a `val` instead of `lazy val`, all benchmarks would run with ZIO's runtime initialized, including its threads. This kind of mistake can introduce significant noise in the benchmarks.


fwbrasil:
Merging to create a follow up PR.

PR #346, opened 2024-05-09
benchs: registry mechanism

fwbrasil:
Benchmarks currently need to be manually registered in `BenchTest`, which is error-prone since we can forget to add new benchmarks. For example, `DeepBindMapBench` was missing. This PR introduces a `Registry` mechanism that loads all benchmark classes automatically to execute tests via reflection. 

In addition to tests, I'm planning to use this mechanism in the actual benchmarks to generate more realistic JIT type profiles. I'll provide more background in the PR.


kyo-bench/src/test/scala/kyo/benchTest/CatsBenchTest.scala
```diff
@@ -1,4 +1,6 @@
-package kyo.bench
+package kyo.benchTest
```
fwbrasil, line 1: Had to move to a separate package otherwise `Registry` for some reason only returns test classes.

kyo-bench/src/main/scala/kyo/bench/Bench.scala
```diff
@@ -21,6 +21,8 @@ import zio.UIO
 @BenchmarkMode(Array(Mode.Throughput))
 abstract class Bench[T]:
 
+    val expectedResult: T
```
fwbrasil, line 24: So we don't need to specify in tests.

hearnadam (approved):
Very clever!

PR #348, opened 2024-05-10
(almost) allocationless tags ⚡

### LINKED ISSUE #339: Using Streams, IOs and Fibers cause an exception to occur
Hello!
First I want to thank you for all the help you provided me  in trying to create a Virgil interpreter for Kyo. It was really helpful, insightful and interesting to see Kyo first hand in action. I ended up running into a problem when Streams were involved in order to fetch data from Cassandra which also involved using Fibers underneath the hood. This was the stacktrace that I got:

 ```
Exception in thread "main" kyo.package$bug$KyoBugException: Unexpected effect 'Streams[=_$Output]' found while handling '{IOs & fibersInternal::FiberGets}'. Please file a ticket.
        at kyo.package$bug$KyoBugException$.apply(package.scala:70)
        at kyo.package$bug$.apply(package.scala:94)
        at kyo.scheduler.IOTask$$anon$2.apply(IOTask.scala:21)
        at kyo.scheduler.IOTask$$anon$2.apply(IOTask.scala:72)
        at kyo.scheduler.IOTask.eval(IOTask.scala:46)
        at kyo.scheduler.IOTask.run(IOTask.scala:83)
        at kyo.scheduler.Worker.runTask(Worker.scala:134)
        at kyo.scheduler.Worker.run(Worker.scala:112)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1570)
```

I have pushed all the code here: https://github.com/calvinlfer/virgil-kyo-experiment/blob/main/src/main/scala/io/kaizensolutions/virgil/CassandraPlay.scala 

I am currently working on trying to get a small reproducer without needing to bring up Cassandra.

### PR DESCRIPTION
**Pull request by fwbrasil**
Fixes #339.

Kyo uses tags for effect handling. Izumi's performance has been an important limitation, especially to provide sub-type checking it hot paths. This PR introduces a new `Tag` implementation that bundles the type information in a string format amenable to type comparisons without requiring allocations. Since the string is generated at compile-time, it becomes a constant in the class pool and doesn't require allocations.

The only scenario where this new `Tag` can allocate is when one of the type parameters has an implicit tag. For example:

```scala
def test[T: Tag] = Tag[List[T]]
```

In this case, the macro generates a tree that will concatenate the string representing `Tag[T]` into `Tag[List[T]]`. JIT compilers can many times avoid such allocation, though.

This change introduces an important restriction: only types with concrete class symbols (traits, classes, objects) are supported. We could extend the implementation to support intersection and union types as well but I think allowing these types can be confusing to users since they can be produced by inference when a user makes a mistake.

The performance is ~10x in comparison to Izumi:

image

The main reason is **zero allocation**:

image

https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/7010fc054376ce94e3497de2fbeb3030/raw/0ba5e81e932185ef4b10c75201e0ace98aac4a33/jmh-result.json#details

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -102,7 +102,6 @@ lazy val `kyo-core` =
         .in(file("kyo-core"))
         .settings(
             `kyo-settings`,
-            libraryDependencies += "dev.zio"       %%% "izumi-reflect"   % "2.3.9",
             libraryDependencies += "com.lihaoyi"   %%% "pprint"          % "0.9.0",
             libraryDependencies += "org.jctools"     % "jctools-core"    % "4.0.3",
             libraryDependencies += "org.slf4j"       % "slf4j-api"       % "2.0.13",
@@ -264,6 +263,7 @@ lazy val `kyo-bench` =
                     )
                 }
             },
+            libraryDependencies += "dev.zio"             %% "izumi-reflect"       % "2.3.9",
             libraryDependencies += "org.typelevel"       %% "cats-effect"         % "3.5.4",
             libraryDependencies += "org.typelevel"       %% "log4cats-core"       % "2.7.0",
             libraryDependencies += "org.typelevel"       %% "log4cats-slf4j"      % "2.7.0",
--- a/kyo-core/jvm/src/test/scala/kyoTest/coreBytecodeSizeTest.scala
+++ b/kyo-core/jvm/src/test/scala/kyoTest/coreBytecodeSizeTest.scala
@@ -36,17 +36,10 @@ class coreBytecodeSizeTest extends KyoTest:
     "handle" in runJVM {
         val map = methodBytecodeSize[TestHandle]
         assert(map == Map(
-<<<<<<< HEAD
             "test"        -> 27,
             "resultLoop"  -> 91,
-            "handleLoop"  -> 276,
+            "handleLoop"  -> 253,
             "_handleLoop" -> 10
-=======
-            "test"        -> 28,
-            "resultLoop"  -> 94,
-            "handleLoop"  -> 257,
-            "_handleLoop" -> 12
->>>>>>> d5847e0a ((almost) allocationless tags)
         ))
     }
 
--- a/kyo-core/jvm/src/test/scala/kyoTest/coreBytecodeSizeTest.scala
+++ b/kyo-core/jvm/src/test/scala/kyoTest/coreBytecodeSizeTest.scala
@@ -36,10 +36,17 @@ class coreBytecodeSizeTest extends KyoTest:
     "handle" in runJVM {
         val map = methodBytecodeSize[TestHandle]
         assert(map == Map(
+<<<<<<< HEAD
             "test"        -> 27,
             "resultLoop"  -> 91,
             "handleLoop"  -> 276,
             "_handleLoop" -> 10
+=======
+            "test"        -> 28,
+            "resultLoop"  -> 94,
+            "handleLoop"  -> 257,
+            "_handleLoop" -> 12
+>>>>>>> d5847e0a ((almost) allocationless tags)
         ))
     }
 
--- a/kyo-core/shared/src/main/scala/kyo/package.scala
+++ b/kyo-core/shared/src/main/scala/kyo/package.scala
@@ -72,19 +72,19 @@ package object kyo:
         inline def failTag[T, U](
             inline actual: Tag[U]
         ): Nothing =
-            bug(s"Unexpected effect '${actual.parse}' found in 'pure'.")
+            bug(s"Unexpected effect '${actual.show}' found in 'pure'.")
 
-        inline def failTag[T, U](
-            inline actual: Tag[U],
-            inline expected: Tag[T]
+        inline def failTag(
+            inline actual: Tag[?],
+            inline expected: Tag[?]*
         ): Nothing =
-            bug(s"Unexpected effect '${actual.parse}' found while handling '${expected.parse}'.")
+            bug(s"Unexpected effect '${actual.show}' found while handling '${expected.map(_.show).mkString(" & ")}'.")
 
         inline def checkTag[T, U](
             inline actual: Tag[U],
             inline expected: Tag[T]
         ): Unit =
-            if actual != expected then
+            if actual =!= expected then
                 failTag(actual, expected)
 
         def when(cond: Boolean)(msg: String): Unit =
--- a/kyo-core/shared/src/main/scala/kyo/tags.scala
+++ b/kyo-core/shared/src/main/scala/kyo/tags.scala
@@ -154,9 +154,13 @@ object Tag:
 
             tpe.dealias match
                 case AndType(_, _) | OrType(_, _) =>
-                    report.errorAndAbort(s"Unsupported Tag type ${tpe.show}")
+                    report.errorAndAbort(
+                        s"Unsupported intersection or union type in Tag: ${tpe.show}. Only class types and type parameters are allowed."
+                    )
                 case tpe if tpe.typeSymbol.fullName == "scala.Nothing" =>
-                    report.errorAndAbort(s"Unsupported Tag type ${tpe.show}")
+                    report.errorAndAbort(
+                        s"Tag cannot be created for Nothing as it has no values. Use a different type or add explicit type parameters if needed."
+                    )
                 case tpe if tpe.typeSymbol.isClassDef =>
                     val sym  = tpe.typeSymbol
                     val path = tpe.dealias.baseClasses.map(_.fullName).mkString(";") + ";"
@@ -184,12 +188,12 @@ object Tag:
                             Expr.summon[Tag[tpe]] match
                                 case None =>
                                     report.errorAndAbort(
-                                        s"Please provide an implicit ${TypeRepr.of[Tag[tpe]].show}"
+                                        report.errorAndAbort(s"Please provide an implicit kyo.Flat[${TypeRepr.of[tpe].show}] parameter.")
                                     )
                                 case Some(value) =>
                                     '{ $value.tpe }
                         case _ =>
-                            report.errorAndAbort(s"Unsupported Tag type ${tpe.show}")
+                            report.errorAndAbort(s"Tag only supports class types and type parameters, but got: ${tpe.show}")
             end match
         end encodeType
 
--- a/kyo-core/shared/src/main/scala/kyo/tags.scala
+++ b/kyo-core/shared/src/main/scala/kyo/tags.scala
@@ -16,16 +16,16 @@ object Tag:
         def show = t1.tpe.takeWhile(_ != ';')
 
         def <:<[U](t2: Tag[U]): Boolean =
-            t1.tpe == t2.tpe || isSubtype(t1, t2)
+            t1 =:= t2 || isSubtype(t1, t2)
 
         def =:=[U](t2: Tag[U]): Boolean =
-            t1.tpe == t2.tpe
+            (t1.tpe eq t2.tpe) || t1.tpe == t2.tpe
 
         def =!=[U](t2: Tag[U]): Boolean =
-            t1.tpe != t2.tpe
+            (t1.tpe ne t2.tpe) || t1.tpe != t2.tpe
 
         def >:>[U](t2: Tag[U]): Boolean =
-            t1.tpe == t2.tpe || isSubtype(t2, t1)
+            t1 =:= t2 || isSubtype(t2, t1)
 
     end extension
 
--- a/kyo-core/shared/src/main/scala/kyo/tags.scala
+++ b/kyo-core/shared/src/main/scala/kyo/tags.scala
@@ -1,5 +1,6 @@
 package kyo
 
+import scala.annotation.switch
 import scala.annotation.tailrec
 import scala.quoted.*
 
@@ -74,7 +75,7 @@ object Tag:
             @tailrec def loop(idx: Int): Int =
                 if idx >= tag.length() then -1
                 else
-                    tag.charAt(idx) match
+                    (tag.charAt(idx): @switch) match
                         case ';'             => idx + 1
                         case '[' | ',' | ']' => -1
                         case _               => loop(idx + 1)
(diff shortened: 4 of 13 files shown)


build.sbt
```diff
@@ -102,7 +102,6 @@ lazy val `kyo-core` =
         .in(file("kyo-core"))
         .settings(
             `kyo-settings`,
-            libraryDependencies += "dev.zio"       %%% "izumi-reflect"   % "2.3.9",
```
fwbrasil, line 105: Izumi dependency only for benchmarks now.

kyo-core/jvm/src/test/scala/kyoTest/coreBytecodeSizeTest.scala
```diff
@@ -38,7 +38,7 @@ class coreBytecodeSizeTest extends KyoTest:
         assert(map == Map(
             "test"        -> 28,
             "resultLoop"  -> 94,
-            "handleLoop"  -> 280,
+            "handleLoop"  -> 257,
```
fwbrasil, line 41: fewer bytecodes to call `=:=`

kyo-core/shared/src/main/scala/kyo/package.scala
```diff
@@ -72,19 +72,19 @@ package object kyo:
...
         ): Nothing =
-            bug(s"Unexpected effect '${actual.parse}' found in 'pure'.")
+            bug(s"Unexpected effect '${actual.show}' found in 'pure'.")
 
-        inline def failTag[T, U](
-            inline actual: Tag[U],
-            inline expected: Tag[T]
+        inline def failTag(
+            inline actual: Tag[?],
+            inline expected: Tag[?]*
```
fwbrasil, line 79: Since the new `Tag` doesn't support intersection types, I'm simulating them here for error reporting via a vararg param.

kyo-core/shared/src/main/scala/kyo/tags.scala
```diff
@@ -1,41 +1,206 @@
 package kyo
 
-import izumi.reflect.dottyreflection.TypeInspections
-import izumi.reflect.macrortti.LightTypeTag
+import scala.annotation.tailrec
 import scala.quoted.*
 
-opaque type Tag[T] = String
+case class Tag[T](tpe: String) extends AnyVal
```
fwbrasil, line 6: Using `AnyVal` solved some issues I was having with type inference.

kyo-core/shared/src/main/scala/kyo/tags.scala
```diff
@@ -1,41 +1,206 @@
...
 object Tag:
 
-    given canEqual[T, U]: CanEqual[Tag[T], Tag[U]] = CanEqual.derived
+    import internal.*
+
+    inline given apply[T >: Nothing]: Tag[T] = ${ tagImpl[T] }
+
+    extension [T](t1: Tag[T])
+
+        def show = t1.tpe.takeWhile(_ != ';')
```
fwbrasil, line 16: We should eventually improve this to show the entire type signature. This code only prints the outer class name, no type parameters. It's enough for the error reporting use case, though.

kyo-core/shared/src/test/scala/kyoTest/envsTest.scala
```diff
@@ -47,22 +47,8 @@ class envsTest extends KyoTest:
...
-            def test[T](v: T < (Envs[Int] & Envs[Double])) =
-                v
-            val a                                = test(Envs.get)
-            val b: (Int & Double) < Envs[Double] = Envs.run(1)(a)
-            val c: (Int & Double) < Any          = Envs.run(1.2d)(b)
-            assert(c.pure == 1)
-            ()
-        }
+    "intersection type env" in {
+        assertDoesNotCompile("Envs.get[Int & Double]")
```
fwbrasil, line 51: Doesn't compile anymore since tags don't allow intersection types.

kyo-tapir/src/main/scala/kyo/routes.scala
```diff
@@ -9,7 +8,7 @@ import sttp.tapir.server.ServerEndpoint
 import sttp.tapir.server.netty.NettyKyoServer
 import sttp.tapir.server.netty.NettyKyoServerBinding
 
-type Route = ServerEndpoint[Any, KyoSttpMonad.M]
+case class Route(endpoint: ServerEndpoint[Any, KyoSttpMonad.M]) extends AnyVal
```
fwbrasil, line 11: I had to change this since `KyoSttpMonad.M` isn't a concrete type.

kyo-tapir/src/main/scala/kyo/routes.scala
```diff
@@ -22,19 +21,20 @@ object Routes:
...
         Sums.run[Route].apply[Unit, Fibers & S](v).map { (routes, _) =>
-            IOs(server.addEndpoints(routes.toSeq.toList).start()): NettyKyoServerBinding < (Fibers & S)
+            IOs(server.addEndpoints(routes.toSeq.map(_.endpoint).toList).start()): NettyKyoServerBinding < (Fibers & S)
         }
     end run
 
     def add[A: Tag, I, E: Tag: ClassTag, O: Flat](e: Endpoint[A, I, E, O, Any])(
         f: I => O < (Fibers & Envs[A] & Aborts[E])
     ): Unit < Routes =
-        Sums.add(List(
```
fwbrasil, line 32: There was a bug here introduced by https://github.com/getkyo/kyo/pull/267

kyo-core/shared/src/main/scala/kyo/tags.scala
```diff
@@ -1,41 +1,206 @@
...
+        // Example Tag
+        //
+        // class Super
+        // class Param1 extends Super
+        // class Param2 extends Super
+        // class Test[-T, +U]
+        //
+        // Tag[Test[Param1, Param2]]
+        //
+        // Test;Super;java.lang.Object;scala.Matchable;scala.Any;[-Param1;Super;java.lang.Object;scala.Matchable;scala.Any;,+Param2;Super;java.lang.Object;scala.Matchable;scala.Any;]
```
fwbrasil, line 43: This format has some opportunity for optimization. We might be able to omit `java.lang.Object;scala.Matchable;scala.Any` and assume they're always present but I'm not sure about `Matcheable`.

kyo-core/shared/src/main/scala/kyo/tags.scala
```diff
@@ -1,41 +1,206 @@
...
+                        else
+                            loop(nextSubParam, nextSuperParam)
+                        end if
+                    }
+                end loop
+                loop(subStart + 1, superStart + 1)
+            end if
+        end checkParams
+
+        def nextParam(tag: String, idx: Int): Int =
```
fwbrasil, line 124: This scan consumes a significant portion of the CPU usage in the benchmark. We can probably optimize it by encoding the length of elements in the string representation. Profiling session:

flame_graph

kyo-core/shared/src/test/scala/kyoTest/envsTest.scala
```diff
@@ -47,22 +47,8 @@ class envsTest extends KyoTest:
...
 
-    "intersection type env" - {
-        "explicit" in pendingUntilFixed {
-            val a = Envs.run(1)(Envs.get[Int & Double])
-            val b = Envs.run(1.2d)(a)
-            assert(b.pure == 1)
-            ()
-        }
-        "method inference" in pendingUntilFixed {
-            def test[T](v: T < (Envs[Int] & Envs[Double])) =
```
hearnadam, line 58: Isn't this still relevant?

hearnadam (commented):
Very impressive performance improvement!

The change looks pretty good overall. Will take a closer look at the tests later.

Does izumi have tests we can copy? It would be good to highlight the overlap in functionality, as well as where Kyo tags may won't work (should be rare in practice).

I wasn't able to tell, but can you still invoke `==` on two tags? Can we somehow disallow this?

kyo-core/shared/src/main/scala/kyo/tags.scala
```diff
@@ -1,41 +1,206 @@
...
 
-opaque type Tag[T] = String
+case class Tag[T](tpe: String) extends AnyVal
 
 object Tag:
 
-    given canEqual[T, U]: CanEqual[Tag[T], Tag[U]] = CanEqual.derived
+    import internal.*
+
+    inline given apply[T >: Nothing]: Tag[T] = ${ tagImpl[T] }
```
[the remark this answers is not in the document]
fwbrasil: Good question! I think we actually shouldn't allow `Tag[Nothing]`. In Kyo's usage of tags, it's essentially an inconsistent state since tags are meant to represent concrete types, and `Nothing`, although it has a class symbol, has no runtime representation. I've changed the macro to fail for `Tag[Nothing]` and added a test for it. It caught an inssue in `streamsTest`! I've removed the type constraint since it isn't strict, and the macro will detect it anyway.

kyo-core/shared/src/main/scala/kyo/tags.scala
```diff
@@ -1,41 +1,206 @@
...
+
+    extension [T](t1: Tag[T])
+
+        def show = t1.tpe.takeWhile(_ != ';')
+
+        def <:<[U](t2: Tag[U]): Boolean =
+            t1.tpe == t2.tpe || isSubtype(t1, t2)
```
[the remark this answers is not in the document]
fwbrasil: Nice catch! It doesn't, fixed.

kyo-core/shared/src/main/scala/kyo/tags.scala
```diff
@@ -1,41 +1,206 @@
...
+
+        def =:=[U](t2: Tag[U]): Boolean =
+            t1.tpe == t2.tpe
+
+        def =!=[U](t2: Tag[U]): Boolean =
+            t1.tpe != t2.tpe
+
+        def >:>[U](t2: Tag[U]): Boolean =
```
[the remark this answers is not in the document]
fwbrasil: I prefer the freedom to order operands in both directions, it sometimes make the code more linear.

kyo-core/shared/src/main/scala/kyo/tags.scala
```diff
@@ -1,41 +1,206 @@
...
+                        end if
+                    }
+                end loop
+                loop(subStart + 1, superStart + 1)
+            end if
+        end checkParams
+
+        def nextParam(tag: String, idx: Int): Int =
+            @tailrec def loop(opens: Int, idx: Int): Int =
+                tag.charAt(idx) match
```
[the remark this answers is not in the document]
fwbrasil: Thanks! I need to get used to this annotation. I had to avoid the pattern guards to add the annotation but it should still improve performance. I added it in a couple of other places as well.

kyo-core/shared/src/test/scala/kyoTest/envsTest.scala
fwbrasil: Both scenarios fail because of the intersection tag and scenarios that test compilation failures are relatively fragile since they can break for different reasons as the codebase evolves. I don't see much value in keeping them. I'd say this is more of a `wontfix` for now but we can review if we explore supporting intersection and union types in tags.

fwbrasil:
Thanks for the review! 🙏

> Does izumi have tests we can copy? It would be good to highlight the overlap in functionality, as well as where Kyo tags may won't work (should be rare in practice).

Izumi can represent many more kinds types than Kyo's tags but I think it's actually a good thing to have something more constrained. There are a few tests for unsupported types.

> I wasn't able to tell, but can you still invoke `==` on two tags? Can we somehow disallow this?

It's only possible to restrict it if the compiler is configured with strict equality enabled, which is the case in Kyo's codebase. If strict equality is enabled, `==` won't be available since `Tag` doesn't provide a `CanEqual`.

kyo-core/shared/src/main/scala/kyo/tags.scala
```diff
@@ -1,41 +1,214 @@
...
+                        case '[tpe] =>
+                            Expr.summon[Tag[tpe]] match
+                                case None =>
+                                    report.errorAndAbort(
+                                        s"Please provide an implicit ${TypeRepr.of[Tag[tpe]].show}"
+                                    )
+                                case Some(value) =>
+                                    '{ $value.tpe }
+                        case _ =>
+                            report.errorAndAbort(s"Unsupported Tag type ${tpe.show}")
```
hearnadam, line 192: Can we make the error messages unique to assist with debugging issues? When users submit tickets it may be useful to know which case they are hitting

kyo-core/shared/src/test/scala/kyoTest/streamsTest.scala
```diff
@@ -321,7 +321,7 @@ class streamsTest extends KyoTest:
         "none" in {
             assert(
                 Streams.initSeq(Seq(1, 2, 3)).collect {
-                    case v if false => ???
+                    case v if false => (): Unit < Streams[Int]
```
hearnadam, line 324: Do we need a generalized unit method to improve inference?

```
object Kyo:
  def unit[S]: Unit < S = ()
```

kyo-core/shared/src/main/scala/kyo/tags.scala
hearnadam: Perhaps worth a pending test!

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/tags.scala
fwbrasil: Makes sense! I've just added it.

kyo-core/shared/src/test/scala/kyoTest/streamsTest.scala
fwbrasil: I'd say this is an edge case for testing. I think this compilation error would happen most often when users make mistakes for example not composing effects correctly.

hearnadam (approved):

fwbrasil:
Thanks for the reviews @hearnadam! I'll merge this and work on support for union and intersection types in a separate PR.

PR #349, opened 2024-05-11
flat check fixes

fwbrasil:
Fixes https://github.com/getkyo/kyo/issues/327.

Kyo's `Flat` check was designed to have a complex derivation mechanism to provide better error messages in case of a mismatch between the effects a method supports vs the pending effects of its parameters. This derivation mechanism is unsound. This PR removes it.

This results in a regression in usability since error messages don't explicitly show where the mismatch is anymore. It seems a reasonable tradeoff that we can address later.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Flat.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Flat.scala
@@ -2,19 +2,10 @@ package kyo
 
 import internal.FlatImplicits
 
-sealed trait Flat[-T]
+opaque type Flat[T] = Null
 
 object Flat extends FlatImplicits:
-
-    inline def derive[T: Flat, S]: Flat[T < S] = Flat.unsafe.bypass
-
-    private val cached = new Flat[Any] {}
-
-    given unit[S]: Flat[Unit < S] = unsafe.bypass[Unit < S]
-
     object unsafe:
-
-        inline given bypass[T]: Flat[T] =
-            cached.asInstanceOf[Flat[T]]
+        inline given bypass[T]: Flat[T] = null
     end unsafe
 end Flat
--- a/kyo-core/shared/src/main/scala/kyo/internal/FlatImplicits.scala
+++ b/kyo-core/shared/src/main/scala/kyo/internal/FlatImplicits.scala
@@ -3,17 +3,9 @@ package kyo.internal
 import kyo.*
 import scala.quoted.*
 
-trait FlatImplicits0:
+trait FlatImplicits:
     inline given infer[T]: Flat[T] =
         ${ FlatImplicits.inferMacro[T] }
-
-trait FlatImplicits1 extends FlatImplicits0:
-    given product[T <: Product]: Flat[T] =
-        Flat.unsafe.bypass[T]
-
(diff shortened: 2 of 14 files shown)


kyo-core/shared/src/main/scala/kyo/Flat.scala
```diff
@@ -2,19 +2,10 @@ package kyo
 
 import internal.FlatImplicits
 
-sealed trait Flat[-T]
+opaque type Flat[T] = Null
```
fwbrasil, line 5: This encoding saves a few bytecodes in `core`.

kyo-core/shared/src/main/scala/kyo/Flat.scala
```diff
@@ -2,19 +2,10 @@ package kyo
...
 
 object Flat extends FlatImplicits:
-
-    inline def derive[T: Flat, S]: Flat[T < S] = Flat.unsafe.bypass
-
-    private val cached = new Flat[Any] {}
-
-    given unit[S]: Flat[Unit < S] = unsafe.bypass[Unit < S]
```
fwbrasil, line 13: I think this was necessary only for Scala 2.

kyo-core/shared/src/main/scala/kyo/internal/FlatImplicits.scala
```diff
@@ -43,33 +35,27 @@ object FlatImplicits:
...
-                            s"Cannot prove ${code(print(t))} isn't nested. Provide an implicit evidence ${code(s"kyo.Flat[${print(t)}]")}."
+                            s"Cannot prove ${code(print(t))} isn't nested. " +
+                                s"This error can be reported an unsupported pending effect is passed to a method. " +
+                                s"If that's not the case, provide an implicit evidence ${code(s"kyo.Flat[${print(t)}]")}."
                         )
-            else
-                '{ Flat.unsafe.bypass[T] }
 
         t match
             case Kyo(Kyo(nt, s1), s2) =>
```
fwbrasil, line 61: This case shouldn't happen naturally anymore but I think it's good to keep the logic since we'll likely try to improve the error messages in the future.

kyo-core/shared/src/main/scala/kyo/Flat.scala
fwbrasil: I also tried using the new `erased` modifier but the build times increased too much.

hearnadam: How about `AnyVal`? I don't love null, given that it may be difficult to debug if any methods are invoked on the bypass instance.

hearnadam (approved):
Very nice! Seems like a huge improvement!

kyo-core/shared/src/main/scala/kyo/Flat.scala
fwbrasil: It could be an issue if someone calls `flat.toString` for example but that wouldn't make much sense. Ideally, this type should have no runtime representation with `erased`. Using `Null` here reduces the core size because it gets inlined as a `null` constant. I had used `Unit` initially but just pointing to an actual object requires more instructions.

PR #356, opened 2024-05-12
move effect command from member type to type parameter

fwbrasil:
I'm working on a follow up of https://github.com/getkyo/kyo/pull/348 to rely on the new `Tag` in `Aborts`. Something that has been challenging when developing effects is dealing with variance and the `Command` defined as type member:

```scala
class Aborts[-V] extends Effect[Aborts[V]]:
    type Command[T] = Left[V, ?] // compilation error because V is contravariant
```

This fails to compile because `V` is contravariant and type members are invariant. It's possible to work it around by making `Command` opaque but then it's necessary to provide bridge methods.

This PR changes the encoding of effects to use a type parameter for the command instead of a type member, which solves this encoding issue with variance.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,6 +10,7 @@ val compilerOptions = Seq(
     "-feature",
     "-unchecked",
     "-deprecation",
+    "-Ykind-projector",
     "-language:implicitConversions",
     "-Wvalue-discard",
     "-Wunused:all",
@@ -85,7 +86,8 @@ lazy val `kyo-scheduler` =
             scalacOptions --= Seq(
                 "-Wvalue-discard",
                 "-Wunused:all",
-                "-language:strictEquality"
+                "-language:strictEquality",
+                "-Ykind-projector"
             ),
             scalacOptions += "-Xsource:3",
             crossScalaVersions                      := List(scala3Version, scala212Version, scala213Version),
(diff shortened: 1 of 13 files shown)


build.sbt
```diff
@@ -10,6 +10,7 @@ val compilerOptions = Seq(
     "-feature",
     "-unchecked",
     "-deprecation",
+    "-Ykind-projector",
```
fwbrasil, line 13: To make it easier to define commands.

kyo-core/shared/src/main/scala/kyo/vars.scala
```diff
@@ -54,3 +49,12 @@ object Vars:
     def run[V >: Nothing]: RunDsl[V] = new RunDsl[V]
 
 end Vars
+
+// Moving to Vars.internal crashes the compiler
```
fwbrasil, line 53: I get a cyclic type error for `Flat` if I do this, which doesn't make much sense. It seems a compiler bug.

fwbrasil:
I'm merging this to unblock other work but please feel free to provide feedback!

fwbrasil:
FYI This change introduced other complexities and didn't fully solve the variance issue. I've reverted it on main.

PR #357, opened 2024-05-12
move Tag to separate module

fwbrasil:
Moves the new `Tag` to a separate module so it can be reused in isolation by other projects.


fwbrasil:
I've ported `Aborts` and `Envs` to the new `Tag` but there was a performance regression due to the tag subtype checking cost. I'm merging this to work on optimizations.

PR #359, opened 2024-05-13
Tag optimizations

fwbrasil:
I'm working on reimplementing `Aborts` and `Envs` based on the new `Tag` but there's a performance regression. This PR introduces optimizations to address it:

1. Each type in a segment is now prefixed with its size encoded as `char`, which avoids scans to find the next element.
2. A hash of the type name is also encoded as a `char` after the length, which enables fast mismatches. For example, when checking `scala.Int` against `scala.Double`, the previous implementation would first compare `scala.` to then find out the type doesn't match. Hash conflicts aren't a concern since, in case of match, the entire type string is checked for equality.
3. Introduces a mechanism to compact the representation of common types.

This change introduces a restriction: type names with more than 94 characters aren't supported. I had some trouble with encoding arbitrary char values into strings because the UTF-8/Latin string encodings produce garbage in some cases since there are control characters. I decided to use the ASCII range of printable chars, which should be the same regardless of the string encoding. It seems a reasonable tradeoff, especially since the issue is detected at compile time.

Benchmark:

image
https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/6614ae6cf60bccdc5aa52eebc5196213/raw/785a7a001111974fb986eed4cfe7f77079f1dba3/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/6614ae6cf60bccdc5aa52eebc5196213/raw/785a7a001111974fb986eed4cfe7f77079f1dba3/jmh-result-candidate.json

### DIFF
--- a/kyo-tag/shared/src/main/scala/kyo/tags.scala
+++ b/kyo-tag/shared/src/main/scala/kyo/tags.scala
@@ -14,7 +14,9 @@ object Tag:
 
     extension [T](t1: Tag[T])
 
-        def show = t1.tpe.drop(2).takeWhile(_ != ';')
+        def show =
+            val decoded = t1.tpe.drop(2).takeWhile(_ != ';')
+            fromCompact.getOrElse(decoded, decoded)
 
         def <:<[U](t2: Tag[U]): Boolean =
             t1 =:= t2 || isSubtype(t1, t2)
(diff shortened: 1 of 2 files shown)


kitlangton:
> type names with more than 94 characters

An unprovoked attack on Java programmers!

How will we ever port [Fizz Buzz Enterprise Edition](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition) to Kyo?

---

jk. Super impressive, Flavio!

fwbrasil:
> > type names with more than 94 characters
> 
> An unprovoked attack on Java programmers!
> 
> How will we ever port [Fizz Buzz Enterprise Edition](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition) to Kyo?
> 
> jk. Super impressive, Flavio!

haha it's hard to tell if it's a limitation or a feature :P

PR #361, opened 2024-05-13
workaround to declare effects as opaque

fwbrasil:
As mentioned in https://github.com/getkyo/kyo/pull/360, I've identified an encoding that allows us to declare effects as `opaque` without breaking the direct syntax. It's not ideal but at least unblocks https://github.com/getkyo/kyo/pull/335.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/aborts.scala
+++ b/kyo-core/shared/src/main/scala/kyo/aborts.scala
@@ -4,13 +4,13 @@ import kyo.core.*
 import scala.reflect.ClassTag
 import scala.util.control.NonFatal
 
-// Can't use an opaque type because
-// kyo-direct tests crash the compiler.
-type Aborts[-V] //= DoAbort
+type Aborts[-V] >: Aborts.Effects[V] <: Aborts.Effects[V]
 
 object Aborts:
 
     import internal.*
+
+    opaque type Effects[-V] = DoAbort
 
     def fail[V](v: V): Nothing < Aborts[V] =
         DoAbort.suspend(v).asInstanceOf[Nothing < Aborts[V]]


kyo-core/shared/src/main/scala/kyo/options.scala
```diff
@@ -1,11 +1,16 @@
 package kyo
 
-type Options = Aborts[None.type]
+type Options >: Options.Effects <: Options.Effects
 
 object Options:
 
+    private class Empty
```
fwbrasil, line 7: I've added a separate type to abort with since users can use `Aborts.fail(None)`, which could conflict with `Options`. This is unrelated to the opaque type change.

hearnadam: I don't mind if `Aborts.fail(None) == Options.empty`...

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/options.scala
fwbrasil: Thinking it through, the change to `Empty` is actually necessary if the effect is `opaque`. Example:

```scala
val a: Either[None, Int] < Options = Aborts.run[None.type](Options.get(Option.empty[Int]))
val b: Option[Either[None, Int]] = Options.run(a)
```

The `Aborts.run` would handle `Options` since the tag will match but keep it pending since it's `opaque`. If it's not `opaque`, that's not an issue. I'm not sure there's much benefit in keeping `Options` as `opaque`, though. It'd allow us to evolve the effect with different internal encodings, even not based on `Aborts`, but it doesn't seem very likely. I'll revert it.

Issue #347, opened 2024-05-09
Fibers Regression in v0.9.3


hearnadam:
[EasyRacer](https://github.com/jamesward/easyracer/) is facing several regressions when attempting to update to v0.9.3.

You can see the test run, as each scenario's duration here:
v0.9.3
https://github.com/jamesward/easyracer/actions/runs/9020984988/job/24787301439

v0.9.2
https://github.com/jamesward/easyracer/actions/runs/9020946688/job/24787191885

As discussed offline, I believe this may be a bug in the new scheduler, though have no specific evidence of this.

Issue #329, opened 2024-05-06
Implement HTTP Client Benchmarks for ZIO


fwbrasil:
background: https://github.com/getkyo/kyo/pull/328

A solution for this issue must ensure the same HTTP client instance is reused across benchmark iterations.

PR #364, opened 2024-05-14
http client race bench

fwbrasil:
This benchmark is similar to easyracer's scenario in https://github.com/getkyo/kyo/issues/347.

Please ignore ZIO's results for now since it's disabled. See https://github.com/getkyo/kyo/issues/329 for more context.

image
https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/8e5f0b82abc1aff56bdea686b3008be4/raw/2975084ca13768ca7321f0401f344116ba07a3d4/jmh-result.json

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/HttpClientRaceContentionBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/HttpClientRaceContentionBench.scala
@@ -1,0 +1,71 @@
+package kyo.bench
+
+import org.http4s.ember.client.EmberClientBuilder
+
+class HttpClientRaceContentionBench
+    extends Bench.ForkOnly("pong"):
+
+    val port        = 9999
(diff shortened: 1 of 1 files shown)


kyo-bench/src/main/scala/kyo/bench/HttpClientRaceContentionBench.scala
```diff
@@ -0,0 +1,71 @@
+package kyo.bench
+
+import org.http4s.ember.client.EmberClientBuilder
+
+class HttpClientRaceContentionBench
+    extends Bench.ForkOnly("pong"):
+
+    val port        = 9999
+    val concurrency = 100
```
fwbrasil, line 9: We can't use 10k like the scenario here since the benchmark would take too long.

hearnadam (approved):

PR #365, opened 2024-05-14
kyo-sttp: avoid translating Interrupted exceptions

fwbrasil:
Sttp has an expensive mechanism to [translate exceptions](https://github.com/softwaremill/sttp/blob/v3/core/src/main/scalajvm/sttp/client3/SttpClientExceptionExtensions.scala) to its own types. Since the method is [used as a partial function](https://github.com/softwaremill/sttp/blob/a51c3bb1d047418d1bfd7fe256632b227c15553e/core/src/main/scala/sttp/client3/SttpClientException.scala#L31), each exception handling produces two exceptions: one for `PartialFunction.isDefinedAt` and another for `PartialFunction.apply`. This is quite expensive because the exception types aren't marked as `NoStackTrace` and produce new stack traces on each instantiation. I'll report this issue upstream.

This PR introduces a workaround to avoid translating fiber interruptions specifically since it's the scenario present in https://github.com/getkyo/kyo/issues/347.

I'm not able to show a performance improvement with the new benchmark in https://github.com/getkyo/kyo/pull/364 because there is another larger performance issue that I'm investigating.

### DIFF
--- a/kyo-sttp/shared/src/main/scala/kyo/internal/KyoSttpMonad.scala
+++ b/kyo-sttp/shared/src/main/scala/kyo/internal/KyoSttpMonad.scala
@@ -21,7 +21,15 @@ object KyoSttpMonad:
             protected def handleWrappedError[T](rt: T < Fibers)(
                 h: PartialFunction[Throwable, T < Fibers]
             ) =
-                IOs.catching(rt)(h)
+                IOs.catching(rt) {
+                    case ex if ex eq Fibers.Interrupted =>
+                        Fibers.interrupted
+                    case ex if h.isDefinedAt(ex) =>
+                        h(ex)
+                }
+
+            override def handleError[T](rt: => T < Fibers)(h: PartialFunction[Throwable, T < Fibers]) =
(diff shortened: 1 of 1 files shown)


kyo-sttp/shared/src/main/scala/kyo/internal/KyoSttpMonad.scala
```diff
@@ -21,7 +21,15 @@ object KyoSttpMonad:
...
             ) =
-                IOs.catching(rt)(h)
+                IOs.catching(rt) {
+                    case ex if ex eq Fibers.Interrupted =>
+                        Fibers.interrupted
+                    case ex if h.isDefinedAt(ex) =>
+                        h(ex)
+                }
+
+            override def handleError[T](rt: => T < Fibers)(h: PartialFunction[Throwable, T < Fibers]) =
```
fwbrasil, line 31: This method is used internally by sttp. I had to override it as well to avoid the translation in some codepaths.

hearnadam (approved):

PR #367, opened 2024-05-14
cache stream handlers

fwbrasil:
Avoids handler allocations in `Streams`. The benchmarks show no improvement but it's because the JIT is able to avoid the handler allocations. I think it's still worth making the change since that might not be the case in production code.

image

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/core.scala
+++ b/kyo-core/shared/src/main/scala/kyo/core.scala
@@ -130,35 +130,35 @@ object core:
         def accepts[T](st: State, command: Command[T]): Boolean =
             true
 
-        def done[T](st: State, v: T): Result[T] < S
+        def done[T](st: State, v: T)(using Tag[E]): Result[T] < S
 
-        def failed(st: State, ex: Throwable): Nothing < (E & S) = throw ex
+        def failed(st: State, ex: Throwable)(using Tag[E]): Nothing < (E & S) = throw ex
 
         def resume[T, U: Flat, S2](
             st: State,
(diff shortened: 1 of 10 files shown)


kyo-core/shared/src/main/scala/kyo/core.scala
```diff
@@ -130,35 +130,35 @@ object core:
         def accepts[T](st: State, command: Command[T]): Boolean =
             true
 
-        def done[T](st: State, v: T): Result[T] < S
+        def done[T](st: State, v: T)(using Tag[E]): Result[T] < S
```
fwbrasil, line 133: These changes allows cached handlers to perform new suspensions of the effect by using the provided `Tag`. It's a downside that we'll have to specify the parameter in all handlers but I think it's worth it since it enables allocationless effect handling in more scenarios.

kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -143,7 +51,7 @@ object Stream:
                     def resume[T, U: Flat, S3](
                         command: Chunk[V],
                         k: T => U < (Streams[V] & S3)
-                    ) =
+                    )(using Tag[Streams[V]]) =
                         command.collectUnit(pf).andThen {
```
fwbrasil, line 55: Some of the methods that depend on a function and an internal state can't be cached because we'd need to allocate tuples on each effect handling, which is more expensive than allocating the handler.

hearnadam (approved):

PR #371, opened 2024-05-16
avoid `lazy val` to access `KyoSttpMonad`

fwbrasil:
Scala 3 automatically uses a `lazy val` in the desugaring of `given` statements. This isn't a major optimization but it avoids the `lazy val` by using `inline`.


fwbrasil:
merging to post a follow up PR

PR #372, opened 2024-05-16
scheduler: ensure workers stop when inactivated

fwbrasil:
Kyo's concurrency regulator can reduce concurrency and make workers inactive. Inactive workers are automatically drained in multiple code paths and don't receive new workload but, if a worker has a long-running task, it can continue running until it finishes. This change makes workers exit as soon as possible.

### DIFF
--- a/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Worker.scala
+++ b/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Worker.scala
@@ -10,6 +10,7 @@ import scala.util.control.NonFatal
 
 final private class Worker(
     id: Int,
+    stop: Int => Boolean,
     exec: Executor,
     scheduleTask: (Task, Worker) => Unit,
     stealTask: Worker => Task,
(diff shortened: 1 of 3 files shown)


kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Worker.scala
```diff
@@ -90,18 +91,19 @@ final private class Worker(
         available
     }
 
-    def isStalled(cycles: Long): Boolean =
+    private def isStalled(cycles: Long): Boolean =
```
fwbrasil, line 94: Minor refactoring. This method didn't need to be public.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Worker.scala
```diff
@@ -90,18 +91,19 @@ final private class Worker(
...
 
-    def isBlocked(): Boolean = {
-        val mount = this.mount
-        mount != null && {
-            val state = mount.getState().ordinal()
-            state == Thread.State.BLOCKED.ordinal() ||
-            state == Thread.State.WAITING.ordinal() ||
-            state == Thread.State.TIMED_WAITING.ordinal()
+    private def isBlocked(): Boolean =
+        running && {
```
fwbrasil, line 98: a minor unrelated fix: the top utility was incorrectly reporting inactive workers as blocked.

fwbrasil:
merging to finalize the fix for https://github.com/getkyo/kyo/issues/347

PR #373, opened 2024-05-16
avoid `core.eval` for error handling

fwbrasil:
This is the main cause of the regression in https://github.com/getkyo/kyo/issues/347. The `eval` method introduces much more overhead than the previous dedicated loop. The issue is that function applications don't get inlined and producing a thunk to provide to `resume` introduces more indirection, preventing JIT optimizations.

I'll post benchmark results tomorrow.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/core.scala
+++ b/kyo-core/shared/src/main/scala/kyo/core.scala
@@ -106,6 +106,29 @@ object core:
         end handle
     end extension
 
+    inline def catching[T, S, U >: T, S2](v: => T < S)(
+        inline pf: PartialFunction[Throwable, U < S2]
+    ): U < (S & S2) =
+        def catchingLoop(v: U < (S & S2)): U < (S & S2) =
+            v match
+                case kyo: Suspend[MX, Any, U, S & S2] @unchecked =>
+                    new Continue[MX, Any, U, S & S2](kyo):
+                        def apply(v: Any < (S & S2), s: Safepoint[S & S2], l: Locals.State) =
+                            try
+                                catchingLoop(kyo(v, s, l))
+                            catch
+                                case ex: Throwable if (NonFatal(ex) && pf.isDefinedAt(ex)) =>
+                                    pf(ex)
+                case _ =>
+                    v.asInstanceOf[T]
+        try
+            catchingLoop(v)
+        catch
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/ios.scala
```diff
@@ -36,22 +36,15 @@ sealed trait IOs extends Effect[IOs]:
...
-        eval(v.map(Success(_): Try[T]))((s, k) =>
-            try k()
-            catch
-                case ex if NonFatal(ex) =>
-                    Failure(ex)
-        )
+        core.catching(v.map(Success(_): Try[T])) {
+            case ex if NonFatal(ex) =>
+                Failure(ex)
+        }
```
hearnadam, line 42: There is no longer a need to match on `NonFatal` here - that's already checked in the loop.

kyo-core/shared/src/main/scala/kyo/core.scala
```diff
@@ -106,6 +106,29 @@ object core:
...
+                            catch
+                                case ex: Throwable if (NonFatal(ex) && pf.isDefinedAt(ex)) =>
+                                    pf(ex)
+                case _ =>
+                    v.asInstanceOf[T]
+        try
+            catchingLoop(v)
+        catch
+            case ex: Throwable if (NonFatal(ex) && pf.isDefinedAt(ex)) =>
+                pf(ex)
```
hearnadam, line 128: This outer try/catch appears unnecessary if you make `v` a function, but from the description that seems intentional.

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/core.scala
fwbrasil: It's placed to catch exceptions in the evaluation of the `v` thunk since it's a by-name parameter. Making it a no-arg function is equivalent to a by-name param in the generated bytecode and we'd still need to evaluate the function to decide if it's necessary to suspend or not (`catchingLoop`). Although these two catches seem close in the method, they're placed in very different execution paths, which is helpful to specialize the code for JIT compilation. This seems the key characteristic that enables the better performance in comparison to `eval`, which was making the execution become much more dynamic.

PR #377, opened 2024-05-16
kyo-benchs: migrate TestHttpServer to Vert.x

fwbrasil:
The Java server became a bottleneck in the http contention benchmarks. This PR migrates it to Vert.x since it's highly optimized. We can't use `kyo-tapir` since it'd be difficult to distinguish between performance changes in the client vs server.

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/TestHttpServer.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/TestHttpServer.scala
@@ -1,64 +1,113 @@
 package kyo.bench
 
-import com.sun.net.httpserver.HttpExchange
-import com.sun.net.httpserver.HttpHandler
-import com.sun.net.httpserver.HttpServer
-import java.io.OutputStream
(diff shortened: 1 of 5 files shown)


kyo-bench/src/main/scala/kyo/bench/TestHttpServer.scala
```diff
@@ -1,64 +1,113 @@
...
             )
+            redirect(process.getInputStream, System.out)
+            redirect(process.getErrorStream, System.err)
             s"http://localhost:$port/ping"
         finally
             log(port, "forked")
         end try
     end start
 
+    def redirect(inputStream: InputStream, outputStream: PrintStream): Unit =
```
fwbrasil, line 45: A new mechanism to show that console messages are coming from the forked http server.

fwbrasil:
Merging to finalize the fix for https://github.com/getkyo/kyo/issues/347.

PR #378, opened 2024-05-16
kyo-tag: support type unions and intersections

fwbrasil:
We don't need to support tags for union and intersection types in most scenarios but the ZIO integration needs it in order to integrate environment values between the libraries. This PR introduces support for them but using wrapper APIs so tags still aren't materialized for unions and intersections. The implementation isn't optimized but that should be ok considering the limited use.

### DIFF
--- a/kyo-tag/shared/src/main/scala/kyo/tags.scala
+++ b/kyo-tag/shared/src/main/scala/kyo/tags.scala
@@ -14,23 +14,97 @@ object Tag:
 
     extension [T](t1: Tag[T])
 
-        def show =
+        def show: String =
             val decoded = t1.tpe.drop(2).takeWhile(_ != ';')
             fromCompact.getOrElse(decoded, decoded)
 
-        def <:<[U](t2: Tag[U]): Boolean =
+        infix def <:<[U](t2: Tag[U]): Boolean =
             t1 =:= t2 || isSubtype(t1, t2)
 
-        def =:=[U](t2: Tag[U]): Boolean =
+        infix def =:=[U](t2: Tag[U]): Boolean =
             (t1.tpe eq t2.tpe) || t1.tpe == t2.tpe
 
-        def =!=[U](t2: Tag[U]): Boolean =
+        infix def =!=[U](t2: Tag[U]): Boolean =
             (t1.tpe ne t2.tpe) || t1.tpe != t2.tpe
 
-        def >:>[U](t2: Tag[U]): Boolean =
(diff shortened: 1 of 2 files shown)


kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -14,24 +14,98 @@ object Tag:
 
     extension [T](t1: Tag[T])
 
-        def show =
+        def show: String =
             val decoded = t1.tpe.drop(2).takeWhile(_ != ';')
             fromCompact.getOrElse(decoded, decoded)
 
-        def <:<[U](t2: Tag[U]): Boolean =
+        infix def <:<[U](t2: Tag[U]): Boolean =
```
hearnadam, line 21: The compiler seems quite finicky when it requires the infix keyword...

hearnadam (approved):
Very nice! This will be useful.

I didn't get a chance to look at the macros closely, but I can always followup there later.

I generally have 2 comments:

1. Can we create a unified type + macro? I think it should be possible to use an opaque type to unify the tags, then a macro which 'dispatches' based on if the type is union/intersection/neither. For performance sensitive code, we can specify only the string tags in the type signature.
2. Do you think it's possible to encode this in a string? Or not worth the effort?

kyo-tag/shared/src/main/scala/kyo/tags.scala
fwbrasil: yeah, it started complaining about it out of the blue. I imagine it could be a bad interaction with extension methods.

fwbrasil:
> 1. Can we create a unified type + macro? I think it should be possible to use an opaque type to unify the tags, then a macro which 'dispatches' based on if the type is union/intersection/neither. For performance sensitive code, we can specify only the string tags in the type signature.

The separation of `Tag` not being able to infer union and intersection types is an important property for Kyo's use cases but, yes, I think we could have a single unified implementation. We could even evolve into something more generic that allows us to express more constraints about the types that can be tagged. 

To exemplify, if ZIO wanted to have a `Tag` that is not able to express instances of `ZIO` itself similarly to Kyo's flat check, we could have an API to specify a tag for that:

```scala
type Tag[+T] = Tag.Excluding[ZIO]
```

We'd need to figure out a good type-level API but it seems it could even have more complex rules.

> 2. Do you think it's possible to encode this in a string? Or not worth the effort?

It is, I didn't do that here because I felt it wasn't necessary and increases the complexity of the implementation. We should also refactor the code into multiple files/classes since it's getting too large in a single file.

PR #379, opened 2024-05-16
allow users to define custom effects

fwbrasil:
The new `Tag` solved some safety issues when defining new effects. I think we'll still need to improve the usability of the APIs but I think it's in a state where we can allow users to start using it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2100,6 +2100,104 @@ For further examples, Kyo's [example ledger service](https://github.com/getkyo/k
 
 Coming soon..
 
+## Defining New Effects (Advanced)
+
+While Kyo provides a wide range of built-in effects, there may be situations where you need to define your own custom effects. Defining new effects is an advanced feature that most users typically won't need to deal with, but it can be useful in certain scenarios where you want to extend the functionality of Kyo or create domain-specific effects.
+
+To define a new effect, you need to create a subclass of the `Effect` trait and specify the effect's `Command` type. The `Command` type represents the data associated with the effect's operations. Here's an example of creating a simple `Counter` effect:
+
+```scala
+import kyo.*
+import kyo.core.* // needed to define effects
+
+class Counter extends Effect[Counter]:
+  type Command[T] = (Int, T)
+```
+
+In this example, `Counter` is a new effect that uses `(Int, T)` as its `Command` type, where `Int` represents the amount to increment or decrement the counter, and `T` is the value to be returned after updating the counter. Next, let's define the operations for the `Counter` effect in its companion object:
+
+```scala mdoc:nest
+object Counter extends Counter:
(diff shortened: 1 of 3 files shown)


kyo-core/shared/src/main/scala/kyo/core.scala
```diff
@@ -104,6 +104,26 @@ object core:
                         r.asInstanceOf[Result[T]]
             handleLoop(state, value)
         end handle
+
+        inline def deepHandle[Command[_], S, T: Flat](
```
fwbrasil, line 108: Moved to a public scope.

README.md
```diff
@@ -2100,6 +2100,104 @@ For further examples, Kyo's [example ledger service](https://github.com/getkyo/k
...
+    Counter.dec(2)(10)
+  }
+
+// Typically wrapped in a `run`
+// method by Kyo effects
+val b: Int < Any = 
+  Counter.handle(handler)((), a)
+```
+
+In this example, `a` is a computation prints `5` and `2` and finally returns the value `10`. We handle the `Counter` effect using `Counter.run`, which returns a tuple containing the final count and the result of the computation.
```
hearnadam, line 2156: Maybe we can just show the console output.

README.md
```diff
@@ -2100,6 +2100,104 @@ For further examples, Kyo's [example ledger service](https://github.com/getkyo/k
...
+```scala mdoc:nest
+val a: Int < Counter = 
+  Counter.inc(5) {
+    Counter.dec(2)(10)
+  }
+
+// Typically wrapped in a `run`
+// method by Kyo effects
+val b: Int < Any = 
+  Counter.handle(handler)((), a)
```
hearnadam, line 2153: Perhaps we can define a run method? that way users can understand better the goal.

README.md
```diff
@@ -2100,6 +2100,104 @@ For further examples, Kyo's [example ledger service](https://github.com/getkyo/k
...
+```
+
+The `inc` and `dec` operations suspend the `Counter` effect with the provided amount and the computation `v`. The `inc` operation increments the counter by the specified amount, while the `dec` operation decrements it.
+
+### Effect Handlers
+
+Effect handlers specify how to interpret and execute the operations of an effect. In Kyo, it's a common style pattern to define effect handling using a `run` method in the effect's companion object. Here's an example of defining the run method and the effect handler for the `Counter` effect that prints commands to the console as an example:
+
+```scala mdoc:nest
+val handler = new Handler[[T] =>> (Int, T), Counter, Any]:
```
hearnadam, line 2135: Which object should this go in?

Should we encourage users to model handlers via extension?

fwbrasil:
I'm closing this for now since we're looking into core improvements https://github.com/getkyo/kyo/issues/462

PR #380, opened 2024-05-16
zios: readme

fwbrasil:
An initial readme entry for the effect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2096,6 +2096,62 @@ val a: Unit < Routes =
 
 For further examples, Kyo's [example ledger service](https://github.com/getkyo/kyo/tree/main/kyo-examples/jvm/src/main/scala/kyo/examples/ledger) provides practical applications of these concepts.
 
+### ZIOs: Integration with ZIO
+
+The `ZIOs` effect provides seamless integration between Kyo and the ZIO library. The effect is designed to enable gradual adoption of Kyo within a ZIO codebase. The integration properly suspends side effects and propagates fiber cancellations/interrupts between both libraries.
+
+```scala
+import kyo._
+import zio._
+
+// Use the 'get' method to extract a 'ZIO' effect
+val a: Int < ZIOs =
+  ZIOs.get(ZIO.succeed(42))
+
+// 'get' also supports error handling with 'Aborts'
+val b: Int < (Aborts[String] & ZIOs) =
+  ZIOs.get(ZIO.fail("error"))
+
+// Handle the 'ZIOs' effect to obtain a 'ZIO' effect
+val c: Task[Int] =
+  ZIOs.run(a)
+```
+
+Kyo and ZIO effects can be seamlessly mixed and matched within computations, allowing developers to leverage the power of both libraries. Here are a few examples showcasing this integration:
+
+```scala
+import kyo._
+import zio._
+
+// Note how ZIOs includes the
+// IOs and Fibers effects
+val a: Int < ZIOs = 
+  for
+    v1 <- ZIOs.get(ZIO.succeed(21))
+    v2 <- IOs(21)
+    v3 <- Fibers.init(-42).map(_.get)
+  yield v1 + v2 + v3
+
+// Using fibers from both libraries
+val b: Int < ZIOs = 
+  for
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2096,6 +2096,31 @@ val a: Unit < Routes =
 
 For further examples, Kyo's [example ledger service](https://github.com/getkyo/kyo/tree/main/kyo-examples/jvm/src/main/scala/kyo/examples/ledger) provides practical applications of these concepts.
 
+### ZIOs: Integration with ZIO
+
+The `ZIOs` effect provides seamless integration between Kyo and the ZIO library. The effect is designed to enable gradual adoption of Kyo within a ZIO codebase. It allows you to use ZIO effects within Kyo computations and vice-versa, leveraging the power and expressiveness of both solutions.
```
hearnadam, line 2101: > and vice-versa

We don't yet expose a way to run `Fibers` in the `ZIOs` loop, so I would prefer not to mention this yet. Though it should be safe to port the implementation I built for zio-test to a public method.

README.md
```diff
@@ -2096,6 +2096,31 @@ val a: Unit < Routes =
...
+
+```scala
+import kyo._
+import zio._
+
+// Use the 'get' method to extract a 'ZIO' effect
+val a: Int < (Aborts[Throwable] & ZIOs) =
+  ZIOs.get(ZIO.succeed(42))
```
hearnadam, line 2109: If you don't widen this, it should be inferred as

```scala
val a: Int < ZIOs
```
IMO it's preferable to show that the `succeed` is correctly 'observed'

README.md
```diff
@@ -2096,6 +2096,31 @@ val a: Unit < Routes =
...
+  ZIOs.get(ZIO.fail("error"))
+
+// Handle the 'ZIOs' effect to obtain a 'ZIO' effect
+val c: Task[Int] =
+  ZIOs.run(a)
+```
+
+The `ZIOs` effect is built on top of Kyo's `Fibers` effect, ensuring proper integration with Kyo's concurrency primitives. This means that ZIO effects can be used seamlessly within Kyo computations, benefiting from features like fiber management, interruption, and resource safety.
+
+> Note: Currently, the `ZIOs` effect doesn't support ZIO environments (`E` in `ZIO[R, E, A]`). Please handle ZIO environments before using the `ZIOs.get` method.
```
hearnadam, line 2122: > Please provide all ZIO environment dependencies before using the `ZIOs.get` method.

Maybe worth mentioning we plan an integration.

README.md
```diff
@@ -2096,6 +2096,31 @@ val a: Unit < Routes =
...
+val a: Int < (Aborts[Throwable] & ZIOs) =
+  ZIOs.get(ZIO.succeed(42))
+
+// 'get' also supports error handling with 'Aborts'
+val b: Int < (Aborts[String] & ZIOs) =
+  ZIOs.get(ZIO.fail("error"))
+
+// Handle the 'ZIOs' effect to obtain a 'ZIO' effect
+val c: Task[Int] =
+  ZIOs.run(a)
```
hearnadam, line 2117: I think it would be useful to show using `a` can be mapped with other kyo effects and correctly result in the correct ZIO result.

README.md
fwbrasil: I'm not sure I follow. The effect integrates fibers from both libraries: https://github.com/getkyo/kyo/blob/27eab354698b290b440c0bba37141747d631aed3/kyo-zio/shared/src/main/scala/kyo/zios.scala#L33

hearnadam: I understand that the effect uses fibers. ~~However, since the `run` function does not accept `Fibers`, how can I run this: `val a: Int < Fibers & ZIOs` to a: `Task[Int]`?~~

I realize now that the run function does accept `Fibers` - this is non obvious unless you look closely at the opaque type.

```
ZIOs.run(Fibers.sleep(10.seconds))
```

Might be a good opportunity to add examples 🙂 Optionally we can change the signature to accept `ZIOs & Fibers` - should be semantically equivalent

fwbrasil:
I've addressed the review feedback and added more examples.

hearnadam (approved):

PR #381, opened 2024-05-16
http race contention bench: use a meter to avoid overloading the http server

fwbrasil:
Fixes https://github.com/getkyo/kyo/issues/347.

This is the last PR to fix the regression we identified in the last release. [Benchmark results](https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/27c8abec86e947e9719d41a859deb5d2/raw/814ca4ebb3a1294f8dd7bbec9b54ea0957b92434/jmh-result.json):

image

@hearnadam EasyRacer might also need a meter to avoid overloading the server.

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/HttpClientRaceContentionBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/HttpClientRaceContentionBench.scala
@@ -45,7 +45,7 @@ class HttpClientRaceContentionBench
 
     lazy val kyoClient =
         import kyo.*
-        PlatformBackend.default
+        IOs.run(Meters.initSemaphore(5).map(PlatformBackend.default.withMeter))
 
     val kyoUrl =
         import sttp.client3.*


hearnadam:
Wow! This is very impressive. Perhaps we can add this to the default client backend. Users can always override it with Locals, right?

hearnadam (approved):

fwbrasil:
> Wow! This is very impressive. Perhaps we can add this to the default client backend. Users can always override it with Locals, right?

Kyo is a victim of its own success in this case :) It needs the meter because it's able to perform requests at a much faster pace. I imagine other libraries would have the same trouble if optimized.

I'm not sure we can set a reasonable default for the max concurrency since it's highly dependent on the specific use case. For example, if the benchmark was a single client being served by multiple servers, it might be able to run without a meter. If some production system can handle talking with larger services at a high rate of requests, it should be able to do so with a couple of pods if possible.

A direction we could explore is making the max concurrency adaptive. It seems the regulators in Kyo's scheduler could eventually evolve into a more generic solution for that.

PR #384, opened 2024-05-17
scheduler: minor optimizations

fwbrasil:
I'm working on tracking down the regression in ZIO's benchmarks when I enable the option use Kyo's scheduler and noticed a couple of minor optimization opportunities. There isn't a significant change in benchmark results but I think it's worth it since things can change over time.

### DIFF
--- a/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Queue.scala
+++ b/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Queue.scala
@@ -19,83 +19,101 @@ final private class Queue[T](implicit ord: Ordering[T]) extends AtomicBoolean {
         items
     }
 
-    def add(value: T): Unit =
-        modify {
+    def add(value: T): Unit = {
+        lock()
+        try {
             items += 1
-            queue.enqueue(value)
+            queue += value
             ()
-        }
+        } finally
+            unlock()
+    }
 
     def offer(value: T): Boolean =
-        tryModify {
-            items += 1
-            queue.enqueue(value)
-            true
+        tryLock() && {
+            try {
+                items += 1
+                queue += value
+                true
+            } finally
+                unlock()
         }
 
     def poll(): T =
         if (isEmpty())
             null.asInstanceOf[T]
-        else
-            modify {
+        else {
+            lock()
+            try {
                 if (isEmpty())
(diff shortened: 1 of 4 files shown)


kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Queue.scala
```diff
@@ -19,83 +19,101 @@ final private class Queue[T](implicit ord: Ordering[T]) extends AtomicBoolean {
         items
     }
 
-    def add(value: T): Unit =
-        modify {
+    def add(value: T): Unit = {
+        lock()
+        try {
             items += 1
-            queue.enqueue(value)
```
fwbrasil, line 25: This code was using `addOne` before the cross compilation to Scala 2.12, which doesn't have this method. I had changed it to `enqueue` but the underlying implementation allocates an `ArrayRef` and uses scala collection methods that have more overhead. The `+=` method is equivalent to `addOne`.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Queue.scala
```diff
@@ -19,83 +19,101 @@ final private class Queue[T](implicit ord: Ordering[T]) extends AtomicBoolean {
...
                 items = 0
                 queue.dequeueAll
-            }
+            } finally
+                unlock()
+        }
         tasks.foreach(f)
     }
 
-    private def modify[T](f: => T): T = {
```
fwbrasil, line 90: These methods were marked as `inline` before the cross compilation so I'm inlining them manually.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -176,7 +176,11 @@ final class Scheduler(
 
     private def ensureWorkers() =
         for (idx <- allocatedWorkers until currentWorkers) {
-            workers(idx) = new Worker(idx, _ >= currentWorkers, pool, schedule, steal, () => cycles, clock)
```
fwbrasil, line 179: Avoids dispatching calls to functions in the worker hot path.

hearnadam (approved):

PR #385, opened 2024-05-17
scala 3.4.2

fwbrasil:


kyo-core/shared/src/main/scala/kyo/loops.scala
```diff
@@ -17,18 +18,23 @@ object Loops:
     private val _continueUnit = Continue[Unit](())
 
     inline def continue[T]: Result[Unit, T] = _continueUnit
-    inline def done[T]: Result[T, Unit]     = ()
```
fwbrasil, line 20: The compiler doesn't line the `inline` overloaded methods anymore. This should be ok since the method will be inlined by the JIT compiler.

kyo-core/shared/src/test/scala/kyoTest/varsTest.scala
```diff
@@ -45,7 +45,7 @@ class varsTest extends KyoTest:
                         assert(outerValue == 1)
                     }
             }
-        }
+        }.pure
```
fwbrasil, line 48: Interesting that this was compiling before, it shouldn't.

hearnadam: Did this only happen after adding the `Int` type?

hearnadam (approved):

kyo-core/shared/src/test/scala/kyoTest/varsTest.scala
fwbrasil: Actually I could have removed the `Int` annotations. It was my initial attempt but then I saw `pure` was missing.

PR #386, opened 2024-05-18
kyo-scheduler-zio

fwbrasil:
Provides a module integrating Kyo's scheduler as an alternative for `ZScheduler` in ZIO applications.

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/Bench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/Bench.scala
@@ -25,24 +25,13 @@ import zio.UIO
 @BenchmarkMode(Array(Mode.Throughput))
 abstract class Bench[T](val expectedResult: T):
 
-    @Param(Array("false"))
-    var replaceZioExecutor = false
-
     lazy val zioRuntime =
-        import zio.*
+        val replaceZioExecutor =
+            System.getProperty("replaceZioExecutor", "false") == "true"
         if !replaceZioExecutor then
             zio.Runtime.default.unsafe
         else
-            given Unsafe = Unsafe.unsafe(identity)
-            val exec =
-                new Executor:
-                    val scheduler                     = kyo.scheduler.Scheduler.get
-                    def metrics(using unsafe: Unsafe) = None
-                    def submit(runnable: Runnable)(implicit unsafe: Unsafe): Boolean =
-                        scheduler.schedule(kyo.scheduler.Task(runnable.run()))
-                        true
-            val kExecutorLayer = Runtime.setExecutor(exec) ++ Runtime.setBlockingExecutor(exec)
-            Runtime.unsafe.fromLayer(kExecutorLayer).unsafe
+            KyoSchedulerRuntime.default.unsafe
         end if
     end zioRuntime
 end Bench
--- a/kyo-scheduler-zio/jvm/src/main/scala/kyo/KyoSchedulerRuntime.scala
+++ b/kyo-scheduler-zio/jvm/src/main/scala/kyo/KyoSchedulerRuntime.scala
@@ -1,0 +1,31 @@
+package kyo
+
+import kyo.scheduler.*
+import zio.*
+import zio.internal.ExecutionMetrics
+import zio.internal.FiberRuntime
+
+object KyoSchedulerRuntime {
+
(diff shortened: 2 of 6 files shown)


kyo-bench/src/main/scala/kyo/bench/Bench.scala
```diff
@@ -25,24 +25,13 @@ import zio.UIO
 @BenchmarkMode(Array(Mode.Throughput))
 abstract class Bench[T](val expectedResult: T):
 
-    @Param(Array("false"))
-    var replaceZioExecutor = false
```
fwbrasil, line 29: I've moved this config to a system property to make it easier to compare benchmark results. The JMH visualizer considers a benchmark not the same if there's any difference in configuration between runs.

kyo-scheduler-zio/jvm/src/main/scala/kyo/KyoSchedulerRuntime.scala
```diff
@@ -0,0 +1,31 @@
...
+                }
+            }
+        Unsafe.unsafe { implicit u =>
+            Runtime.default.unsafe.run {
+                for {
+                    scope   <- Scope.make
+                    env     <- (Runtime.setExecutor(exec) ++ Runtime.setBlockingExecutor(exec)).build(scope)
+                    runtime <- ZIO.runtime[Any].provideEnvironment(env)
+                } yield runtime
+            }.getOrThrowFiberFailure()
```
fwbrasil, line 28: I found this approach that avoids the regression reported in https://github.com/zio/zio/pull/8869. I'm not sure about the implications of how it manages the scope but it might be ok since Kyo's default scheduler isn't meant to be shut down.

kyo-scheduler-zio/jvm/src/main/scala/kyo/KyoSchedulerRuntime.scala
```diff
@@ -0,0 +1,31 @@
+package kyo
```
hearnadam, line 1: perhaps `kyo.zio` namespace?

kyo-scheduler-zio/jvm/src/main/scala/kyo/KyoSchedulerRuntime.scala
```diff
@@ -0,0 +1,31 @@
...
+                def submit(runnable: Runnable)(implicit unsafe: Unsafe): Boolean = {
+                    scheduler.schedule(kyo.scheduler.Task(runnable.run()))
+                    true
+                }
+            }
+        Unsafe.unsafe { implicit u =>
+            Runtime.default.unsafe.run {
+                for {
+                    scope   <- Scope.make
+                    env     <- (Runtime.setExecutor(exec) ++ Runtime.setBlockingExecutor(exec)).build(scope)
```
hearnadam, line 25: It should be safe to use global scope here:

https://github.com/zio/zio/blob/series/2.x/core/shared/src/main/scala/zio/Scope.scala#L130-L139

hearnadam (approved):

kyo-scheduler-zio/jvm/src/main/scala/kyo/KyoSchedulerRuntime.scala
fwbrasil: I think a `zio` package in Kyo's codebase could get confusing since the packages can clash if in the same scope.

hearnadam: Good point

kyo-scheduler-zio/jvm/src/main/scala/kyo/KyoSchedulerRuntime.scala
fwbrasil: updated

kyo-scheduler-zio/jvm/src/main/scala/kyo/KyoSchedulerRuntime.scala
fwbrasil: I've renamed the object to from `KyoSchedulerRuntime` to `KyoSchedulerZioRuntime`. It should help avoid confusion since the previous name could indicate it's a runtime for Kyo itself.

PR #387, opened 2024-05-19
Enable choosing a `Position` with/without owning scope

hearnadam:


kyo-core/shared/src/main/scala/kyo/internal/Position.scala
```diff
@@ -9,13 +9,19 @@ object Position:
...
     private def fileNameWithLine(using Quotes): Expr[Position] =
         val expansion = quotes.reflect.Position.ofMacroExpansion
         val name      = expansion.sourceFile.name
         val line      = expansion.startLine + 1
-        Expr(s"$name:$line - $owner")
+        Expr(s"$name:$line")
     end fileNameWithLine
 
+    private def withOwner(using Quotes): Expr[Position] =
+        Expr(s"${fileNameWithLine.valueOrAbort} - $owner")
```
hearnadam, line 23: I've never used this operator, but it seems correct.

kyo-core/shared/src/main/scala/kyo/internal/Position.scala
```diff
@@ -9,13 +9,19 @@ object Position:
 
     inline given derive: Position = ${ Position.fileNameWithLine }
 
+    object WithOwner:
+        inline given derive: Position = ${ Position.withOwner }
```
hearnadam, line 13: I'm biased, but I don't think we need owner by default. Up to you @fwbrasil

PR #391, opened 2024-05-20
Experimental Aborts.fold handler

hearnadam:
A little handler experiment I am working on to understand handlers deeper.

I am concerned that this will cause a too high of allocation rate, but I intentionally used extension to enable users to potentially avoid this...

Consider something like:
```scala

val FoldResponse = Aborts.fold(HttpResponse.default)
...

FoldResponse(v)
```


hearnadam:
Since this seems pretty allocation heavy, I will later focus on how we can provide better utilities on the existing handlers.

PR #393, opened 2024-05-20
fix zio versions

fwbrasil:
I noticed that `kyo-zio` is using an `RC` version. Maybe we need to make another release with correct version?


hearnadam (approved):
Let's cut a 0.10.1

PR #394, opened 2024-05-20
avoid implicit conversions

fwbrasil:
Implicit conversions might end up getting removed from the language. Kyo has very little use for them but the compiler option enabling the feature is set globally, which can end up leading to more implicit conversions getting introduced into the codebase, which could make a migration to new Scala versions more challenging.

This PR disables the feature globally by default.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/core.scala
+++ b/kyo-core/shared/src/main/scala/kyo/core.scala
@@ -241,7 +241,7 @@ object core:
             val tag     = s.tag
         end Continue
 
-        implicit inline def fromKyo[T, S](v: Kyo[T, S]): T < S =
-            v.asInstanceOf[T < S]
+        inline def fromKyo[T, S](inline v: Kyo[T, S]): T < S =
+            v
     end internal
 end core
--- a/kyo-core/shared/src/main/scala/kyo/internal/Param.scala
+++ b/kyo-core/shared/src/main/scala/kyo/internal/Param.scala
@@ -1,6 +1,7 @@
 package kyo.internal
 
 import scala.collection.mutable.LinkedHashMap
+import scala.language.implicitConversions
 import scala.quoted.*
 
 case class Param[T](code: String, value: T) derives CanEqual
--- a/kyo-core/shared/src/test/scala/kyoTest/KyoTest.scala
+++ b/kyo-core/shared/src/test/scala/kyoTest/KyoTest.scala
@@ -7,6 +7,7 @@ import org.scalatest.concurrent.ScalaFutures.*
 import org.scalatest.freespec.AsyncFreeSpec
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+import scala.language.implicitConversions
 import scala.util.Try
 
 class KyoTest extends AsyncFreeSpec with NonImplicitAssertions:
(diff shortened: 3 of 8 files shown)


kyo-core/shared/src/main/scala/kyo/core.scala
```diff
@@ -241,7 +241,7 @@ object core:
             val tag     = s.tag
         end Continue
 
-        implicit inline def fromKyo[T, S](v: Kyo[T, S]): T < S =
-            v.asInstanceOf[T < S]
+        inline def fromKyo[T, S](inline v: Kyo[T, S]): T < S =
```
fwbrasil, line 244: changes this case to a regular method

kyo-core/shared/src/test/scala/kyoTest/KyoTest.scala
```diff
@@ -7,6 +7,7 @@ import org.scalatest.concurrent.ScalaFutures.*
 import org.scalatest.freespec.AsyncFreeSpec
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
+import scala.language.implicitConversions
```
fwbrasil, line 10: There's an implicit conversion in this class from `T` to `Future[T]`. I think it's ok to keep it for now and remove it later if we need to.

kyo-core/shared/src/test/scala/kyoTest/internal/PositionTest.scala
```diff
@@ -24,7 +24,7 @@ class PositionTest extends KyoTest:
         assert(e == "PositionTest.scala:21 - ")
     }
     "WithOwner" - {
-        import Position.WithOwner.given Position
+        inline given Position = Position.WithOwner.derive
```
fwbrasil, line 27: Unrelated compilation warning fix. The compiler was reporting the implicit as not used.

kyo-core/shared/src/main/scala/kyo/internal/Param.scala
```diff
@@ -1,6 +1,7 @@
 package kyo.internal
 
 import scala.collection.mutable.LinkedHashMap
+import scala.language.implicitConversions
```
fwbrasil, line 4: This case would be harder to support without implicit conversions. I think it's ok to keep it as is for now.

kyo-core/shared/src/main/scala/kyo/core.scala
hearnadam: Perhaps we make this an extension method?

kyo-core/shared/src/test/scala/kyoTest/internal/PositionTest.scala
hearnadam: Thanks

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/core.scala
fwbrasil: I think it'd be odd as an extension method:

```scala
(
      new KyoIO[T, S]:
          def apply(v: Unit, s: Safepoint[IOs & S], l: Locals.State) =
                  f
).fromKyo
```

vs 

```scala
fromKyo (
      new KyoIO[T, S]:
          def apply(v: Unit, s: Safepoint[IOs & S], l: Locals.State) =
                  f
)
```

PR #398, opened 2024-05-21
Fix unknown option compiler warning

hearnadam:
`bad option '-Xsource:3' was ignored`


hearnadam:
Not sure that fixed it. Will followup later.

PR #399, opened 2024-05-21
scheduler: fix console top to show deltas

fwbrasil:


kyo-scheduler/jvm/src/main/scala/kyo/scheduler/util/Top.scala
```diff
@@ -24,16 +24,22 @@ class Top(
...
-        close()
-        mBeanServer.registerMBean(new StandardMBean(this, classOf[TopMBean]), objectName)
-    }
+    private var lastConsoleStatus: Scheduler.Status = null
 
     if (enableTopConsoleMs > 0)
         timer.schedule(enableTopConsoleMs.millis) {
-            println(Top.print(status()))
+            val currentStatus = status()
+            if (lastConsoleStatus != null) {
```
hearnadam, line 32: `ne` 😛

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/util/Top.scala
```diff
@@ -24,16 +24,22 @@ class Top(
...
-    if (enableTopJMX) {
-        close()
-        mBeanServer.registerMBean(new StandardMBean(this, classOf[TopMBean]), objectName)
-    }
+    private var lastConsoleStatus: Scheduler.Status = null
 
     if (enableTopConsoleMs > 0)
         timer.schedule(enableTopConsoleMs.millis) {
-            println(Top.print(status()))
+            val currentStatus = status()
```
hearnadam, line 31: I assume `status()` will never return `null`?

If not, should we have a `Status.empty` to prevent gaps in reporting? That would prevent the need for null on line 27 as well.

hearnadam (approved):

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/util/Top.scala
fwbrasil: I'm not sure there's much value in reporting the initial measurement since it'd present accumulated counters up to the point when the observation started.

hearnadam: Right, wasn't sure if status could also be null for any other reason.

PR #404, opened 2024-05-22
Make kyo build with Intellij (#138)

### LINKED ISSUE #138: IntelliJ IDEA support
Kyo has been developed so far exclusively with VSCode and Metals. Since IntelliJ IDEA uses a custom compiler, we need to validate if there are any problems. A few things that seem important to check:

1. If Kyo's codebase itself works without problems
2. Type inference for Kyo computations
3. I recently added the [first macro](https://github.com/getkyo/kyo/blob/main/kyo-core/shared/src/main/scala-3/kyo/internal/FlatImplicits.scala) in Kyo's core to avoid issues with [nested effects](https://github.com/getkyo/kyo#nested-effects). I'm not sure how IntelliJ handles that.
4. Verify the direct syntax functionality

987Nabil:
The removal of private is not 100% necessary, but intellij load cross projects by default as Scala 2. And when compiling Scala 2 it fails with an error that the class is exposed via a public method but it self is private. Which is actually a smell I guess.
Maybe both should be package private?

fixes #138 
/claim #138

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,3 @@
-import sys.process.*
-
 val scala3Version   = "3.4.2"
 val scala212Version = "2.12.19"
 val scala213Version = "2.13.14"
@@ -17,7 +15,7 @@ val compilerOptions = Seq(
     "-language:strictEquality"
 )
 
-scalaVersion                       := scala3Version
+ThisBuild / scalaVersion           := scala3Version
 ThisBuild / sonatypeCredentialHost := "s01.oss.sonatype.org"
 sonatypeRepository                 := "https://s01.oss.sonatype.org/service/local"
 sonatypeProfileName                := "io.getkyo"
@@ -97,11 +95,10 @@ lazy val `kyo-scheduler` =
         )
         .jsSettings(`js-settings`)
 
-lazy val `kyo-scheduler-zio` =
-    crossProject(JVMPlatform)
+def `kyo-scheduler-zio-base` =
+    sbtcrossproject.CrossProject("kyo-scheduler-zio", file("kyo-scheduler-zio"))(JVMPlatform)
         .withoutSuffixFor(JVMPlatform)
         .crossType(CrossType.Full)
-        .in(file("kyo-scheduler-zio"))
         .dependsOn(`kyo-scheduler`)
         .settings(
             `kyo-settings`,
@@ -110,11 +107,18 @@ lazy val `kyo-scheduler-zio` =
                 "-Wunused:all",
                 "-language:strictEquality"
             ),
-            scalacOptions += "-Xsource:3",
             libraryDependencies += "dev.zio"       %%% "zio"       % zioVersion,
-            libraryDependencies += "org.scalatest" %%% "scalatest" % "3.2.16" % Test,
-            crossScalaVersions                      := List(scala3Version, scala212Version, scala213Version)
+            libraryDependencies += "org.scalatest" %%% "scalatest" % "3.2.16" % Test
         )
+
+lazy val `kyo-scheduler-zio-3` = `kyo-scheduler-zio-base`.settings(
+    crossScalaVersions := List(scala3Version)
+)
+
+lazy val `kyo-scheduler-zio` = `kyo-scheduler-zio-base`.settings(
+    scalacOptions += "-Xsource:3",
+    crossScalaVersions := List(scala3Version, scala212Version, scala213Version)
+)
 
 lazy val `kyo-tag` =
     crossProject(JSPlatform, JVMPlatform)
@@ -273,7 +277,7 @@ lazy val `kyo-bench` =
         .enablePlugins(JmhPlugin)
         .dependsOn(`kyo-core`)
         .dependsOn(`kyo-sttp`)
-        .dependsOn(`kyo-scheduler-zio`)
+        .dependsOn(`kyo-scheduler-zio-3`)
         .settings(
             `kyo-settings`,
             // Forks each test suite individually
@@ -281,7 +285,6 @@ lazy val `kyo-bench` =
                 val javaOptionsValue = javaOptions.value.toVector
                 val envsVarsValue    = envVars.value
                 (Test / definedTests).value map { test =>
-                    import sbt.dsl.LinterLevel.Ignore
                     Tests.Group(
                         name = test.name,
                         tests = Seq(test),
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -110,12 +107,19 @@ lazy val `kyo-scheduler-zio` =
...
         )
 
+lazy val `kyo-scheduler-zio-3` = `kyo-scheduler-zio-base`.settings(
+    crossScalaVersions := List(scala3Version)
+)
+
+lazy val `kyo-scheduler-zio` = `kyo-scheduler-zio-base`.settings(
+    scalacOptions += "-Xsource:3",
+    crossScalaVersions := List(scala3Version, scala212Version, scala213Version)
+)
```
hearnadam, line 121: I think this should be possible without separate values with `++= if (_) Seq("-Xsource:3") else Nil`

build.sbt
```diff
@@ -356,5 +359,5 @@ lazy val `js-settings` = Seq(
     Compile / doc / sources                     := Seq.empty,
     fork                                        := false,
     jsEnv                                       := new NodeJSEnv(NodeJSEnv.Config().withArgs(List("--max_old_space_size=5120"))),
-    libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.5.0"
+    libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.5.0" % "provided"
```
hearnadam, line 362: Nice catch, thanks

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/InternalClock.scala
```diff
@@ -4,7 +4,7 @@ import java.util.concurrent.Executor
 import java.util.concurrent.locks.LockSupport
 import kyo.stats.internal.MetricReceiver
 
-final private case class InternalClock(executor: Executor) {
+final case class InternalClock(executor: Executor) {
```
hearnadam, line 7: As you mentioned in description, I think you can make both package private.

fwbrasil:
Nice!! That was quick :) Have you been able to check the other items described in #404?

987Nabil:
@fwbrasil the link to the macro is broken. I checked a few files and could at least not spot any falsly reported errors. Do you have a code part where you would doubt the inference might not be working right?

There are some issues, that are out of scope for this issue for sure, since there are still know issues with intellij when it comes for example to transparent inline. And as much as I wish I could, I won't be able to fix intellij with this PR 😉

build.sbt
987Nabil: Yes, we can exclude Scala 3 here

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/InternalClock.scala
987Nabil: Will do

fwbrasil:
@987Nabil Yes, the issue wasn't meant to fix the IntelliJ compiler :) The bounty's scope differs from this PR, though. Loading Kyo's codebase is a good initial step but, as described in the ticket, it's necessary to validate if the types of Kyo computations can be inferred, if the direct syntax is functional, and if the [Flat macro](https://github.com/getkyo/kyo?tab=readme-ov-file#nested-effects) works (sorry for the broken link). We might need to open tickets upstream if there are issues.

987Nabil:
Okay, got it. I'll take a deeper look. If I see issues, I'll be opening tickets or check of there are already some and link them

987Nabil:
@fwbrasil 
Type inference of `&` is [know to be broken](https://youtrack.jetbrains.com/issue/SCL-21732/Scala-3-Intersection-types-are-not-inferred). So auto generating types or hovering will show some garbage.

I tried the first example in nested effects, but the error message is a different one. I guess the docs are wrong here? Though the docs message seems more useful?

Screenshot 2024-05-23 at 18 26 56

This fits though.
Screenshot 2024-05-23 at 18 40 53

The ops for direct style do not have any or the right tool tip. I do not understand it good enough yet though to open an issue for it.
Screenshot 2024-05-23 at 22 41 44
Screenshot 2024-05-23 at 22 42 54

Interesting is, that the type of the variable on the lhs of the await shows the correct type
Screenshot 2024-05-23 at 22 45 09

fwbrasil:
Thank you for the additional tests @987Nabil! 🙏

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/InternalClock.scala
fwbrasil: Is this still pending?

987Nabil:
Sure @fwbrasil I hope we will see more Scala 3 improvements in intellij. A quality like it has for Java or Kotlin would be awesome.

Anyhow, is there anything else missing?

fwbrasil:
@987Nabil there's this comment https://github.com/getkyo/kyo/pull/404/files#r1614005377

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/InternalClock.scala
987Nabil: done

build.sbt
987Nabil: done

987Nabil:
@fwbrasil sorry. Forgot about it. But fixed now :)

build.sbt
```diff
@@ -110,12 +107,19 @@ lazy val `kyo-scheduler-zio` =
...
                 "-language:strictEquality"
             ),
-            scalacOptions += "-Xsource:3",
             libraryDependencies += "dev.zio"       %%% "zio"       % zioVersion,
-            libraryDependencies += "org.scalatest" %%% "scalatest" % "3.2.16" % Test,
-            crossScalaVersions                      := List(scala3Version, scala212Version, scala213Version)
+            libraryDependencies += "org.scalatest" %%% "scalatest" % "3.2.16" % Test
         )
 
+lazy val `kyo-scheduler-zio-3` = `kyo-scheduler-zio-base`.settings(
```
hearnadam, line 114: We don't nee this anymore, right?

fwbrasil:
All good! Congrats on your first contribution 👏🏼

hearnadam:
@987Nabil I think this change to the build is causing some issues. locally I am now seeing issues related to Heap space and GC is running aggressively. It also looks like you left the scheduler-zio-base / -3 but not sure that's necessary.

Can you take a look?

ghostdogpr:
Is it supposed to work using sbt? I'm getting this when I import the project.
Screenshot 2024-05-26 at 5 10 48 PM

If I pick bsp it works, but it was working before the PR.

987Nabil:
@hearnadam I'll check this. 

@ghostdogpr I had import issues using sbt before. That's what I fixed. Idk why your build is failing now. But I'll check again

PR #409, opened 2024-05-22
stats improvements

fwbrasil:
## Problem

Kyo's `Stats` was designed to be a lightweight wrapper for OpenTelemetry but the approach has some important downsides:

1. **Safety**: Users need to manually call `close` on metrics to release them when not used anymore. It's not difficult to introduce a memory leak with the current approach.
2. **Introspection**: OTel metrics don't provide a way to obtain the current value of metrics because they can be a no-op implementation internally if not enabled. There are situations where using reading values from metrics can be useful. In the scheduler code, I had to avoid using metrics directly because of this limitation.
3. **Integration**: I want to integrate with https://github.com/sqshq/sampler to provide terminal dashboards for Kyo metrics but that's not possible because metrics can be disabled with a no-op internal implementation so the information can be not even available.

## Solution

This PR introduces a new approach where `StatsRegistry` holds weak references to all metrics and garbage collects empty references periodically. The integration with OTel is done via a periodic synchronization with a service-loaded `StatsExporter`.

This change introduces a performance penalty: metrics will always be collected even if no exporters are present. I think it's a reasonable tradeoff and the overhead of metrics is typically not very significant.


kyo-core/shared/src/main/scala/kyo/stats/attributes.scala
```diff
@@ -3,14 +3,14 @@ package kyo.stats
 import kyo.stats.Attributes.AsAttribute
 import scala.annotation.implicitNotFound
 
-case class Attributes(get: List[Attributes.Attribute]) extends AnyVal {
```
fwbrasil, line 6: Used only for traces now, so I moved it to from `kyo-scheduler` to `kyo-core`, which uses Scala 3 syntax.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/InternalClock.scala
```diff
@@ -30,15 +29,9 @@ final private case class InternalClock(executor: Executor) {
 
     def currentMillis(): Long = millis
 
-    def stop(): Unit = {
-        gauge.close()
+    def stop(): Unit =
```
fwbrasil, line 32: The gauge will now be automatically collected when there are no strong references to it anymore.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/regulator/Regulator.scala
```diff
@@ -43,7 +41,7 @@ abstract class Regulator(
 
     protected def measure(v: Long): Unit = {
         probesCompleted.increment()
-        stats.measurement.observe(v.toDouble)
+        stats.measurement.observe(Math.max(0, v.toDouble))
```
fwbrasil, line 44: Workaround because HdrHistogram doesn't support negative values and there's a test that reproduces negative measurements.

kyo-stats-otel/shared/src/main/scala/kyo/stats/otel/OTelAttributes.scala
```diff
@@ -6,15 +6,15 @@ import kyo.stats.Attributes
 import kyo.stats.Attributes.Attribute
```
fwbrasil, line 6: `kyo-stats-otel` wan't getting cross-compiled to Scala 2

kyo-stats-registry/shared/src/main/scala/kyo/stats/internal/UnsafeCounter.scala
```diff
@@ -0,0 +1,19 @@
+package kyo.stats.internal
+
+import java.util.concurrent.atomic.LongAdder
+
+class UnsafeCounter {
+    private var last  = 0L
```
hearnadam, line 6: Do we need to worry about rollover for long running applications? I realize it's mostly unavoidable.

hearnadam (approved):

PR #411, opened 2024-05-23
blocking benchmarks

fwbrasil:
I'm testing the scheduler's scalability and noticed an issue with how it handles blocking. The main issue is the `schedule` method draining blocked workers over and over.

I'm adding these benchmarks to guide the optimization. Current results:

image
https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/5d0d73efeb75e2ed6f5edb0b5a47bdac/raw/c8773ddbecc1c66f378af2852e326968d604e2dd/jmh-result.json

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/BlockingBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/BlockingBench.scala
@@ -1,0 +1,27 @@
+package kyo.bench
+
+import java.util.concurrent.locks.LockSupport
+
+class BlockingBench extends Bench.ForkOnly(()):
+
+    def block(): Unit =
+        LockSupport.parkNanos(100000) // 0.1 ms
+
+    override def kyoBenchFiber() =
+        import kyo.*
+
+        IOs(block())
+    end kyoBenchFiber
+
+    def catsBench() =
+        import cats.effect.*
+
+        IO.blocking(block())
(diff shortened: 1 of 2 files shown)


kyo-bench/src/main/scala/kyo/bench/BlockingBench.scala
```diff
@@ -0,0 +1,27 @@
+package kyo.bench
+
+import java.util.concurrent.locks.LockSupport
+
+class BlockingBench extends Bench.ForkOnly(()):
+
+    def block(): Unit =
+        LockSupport.parkNanos(100000) // 0.1 ms
```
hearnadam, line 8: Can parkNanos be trivially interrupted? Perhaps we should have a blocking interruption benchmark, though it may not be very fair.

hearnadam (approved):

kyo-bench/src/main/scala/kyo/bench/BlockingBench.scala
fwbrasil: Yes, Java's spec allows spurious wake ups but it shouldn't be a significant factor since that typically happens when the OS scheduler is under pressure, which isn't the case here since the workload is basically blocking threads.

Agree about introducing an interruption benchmark! I'll add it to my todo list.

PR #412, opened 2024-05-23
scheduler: improve blocking handling

fwbrasil:
As demonstrated by https://github.com/getkyo/kyo/pull/411, the scheduler's handling of blocking under contention introduces significant overhead. 

This PR introduces a few changes to address it:

1. Worker cycling used to happen every `timeSliceMs` but that's not enough to ensure liveness in case workers get blocked. I've introduced a `cycleNs` configuration that determines the interval between cycles with a default of 0.1 ms.
2. The stalling detection used to consider only the number of cycles that have passed, which is not ideal given the coarse granularity of updates every `timeSliceMs` (5ms by default). The mechanism is now based on `clock.currentMillis`, which provides more precision to perform preemptions.
3. These changes introduce some overhead so I introduced some additional optimizations to offset the regression.

The blocking contention benchmark had a major improvement but several other benchmarks had some regression. I think that's acceptable given that Kyo is still way ahead of alternatives even with this change.

image

https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/1e7f194cd8c8ac460f74c656f19dd626/raw/935381651578dab6e70b13b3722cabc66d165e2a/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/1e7f194cd8c8ac460f74c656f19dd626/raw/935381651578dab6e70b13b3722cabc66d165e2a/jmh-result-candidate.json


kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Queue.scala
```diff
@@ -96,17 +96,18 @@ final private class Queue[T](implicit ord: Ordering[T]) extends AtomicBoolean {
...
-    def drain(f: T => Unit): Unit = {
-        val tasks = {
-            lock()
-            try {
-                items = 0
-                queue.dequeueAll
-            } finally
-                unlock()
+    def drain(f: T => Unit): Unit =
+        if (!isEmpty()) {
```
fwbrasil, line 100: Minor refactoring: moved the `isEmpty` check from `Worker.drain` to here.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -74,55 +68,54 @@ final class Scheduler(
...
-                        val l = candidate.load()
-                        if (l < minLoad) {
-                            minLoad = l
-                            worker = candidate
-                        }
+        val nowMs          = clock.currentMillis()
+        var worker: Worker = null
+        if (submitter == null) {
+            worker = Worker.current()
+            if (worker != null && !worker.checkAvailability(nowMs))
```
fwbrasil, line 75: This check was missing. If the current worker is not available, the task shouldn't be enqueued locally.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -74,55 +68,54 @@ final class Scheduler(
         ExecutionContext.fromExecutor(asExecutor)
 
     private def schedule(task: Task, submitter: Worker): Unit = {
-        val cycles = this.cycles
-        @tailrec def loop(tries: Int = 0): Unit = {
```
fwbrasil, line 78: This method used to loop only when the current worker wasn't available (missing check below) or the candidate became unavailable between its election and the enqueuing, which isn't a relevant edge case.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -74,55 +68,54 @@ final class Scheduler(
...
     private def steal(thief: Worker): Task = {
-        val cycles         = this.cycles
+        val nowMs          = clock.currentMillis()
         val currentWorkers = this.currentWorkers
         var worker: Worker = null
         var maxLoad        = 1
-        var position       = 0
-        while (position < currentWorkers) {
+        var position       = XSRandom.nextInt(currentWorkers)
+        var stride         = Math.min(currentWorkers, stealStride)
```
fwbrasil, line 113: I've introduced a `stride` mechanism similar to `schedule` to reduce the amount of work during stealing. The previous impl could also become too expensive if there are too many blocked workers.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -37,12 +37,6 @@ final class Scheduler(
     @volatile private var allocatedWorkers = 0
     @volatile private var currentWorkers   = coreWorkers
 
-    val a1, a2, a3, a4, a5, a6, a7 = 0L // padding
-
-    @volatile private var cycles = 0L
```
fwbrasil, line 42: No `cycles` anymore since preemption is now done based on `clock.currentMillis`.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -258,34 +252,37 @@ object Scheduler {
...
         timeSliceMs: Int,
+        cycleNs: Int,
         enableTopJMX: Boolean,
         enableTopConsoleMs: Int
     )
     object Config {
         val default: Config = {
             val cores             = Runtime.getRuntime().availableProcessors()
-            val coreWorkers       = Math.max(1, Flag("coreWorkers", cores))
+            val coreWorkers       = Math.max(1, Flag("coreWorkers", cores * 10))
```
fwbrasil, line 266: Start with more workers to avoid having to wait for the concurrency regulator to adapt.

kyo-bench/src/main/scala/kyo/bench/BlockingContentionBench.scala
```diff
@@ -4,7 +4,7 @@ import java.util.concurrent.locks.LockSupport
 
 class BlockingContentionBench extends Bench.ForkOnly(()):
 
-    val concurrency = Runtime.getRuntime().availableProcessors() * 10
+    val concurrency = Runtime.getRuntime().availableProcessors() * 30
```
fwbrasil, line 7: Increase to avoid matching the new default for `coreWorkers`.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
hearnadam: Won't this drain the worker? Should we have a `checkAvailability` which doesn't drain?

hearnadam (approved):

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Worker.scala
```diff
@@ -74,47 +69,41 @@ abstract private class Worker(
...
-            mount != null && {
-                val state = mount.getState().ordinal()
-                state == Thread.State.BLOCKED.ordinal() ||
-                state == Thread.State.WAITING.ordinal() ||
-                state == Thread.State.TIMED_WAITING.ordinal()
-            }
+    private def checkStalling(nowMs: Long): Boolean = {
+        val task    = currentTask
+        val start   = taskStartMs
+        val stalled = task != null && start > 0 && start < nowMs - timeSliceMs
```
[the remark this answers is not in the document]
fwbrasil: ouch I always forget this, I'll fix it. Please feel free to point it out! Old habits die hard :)

Issue #213, opened 2024-03-22
Kyo Stack Traces


fwbrasil:
Early versions of Kyo used to have a mechanism to present better stack traces, similarly to other effect systems, but evolving Kyo while considering how to do that efficiently got difficult so I decided to remove the feature.

At the time, the solution used to use an allocationless macro to generate frame information as constant strings at compile time. It was quite efficient but I felt it was too intrusive. The frame implicit has to be included in virtually all of Kyo's methods since it should come from the user's code.

I think we could go with a solution similar to cats-effect by dynamically inspecting the JVM's stack trace to capture frames. We should use [StackWalker](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/StackWalker.html) to find the origin frame in the user code given that it's much more lightweight. The mechanism would introduce more overhead than the previous static one so I think it'd need to be disabled by default.

fwbrasil:
I played a bit with `StackWalker` but the overhead is too high, it wouldn't be viable to use in a production environment. I'm going with a macro similar to ZIO's https://github.com/getkyo/kyo/pull/413

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/572

PR #413, opened 2024-05-24
trace macro

fwbrasil:
I'm working on https://github.com/getkyo/kyo/issues/213. This PR introduces a macro to generate `Trace`s including a snippet of the method call. The macro isn't very robust because I couldn't find a way to obtain the snippet via the macro APIs. `Position.ofMacroExpansion` returns the position of the `Trace` macro expansion, not the method call. The implementation uses a heuristic to extract the call directly from the source code string. Since the snippet is informative, I think it's a reasonable risk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-1119### Please visit https://getkyo.io for an indexed version of this documentation.
+### Please visit https://getkyo.io for an indexed version of this documentation.
 ##
 
 <img src="https://raw.githubusercontent.com/getkyo/kyo/master/kyo.png" width="200" alt="Kyo">
--- a/kyo-sttp/shared/src/test/scala/kyoTest/requestsTest.scala
+++ b/kyo-sttp/shared/src/test/scala/kyoTest/requestsTest.scala
@@ -9,7 +9,7 @@ class requestsTest extends KyoTest:
 
     class TestBackend extends Requests.Backend:
         var calls = 0
-        def send[T](r: Request[T, Any])(using Trace) =
+        def send[T](r: Request[T, Any]) =
             calls += 1
             Response.ok(Right("mocked")).asInstanceOf[Response[T]]
     end TestBackend


fwbrasil:
I've added the `Trace` implicits to the public APIs. There's a regression in benchmarks but it seems reasonable. The benchmarks showing improvement is just noise.

image


https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/a94b20bb5f49d906f9cb04aad64b0988/raw/85d0098feb1b1b97b12fd4a0d500abd5c7e6882d/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/a94b20bb5f49d906f9cb04aad64b0988/raw/85d0098feb1b1b97b12fd4a0d500abd5c7e6882d/jmh-result-candidate.json

kyo-core/shared/src/test/scala/kyoTest/internal/TraceTest.scala
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    "one pram" in {
+        def test(i: Int)(using t: Trace) = t.snippet
+        assert(test(42) == "test(42)")
+    }
+
+    "two params" in {
+        def test(a: Int, b: String)(using t: Trace) = t.snippet
+        assert(test(1, "hello") == "test(1, \"hello\")")
```
hearnadam, line 27: might be easier to maintain/read tests if we use triple quotes.

kyo-core/shared/src/test/scala/kyoTest/internal/TraceTest.scala
```diff
@@ -0,0 +1,284 @@
...
+    "multiple param lists" in {
+        def test(a: Int)(b: String)(using t: Trace) = t.snippet
+        assert(test(1)("hello") == "test(1)(\"hello\")")
+    }
+
+    "infix method" in pendingUntilFixed {
+        object Test:
+            infix def test(v: Int)(using t: Trace) = t.snippet
+
+        assert((Test test 42) == "Test test 42")
```
hearnadam, line 39: How would this be serialized differently?

Maybe an extension infix method might be a good test as well.

README.md
```diff
@@ -1076,7 +1077,7 @@ def example(db: Database): Int < IOs =
 // Another 'Cut' implementation
 val countTimesTen =
   new Cut[Database, Int, IOs] {
-    def apply[S](v: Database < S)(f: Database => Int < IOs) =
+    def apply[S](v: Database < S)(f: Database => Int < IOs)(using Trace) =
```
hearnadam, line 1080: Why does the readme need to mention `Trace` at all?

kyo-core/shared/src/test/scala/kyoTest/internal/TraceTest.scala
```diff
@@ -0,0 +1,284 @@
+package kyoTest.internal
+
+import kyo.internal.Trace
+import kyoTest.KyoTest
+
+class TraceTest extends KyoTest:
+
+    "show" in {
+        def test(i: Int)(using t: Trace) = t.show
```
hearnadam, line 9: Can you add a `.show` with several layers of nesting? I would like to see how the overall message appears.

README.md
```diff
@@ -1,4 +1,4 @@
-### Please visit https://getkyo.io for an indexed version of this documentation.
+1119### Please visit https://getkyo.io for an indexed version of this documentation.
```
hearnadam, line 1: accident?

fwbrasil: yep!

README.md
fwbrasil: I thought it'd be nice to ensure aspects propagate traces but, rethinking it now, it doesn't seem to make sense. We do want to show the trace based on the user's provided implementation for `apply`.

kyo-core/shared/src/test/scala/kyoTest/internal/TraceTest.scala
fwbrasil: > How would this be serialized differently?

Note that the test is pending. The macro is currently generating an empty string for this scenario.

> Maybe an extension infix method might be a good test as well.

I'll add a pending test for it as well.

kyo-core/shared/src/test/scala/kyoTest/internal/TraceTest.scala
fwbrasil: do you think it's still necessary after the clarifications in private? If not, can you provide an example of what you have in mind?

fwbrasil:
I should have clarified something: this change is only an initial step to address https://github.com/getkyo/kyo/issues/213. The trace information is stored in all Kyo computations, but we still need to change the places where effects are evaluated (`core`, `IOs`, `IOTask`, `ZIOs`) to accumulate the traces and provide a mechanism similar to stack traces. I might first explore a way to avoid custom effect handling in all these different places. Ideally, they shouldn't rely on internal core classes.

Meanwhile, once this change is merged, we can already start showing frame information for example in the scheduler top. I imagine we'll find some other use cases as well.

kyo-core/shared/src/test/scala/kyoTest/internal/TraceTest.scala
hearnadam: Nope

hearnadam:
Changes LGTM, do you still want to hold off on merging for another release?

fwbrasil:
@hearnadam I'm planning to merge this once the build passes since we've made the release already. Maintaining the branch generates extra work and I want to work on follow ups on top of this change.

PR #414, opened 2024-05-25
use: scala-js-macrotask-executor as ExecutionContext

### LINKED ISSUE #302: `Kyo-Scheduler`: `scala-js-macrotask-executor` as `ExecutionContext` on JS platform
`kyo-scheduler/js` is using the default `scala.scalajs.concurrent.QueueExecutionContext` of `scalajs`. Although not included in the `scala-js` official repo, [scala-js-macrotask-executor](https://github.com/scala-js/scala-js-macrotask-executor) is usually recommended as the optimal `ExecutionContext` implementation for use in any Scala.js project.

If there is no specific requirements, I think we can safely use `scala-js-macrotask-executor` instead of `scala.scalajs.concurrent.QueueExecutionContext`

hearnadam:
/claims #302
/resolves #302 

re: your comment @fwbrasil 
- https://github.com/getkyo/kyo/issues/302#issuecomment-2107925967

I think the artifacts were not published for non-js targets after `1.1.0`. If we add this only under JS settings, all is well.

Regarding build time, I think that was a bug potentially having to do with adding the dependency for all targets. For a full clean + build of core, the build time is now 28s on my machine.

Screenshot 2024-05-24 at 7 48 22 PM

I also went ahead and updated this as the executor under kyoTest. Do you think we need a direct dependency under kyo-core? I think transitive should be okay...

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -97,7 +97,10 @@ lazy val `kyo-scheduler` =
             libraryDependencies += "org.scalatest" %%% "scalatest"       % "3.2.16" % Test,
             libraryDependencies += "ch.qos.logback"  % "logback-classic" % "1.5.5"  % Test
         )
-        .jsSettings(`js-settings`)
+        .jsSettings(
+            `js-settings`,
+            libraryDependencies += "org.scala-js" %%% "scala-js-macrotask-executor" % "1.1.1",
+        )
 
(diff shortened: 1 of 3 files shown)


fwbrasil:
Thanks! 😊

PR #415, opened 2024-05-25
Updating README to use Scala 3 syntax

### LINKED ISSUE #218: Update readme to use Scala 3 syntax
The codebase is already fully ported to Scala 3's significant indentation. Although I'm not very convinced of the benefits of the new syntax, I think it's likely that it'll become the main Scala dialect over time. The similarity with Python should also help with the adoption of Kyo's new LLM module to be released.

mobley-trent:
Closes #218
/claim #218

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,20 +207,20 @@ The order in which you handle effects in Kyo can significantly influence both th
 import kyo._
 import scala.util._
 
-def optionsFirst(a: Int < (Options & Aborts[Exception])): Either[Exception, Option[Int]] = {
+def optionsFirst(a: Int < (Options & Aborts[Exception])): Either[Exception, Option[Int]] = 
   val b: Option[Int] < Aborts[Exception] = 
     Options.run(a)
   val c: Either[Exception, Option[Int]] < Any = 
     Aborts.run[Exception](b)
   c.pure
-}
-def abortsFirst(a: Int < (Options & Aborts[Exception])): Option[Either[Exception, Int]] = {
+
+def abortsFirst(a: Int < (Options & Aborts[Exception])): Option[Either[Exception, Int]] = 
   val b: Either[Exception, Int] < Options =
     Aborts.run[Exception](a)
   val c: Option[Either[Exception, Int]] < Any = 
     Options.run(b)
   c.pure
-}
+
 
 // The sequence in which effects are handled has a significant impact on the outcome. 
 // This is especially true for effects that can short-circuit the computation.


README.md
```diff
@@ -207,20 +207,20 @@ The order in which you handle effects in Kyo can significantly influence both th
...
 import scala.util._
 
-def optionsFirst(a: Int < (Options & Aborts[Exception])): Either[Exception, Option[Int]] = {
+def optionsFirst(a: Int < (Options & Aborts[Exception])): Either[Exception, Option[Int]] = 
   val b: Option[Int] < Aborts[Exception] = 
     Options.run(a)
   val c: Either[Exception, Option[Int]] < Any = 
     Aborts.run[Exception](b)
   c.pure
-}
```
fwbrasil, line 216: Thanks for picking this up! Please take a look at Kyo's codebase to follow the same formatting. For example, method bodies with more than 4 lines have an [end maker](https://github.com/getkyo/kyo/blob/326f7588402cbe0a86133615db98b8c9cafed3bd/.scalafmt.conf#L8). I wonder if there'd be a way to make scalafmt format the code in the readme since it'd be more future-proof.

hearnadam: I think this bounty should involve configuring scalafmt on the readme.

fwbrasil: @mobley-trent let us know if you'd be ok with the scope change. I could also increase the bounty.

mobley-trent: @fwbrasil it's perfectly fine. I'm currently looking into how I can integrate scalafmt with mdoc during the build process.

987Nabil: @mobley-trent You could put the code in scala files and scalafmt will format them anyway. And read in the files. We do this in zio projects. For example like this.
https://github.com/zio/zio-http/blob/342ac4b578a5d7fd11229bcea3b546f632ff67b8/zio-http-docs/src/main/scala/utils.scala#L39

README.md
```diff
@@ -257,26 +257,10 @@ Kyo provides direct syntax for a more intuitive and concise way to express compu
...
-  }
-
-// Equivalent desugared
-val b: String < (Aborts[Exception] & Options) =
-  Options.get(Some("hello")).map { b =>
-    Aborts.get(Right("world")).map { c =>
-      b + " " + c
-    }
-  }
+printSource("readme-examples/shared/src/main/scala/kyo/Example1.scala")
```
mobley-trent, line 263: Hi @987Nabil I went with the approach used in zio-http but I noticed that this function only works when I use the absolute path to the file containing the example for some reason. Also I noticed that the output compiles in `target\README-out.md`. Should I copy this file to the main directory as the new README ?

987Nabil: I am not the expert on the doc generation of zio-http, but seems reasonable :)

mobley-trent:
@fwbrasil @987Nabil I ended up going with the original approach by formatting the code snippets inside scala files using `sbt sscalafmt`. I shall proceed to remove the `readme-examples module` if you find the work satisfactory.

README.md
mobley-trent: @fwbrasil I would appreciate your feedback

readme-examples/shared/src/main/scala/kyo/utils.scala
```diff
@@ -0,0 +1,43 @@
+package kyo.readmeExamples
+
+import scala.io.Source
+
+object utils:
```
fwbrasil, line 5: @mobley-trent apologies for the delay to review! Can you clarify how the solution works? Do we need to use this implementation manually to format the readme code?

mobley-trent: @fwbrasil this was actually a solution proposed by @987Nabil but I went on to use your original approach of putting the code inside scala files and running scalafmt. I actually meant to remove the rest of the redundant files

fwbrasil:
Thank you @mobley-trent! I’m sorry again for the delay in review. Congrats on your first contribution 👏🏼

PR #417, opened 2024-05-25
scheduler: dispatch java interrupts

fwbrasil:
When a worker gets blocked, Kyo's scheduler attempts to preempt the fiber but never dispatches a Java interrupt to the thread. This can be an issue in systems that rely on interruptions in blocking code since it can lead to significant accumulation of state, which can produce degradation.

I hadn't included this initially because Java interrupts are normally a headache but this seems to be a common practice in effect systems. In case of issues, it's also possible to disable the behavior by setting the config to `-1`.

### DIFF
--- a/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
+++ b/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
@@ -171,7 +171,7 @@ final class Scheduler(
     private def ensureWorkers() =
         for (idx <- allocatedWorkers until currentWorkers) {
             workers(idx) =
-                new Worker(idx, pool, schedule, steal, clock, timeSliceMs) {
+                new Worker(idx, pool, schedule, steal, clock, timeSliceMs, javaInterruptAfterMs) {
                     def shouldStop(): Boolean = idx >= currentWorkers
                 }
             allocatedWorkers += 1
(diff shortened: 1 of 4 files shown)


kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Worker.scala
```diff
@@ -84,6 +85,11 @@ abstract private class Worker(
         val stalled = (task ne null) && start > 0 && start < nowMs - timeSliceMs
         if (stalled) {
             task.doPreempt()
+            if (javaInterruptAfterMs > 0) {
+                val thread = mount
+                if (thread != null && (currentTask eq task) && start < nowMs - javaInterruptAfterMs && !thread.isInterrupted())
```
fwbrasil, line 90: `currentTask eq task` just to double check that the task is still running

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -257,22 +257,24 @@ object Scheduler {
...
+            val cores                = Runtime.getRuntime().availableProcessors()
+            val coreWorkers          = Math.max(1, Flag("coreWorkers", cores * 10))
+            val minWorkers           = Math.max(1, Flag("minWorkers", coreWorkers.toDouble / 2).intValue())
+            val maxWorkers           = Math.max(minWorkers, Flag("maxWorkers", coreWorkers * 100))
+            val scheduleStride       = Math.max(1, Flag("scheduleStride", cores))
+            val stealStride          = Math.max(1, Flag("stealStride", cores * 4))
+            val virtualizeWorkers    = Flag("virtualizeWorkers", false)
+            val timeSliceMs          = Flag("timeSliceMs", 10)
+            val cycleNs              = Flag("cycleNs", 100000)
+            val javaInterruptAfterMs = Flag("javaInterruptAfterMs", timeSliceMs * 4)
```
fwbrasil, line 275: Default of 40ms

kyo-scheduler/jvm/src/test/scala/kyo/scheduler/SchedulerTest.scala
```diff
@@ -123,6 +123,19 @@ class SchedulerTest extends AnyFreeSpec with NonImplicitAssertions {
         }
     }
 
+    "dispatches java interrupt" in withScheduler { scheduler =>
+        val cdl = new CountDownLatch(1)
+        scheduler.schedule(TestTask(_run = () => {
```
hearnadam, line 128: I am concerned this may dispatch an interrupt to a blocked, but working task. Can we add a test for computing something blocking, eg md5 hashes for some time period greater than the default java interrupt time?

hearnadam (commented):
Thanks for the fast PR!

kyo-scheduler/jvm/src/test/scala/kyo/scheduler/SchedulerTest.scala
fwbrasil: Yes, it'll first try preempting the task and then dispatching the thread interrupt if the preemption didn't work. I think we could dispatch the thread interrupt only if it's blocked. Detecting blocking isn't fail-proof, though. There are cases where the blocking happens in native code, which is not possible to detect by inspecting the thread's state.

hearnadam (approved):

PR #420, opened 2024-05-26
kyo-tapir: fix netty future integration to consider Void completions

fwbrasil:
Netty's future represents computations with result type `Void`, which is encoded at runtime as a `null` value. The code that integrates Kyo's fibers with Netty's wasn't considering this scenario and lifting the result directly to `T < IOs`. This PR fixes that by completing with a pending `IOs(null)` if Netty's Future completes with `null`, indicating the result of a `Void` computation.

I've also added a NPE guard in `fiber.complete` to make it easier to detect this issue since this is a low-level API meant for integration with other libraries.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
+++ b/kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
@@ -114,11 +114,16 @@ private[kyo] class IOPromise[T](state: State[T])
             true
         }
 
+    private val nullCompletion = IOs(null)
+
     final def complete(v: T < IOs): Boolean =
+        val r =
+            if !isNull(v) then v
+            else nullCompletion.asInstanceOf[T < IOs]
         @tailrec def loop(): Boolean =
             get() match
                 case p: Pending[T] @unchecked =>
-                    complete(p, v) || loop()
+                    complete(p, r) || loop()
                 case _ =>
                     false
         loop()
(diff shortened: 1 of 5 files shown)


fwbrasil:
I remembered that we also have `Fibers.fromCompletionStage` now that would need to be fixed. I've decided to allow `null` completions instead by changing `IOPromise.complete` so it's fixed for both cases and we can avoid future issues with other integrations in case we forget to handle this scenario.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -114,11 +114,16 @@ private[kyo] class IOPromise[T](state: State[T])
             true
         }
 
+    private val nullCompletion = IOs(null)
+
     final def complete(v: T < IOs): Boolean =
+        val r =
+            if !isNull(v) then v
```
hearnadam, line 121: Can you remind me why we need this function vs `ne null`?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
hearnadam: Ah right, thanks. I notice that function is using `==`... I'll sweep the code base later to switch all null checks to `eq/ne`

fwbrasil: Cool! I wonder if there's a way to configure a linter for that.

Issue #203, opened 2024-03-16
Do we really need layers?


fwbrasil:
I've been thinking about the usability of `Layer` and, although it ended up an interesting and more generic solution than the original in ZIO, I'm not sure its level of abstraction works well:

- Dealing with generic type parameters is a common difficulty of newcomers. Types with a single parameter like in collections are easier but an effect-level arrow with two type parameters is a quite high-level concept to grasp in addition to the pending type.
- Learning horizontal and vertical composition also doesn't seem trivial. I showed this [excellent blog post](https://www.scalamatters.io/post/writing-modular-applications-using-the-kyo-library) to a couple of people that were learning Kyo and it felt very foreign to them.
- The type inference of layers is problematic. Users have to specify types explicitly in some situations and there's also an issue that layers can produce a value of type `T < Nothing`, which is invalid. 
- Although we limited layers only to effects where ordering isn't an issue, users might use layers with other effects that can become problematic.

Instead of having a separate API, I think we could leverage Kyo's regular pending type to express the dependency graph and then provide a macro to derive the environment automatically:

```scala
trait Service1 {
  def a: Int
}
object Service1 {
  val init: Service1 < IOs = 
    IOs(Live()))
  case class Live() extends Service1
}

trait Service2 {
  def b: Int
}
object Service2 {
  val init: Service2 < (Envs[Service1] & IOs) = 
    Envs[Service1].get.map(Live(_))
  case class Live(s1: Service1) extends Service2
}

// example computation
val io: Int < (Envs[Service1] & Envs[Service2] & IOs) = ???

// typically in a main file
val res: Int < IOs = 
  Envs.run(Service1.init, Service2.init)(io)
```

The `init` of each service would express the value it provides and its dependencies via regular pending `Envs` effects. The `Envs.run` method would need a macro to inspect the dependencies, build the graph, and provide the initialization of the services.

@jiveshungerford @kitlangton I'd love to hear your thoughts on this if you have some time!

johnhungerford:
[...]

PR #422, opened 2024-05-27
add: type `TypeMap[A]` enabling `Tag[T]` -> instance `T`

hearnadam:
This will serve as the basis for both Layers + Envs that supports intersections in #203. Inspired by ZEnvironment, but hopefully more lightweight...

@kitlangton plans to PR the Envs with intersection support next.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/env.scala
+++ b/kyo-core/shared/src/main/scala/kyo/env.scala
@@ -1,0 +1,47 @@
+package kyo
+
+import kyo.Tag.Intersection
+import scala.collection.immutable.HashMap
+
+opaque type Env[+R] = HashMap[Tag[?], Any]
+extension [R](self: Env[R])
+
+    private def fatal(using t: Tag[?]): Nothing =
+        throw new RuntimeException(s"fatal: kyo.Env of contents [$self] missing value of type: [${t.show}].")
+
+    def get[A >: R](using Tag[A]): A =
+        getOrElse(fatal)
+
+    def getOrElse[A >: R](default: => A)(using t: Tag[A]): A =
+        val result = self.getOrElse(t, null)
+        if isNull(result) then
+            default
+        else
+            result.asInstanceOf[A]
+        end if
+    end getOrElse
+
+    def add[A: Tag](a: A): Env[R & A] =
+        self.updated(Tag[A], a)
+
+    def union[R0](that: Env[R0]): Env[R & R0] =
+        self ++ that
+
+    def size: Int        = self.size
+    def isEmpty: Boolean = self.isEmpty
+
+    private[kyo] def tag: Intersection[?] = Intersection(self.keySet.toIndexedSeq)
+end extension
+
+object Env:
+    val empty: Env[Any] = HashMap.empty
+
+    def apply[A: Tag](a: A): Env[A] =
+        HashMap(Tag[A] -> a)
+    def apply[A: Tag, B: Tag](a: A, b: B): Env[A & B] =
+        HashMap(Tag[A] -> a, Tag[B] -> b)
+    def apply[A: Tag, B: Tag, C: Tag](a: A, b: B, c: C): Env[A & B & C] =
+        HashMap(Tag[A] -> a, Tag[B] -> b, Tag[C] -> c)
+    def apply[A: Tag, B: Tag, C: Tag, D: Tag](a: A, b: B, c: C, d: D): Env[A & B & C & D] =
+        HashMap(Tag[A] -> a, Tag[B] -> b, Tag[C] -> c, Tag[D] -> d)
+end Env
--- a/kyo-core/shared/src/main/scala/kyo/env.scala
+++ b/kyo-core/shared/src/main/scala/kyo/env.scala
@@ -6,29 +6,20 @@ import scala.collection.immutable.HashMap
 opaque type Env[+R] = HashMap[Tag[?], Any]
 extension [R](self: Env[R])
 
-    private def fatal(using t: Tag[?]): Nothing =
+    private inline def fatal(using t: Tag[?]): Nothing =
         throw new RuntimeException(s"fatal: kyo.Env of contents [$self] missing value of type: [${t.show}].")
 
-    def get[A >: R](using Tag[A]): A =
-        getOrElse(fatal)
+    inline def get[A >: R](using inline t: Tag[A]): A =
+        self.getOrElse(t, fatal).asInstanceOf[A]
 
-    def getOrElse[A >: R](default: => A)(using t: Tag[A]): A =
-        val result = self.getOrElse(t, null)
-        if isNull(result) then
-            default
-        else
-            result.asInstanceOf[A]
-        end if
-    end getOrElse
-
-    def add[A: Tag](a: A): Env[R & A] =
-        self.updated(Tag[A], a)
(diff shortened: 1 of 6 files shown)


kitlangton:
Awesome! One thing that crossed my mind: Should this potentially be nested in some sub-package or within the Envs companion object? It could compete with `Envs` for autocomplete if it's exposed at the top-level.

kyo-core/shared/src/main/scala/kyo/env.scala
```diff
@@ -0,0 +1,47 @@
...
+extension [R](self: Env[R])
+
+    private def fatal(using t: Tag[?]): Nothing =
+        throw new RuntimeException(s"fatal: kyo.Env of contents [$self] missing value of type: [${t.show}].")
+
+    def get[A >: R](using Tag[A]): A =
+        getOrElse(fatal)
+
+    def getOrElse[A >: R](default: => A)(using t: Tag[A]): A =
+        val result = self.getOrElse(t, null)
```
fwbrasil, line 16: Could `default` be passed to `getOrElse` directly?

kyo-core/shared/src/main/scala/kyo/env.scala
```diff
@@ -0,0 +1,47 @@
...
+            default
+        else
+            result.asInstanceOf[A]
+        end if
+    end getOrElse
+
+    def add[A: Tag](a: A): Env[R & A] =
+        self.updated(Tag[A], a)
+
+    def union[R0](that: Env[R0]): Env[R & R0] =
```
fwbrasil, line 27: What do you think about naming this method `add` as well?

kyo-core/shared/src/main/scala/kyo/env.scala
```diff
@@ -0,0 +1,47 @@
...
+        self ++ that
+
+    def size: Int        = self.size
+    def isEmpty: Boolean = self.isEmpty
+
+    private[kyo] def tag: Intersection[?] = Intersection(self.keySet.toIndexedSeq)
```
fwbrasil, line 33: I love how well the decision not to optimize intersection and union tags is paying off here! We should probably make the constructor private[kyo], though

kyo-core/shared/src/main/scala/kyo/package.scala
```diff
@@ -63,7 +63,7 @@ package object kyo:
         ()
 
     private[kyo] inline def isNull[T](v: T): Boolean =
-        v.asInstanceOf[AnyRef] == null
+        v.asInstanceOf[AnyRef] eq null
```
fwbrasil, line 66: thanks! :)

fwbrasil:
This data structure seems quite useful! It's essentially a map with types as keys 😍 I imagine we might find other use cases for it as well.

kyo-core/shared/src/main/scala/kyo/env.scala
```diff
@@ -0,0 +1,47 @@
...
+        val result = self.getOrElse(t, null)
+        if isNull(result) then
+            default
+        else
+            result.asInstanceOf[A]
+        end if
+    end getOrElse
+
+    def add[A: Tag](a: A): Env[R & A] =
+        self.updated(Tag[A], a)
```
fwbrasil, line 25: Can you receive the `Tag` directly as an implicit parameter instead? I'm not sure if there'd be a noticeable performance difference here but avoiding the macro indirection should reduce the size of the generated bytecode.

kyo-core/shared/src/test/scala/kyoTest/envTest.scala
```diff
@@ -0,0 +1,109 @@
+package kyoTest
+
+import kyo.*
+
+class envTest extends KyoTest:
```
fwbrasil, line 5: Can you add tests showing that getting a type not present in the `Env` fails to compile?

kyo-core/shared/src/main/scala/kyo/env.scala
kitlangton: A is basically guaranteed to exist in the map, as long as we expose the right interface (and people don't unsafely cast). So I don't think we need getOrElse?

fwbrasil:
> Awesome! One thing that crossed my mind: Should this potentially be nested in some sub-package or within the Envs companion object? It could compete with `Envs` for autocomplete if it's exposed at the top-level.

Good point! It feels like this is more generic than just for the environments use case. I think it could be in the `kyo` package but maybe with a more generic name.

kyo-core/shared/src/main/scala/kyo/env.scala
hearnadam: I think we can remove it. Realistically, this method is only useful for internal usage. Later we can re-expose as package private.

kyo-core/shared/src/main/scala/kyo/env.scala
hearnadam: The problem is that this not simply additive in the case of overlapping values. I think that may confuse users.

kyo-core/shared/src/main/scala/kyo/env.scala
hearnadam: Sure, will do.

kitlangton:
> Good point! It feels like this is more generic than just for the environments use case. I think it could be in the `kyo` package but maybe with a more generic name.

It's sort of a type-keyed map... `TypeMap`?

kyo-core/shared/src/test/scala/kyoTest/envTest.scala
hearnadam: Done.

kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -0,0 +1,38 @@
...
+    private inline def fatal(using t: Tag[?]): Nothing =
+        throw new RuntimeException(s"fatal: kyo.TypeMap of contents [$self] missing value of type: [${t.show}].")
+
+    inline def get[A >: R](using inline t: Tag[A]): A =
+        self.getOrElse(t, fatal).asInstanceOf[A]
+
+    inline def add[A: Tag](a: A)(using inline t: Tag[A]): TypeMap[R & A] =
+        self.updated(t, a)
+
+    def union[R0](that: TypeMap[R0]): TypeMap[R & R0] =
```
fwbrasil, line 18: Could it be `inline` as well?

fwbrasil (approved):

kyo-core/shared/src/test/scala/kyoTest/typeMapTest.scala
```diff
@@ -0,0 +1,116 @@
+package kyoTest
+
+import kyo.*
+
+class typeMapTest extends KyoTest:
```
fwbrasil, line 5: Can you add tests to show that it's not possible to call `get` and `add` with intersection and union type parameters?

hearnadam: Done.

hearnadam:
I think we should merge this if you think it's ready @fwbrasil. We can followup to:

1. add `.prune`
2. potential: require distinct values in `.apply`?
3. make Intersection.apply package private

fwbrasil:
Love this! Thanks folks 🙏

PR #423, opened 2024-05-27
Caliban integration

### LINKED ISSUE #216: Resolvers: GraphQL via Caliban
GraphQL is a powerful framework for building flexible HTTP clients and servers. Its query representation allows for expressing a wide range of sophisticated access patterns in a straightforward and consistent manner, a notable advantage over traditional REST approaches. For example, a single GraphQL query can retrieve multiple resources, as well as specific fields of those resources, reducing the need for multiple network requests and enabling more efficient data retrieval.

Given the developments in AI, I think the relevance of solutions like GraphQL can increase significantly. It's possible that a large portion of software user interfaces will transition to highly personalized, AI-driven interactions. The flexible and dynamically discoverable nature of GraphQL could make it a great primitive to develop software to be used by AI systems.

[Caliban](https://ghostdogpr.github.io/caliban/) seems a good choice for Kyo's integration for GraphQL. It seems generic and adaptable to different effect systems and its query definition mechanism seems to enable the kind of simple-to-use experience Kyo aims for. Ideally, the new module should integrate with `kyo-tapir`.

ghostdogpr:
/fixes #216
/claim #216

### Schema derivation
You can now use kyo effects in your resolvers and schemas are derived with 2 different givens:
- if your kyo effect is a subtype of `(Aborts[Throwable] & ZIOs)`, it works out of the box
- if your kyo effect is something else, it will require using `SchemaDerivation[Runner[S]]` to derive the schema, and you will need to provide the `Runner` on `run`

```scala
// works out of the box
case class Query(
    k1: Int < Aborts[Throwable],
    k2: Int < ZIOs,
    k3: Int < (Aborts[Throwable] & ZIOs),
    k4: Int < IOs
) derives Schema.SemiAuto

// other kyo effects require extending `SchemaDerivation`
object schema extends SchemaDerivation[Runner[Vars[Int] & Consoles]]
case class Query(k: Int < (Vars[Int] & Consoles)) derives schema.SemiAuto
```

### Resolvers effect
- you can use `Resolvers.run { Resolvers.get(api) }` to simply turn a caliban `GraphQL` api into a running server
- when a runner is required, you can use `Resolvers.run(runner) { Resolvers.get(api) }`
- there are also variants that take a `NettyKyoServer` and a ZIO `Runtime`

```scala
// works out of the box
Resolvers.run { Resolvers.get(api) }

// when using a Runner
val runner = new Runner[Env]:
    def apply[T: Flat](v: T < Env): Task[T] = ZIOs.run(Consoles.run(Vars.run(0)(v)))

Resolvers.run(runner) { Resolvers.get(api) }
```

Note that `Resolvers.get` returns Caliban's `HttpInterpreter` which allows customizing configuration, adding middleware, changing the endpoints (e.g. path), etc

### Further work
- [x] Documentation (planning to do this after I get some reviews first)
- [ ] Planning to open an issue afterwards to add Streaming and WebSocket support once it's supported by the tapir integration


ghostdogpr:
I found a new approach that supports the ZIO environment and using arbitrary Kyo effects 🎉

kyo-caliban/src/test/scala/kyo/resolversTest.scala
```diff
@@ -0,0 +1,134 @@
...
+    def runZIO[T](v: Task[T]): T =
+        zio.Unsafe.unsafe(implicit u =>
+            zio.Runtime.default.unsafe.run(v).getOrThrow()
+        )
+
+    case class Query(
+        k1: Int < Aborts[Throwable],
+        k2: Int < ZIOs,
+        k3: Int < (Aborts[Throwable] & ZIOs),
+        k4: Int < IOs
```
fwbrasil, line 26: Can you add a field with pending `Fibers` as well?

kyo-caliban/src/test/scala/kyo/resolversTest.scala
```diff
@@ -0,0 +1,134 @@
...
+        type Env = Vars[Int] & Consoles
+        object schema extends SchemaDerivation[Runner[Env]]
+
+        case class Query(k: Int < Env) derives schema.SemiAuto
+
+        val api = graphQL(RootResolver(Query(
+            for
+                _ <- Vars.update[Int](_ + 1)
+                v <- Vars.get[Int]
+                _ <- Consoles.println(v)
```
fwbrasil, line 58: I'd prefer to avoid printing to the console in tests. Maybe showing `Envs` use would be interesting here instead?

kyo-caliban/src/main/scala/kyo/resolvers.scala
```diff
@@ -0,0 +1,65 @@
...
+import kyo.internal.KyoSttpMonad
+import sttp.tapir.Codec.JsonCodec
+import sttp.tapir.Endpoint
+import sttp.tapir.capabilities.NoStreams
+import sttp.tapir.server.ServerEndpoint
+import sttp.tapir.server.netty.*
+import zio.*
+import zio.stream.ZStream
+
+opaque type Resolvers = Aborts[Throwable] & ZIOs
```
fwbrasil, line 15: FYI the direct syntax has a bug that crashes the compiler if effects are defined as `opaque`. I think it's ok, though. We should fix the direct syntax instead.

kyo-caliban/src/main/scala/kyo/resolvers.scala
```diff
@@ -0,0 +1,65 @@
...
+object Resolvers:
+
+    private given StreamConstructor[Nothing] =
+        (_: ZStream[Any, Throwable, Byte]) => throw new Throwable("Streaming is not supported")
+
+    def run[T, S](v: HttpInterpreter[Any, CalibanError] < (Resolvers & S)): NettyKyoServerBinding < (ZIOs & Aborts[Throwable] & S) =
+        run[T, S](NettyKyoServer())(v)
+
+    def run[T, S](server: NettyKyoServer)(v: HttpInterpreter[Any, CalibanError] < (Resolvers & S))
+        : NettyKyoServerBinding < (ZIOs & Aborts[Throwable] & S) =
```
fwbrasil, line 26: Should it be `Aborts[CalibanError]` instead?

kyo-caliban/src/main/scala/kyo/resolvers.scala
```diff
@@ -0,0 +1,65 @@
...
+
+    def run[R, T, S](runner: Runner[R])(v: HttpInterpreter[Runner[R], CalibanError] < (Resolvers & S))(using
+        tag: Tag[Runner[R]]
+    ): NettyKyoServerBinding < (ZIOs & Aborts[Throwable] & S) =
+        run[R, T, S](NettyKyoServer(), runner)(v)
+
+    def run[R, T, S](server: NettyKyoServer, runner: Runner[R])(v: HttpInterpreter[Runner[R], CalibanError] < (Resolvers & S))(using
+        tag: Tag[Runner[R]]
+    ): NettyKyoServerBinding < (ZIOs & Aborts[Throwable] & S) =
+        ZIOs.get(ZIO.runtime[Any]).flatMap(runtime => run(server, runtime.withEnvironment(ZEnvironment(runner)))(v))
```
fwbrasil, line 37: s/flatMap/map/g :)

kyo-caliban/src/main/scala/kyo/resolvers.scala
```diff
@@ -0,0 +1,65 @@
...
+        ZIOs.get(api.interpreter.map(HttpInterpreter(_)))
+
+    private def convertEndpoint[R, I](
+        endpoint: ServerEndpoint.Full[Unit, Unit, I, TapirResponse, CalibanResponse[NoStreams.BinaryStream], NoStreams, [x] =>> RIO[R, x]],
+        runtime: Runtime[R]
+    ): ServerEndpoint[Any, KyoSttpMonad.M] =
+        ServerEndpoint[Unit, Unit, I, TapirResponse, CalibanResponse[NoStreams.BinaryStream], Any, KyoSttpMonad.M](
+            endpoint.endpoint.asInstanceOf[Endpoint[Unit, I, TapirResponse, CalibanResponse[NoStreams.BinaryStream], Any]],
+            _ => _ => Right(()),
+            _ => _ => req => Unsafe.unsafe { implicit u => runtime.unsafe.run(endpoint.logic(zioMonadError)(())(req)).getOrThrow() }
```
fwbrasil, line 62: I'm taking a look if we could avoid unsafe here by allowing a type parameter `S` in `NettyKyoServer`.

fwbrasil: I wanted to allow `ZIOs` with `NettyKyoServer` but it's not trivial because `MonadAsyncError` requires `IOs.ensure` that can't accept arbitrary effects. I think we could eventually integrate ZIO's `Scope` with Kyo's `Resources` to address that, though. What are the implications of using unsafe here? I imagine it'd use thread blocking and not propagate cancellations/interrupts?

kyo-caliban/src/test/scala/kyo/resolversTest.scala
ghostdogpr: Done 👍

kyo-caliban/src/test/scala/kyo/resolversTest.scala
ghostdogpr: Sure, I picked that one arbitrarily. Changed to `Envs`.

kyo-caliban/src/main/scala/kyo/resolvers.scala
ghostdogpr: That's right! I was hesitating whether I should expose it but yeah, why not use the most specific error type. Changed it.

kyo-caliban/src/main/scala/kyo/resolvers.scala
ghostdogpr: Old habits die hard 😄 Fixed.

kyo-caliban/src/main/scala/kyo/resolvers.scala
ghostdogpr: I agree it's not great since it will just block. I will see what I can do with other `run` variants such as `runToFuture` or `fork` 🤔

ghostdogpr: I change the implementation to re-use `KyoSttpMonad.async` with `runToFuture`. Let me know what you think 🙏

fwbrasil: Nice! I didn't know ZIO had a `Future` subclass with cancellation. We should have something similar.

fwbrasil (approved):
LGTM! Thank you for working on this @ghostdogpr, it'll be quite useful :)

kyo-caliban/src/main/scala/kyo/resolvers.scala
```diff
@@ -0,0 +1,74 @@
...
+import sttp.tapir.server.netty.*
+import zio.*
+import zio.stream.ZStream
+
+opaque type Resolvers = Aborts[CalibanError] & ZIOs
+
+object Resolvers:
+
+    private given StreamConstructor[Nothing] =
+        (_: ZStream[Any, Throwable, Byte]) => throw new Throwable("Streaming is not supported")
```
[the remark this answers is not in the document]
fwbrasil: I think it'd require `inline given` as well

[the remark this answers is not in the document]
ghostdogpr: But we don't want a compile error here, we need to provide a `StreamConstructor` for `NoStreams`, which is the only stream implementation we can use now. Once streaming is supported by `NettyKyoServer`, this can be removed.

kyo-caliban/src/main/scala/kyo/resolvers.scala
```diff
@@ -0,0 +1,74 @@
...
+        run[T, S](NettyKyoServer())(v)
+
+    def run[T, S](server: NettyKyoServer)(v: HttpInterpreter[Any, CalibanError] < (Resolvers & S))
+        : NettyKyoServerBinding < (ZIOs & Aborts[CalibanError] & S) =
+        ZIOs.get(ZIO.runtime[Any]).map(runtime => run(server, runtime)(v))
+
+    def run[R, T, S](runner: Runner[R])(v: HttpInterpreter[Runner[R], CalibanError] < (Resolvers & S))(using
+        tag: Tag[Runner[R]]
+    ): NettyKyoServerBinding < (ZIOs & Aborts[CalibanError] & S) =
+        run[R, T, S](NettyKyoServer(), runner)(v)
```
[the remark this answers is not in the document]
ghostdogpr: I did the same thing as in the existing `Routes` effect.

kyo-caliban/src/main/scala/kyo/resolvers.scala
```diff
@@ -0,0 +1,74 @@
...
+    ): HttpInterpreter[R, CalibanError] < Resolvers =
+        ZIOs.get(api.interpreter.map(HttpInterpreter(_)))
+
+    private def convertEndpoint[R, I](
+        endpoint: ServerEndpoint.Full[Unit, Unit, I, TapirResponse, CalibanResponse[NoStreams.BinaryStream], NoStreams, [x] =>> RIO[R, x]],
+        runtime: Runtime[R]
+    ): ServerEndpoint[Any, KyoSttpMonad.M] =
+        ServerEndpoint[Unit, Unit, I, TapirResponse, CalibanResponse[NoStreams.BinaryStream], Any, KyoSttpMonad.M](
+            endpoint.endpoint.asInstanceOf[Endpoint[Unit, I, TapirResponse, CalibanResponse[NoStreams.BinaryStream], Any]],
+            _ => _ => Right(()),
```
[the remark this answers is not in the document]
ghostdogpr: Good point, done!

kyo-caliban/src/main/scala/kyo/resolvers.scala
```diff
@@ -0,0 +1,74 @@
...
+import caliban.interop.tapir.TapirAdapter.*
+import kyo.internal.KyoSttpMonad
+import scala.concurrent.ExecutionContext
+import sttp.monad.Canceler
+import sttp.tapir.Codec.JsonCodec
+import sttp.tapir.Endpoint
+import sttp.tapir.capabilities.NoStreams
+import sttp.tapir.server.ServerEndpoint
+import sttp.tapir.server.netty.*
+import zio.*
```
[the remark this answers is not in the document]
ghostdogpr: Done

kyo-caliban/src/main/scala/kyo/resolvers.scala
```diff
@@ -0,0 +1,74 @@
...
+        responseValueCodec: JsonCodec[ResponseValue]
+    ): HttpInterpreter[R, CalibanError] < Resolvers =
+        ZIOs.get(api.interpreter.map(HttpInterpreter(_)))
+
+    private def convertEndpoint[R, I](
+        endpoint: ServerEndpoint.Full[Unit, Unit, I, TapirResponse, CalibanResponse[NoStreams.BinaryStream], NoStreams, [x] =>> RIO[R, x]],
+        runtime: Runtime[R]
+    ): ServerEndpoint[Any, KyoSttpMonad.M] =
+        ServerEndpoint[Unit, Unit, I, TapirResponse, CalibanResponse[NoStreams.BinaryStream], Any, KyoSttpMonad.M](
+            endpoint.endpoint.asInstanceOf[Endpoint[Unit, I, TapirResponse, CalibanResponse[NoStreams.BinaryStream], Any]],
```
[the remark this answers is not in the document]
ghostdogpr: I couldn't find a way. The problem is with the last type parameter, `R` which should be `Any`. But we're using `NoStreams` because Caliban does require a `Stream` implementation and it's the only once we can use for now. The cast can also be removed once streams are supported in `NettyKyoServer`.

ghostdogpr:
Documentation added! I'll create the additional issue for Streaming/WebSocket support after this is merged.

README.md
```diff
@@ -2147,6 +2149,57 @@ val d: Task[Int] =
...
+The first integration is that you can use Kyo effects inside your Caliban schemas.
+- If your Kyo effects is `(Aborts[Throwable] & ZIOs)` or a subtype of it, a Caliban `Schema` can be derived automatically.
+- If your Kyo effect is something else, a Caliban schema can be derived if it has a `Runner` for that effect as part of ZIO environment.
+
+```scala
+// this works by just importing kyo.*
+case class Query(k: Int < Aborts[Throwable]) derives Schema.SemiAuto
+
+// for other effects, you need to extend `SchemaDerivation[Runner[YourEnv]]`
+type Env = Vars[Int] & Envs[String]
```
hearnadam, line 2165: I would prefer not to overload the term `Env` - potentially confusing with `Envs[A]` or ZIO's `R` (which is often type aliased to `Env`....

Can we use `Effects`? or `Pending`?

README.md
```diff
@@ -2147,6 +2149,57 @@ val d: Task[Int] =
 
 > Note: Support for ZIO environments (`R` in `ZIO[R, E, A]`) is currently in development. Once implemented, it will be possible to use ZIO effects with environments directly within Kyo computations.
 
+### Resolvers: GraphQL Server via Caliban
+
+`Resolvers` integrates with the [Caliban](https://github.com/ghostdogpr/caliban) library to help setup GraphQL servers.
+
+The first integration is that you can use Kyo effects inside your Caliban schemas.
+- If your Kyo effects is `(Aborts[Throwable] & ZIOs)` or a subtype of it, a Caliban `Schema` can be derived automatically.
```
hearnadam, line 2157: Might be worth noting that ZIOs includes Fibers/IOs.

hearnadam (approved):
Nice work!

README.md
ghostdogpr: Changed to `CustomEffects`

README.md
ghostdogpr: Added 👍

ghostdogpr:
Created follow-up issues: https://github.com/getkyo/kyo/issues/441 https://github.com/getkyo/kyo/issues/442

fwbrasil:
I'm happy to see this merged!!!! Thanks folks!!!!

PR #424, opened 2024-05-27
Update `Envs` to use type level map

kitlangton:
I'll rebase this atop @hearnadam's changes once they're in.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/envs.scala
+++ b/kyo-core/shared/src/main/scala/kyo/envs.scala
@@ -38,12 +38,16 @@ object Envs:
       *
       * The environment may either be a value or a [[TypeMap]]
       */
-    inline def run[In, V, T: Flat, S, VS, VR](env: In)(value: T < (Envs[VS] & S))(
-        using
-        inline runInput: RunInput[In] { type Out = TypeMap[V] },
-        inline hasEnvs: HasEnvs[V, VS] { type Remainder = VR }
+    inline def runTypeMap[V, T: Flat, S, VS, VR](env: TypeMap[V])(value: T < (Envs[VS] & S))(
+        using HasEnvs[V, VS] { type Remainder = VR }
     ): T < (S & VR) =
-        envs[V].handle(handler[V])(runInput(env), value).asInstanceOf[T < (S & VR)]
+        envs[V].handle(handler[V])(env, value).asInstanceOf[T < (S & VR)]
+    end runTypeMap
+
+    inline def run[V: Tag, T: Flat, S, VS, VR](env: V)(value: T < (Envs[VS] & S))(
+        using HasEnvs[V, VS] { type Remainder = VR }
+    ): T < (S & VR) =
+        envs[V].handle(handler[V])(TypeMap(env), value).asInstanceOf[T < (S & VR)]
     end run
 
     trait RunInput[-In]:
--- a/kyo-core/shared/src/main/scala/kyo/envs.scala
+++ b/kyo-core/shared/src/main/scala/kyo/envs.scala
@@ -70,8 +70,18 @@ object Envs:
             def resume[T, U: Flat, S2](st: TypeMap[Any], command: Tag[T], k: T => U < (Envs[Any] & S2))(using Tag[Envs[Any]]) =
                 Resume(st, k(st.get(using command.asInstanceOf[Tag[Any]]).asInstanceOf[T]))
 
+    /** An effect `Envs[VS]` includes a dependency on `V`, and once `V` has been handled, `Envs[VS]` should be replaced by `Out`
+      *
+      * @tparam V
+      *   the dependency included in `VS`
+      * @tparam VS
+      *   all of the `Envs` dependencies represented by type intersection
+      */
     sealed trait HasEnvs[V, +VS]:
+        /** Remaining effect type, once the `V` dependency has been provided
+          */
         type Remainder
+    end HasEnvs
 
     trait LowPriorityHasEnvs:
         given hasEnvs[V, VR]: HasEnvs[V, V & VR] with
--- a/kyo-core/shared/src/main/scala/kyo/envs.scala
+++ b/kyo-core/shared/src/main/scala/kyo/envs.scala
@@ -57,15 +57,6 @@ object Envs:
         envs[V].handle(handler[V])(env, value).asInstanceOf[T < (S & VR)]
     end provide
 
-    def provideReverse[V >: Nothing, T: Flat, S, VS, VR](value: T < (Envs[VS] & S))(env: TypeMap[V])(
-        using
-        HasEnvs[V, VS] { type Remainder = VR },
-        Tag.Intersection[V]
-    ): T < (S & VR) =
-        given Tag[Envs[V]] = makeEnvsTag[V]
-        envs[V].handle(handler[V])(env, value).asInstanceOf[T < (S & VR)]
-    end provideReverse
-
     private def handler[V](using intersection: Tag.Intersection[V]) =
         new ResultHandler[TypeMap[V], Tag, Envs[V], Id, Any]:
             override def accepts[T](st: TypeMap[V], command: Tag[T]): Boolean =
--- a/kyo-core/shared/src/main/scala/kyo/envs.scala
+++ b/kyo-core/shared/src/main/scala/kyo/envs.scala
@@ -21,19 +21,19 @@ object Envs:
     def get[V](using tag: Tag[V]): V < Envs[V] =
         envs[V].suspend[V](tag)
 
-    inline def run[V: Tag, T: Flat, S, VS, VR](env: V)(value: T < (Envs[VS] & S))(
+    def run[V >: Nothing: Tag, T: Flat, S, VS, VR](env: V)(value: T < (Envs[VS] & S))(
         using HasEnvs[V, VS] { type Remainder = VR }
     ): T < (S & VR) =
         envs[V].handle(handler[V])(TypeMap(env), value).asInstanceOf[T < (S & VR)]
     end run
 
-    inline def runTypeMap[V, T: Flat, S, VS, VR](env: TypeMap[V])(value: T < (Envs[VS] & S))(
+    def runTypeMap[V >: Nothing, T: Flat, S, VS, VR](env: TypeMap[V])(value: T < (Envs[VS] & S))(
         using HasEnvs[V, VS] { type Remainder = VR }
     ): T < (S & VR) =
         envs[V].handle(handler[V])(env, value).asInstanceOf[T < (S & VR)]
     end runTypeMap
 
-    class UseDsl[V]:
+    class UseDsl[V >: Nothing]:
         inline def apply[T, S](inline f: V => T < S)(
             using
             inline intersection: Tag.Intersection[V],
@@ -43,7 +43,7 @@ object Envs:
         end apply
     end UseDsl
 
-    def use[V]: UseDsl[V] =
+    def use[V >: Nothing]: UseDsl[V] =
         new UseDsl[V]
 
     private def handler[V]: ResultHandler[TypeMap[V], Tag, Envs[V], Id, Any] =
--- a/kyo-core/shared/src/main/scala/kyo/envs.scala
+++ b/kyo-core/shared/src/main/scala/kyo/envs.scala
@@ -15,8 +15,8 @@ object Envs:
     // However, then handler.tag != kyo.tag in the handle implementation.
     // So we have to use this type erased tag and rely on our `accepts` implementation
     private trait EnvsErased
-    private val envsTag: Tag[EnvsErased]     = Tag[EnvsErased]
-    private def makeEnvsTag[V]: Tag[Envs[V]] = envsTag.asInstanceOf[Tag[Envs[V]]]
+    private val envsTag: Tag[EnvsErased]            = Tag[EnvsErased]
+    private inline def makeEnvsTag[V]: Tag[Envs[V]] = envsTag.asInstanceOf[Tag[Envs[V]]]
 
     def get[V](using tag: Tag[V]): V < Envs[V] =
         envs[V].suspend[V](tag)(using makeEnvsTag[V])
@@ -63,7 +63,7 @@ object Envs:
     private val cachedHandler =
         new ResultHandler[TypeMap[Any], Tag, Envs[Any], Id, Any]:
             override def accepts[T](st: TypeMap[Any], command: Tag[T]): Boolean =
-                st.tag <:< command
+                st <:< command
 
             def done[T](st: TypeMap[Any], v: T)(using Tag[Envs[Any]]) = v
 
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -3,20 +3,32 @@ package kyo
...
-    def get[V](using Tag[Envs[V]]): V < Envs[V] =
-        envs[V].suspend[V](())
+    // Envs.get[String] would normally use a tag of Envs[String]
+    // Yet we may provide a Layer with an output of String & Int.
+    // This would usually give the handler a tag of Envs[String & Int]
+    // However, then handler.tag != kyo.tag in the handle implementation.
+    // So we have to use this type erased tag and rely on our `accepts` implementation
+    private trait EnvsErased
+    private val envsTag: Tag[EnvsErased]     = Tag[EnvsErased]
+    private def makeEnvsTag[V]: Tag[Envs[V]] = envsTag.asInstanceOf[Tag[Envs[V]]]
```
kitlangton, line 19: This here was the tricky part.

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -26,35 +38,40 @@ object Envs:
...
-      *   the dependency included in `VS`
-      * @tparam VS
-      *   all of the `Envs` dependencies represented by type intersection
-      */
-    sealed trait HasEnvs[V, VS]:
-        /** Remaining effect type, once the `V` dependency has been provided
-          */
+        new ResultHandler[TypeMap[Any], Tag, Envs[Any], Id, Any]:
+            override def accepts[T](st: TypeMap[Any], command: Tag[T]): Boolean =
+                st.tag <:< command
```
kitlangton, line 66: Perhaps we should make tag a lazy val? Or I can make the state a tuple of the type map and its intersection tag.

fwbrasil: Agree! Avoiding the allocations to obtain the tag is important here since it's a hot path. I think a lazy val would work well since the tuple allocation should be more expensive.

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -26,22 +38,37 @@ object Envs:
...
+            envs[V].handle(handler[V])(TypeMap(env), value).asInstanceOf[T < (S & VR)]
+        end apply
     end RunDsl
 
     def run[V >: Nothing]: RunDsl[V] =
         new RunDsl[V]
 
-    private def handler[V]: ResultHandler[V, Const[Unit], Envs[V], Id, Any] =
-        cachedHandler.asInstanceOf[ResultHandler[V, Const[Unit], Envs[V], Id, Any]]
+    def provide[V >: Nothing, T: Flat, S, VS, VR](env: TypeMap[V])(value: T < (Envs[VS] & S))(
```
fwbrasil, line 51: How about runMap? Effect handling methods are typically named run*.

kyo-core/shared/src/main/scala/kyo/envs.scala
fwbrasil: We should try to find a safer encoding later but just the fact that the effect handling still doesn't require allocations seems worth the workaround :)

fwbrasil:
Would we still need a separate API for layers?

hearnadam:
Yes. Layers will support memoization and macro for ordering dependencies

kyo-core/shared/src/main/scala/kyo/envs.scala
kitlangton: The new State parameter of the handler is brilliant! The duality between inputs to recursive functions and state is often worth exploiting.

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -3,45 +3,77 @@ package kyo
...
+      *
+      * The environment may either be a value or a [[TypeMap]]
+      */
+    inline def run[In, V, T: Flat, S, VS, VR](env: In)(value: T < (Envs[VS] & S))(
+        using
+        inline runInput: RunInput[In] { type Out = TypeMap[V] },
+        inline hasEnvs: HasEnvs[V, VS] { type Remainder = VR }
+    ): T < (S & VR) =
+        envs[V].handle(handler[V])(runInput(env), value).asInstanceOf[T < (S & VR)]
+    end run
```
kitlangton, line 47: I took some inspiration from `ZLayers.fromFunction` to allow the same `run` method to be called with either `TypeMap`s or values. Perhaps this is a bit too sneaky 😜, but I just wanted to test it out. Let me know what you think. I'm happy to break it into multiple methods.

The only issue is that we'll eventually need another method overload for Layers. Then again, perhaps having `run`,`runTypeMap`, and `runLayer` isn't the end of the world. I'd love to know your thoughts, as I haven't yet fully absorbed the _kyo style_ 😎.

kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -10,7 +10,7 @@ extension [A](self: TypeMap[A])
         throw new RuntimeException(s"fatal: kyo.TypeMap of contents [$self] missing value of type: [${t.show}].")
 
     inline def get[B >: A](using inline t: Tag[B]): B =
-        self.getOrElse(t, fatal).asInstanceOf[B]
+        self.collectFirst { case (tag, value) if tag <:< t => value }.getOrElse(fatal).asInstanceOf[B]
```
kitlangton, line 13: By doing this we don't have to explicitly widen subtypes. This is basically how ZEnvironment works.
However, since this type map is immutable and usually provided once at the _top of the application,_ ZEnvironment also keeps a cache of successful lookups, so only the first lookup needs to be O(n).

kyo-core/shared/src/main/scala/kyo/envs.scala
kitlangton: Update: I broke these back out into separate methods for now 😛 . But I'm still curious to know your thoughts regarding that general degree of API _cleverness_.

kyo-core/shared/src/main/scala/kyo/envs.scala
kitlangton: It also kind of reminds me of how Futures in Rust are compiled down to state machines. It really cleans things up! 👏

kyo-core/shared/src/main/scala/kyo/typeMap.scala
hearnadam: Can you explain how the first lookup will be `O(1)`? I will cut a PR to add unit tests + attempt a slightly more optimized option.

kitlangton: Oh, no, the first lookup needs to find the subtype by walking through each item (unless there's some other clever algorithm we can come up with). But once it finds the value, it'll put it in the cache.

kyo-core/shared/src/main/scala/kyo/envs.scala
fwbrasil: yep! It's 🤯 to think that the effect handling machinery can execute quite efficiently while still being pure without any mutability

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -3,45 +3,61 @@ package kyo
...
-    type Command[T] = Unit
+    type Command[T] = Tag[T]
 
 object Envs:
     private case object envs extends Envs[Any]
     private def envs[V]: Envs[V] = envs.asInstanceOf[Envs[V]]
 
-    def get[V](using Tag[Envs[V]]): V < Envs[V] =
-        envs[V].suspend[V](())
+    // Envs.get[String] would normally use a tag of Envs[String]
```
fwbrasil, line 12: Can you add `TODO` at the start?

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -3,45 +3,61 @@ package kyo
...
+    // However, then handler.tag != kyo.tag in the handle implementation.
+    // So we have to use this type erased tag and rely on our `accepts` implementation
+    private trait EnvsErased
+    private val envsTag: Tag[EnvsErased]   = Tag[EnvsErased]
+    private inline given [V]: Tag[Envs[V]] = envsTag.asInstanceOf[Tag[Envs[V]]]
+
+    def get[V](using tag: Tag[V]): V < Envs[V] =
+        envs[V].suspend[V](tag)
+
+    inline def run[V: Tag, T: Flat, S, VS, VR](env: V)(value: T < (Envs[VS] & S))(
```
fwbrasil, line 24: We typically don't  `inline` effect handling because it produces a large tree.

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -3,45 +3,61 @@ package kyo
...
+        using HasEnvs[V, VS] { type Remainder = VR }
+    ): T < (S & VR) =
+        envs[V].handle(handler[V])(TypeMap(env), value).asInstanceOf[T < (S & VR)]
```
fwbrasil, line 27: I'd prefer to avoid the `TypeMap` map allocation. Could this method use a dedicated handler?

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -3,45 +3,61 @@ package kyo
...
+    end run
+
+    inline def runTypeMap[V, T: Flat, S, VS, VR](env: TypeMap[V])(value: T < (Envs[VS] & S))(
```
fwbrasil, line 30: remove `inline` as well

kyo-core/shared/src/main/scala/kyo/envs.scala
```diff
@@ -3,45 +3,61 @@ package kyo
...
+            inline intersection: Tag.Intersection[V],
+            inline tag: Tag[V]
         ): T < (Envs[V] & S) =
-            envs[V].suspend[V, T, S]((), f)
+            envs[V].suspend[V, T, S](tag, f)
+        end apply
     end UseDsl
 
-    def use[V >: Nothing]: UseDsl[V] =
+    def use[V]: UseDsl[V] =
```
fwbrasil, line 46: hmm I think adding `>: Nothing` here had fixed an issue. Was this change necessary to fix a compilation error?

kyo-core/shared/src/main/scala/kyo/typeMap.scala
fwbrasil: Maybe we could fist check if the tag isn't resolved directly from the map first and only check individual items if it returns false? It think it'd provide a fast path in many scenarios.

kyo-tapir/shared/src/main/scala/kyo/routes.scala
```diff
@@ -30,8 +30,10 @@ object Routes:
     ): Unit < Routes =
         Sums.add(
             Route(
-                e.serverSecurityLogic[A, KyoSttpMonad.M](a => Right(a)).serverLogic(a =>
```
fwbrasil, line 33: Did you have type inference issues here?

kyo-core/shared/src/main/scala/kyo/typeMap.scala
hearnadam: Yep! That's why my PR aims to do https://github.com/getkyo/kyo/pull/427

fwbrasil: Cool! I hadn't noticed the initial lookup.

kyo-core/shared/src/main/scala/kyo/envs.scala
kitlangton: Oh, really? No, I just removed it because `Nothing >: Nothing`, so I think it's unnecessary. But Scala be weird sometimes.

kyo-tapir/shared/src/main/scala/kyo/routes.scala
kitlangton: I did! Maybe that's the the Nothing >: thing! Or something else I did.

kitlangton: Let me try to revert the Nothing and see 😄

kyo-core/shared/src/main/scala/kyo/envs.scala
fwbrasil: We do have a few effects that have variations of `run*` so it makes sense to me. We normally aim for canonical APIs and avoid providing multiple ways to do the same thing, though. If `Layers` will provide function-based APIs, we might not need a `run*` that takes a function here.

kyo-core/shared/src/main/scala/kyo/envs.scala
fwbrasil: Yeah, I did notice it changes the type inference in some cases where `Nothing` would be inferred.

kyo-core/shared/src/main/scala/kyo/envs.scala
kitlangton: Hm. I just realized, then we might need to allocate a tuple for the state:

```scala
    private val cachedValueHandler =
        new ResultHandler[Tag[Any], Tag, Envs[Any], Id, Any]:
            override def accepts[T](st: (Any, Tag[Any]), command: Tag[T]): Boolean =
                st._2 <:< command // <--- HERE.      
```

We need both the value and a way to match on the command, because when we suspend, we're using the Tag[ErasedEnv], because we don't know which handler will eventually run the suspension. Is it still worth doing this if there's an allocation either way?

kyo-core/shared/src/main/scala/kyo/envs.scala
kitlangton: Wow! Yeah, adding this back fixed it 😅. Oh, Scala.

kyo-core/shared/src/main/scala/kyo/envs.scala
fwbrasil: Oh, I see. I think we could change `core.handle` to pass the tag to `accepts` so we can avoid having to hold the tag in the state.

fwbrasil: we could do that as a follow up if you like

kyo-core/shared/src/test/scala/kyoTest/envsTest.scala
```diff
@@ -28,7 +29,7 @@ class envsTest extends KyoTest:
         trait Super:
             def i = 42
         case class Sub() extends Super
-        assert(Envs.run[Super](Sub())(Envs.use[Super](_.i)).pure == 42)
```
fwbrasil, line 31: Is this change still necessary after adding `>: Nothing` in `run`?

kyo-core/shared/src/main/scala/kyo/envs.scala
kitlangton: Ah, I think the problem is that the Effect tag needs to be `ErasedEnv` and here:
```scala
            override def accepts[T](st: TypeMap[Any], command: Tag[T]): Boolean =
                st <:< command
```

The TypeMap is both the State and ALSO can be used to check if it contains the Tag[Value] for the suspended Envs.get[Value].

But for the plain value handler, the state would simply be the Value.
```scala
            override def accepts[T](st: Any, command: Tag[T]): Boolean =
                ???
```

And we'd have no way of seeing if should handle this Tag. The tag of the effect is still going to be Tag[ErasedEnv], so that wouldn't help us in this case, even if it got passed in.

Am I making sense? I could also be confused 😅

kyo-core/shared/src/test/scala/kyoTest/envsTest.scala
kitlangton: Oh. This wasn't due to type inference reasons. It's just that we don't need to upcast, because the type map will provide Subtypes to Supertypes automatically.

kitlangton: And I inlined the `RunDsl` 😜 because I always find that pattern of allowing partially applied type parameters to be pretty confusing (esp. to newcomers command-clicking through the codebase). But I'm happy to revert if you'd prefer!

kyo-core/shared/src/main/scala/kyo/envs.scala
fwbrasil: Couldn't we use the regular `Tag[Envs[V]]` instead of `EnvsErased` in this method?

fwbrasil: btw, if this is getting complicated, I'm ok with merging with the allocation! We can address it later.

kyo-core/shared/src/test/scala/kyoTest/envsTest.scala
fwbrasil: makes sense, thanks for the clarification! Yep, I'm not a fan of the `*Dsl` encoding, hopefully we'll eventually be able to use [clause interleaving](https://docs.scala-lang.org/sips/clause-interleaving.html) instead but I think it's still marked as experimental.

kitlangton: Oh, awesome! I'm glad they're adding that.

kyo-core/shared/src/main/scala/kyo/envs.scala
kitlangton: The problem is that we need to use the same tag in the suspension _and_ the handler. And when we suspend, we won't know which handler is going to handle it :)

kitlangton: But definitely worth noting and possibly re-examining later. It'd certainly be nice to avoid. State could also be a union of `Tag[?] | TypeMap[?]` ? But that would add some cost to the lookup.

fwbrasil (approved):

fwbrasil:
Thanks Kit!!! I'm excited to see the layers impl :)

PR #425, opened 2024-05-27
Add kyo issues link to bug message

kitlangton:
I like the bug message! But I thought including a link + cute, tearful emoji might increase the likelihood of receiving issues.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/package.scala
+++ b/kyo-core/shared/src/main/scala/kyo/package.scala
@@ -91,6 +91,6 @@ package object kyo:
             if cond then bug(msg)
 
         def apply(msg: String): Nothing =
-            throw KyoBugException(msg + " Please file a ticket.")
+            throw KyoBugException(msg + " Please open an issue 🥹 https://github.com/getkyo/kyo/issues")
     end bug
 end kyo


hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/package.scala
```diff
@@ -91,6 +91,6 @@ package object kyo:
             if cond then bug(msg)
 
         def apply(msg: String): Nothing =
-            throw KyoBugException(msg + " Please file a ticket.")
+            throw KyoBugException(msg + " Please open an issue 🥹 https://github.com/getkyo/kyo/issues")
```
hearnadam, line 94: could probably remove the https:// to make it a bit shorter (hopefully fitting on people's terminals)...

fwbrasil: I think the full URL is fine, I imagine some editors could even make it clickable that way. I used 🙈 for the dramatic effect in the [scheduler code](https://github.com/getkyo/kyo/blob/c4a8a955c157bb2a36b731d6f1c4907db1d84383/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/package.scala#L16) 😅

fwbrasil (approved):

kyo-core/shared/src/main/scala/kyo/package.scala
kitlangton: > some editors could even make it clickable that way
Yeah, that's what I was imagining. Make it one command-click away :)

PR #427, opened 2024-05-27
TypeMap: add support for subtype lookups

hearnadam:


kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -10,7 +10,23 @@ extension [A](self: TypeMap[A])
...
 
     inline def get[B >: A](using inline t: Tag[B]): B =
-        self.getOrElse(t, fatal).asInstanceOf[B]
+        val b = self.getOrElse(t, null)
+        if !isNull(b) then b.asInstanceOf[B]
+        else
+            var sub: B = null.asInstanceOf[B]
+            val it     = self.iterator
+            try
+                while isNull(sub) do
```
hearnadam, line 19: Using `try/catch` should help with branch prediction, as we should almost always have an element in the iterator due to the type constraints.

kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -10,7 +10,23 @@ extension [A](self: TypeMap[A])
         throw new RuntimeException(s"fatal: kyo.TypeMap of contents [$self] missing value of type: [${t.show}].")
 
     inline def get[B >: A](using inline t: Tag[B]): B =
-        self.getOrElse(t, fatal).asInstanceOf[B]
+        val b = self.getOrElse(t, null)
+        if !isNull(b) then b.asInstanceOf[B]
+        else
+            var sub: B = null.asInstanceOf[B]
+            val it     = self.iterator
```
hearnadam, line 17: Unfortunately this will allocate tuples, but I am not sure that's cheaply avoidable. I think a keyset iterator will still be slower.

hearnadam: We can followup with some benchmarks vs ZEnvironment later.

kyo-core/shared/src/main/scala/kyo/typeMap.scala
fwbrasil: It might help branch prediction but throwing an exception that doesn't use `NoStackTrace` is much more expensive. I think checking `hasNext` instead would be ideal here.

hearnadam: We will throw an exception anyways...Defects/fatal should never occur if someone is not casting.

kyo-core/shared/src/main/scala/kyo/typeMap.scala
fwbrasil: Makes sense. We could later explore a dedicated map implementation for this use case.

kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -10,7 +10,19 @@ extension [A](self: TypeMap[A])
         throw new RuntimeException(s"fatal: kyo.TypeMap of contents [$self] missing value of type: [${t.show}].")
 
     inline def get[B >: A](using inline t: Tag[B]): B =
```
hearnadam, line 12: @fwbrasil do you think this is still worth inlining now?

fwbrasil: Probably not given the method size. There isn't much benefit as well since the implementation doesn't have calls that could become monomorphic if inlined.

fwbrasil (approved):

Issue #220, opened 2024-03-25
OS module


fwbrasil:
I started working on an [integration](https://github.com/getkyo/kyo/tree/main/kyo-os-lib/shared/src/main/scala/kyo) with [os-lib](https://github.com/com-lihaoyi/os-lib) but never had time to finalize it. We need a module in Kyo that provides primitives for dealing with files and issuing system commands.

I'm not sure if we should keep building it on top of `os-lib` since it's not designed to be integrated with effect systems.

hearnadam:
What would be the minimum set of operations this should include?

fwbrasil:
I expect the main use of the module to be by people working with LLMs. They typically need to read and write files, especially CSVs. I think it's essential to provide simple APIs for this use case, but we could also choose to have a broader scope similar to os-lib. Another alternative would be to check with the os-lib developers if they would be open to providing a parameterized API to integrate with effect systems. It should be possible to provide the reusable parameterized API but also "lower" it to continue offering os-lib's impure APIs (basically binding to an identity effect).

johnhungerford:
Why do you expect LLM work to be the main use? Is this a general design goal of kyo? There are a lot of use cases requiring reading and writing files.

For my part, here's what I would need:

1. Read file as bytes or string
2. Write bytes or string to file
3. Append bytes or string to file
4. Stream bytes or string chunks from a file
5. Stream lines from a file
6. Sink a stream of bytes or strings into a file
7. Sink a stream of strings into a file as lines
8. Stream files from a directory
9. Stream files from a directory recursively
10. Move, copy, remove a file/directory

fwbrasil:
> Why do you expect LLM work to be the main use? Is this a general design goal of kyo?

The market for Kyo within the current Scala community seems too limited. The community has been consistently shrinking, a significant portion of it has a strong non-technical bias against me and the project, and capturing market share from cats-effect and ZIO doesn't seem very feasible. I'd say companies are more likely to rewrite in another language than to adopt Kyo.

I wouldn't say that's a design goal but an attempt to find a new market to make the library viable in the long term and ensure it evolves as a healthy OSS project. I'd say there's a good chance LLMs will become the "killer application" for Kyo.

> There are a lot of use cases requiring reading and writing files

Cool! And we have `Streams` now :) I don't think we need to limit the scope of this ticket. My comment was more in the direction of simplifying the initial scope of the module if necessary.

fwbrasil:
Bounty requirements:

[...]

PR #428, opened 2024-05-27
OS module implementation

pablf:
/claim #220
An implementation of primitives for reading and writing files and managing processes. For the files part, there is an extension that work on any type with a `PathLike` typeclass, so it is possible to apply all the methods on strings, list of strings java Paths and combination of them:
```
"somefolder".mkDir // creates directory somefolder
("somefolder/otherfolder" / "somefile.txt").mkFile // creates file "somefolder/otherfolder/somefile.txt"
(List("somefolder", "otherfolder") / "somefile.txt").mkFile // creates file "somefolder/otherfolder/somefile.txt"
```
This can be expanded to other types and be used for integration with other libraries. We can try another approach if you think it's too "magical".

There are a few tests involving streams that throw runtime exceptions but I feel like they should work. Am I doing something not in the intended way or is it a bug of streams? I've tried different ways to avoid them but it always appear.
```
"walk" in run {
            val folder = "folder"
            val path1 = folder / "path1"
            val path2 = folder / "path2"
            val createFolders = for {
                _ <- folder.mkDir
                _ <- path1.mkFile
                _ <- path2.mkFile
                
            } yield ()
            assert (IOs.run(createFolders.andThen(folder.walk.map(_.runSeq))).pure._1.map(_.toString).toSet == Set("folder/path1", "folder/path2"))
        }
```
throws ` kyo.package$bug$KyoBugException: Unexpected effect 'kyo.Streams' found while handling 'kyo.IOs'. Please open an issue` and
```
"read file as string stream" in run {
            val name = "read-file-string.txt"
            val text = "some text"
            for
                _   <- IOs(createFile(name, text))
                res <- Resources.run(name.readStream())
                _   <- IOs(destroyFile(name))
                v   <- Fibers.init(res.runSeq).map(_.get)
            yield assert(v._1 == List("some text"))
            end for
        }
```
throws `kyo.package$bug$KyoBugException: Unexpected effect 'kyo.Streams' found while handling 'kyo.fibersInternal$.FiberGets & kyo.IOs'. Please open an issue`.


kyo-os-lib/shared/src/main/scala/kyo/os.scala
```diff
@@ -0,0 +1,213 @@
...
+import java.lang.Process as JProcess
+import java.lang.ProcessBuilder.Redirect
+import java.nio.charset.Charset
+import java.nio.charset.StandardCharsets
+import java.nio.file.Path
+import java.util.concurrent.TimeUnit
+import kyo.*
+import scala.jdk.CollectionConverters.*
+
+sealed trait Command:
```
fwbrasil, line 13: We use `Command` for values used in effect suspensions, what do you think about renaming this to `ProcessCommand` to avoid cofusion?

kyo-os-lib/shared/src/main/scala/kyo/os.scala
```diff
@@ -0,0 +1,213 @@
...
+    self =>
+    import Command.*
+
+    /** Spawns a new process executing this command
+      */
+    def spawn: Process < (IOs)
```
fwbrasil, line 19: There's no need for the parenthesis when there's only one pending effect.

kyo-os-lib/shared/src/main/scala/kyo/os.scala
```diff
@@ -0,0 +1,213 @@
...
+    /** Spawns a new process and returns its output as a String
+      */
+    def string: String < (IOs) =
+        stream.map { inputStream =>
+            new String(inputStream.readAllBytes())
+        }
+
+    /** Spawns a new process and returns its output as a Stream
+      */
+    def stream =
```
fwbrasil, line 30: Could you add return types to all public methods?

kyo-os-lib/shared/src/main/scala/kyo/os.scala
```diff
@@ -0,0 +1,213 @@
...
+        stdin: ProcessInput = ProcessInput.Inherit,
+        stdout: ProcessOutput = ProcessOutput.Pipe,
+        stderr: ProcessOutput = ProcessOutput.Pipe,
+        redirectErrorStream: Boolean = false
+    ) extends Command:
+        self =>
+        override def spawn: Process < (IOs) =
+            for
+                process <- IOs {
+                    val builder = new ProcessBuilder(command*)
```
fwbrasil, line 105: It doesn't need to be in the scope of this change but something that would be useful to have is a way to fork a `Function0` into a separate JVM process. We'd need to find the current java binary, setup the classpath, and pass the serialized function to be executed.

kyo-os-lib/shared/src/test/scala/kyoTest/filesTest.scala
```diff
@@ -0,0 +1,236 @@
...
+            yield assert(v == text)
+        }
+        "read file as bytes" in run {
+            val name = "read-file-string"
+            val text = "some text"
+            val eff = useFile(name, text).map { _ =>
+                name.readBytes
+            }
+            for
+                v <- Fibers.init(eff).map(_.get)
```
fwbrasil, line 38: is this fiber indirection necessary?

kyo-os-lib/shared/src/main/scala/kyo/os.scala
```diff
@@ -0,0 +1,213 @@
...
+    def waitFor =
+        spawn.map(_.waitFor)
+
+    def waitFor(timeout: Long, unit: TimeUnit) =
+        spawn.map(_.waitFor(timeout, unit))
+
+    def pipe(that: Command): Command =
+        Command.Piped(self.flatten ++ that.flatten).simplify
+
+    def >>(that: Command): Command =
```
fwbrasil, line 45: We typically avoid symbolic operators in Kyo. What do you think about `andThen`?

build.sbt
```diff
@@ -214,7 +214,9 @@ lazy val `kyo-os-lib` =
         .dependsOn(`kyo-core` % "test->test;compile->compile")
         .settings(
             `kyo-settings`,
-            libraryDependencies += "com.lihaoyi" %% "os-lib" % "0.10.1"
+            scalacOptions --= Seq(
+                "-language:strictEquality"
```
fwbrasil, line 218: It'd be nice if we could keep this flag. If you have a type that should allow equality checks, it's possible to add `derives CanEqual` to it. In case you want to disable strict equality for a scope, it's possible to use `given [T]: CanEqual[T, T] = CanEqual.derived`.

kyo-core/shared/src/main/scala/kyo/streams.scala
```diff
@@ -78,6 +78,24 @@ object Stream:
             Streams[V].handle(handler)((), s)
         end transform
 
+        def transformChunks[V2: Flat, S2](f: Chunk[V] => Chunk[V2] < S2)(
```
fwbrasil, line 81: could you add a test for this?

kyo-os-lib/shared/src/main/scala/kyo/files.scala
```diff
@@ -1,139 +1,405 @@
...
+import java.io.IOException
+import java.nio.*
+import java.nio.channels.FileChannel
+import java.nio.charset.*
+import java.nio.file.*
+import java.nio.file.Path as JPath
+import java.nio.file.attribute.BasicFileAttributes
 import kyo.*
-import os.*
+import kyo.Flat.unsafe.*
```
fwbrasil, line 11: The flat check is crucial to ensure the correct execution of Kyo computations. I imagine you only need to pass the `Flat` implicit where it's required.

kyo-os-lib/shared/src/main/scala/kyo/files.scala
```diff
@@ -1,139 +1,405 @@
...
+            createFolders: Boolean = true
+        ): Unit < IOs =
+            val opts = (if atomicMove then List(StandardCopyOption.ATOMIC_MOVE) else Nil) ++ (if replaceExisting then
+                                                                                                  List(StandardCopyOption.REPLACE_EXISTING)
+                                                                                              else Nil)
+            toJPath.getParent().exists.map { parentExists =>
+                (parentExists, createFolders) match
+                    case (true, _)     => IOs(Files.move(toJPath, to.toJPath, opts*)).map(_ => ())
+                    case (false, true) => toJPath.getParent().mkDir.andThen(IOs(Files.move(toJPath, to.toJPath, opts*)).map(_ => ()))
+                    case _             => IOs {}
```
fwbrasil, line 297: you can return`()` directly

kyo-os-lib/shared/src/main/scala/kyo/files.scala
```diff
@@ -1,139 +1,405 @@
...
+            atomicMove: Boolean = false,
+            createFolders: Boolean = true
+        ): Unit < IOs =
+            val opts = (if atomicMove then List(StandardCopyOption.ATOMIC_MOVE) else Nil) ++ (if replaceExisting then
+                                                                                                  List(StandardCopyOption.REPLACE_EXISTING)
+                                                                                              else Nil)
+            toJPath.getParent().exists.map { parentExists =>
+                (parentExists, createFolders) match
+                    case (true, _)     => IOs(Files.move(toJPath, to.toJPath, opts*)).map(_ => ())
+                    case (false, true) => toJPath.getParent().mkDir.andThen(IOs(Files.move(toJPath, to.toJPath, opts*)).map(_ => ()))
```
fwbrasil, line 296: `.map(_ => ())` => `.unit`

kyo-os-lib/shared/src/main/scala/kyo/os.scala
```diff
@@ -0,0 +1,213 @@
...
+    def stdin: OutputStream                                   = process.getOutputStream
+    def stdout: InputStream                                   = process.getInputStream
+    def stderr: InputStream                                   = process.getErrorStream
+    def waitFor: Int < IOs                                    = IOs(process.waitFor())
+    def waitFor(timeout: Long, unit: TimeUnit): Boolean < IOs = IOs(process.waitFor(timeout, unit))
+    def exitValue: Int < IOs                                  = IOs(process.exitValue())
+    def destroy: Unit < IOs                                   = IOs(process.destroy())
+    def destroyForcibly: JProcess < IOs                       = IOs(process.destroyForcibly())
+    def isAlive: Boolean < IOs                                = IOs(process.isAlive())
+    def close                                                 = IOs.run(destroy).pure
```
fwbrasil, line 184: Can we model the use of `Process` in terms of `Resources` instead of exposing a `close` method?

kyo-os-lib/shared/src/main/scala/kyo/files.scala
```diff
@@ -1,139 +1,405 @@
...
+                            s <- IOs {
+                                val arr = new Array[Byte](b)
+                                buf.get(arr)
+                                new String(arr, charset)
+                            }
+                            _            <- channel.put(Chunks.init(s))
+                            newBytesRead <- IOs(fileChannel.read(buf))
+                        yield Loops.continue(newBytesRead)
+                }
+                _ <- IOs(f)
```
fwbrasil, line 198: There's no need for this `IOs` suspension given that the loop already suspends side effects.

fwbrasil:
@pablf Thank you for the contribution! 

> This can be expanded to other types and be used for integration with other libraries. We can try another approach if you think it's too "magical".

We typically avoid introducing features via implicit extension. Could you use an approach like before with a `Files(path)`?

> There are a few tests involving streams that throw runtime exceptions but I feel like they should work. Am I doing something not in the intended way or is it a bug of streams? I've tried different ways to avoid them but it always appear.

I can't spot anything wrong in the code, it might be a bug in Kyo. Could you try to reproduce it isolation?

fwbrasil:
> @fwbrasil Great to be onboard! This reproduces the bug:
> 
> ```
> val stream: Stream[Unit, Int, Any] < IOs =
>     IOs(Seq(1, 2, 3)).map(seq => Streams.initSeq(seq))
> val v = IOs.run(stream.map(_.runSeq)).pure
> ```

Thank you for isolating it! I've opened a PR with a fix: https://github.com/getkyo/kyo/pull/446

kyo-os-lib/shared/src/main/scala/kyo/jvm.scala
```diff
@@ -0,0 +1,24 @@
...
+    def spawn(klass: Class[?], args: List[String] = Nil)(using Trace): Process < IOs =
+        process(klass, args).spawn
+    end spawn
+
+    /** Returns a `ProcessCommand` representing the execution of the `klass` Class in a new JVM process. To finally execute the command, use
+      * `spawn` or use directly `jvm.spawn`.
+      */
+    def process(klass: Class[?], args: List[String] = Nil): ProcessCommand =
+        val javaHome  = System.getProperty("java.home") + File.separator + "bin" + File.separator + "java"
+        val classPath = System.getProperty("java.class.path")
```
hearnadam, line 18: Can either of these throw?

kyo-os-lib/shared/src/main/scala/kyo/os.scala
```diff
@@ -0,0 +1,214 @@
...
+    import ProcessCommand.*
+
+    /** Spawns a new process executing this command
+      */
+    def spawn(using Trace): Process < IOs
+
+    /** Spawns a new process and returns its output as a String
+      */
+    def string(using Trace): String < IOs =
+        stream.map { inputStream =>
```
hearnadam, line 25: Maybe `text`?

kyo-os-lib/shared/src/test/scala/kyo/jvmTest.scala
```diff
@@ -0,0 +1,25 @@
+package kyo
+
+import kyo.*
+
+class jvmTest extends KyoTest:
+
+    "execute new jvm and throw error" in run {
+        assert {
+            IOs.run(jvm.process(classOf[MainClass.type], "some-arg" :: Nil).waitFor).pure == 1
```
hearnadam, line 9: Perhaps it would be useful to have `ExitCode.Failure/Success` constants?

kyo-os-lib/shared/src/main/scala/kyo/files.scala
```diff
@@ -1,86 +1,285 @@
...
+import java.nio.file.attribute.BasicFileAttributes
 import kyo.*
 import kyo.internal.Trace
-import os.*
 import scala.io.*
+import scala.jdk.CollectionConverters.*
+import scala.jdk.StreamConverters.*
 
 class Files(val path: List[String]):
+    def toJPath: JPath               = Paths.get(path.mkString("/"))
```
hearnadam, line 19: Let's use `toJava`

fwbrasil:
> @fwbrasil Stream methods working after #446! I have added a JVM module. Is it the intended functionality? Also applied the rest of the feedback.

Great news! LGTM besides @hearnadam's comments.

fwbrasil:
@pablf @hearnadam is there anything pending to merge this?

fwbrasil:
congrats on your first contribution @pablf! 🥳

PR #431, opened 2024-05-28
fix: `.pure` when used with intersections types

hearnadam:
This is necessary due to `Flat` now being derived from `Tag`s (#366). Not sure if there is a way we can bypass that only in some cases.

Rarely will effects ever need to be run with values that are intersection types, but for Layers (#203), this is definitely necessary.


fwbrasil:
You beat me to it! :) Even if we improve `Flat` as I'm exploring in https://github.com/getkyo/kyo/pull/432, I think it's better if we're able to encode the desired behavior without a macro as this fix does.

Issue #430, opened 2024-05-28
`.pure` should accept `T < Any` where `T` contains an intersection type


hearnadam:
When working with @kitlangton on #203, I noticed that the `Flat` bound on `T` ruins our ability to get the value from Kyos resulting in a `TypeMap[A]` when A was an intersection.

https://github.com/getkyo/kyo/blob/main/kyo-core/shared/src/main/scala/kyo/package.scala#L38-L45

I have an idea of how to solve this, but I need to ensure it is sound.

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/432

PR #432, opened 2024-05-28
tags: improve interop with Union and Intersection

fwbrasil:
The current `Tag` impl has unrelated representations for `Tag` (class types only) and set types (`Union` and `Intersection` of class types). This split has important limitations:

1. The `Envs` effect was [changed](https://github.com/getkyo/kyo/pull/424) by @hearnadam and @kitlangton recently to provide support for intersection types but it [required a hack](https://github.com/getkyo/kyo/blob/921c2f5738098c5a6245ac717d0dd3b9a47b0ab0/kyo-core/shared/src/main/scala/kyo/envs.scala#L12). To avoid the hack we'd need to support set tags in `core.handle` but it's not possible to implement that properly without a common type for the different kinds of tags.
2. The `Flat` check functions similarly to `Tag` since both reject types that don't represent concrete values. We could reimplement it in terms of `Tag` but that's also not possible to do that properly. I think making this change could be a fix for https://github.com/getkyo/kyo/issues/430.

This PR introduces the following types:

1. `Tag`: Still can only represent concrete class types and fails at compile time for unions and intersections
2. `Tag.Union`: Represents union types with concrete types only (`Seq[Tag[Any]]`)
3. `Tag.Intersection`: Same as `Union` but for intersection types
4. `Tag.Set`: Can be both `Union` or `Intersection`
5. `Tag.Full`: A union of all the previous types

With this representation, we can change `core.handle` to take a `Tag.Full` and each effect can chose to to pass a regular `Tag` or any of the others.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/typeMap.scala
+++ b/kyo-core/shared/src/main/scala/kyo/typeMap.scala
@@ -45,6 +45,9 @@ extension [A](self: TypeMap[A])
 end extension
 
 object TypeMap:
+
+    given flat[A]: Flat[TypeMap[A]] = Flat.unsafe.bypass
+
     val empty: TypeMap[Any] = HashMap.empty
 
     def apply[A](a: A)(using ta: Tag[A]): TypeMap[A] =
--- a/kyo-core/shared/src/main/scala/kyo/typeMap.scala
+++ b/kyo-core/shared/src/main/scala/kyo/typeMap.scala
@@ -3,15 +3,15 @@ package kyo
 import kyo.Tag.Intersection
 import scala.collection.immutable.HashMap
 
-opaque type TypeMap[+A] = HashMap[Tag[?], Any]
+opaque type TypeMap[+A] = HashMap[Tag[Any], Any]
 
 extension [A](self: TypeMap[A])
 
-    private inline def fatal(using t: Tag[?]): Nothing =
+    private inline def fatal[T](using t: Tag[T]): Nothing =
         throw new RuntimeException(s"fatal: kyo.TypeMap of contents [$self] missing value of type: [${t.show}].")
 
     def get[B >: A](using t: Tag[B]): B =
-        val b = self.getOrElse(t, null)
+        val b = self.getOrElse(t.erased, null)
         if !isNull(b) then b.asInstanceOf[B]
         else
             var sub: B = null.asInstanceOf[B]
@@ -30,7 +30,7 @@ extension [A](self: TypeMap[A])
     end get
 
     inline def add[B](b: B)(using inline t: Tag[B]): TypeMap[A & B] =
-        self.updated(t, b)
+        self.updated(t.erased, b)
 
     inline def union[B](that: TypeMap[B]): TypeMap[A & B] =
         self ++ that
@@ -40,19 +40,29 @@ extension [A](self: TypeMap[A])
 
     private[kyo] inline def tag: Intersection[?] = Intersection(self.keySet.toIndexedSeq)
 
-    private[kyo] inline def <:<(tag: Tag[?]): Boolean =
+    private[kyo] inline def <:<[T](tag: Tag[T]): Boolean =
         self.keySet.exists(_ <:< tag)
 end extension
 
 object TypeMap:
     val empty: TypeMap[Any] = HashMap.empty
 
-    def apply[A: Tag](a: A): TypeMap[A] =
-        HashMap(Tag[A] -> a)
-    def apply[A: Tag, B: Tag](a: A, b: B): TypeMap[A & B] =
-        HashMap(Tag[A] -> a, Tag[B] -> b)
-    def apply[A: Tag, B: Tag, C: Tag](a: A, b: B, c: C): TypeMap[A & B & C] =
-        HashMap(Tag[A] -> a, Tag[B] -> b, Tag[C] -> c)
-    def apply[A: Tag, B: Tag, C: Tag, D: Tag](a: A, b: B, c: C, d: D): TypeMap[A & B & C & D] =
-        HashMap(Tag[A] -> a, Tag[B] -> b, Tag[C] -> c, Tag[D] -> d)
+    def apply[A](a: A)(using ta: Tag[A]): TypeMap[A] =
+        HashMap(ta.erased -> a)
+    def apply[A, B](a: A, b: B)(using ta: Tag[A], tb: Tag[B]): TypeMap[A & B] =
+        HashMap(ta.erased -> a, tb.erased -> b)
+    def apply[A: Tag, B: Tag, C: Tag](a: A, b: B, c: C)(using ta: Tag[A], tb: Tag[B], tc: Tag[C]): TypeMap[A & B & C] =
+        HashMap(ta.erased -> a, tb.erased -> b, tc.erased -> c)
+    def apply[A: Tag, B: Tag, C: Tag, D: Tag](a: A, b: B, c: C, d: D)(using
+        ta: Tag[A],
+        tb: Tag[B],
+        tc: Tag[C],
+        td: Tag[D]
+    ): TypeMap[A & B & C & D] =
+        HashMap(
+            ta.erased -> a,
+            tb.erased -> b,
+            tc.erased -> c,
+            td.erased -> d
+        )
 end TypeMap
--- a/kyo-core/shared/src/test/scala/kyoTest/typeMapTest.scala
+++ b/kyo-core/shared/src/test/scala/kyoTest/typeMapTest.scala
@@ -76,7 +76,7 @@ class typeMapTest extends KyoTest:
         }
         "non-empty" in {
             val e: TypeMap[String & Boolean] = TypeMap[Boolean](true).asInstanceOf[TypeMap[String & Boolean]]
-            test[String](e, """HashMap(Tag(-4scala.Boolean;,%scala.AnyVal;"b!M;"V!A;) -> true)""", "java.lang.String")
+            test[String](e, """HashMap(!_9;!W3;!U1;!V2; -> true)""", "java.lang.String")
         }
     }
 
--- a/kyo-tag/shared/src/main/scala/kyo/tags.scala
+++ b/kyo-tag/shared/src/main/scala/kyo/tags.scala
@@ -2,109 +2,128 @@ package kyo
 
 import scala.annotation.switch
 import scala.annotation.tailrec
+import scala.collection.immutable
 import scala.quoted.*
 
-case class Tag[T](tpe: String) extends AnyVal
+opaque type Tag[T] = String
 
 object Tag:
+
+    type Full[T] = Tag[T] | Set[T]
+
+    given [T, U]: CanEqual[Full[T], Full[U]] = CanEqual.derived
 
     import internal.*
 
     inline given apply[T]: Tag[T] = ${ tagImpl[T] }
 
-    extension [T](t1: Tag[T])
+    extension [T](t1: Full[T])
 
         def show: String =
-            val decoded = t1.tpe.drop(2).takeWhile(_ != ';')
-            fromCompact.getOrElse(decoded, decoded)
+            t1 match
+                case t1: String =>
+                    val decoded = t1.drop(2).takeWhile(_ != ';')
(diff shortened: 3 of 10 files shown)


kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -3,14 +3,14 @@ package kyo
 import kyo.Tag.Intersection
 import scala.collection.immutable.HashMap
 
-opaque type TypeMap[+A] = HashMap[Tag[?], Any]
+opaque type TypeMap[+A] = HashMap[Tag[Any], Any]
```
fwbrasil, line 6: `Tag` is now an opaque type, which doesn't support `?`

kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -3,14 +3,14 @@ package kyo
...
 extension [A](self: TypeMap[A])
 
-    private inline def fatal(using t: Tag[?]): Nothing =
+    private inline def fatal[T](using t: Tag[T]): Nothing =
         throw new RuntimeException(s"fatal: kyo.TypeMap of contents [$self] missing value of type: [${t.show}].")
 
     def get[B >: A](using t: Tag[B]): B =
-        val b = self.getOrElse(t, null)
+        val b = self.getOrElse(t.erased, null)
```
fwbrasil, line 13: `erased` transforms a `Tag[T]` into `Tag[Any]`

hearnadam: We can probably make this function accept `Tag[Any]`, right? The `.show` should still be accurate.

kyo-core/shared/src/test/scala/kyoTest/typeMapTest.scala
```diff
@@ -76,7 +76,7 @@ class typeMapTest extends KyoTest:
         }
         "non-empty" in {
             val e: TypeMap[String & Boolean] = TypeMap[Boolean](true).asInstanceOf[TypeMap[String & Boolean]]
-            test[String](e, """HashMap(Tag(-4scala.Boolean;,%scala.AnyVal;"b!M;"V!A;) -> true)""", "java.lang.String")
+            test[String](e, """HashMap(!_9;!W3;!U1;!V2; -> true)""", "java.lang.String")
```
fwbrasil, line 79: No `Tag` wrapper in the string because it is an `opaque type` now. I also added entries in `Tag.internal.toCompact` for `Boolean` and `AnyVal`.

kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -2,110 +2,129 @@ package kyo
...
+                case _ =>
+                    tags.forall(_ =:= t2)
 
-            infix def >:>(t2: Intersection[?]): Boolean =
-                t1.tags.forall(tag1 => t2.tags.exists(tag1 >:> _))
-        end extension
     end Intersection
 
+    object Intersection:
+        private[Tag] def raw[T](tags: Seq[String]) = new Intersection[T](tags.asInstanceOf[Seq[Tag[Any]]])
```
fwbrasil, line 125: I was having issues with the macro generating invalid trees when I changed `Tag` to be `opaque`. This is a workaround for it.

kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -2,110 +2,129 @@ package kyo
...
-            infix def <:<[U](t2: Tag[U]): Boolean =
-                t1.tags.exists(_ <:< t2)
-
-            infix def =:=[U](t2: Tag[U]): Boolean =
-                t1.tags.forall(_ =:= t2)
-
-            infix def =!=[U](t2: Tag[U]): Boolean =
-                !(t1 =:= t2)
+    object Union:
+        private[Tag] def raw[T](tags: Seq[String]) = new Union[T](tags.asInstanceOf[Seq[Tag[Any]]])
```
fwbrasil, line 93: I was having issues with the macro generating invalid trees when I changed Tag to be opaque. This is a workaround for it.

kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -121,17 +140,8 @@ object Tag:
         // |--------------------Test segment---------------------|-|-------------------Param1 segment----------------------|+|-------------------Param2 segment----------------------|
         //                                                     variance                                                  variance
 
-        val toCompact = Map(
```
fwbrasil, line 124: Moved to the end of the file.

kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -214,27 +224,27 @@ object Tag:
...
 
-        def tags[T: Type](using q: Quotes)(flatten: q.reflect.TypeRepr => Seq[q.reflect.TypeRepr]): Expr[Seq[Tag[?]]] =
+        def tags[T: Type](using q: Quotes)(flatten: q.reflect.TypeRepr => Seq[q.reflect.TypeRepr]): Expr[Seq[String]] =
             import quotes.reflect.*
             Expr.ofSeq {
-                flatten(TypeRepr.of[T]).foldLeft(Seq.empty[Expr[Tag[?]]]) {
-                    (acc, tpe) =>
-                        tpe.asType match
-                            case '[t] =>
-                                Expr.summon[Tag[t]] match
```
fwbrasil, line 228: I avoided the unnecessary macro indirection here.

kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -2,110 +2,129 @@ package kyo
...
+            !(t1 =:= t2)
 
-    end extension
-
-    case class Union[T](tags: Seq[Tag[?]]) extends AnyVal
+        infix def >:>[U](t2: Full[U]): Boolean =
+            t2 <:< t1
 
-    object Union:
+        def erased: Tag[Any] = t1.asInstanceOf[Tag[Any]]
```
hearnadam, line 48: Isn't this unsafe in the `Full[T]` case?

hearnadam (commented):
Will review later, but this is much improved. this may make the tag useful to other libraries as well!

kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -214,27 +224,27 @@ object Tag:
...
+                        acc :+ encodeType(repr)
                 }
             }
         end tags
 
+        def setImpl[T: Type](using q: Quotes): Expr[Set[T]] =
+            import quotes.reflect.*
+            TypeRepr.of[T] match
+                case AndType(_, _) => intersectionImpl[T]
+                case _             => unionImpl[T]
```
fwbrasil, line 244: This allows representing a type `T` as a `Union(Tag[T] :: Nil)`.

kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -256,29 +266,37 @@ object Tag:
...
                     case tpe           => Seq(tpe)
 
-            '{ Intersection(${ tags(using q)(flatten) }) }
+            '{ Intersection.raw[T](${ tags(using q)(flatten) }) }
         end intersectionImpl
 
-        // printable ASCII chars
-        val maxEncodableInt = ('~' - ' ').toInt
+        // Encodes ints using printable ASCII chars
+        val charsReserved  = immutable.Set('[', ']', '(', ')', ';', ',')
```
fwbrasil, line 273: I found a bug while working on this change: characters that are used to control the subtype check can end up encoded as the size or hash of segment entries, breaking the logic. This new mechanism ensures the reserved chars aren't used to represent `Int`s.

kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -317,9 +335,7 @@ object Tag:
                                 case None =>
                                     failMissing(TypeRepr.of[tpe])
                                 case Some(value) =>
-                                    '{ $value.tpe }
-                        case _ =>
```
fwbrasil, line 321: This case was unreachable.

kyo-tag/shared/src/test/scala/kyoTest/tagsTest.scala
```diff
@@ -10,7 +10,7 @@ import scala.annotation.nowarn
 
 class tagsTest extends AsyncFreeSpec with NonImplicitAssertions:
 
-    def test[T1, T2](using k1: Tag[T1], i1: ITag[T1], k2: Tag[T2], i2: ITag[T2]): Unit =
+    inline def test[T1, T2](using k1: Tag[T1], i1: ITag[T1], k2: Tag[T2], i2: ITag[T2]): Unit =
```
fwbrasil, line 13: `inline` to make the error messages shoul

kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -3,14 +3,14 @@ package kyo
 import kyo.Tag.Intersection
 import scala.collection.immutable.HashMap
 
-opaque type TypeMap[+A] = HashMap[Tag[?], Any]
+opaque type TypeMap[+A] = HashMap[Tag[Any], Any]
 extension [A](self: TypeMap[A])
 
-    private inline def fatal(using t: Tag[?]): Nothing =
+    private inline def fatal[T](using t: Tag[T]): Nothing =
         throw new RuntimeException(s"fatal: kyo.TypeMap of contents [$self] missing value of type: [${t.show}].")
```
hearnadam, line 10: We can probably make this accept `Tag[Any]`, the `.show` should still be accurate, right?

kyo-tag/shared/src/test/scala/kyoTest/tagsTest.scala
```diff
@@ -27,6 +27,22 @@ class tagsTest extends AsyncFreeSpec with NonImplicitAssertions:
                 s"Tag[T2] <:< Tag[T1] is $kresult but ITag[T2] <:< ITag[T1] is $iresult"
             )
         }
+        "T2 =:= T1" in {
```
fwbrasil, line 30: I had forgotten to add these.

kyo-core/shared/src/main/scala/kyo/typeMap.scala
fwbrasil: Yes, but then the callers would have to invoke `erased` to pass the tags.

hearnadam: Ah, since this should rarely happen, this is probably better.

kyo-tag/shared/src/main/scala/kyo/tags.scala
fwbrasil: Good catch! I've pushed a fix providing separate `erased` extensions for `Tag` and `Full`.

kyo-core/shared/src/main/scala/kyo/typeMap.scala
hearnadam: Right, I agree.

kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -2,110 +2,132 @@ package kyo
...
+    end Set
 
-            infix def >:>[U](t2: Tag[U]): Boolean =
-                t1.tags.exists(t2 <:< _)
+    object Set:
+        inline given apply[T]: Set[T] = ${ setImpl[T] }
 
-            infix def <:<(t2: Union[?]): Boolean =
-                t1.tags.forall(tag1 => t2.tags.exists(tag1 <:< _))
+    case class Union[T](tags: Seq[Tag[Any]]) extends Set[T]:
```
hearnadam, line 68: I believe this can extend AnyVal

hearnadam (approved):

kyo-tag/shared/src/main/scala/kyo/tags.scala
fwbrasil: Nice! I thought it wouldn't be possible but I just needed to make `Set` a [universal trait](https://docs.scala-lang.org/overviews/core/value-classes.html).

PR #438, opened 2024-05-30
Initial Layers implementation

hearnadam:
/claim #203 
/split @kitlangton 

This PR creates a Layer implementation with memoization, combinators, and supports arbitary effects.

It's possible we can condense it into two type parameters, but that may need to be a followup exercise.

The remaining work involves developing a macro to combine layers automagically, but that will be a separate PR.

Written by @kitlangton and I.


kyo-core/shared/src/test/scala/kyoTest/envsTest.scala
```diff
@@ -259,17 +259,17 @@ class envsTest extends KyoTest:
 
         "providing the wrong env map" in {
             assertDoesNotCompile("""
-                val kyo: String < Envs[String] = Envs.get[String]
-                val envMap: TypeMap[Int]        = TypeMap(12)
-                Envs.runTypeMap(envMap)(kyo).pure
+            val kyo: String < Envs[String] = Envs.get[String]
+            val envMap: TypeMap[Int]       = TypeMap(12)
```
hearnadam, line 263: Will followup to ensure we are not leaving unnecessary diff.

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,86 @@
...
+extension [In, Out, S](layer: Layer[In, Out, S])
+    @targetName("runS")
+    def run: TypeMap[Out] < (S & Envs[In]) =
+        layer.doRun()
+
+extension [Out, S](layer: Layer[Any, Out, S])
+    def run: TypeMap[Out] < S =
+        layer.doRun().asInstanceOf[TypeMap[Out] < S]
+
+object Layers:
```
hearnadam, line 59: Should we call this `Layer` or `Layers`?

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,86 @@
...
+
+end Layer
+extension [In, Out, S](layer: Layer[In, Out, S])
+    @targetName("runS")
+    def run: TypeMap[Out] < (S & Envs[In]) =
+        layer.doRun()
+
+extension [Out, S](layer: Layer[Any, Out, S])
+    def run: TypeMap[Out] < S =
+        layer.doRun().asInstanceOf[TypeMap[Out] < S]
```
hearnadam, line 57: We can avoid casting if we switch to an encoding which automatically pulls the remaining types (`In`) from a pending `Envs[In]`

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,84 @@
...
+
+import Layers.internal.*
+import kyo.Flat.unsafe
+import kyo.core.*
+import scala.annotation.targetName
+
+sealed trait Layer[-In, +Out, -S]:
+    self =>
+
+    def >>>[In2, Out2, S2](that: Layer[Out & In2, Out2, S2]): Layer[In & In2, Out2, S & S2]       = To(self, that)
```
fwbrasil, line 11: can we have readable names for these instead?

kyo-core/shared/src/main/scala/kyo/layers.scala
fwbrasil: I think `Layers` for now but we can re-consider the decision to use plurals

kyo-core/shared/src/test/scala/kyoTest/layersTest.scala
```diff
@@ -0,0 +1,85 @@
...
+
+            val env = combinedLayer.run.pure
+            assert(env.get[Guppy].name == "Tiny Guppy")
+            assert(env.get[Shark].belly.name == "Tiny Guppy")
+            assert(env.get[MegaShark].belly.belly.name == "Tiny Guppy")
+        }
+    }
+
+    "memoization" - {
+        "shared >>> a ++ shared >>> b" in run {
```
fwbrasil, line 60: Could you add more tests validating the behavior with more complex dependency graphs?

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,84 @@
...
+                                leftResult  <- lhs.doRun(memoMap)
+                                rightResult <- rhs.doRun(memoMap)
+                            yield leftResult.union(rightResult)
+                        }.asInstanceOf[Expected]
+
+                    case To(lhs, rhs) =>
+                        {
+                            for
+                                leftResult  <- lhs.doRun(memoMap)
+                                rightResult <- Envs.runTypeMap(leftResult)(rhs.doRun(memoMap))(using unsafe.bypass, Envs.bypass)
```
fwbrasil, line 35: Do we need the `Flat` bypass here now that `TypeMap` has an evidence for it?

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,84 @@
...
+                    case FromKyo(kyo) =>
+                        kyo().map { result =>
+                            memoMap += (self -> result)
+                            result
+                        }.asInstanceOf[Expected]
+        end match
+    end doRun
+
+end Layer
+extension [In, Out, S](layer: Layer[In, Out, S])
```
fwbrasil, line 48: Can you move these extensions to a `Layer` companion object?

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,84 @@
...
+
+sealed trait Layer[-In, +Out, -S]:
+    self =>
+
+    def >>>[In2, Out2, S2](that: Layer[Out & In2, Out2, S2]): Layer[In & In2, Out2, S & S2]       = To(self, that)
+    def ++[In2, Out2, S2](that: Layer[In2, Out2, S2]): Layer[In & In2, Out & Out2, S & S2]        = And(self, that)
+    def >+>[In2, Out2, S2](that: Layer[Out & In2, Out2, S2]): Layer[In & In2, Out & Out2, S & S2] = self ++ (self >>> that)
+
+    private[kyo] def doRun(
+        memoMap: scala.collection.mutable.Map[Layer[?, ?, ?], Any] = scala.collection.mutable.Map.empty[Layer[?, ?, ?], Any]
```
fwbrasil, line 16: Kyo provides stronger guarantees regarding purity across transformations of a computation. If an effect in a layer initialization invokes its continuation multiple times, this map can break referential transparency. Could this be a recursive method instead using an immutable `Map`?

kyo-core/shared/src/test/scala/kyoTest/layersTest.scala
```diff
@@ -0,0 +1,85 @@
...
+import kyo.*
+// import kyo.syntax.*
+
+// TODO:
+// - Replace Envs with the implementation of EnvMaps
+// - Allow Envs.run to accept a layer...
+// - kyo.provide(food, heat, stomach, spells)
+// - kyo.provide(food, heat, stomach, spells)
+// - Envs.run(1)
+class layersTest extends KyoTest:
```
fwbrasil, line 12: Do we need to tests for cyclic dependencies? Or is that not possible with the encoding?

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,84 @@
+package kyo
+
+import Layers.internal.*
+import kyo.Flat.unsafe
+import kyo.core.*
+import scala.annotation.targetName
+
+sealed trait Layer[-In, +Out, -S]:
```
fwbrasil, line 8: I was hoping we could have only two type parameters but it doesn't seem possible with memoization. I think it's ok.

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,84 @@
...
+
+    def apply[A, B: Tag, S](kyo: => B < (Envs[A] & S)): Layer[A, B, S] =
+        FromKyo { () =>
+            kyo.map { result => TypeMap(result) }
+        }
+
+    def from[A: Tag, B: Tag, S](f: A => B < S): Layer[A, B, S] =
+        apply { Envs.get[A].map(f) }
+
+    def from[A: Tag, B: Tag, C: Tag, S](f: (A, B) => C < S): Layer[A & B, C, S] =
```
fwbrasil, line 76: Kyo normally has overloaded methods for up to four parameters. Can you add them? I think Scala 3 has some tuple magic that would allow us to accept any number of parameters but I haven't tried it yet.

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,84 @@
...
+        apply {
+            for
+                a <- Envs.get[A]
+                b <- Envs.get[B]
+            yield f(a, b)
```
fwbrasil, line 81: You can also use `zip` here: `zip(Envs.get[A], Envs.get[B]).map(f)`

hearnadam: Nice! Didn't know about `zip`. `.map(f)` is not possible because `zip`  produces a tuple...

kyo-core/shared/src/main/scala/kyo/layers.scala
hearnadam: Done.

kyo-core/shared/src/main/scala/kyo/layers.scala
hearnadam: It should be possible to encode it as:

```
Layer[S & Envs[In], Out]
```
If we can unpack the `Envs[A]`...

kyo-core/shared/src/main/scala/kyo/layers.scala
hearnadam: Sure! Any suggestions?

kyo-core/shared/src/main/scala/kyo/layers.scala
hearnadam: Nope, I can use `summon` instead.

kyo-core/shared/src/main/scala/kyo/layers.scala
hearnadam: Kit and I tried a recursive encoding and weren't able to make it work. Would you prefer immutable map + atomic?

kyo-core/shared/src/main/scala/kyo/layers.scala
fwbrasil: Interesting! That'd be great

kyo-core/shared/src/main/scala/kyo/layers.scala
fwbrasil: Maybe to/and/then?

kyo-core/shared/src/main/scala/kyo/layers.scala
fwbrasil: I'll give it a try later today

kyo-core/shared/src/main/scala/kyo/layers.scala
hearnadam: Do you mean just nesting the extensions? or converting to object methods? ie: `Layer.run(layer)`

kitlangton: I think nesting under the companion object. That should ensure they're always resolvable w/ implicit search.

kyo-core/shared/src/test/scala/kyoTest/layersTest.scala
hearnadam: I don't think that makes much sense. I did add a test for when `In =:= Out`, which we don't want to break memoization.

kyo-core/shared/src/main/scala/kyo/layers.scala
hearnadam: Done.

kyo-core/shared/src/main/scala/kyo/layers.scala
hearnadam: `to/and/using`

kitlangton: Isn’t using a keyword? :P

hearnadam: It compiles 😂

calvinlfer:
Wow this is super nice! 
I'm curious about Envs now, with Layers, do you see Envs as a way to ask for things and make it appear as Layer requirements? or does standalone Envs still have its use?

kyo-core/shared/src/main/scala/kyo/layers.scala
fwbrasil: Sorry for the delay, I took a look but I can't see a way to avoid mutability for memoization. What do you think about adding `IOs` suspensions for the parts of the code that interact with the mutable map? It'd address the referential transparency issue but `Layer.run` would depend on `IOs`. Also, don't we need a concurrent map here since layers might be initialized in parallel?

hearnadam: As it stands today, we don't use any par related operators. That is something we could add in the future, but not sure if the complexity is warranted now.

A dependency on IOs should be alright, it will just remove the ability to use layers as a general memoization technique... IMO it's not too big of a deal as it's pretty dependency injection specific

hearnadam:
@fwbrasil @kitlangton I seem to have cracked the code on an encoding with just 2 type parameters. Let me know what you think!

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,91 @@
...
+        case class To[Out1, Out2, S1, S2](lhs: Layer[?, ?], rhs: Layer[?, ?])          extends Layer[Out1 & Out2, S1 & S2]
+        case class FromKyo[In, Out, S](kyo: () => TypeMap[Out] < (Envs[In] & S))(using val tag: Tag[Out]) extends Layer[Out, S]
+    end internal
+
+    val empty: Layer[Any, Any] = FromKyo { () => TypeMap.empty }
+
+    def apply[A: Tag, S](kyo: => A < S): Layer[A, S] =
+        FromKyo { () =>
+            kyo.map { result => TypeMap(result) }
+        }
```
hearnadam, line 68: I'm very happy with the result here! Without the `Envs[B]` specialization, the layer should be transparent to the underlying Kyo computation.

kyo-core/shared/src/main/scala/kyo/layers.scala
hearnadam: Well maybe not...

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,88 @@
...
+                    case To(lhs, rhs) =>
+                        {
+                            for
+                                leftResult  <- lhs.doRun(memoMap)
+                                rightResult <- Envs.runTypeMap(leftResult)(rhs.doRun(memoMap))(using summon, Envs.bypass)
+                            yield rightResult
+                        }.asInstanceOf[Expected]
+
+                    case FromKyo(kyo) =>
+                        kyo().map { result => IOs(memoMap.compute(self, { (_, _) => result })).map(_ => result) }.asInstanceOf[Expected]
```
fwbrasil, line 37: Minor: it's possible to avoid the `map` transformation:

```scala
IOs {
    memoMap.compute(self, { (_, _) => result })
    result
}
```

By the way, couldn't we use `putIfAbsent`?

fwbrasil (approved):

hearnadam:
@fwbrasil the layers API is complete. There is minor polish further required for the error messaging, but the API should be stable - so we can merge this and followup if you are okay with that!

kyo-core/shared/src/main/scala/kyo/internal/LayerMacros.scala
```diff
@@ -0,0 +1,238 @@
...
+        }
+end Validated
+
+enum GraphError[Key, Value]:
+    case MissingInput(input: Key, parent: Option[Node[Key, Value]])
+    case CircularDependency(node: Node[Key, Value]) // TODO: Maybe also trace the path.
+    case AmbiguousInputs(target: Key, found: Set[Node[Key, Value]], parent: Option[Node[Key, Value]])
+end GraphError
+
+final case class Graph[Key, Value](nodes: Set[Node[Key, Value]])(equals: (Key, Key) => Boolean):
```
hearnadam, line 198: @kitlangton should we make the whole graph/validated private? Is there any reason we can't?

kyo-core/shared/src/main/scala/kyo/package.scala
```diff
@@ -7,6 +7,13 @@ package object kyo:
 
     object Kyo
 
+    extension [T, S, R](inline kyo: T < (S & Envs[R]))
+        inline def provide(inline layers: Layer[?, ?]*): T < (S & IOs) = {
```
fwbrasil, line 11: Can we move the extension to `kyo-combinators` and provide a `Envs.runLayer`? Maybe we could make `Envs.runTypeMap` not public as well?

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,96 @@
...
+    end doRun
+
+end Layer
+
+object Layer:
+    import Envs.HasEnvs
+
+    extension [In, Out, S](layer: Layer[Out, Envs[In] & S])
+        def run[In1, R](using HasEnvs[In1, In] { type Remainder = R }): TypeMap[Out] < (S & R & IOs) =
+            layer.doRun().asInstanceOf
```
fwbrasil, line 55: do we need this method to be part of the public API? It feels like `Layers` is the proper API and `TypeMap` is more of a implementation detail. Also, please add the type annotation to `asInstanceOf`. It's possible the compiler ends up inferring `Nothing` in case we change this code.

kyo-core/shared/src/test/scala/kyoTest/layersTest.scala
```diff
@@ -0,0 +1,257 @@
...
+            assertDoesNotCompile("""Layers.make[Boolean](a, b)""")
+        }
+        "missing output" in {
+            val a = Layers(42)
+            val b = Layers.from((i: Int) => i.toDouble)
+            val c = Layers.from((d: Double) => d.toLong)
+
+            assertDoesNotCompile("""Layers.make[String](a, b, c)""")
+        }
+        "complex layer" in run {
```
fwbrasil, line 186: pretty cool!!

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,96 @@
+package kyo
+
+import Layers.internal.*
+import java.util.concurrent.ConcurrentHashMap as JConcurrentHashMap
+import kyo.core.*
+import scala.annotation.targetName
+
+sealed trait Layer[+Out, -S]:
```
fwbrasil, line 8: Kyo's codebase favors `abstract class` over `trait` where possible to avoid interface dispatches, which have some additional performance overhead. Can you check if this could be `abstract class`? Also, please mark the methods as `final` so the JIT can more easily optimize invocations.

kyo-core/shared/src/main/scala/kyo/layers.scala
```diff
@@ -0,0 +1,96 @@
...
+            zip(Envs.get[A], Envs.get[B], Envs.get[C])
+                .map { case (a, b, c) => f(a, b, c) }
+        }
+
+    def from[A: Tag, B: Tag, C: Tag, D: Tag, E: Tag, S](f: (A, B, C, D) => E < S): Layer[E, Envs[A & B & C & D] & S] =
+        apply {
+            zip(Envs.get[A], Envs.get[B], Envs.get[C], Envs.get[D]).map { case (a, b, c, d) => f(a, b, c, d) }
+        }
+
+    transparent inline def make[Target](inline layers: Layer[?, ?]*): Layer[Target, ?] =
```
fwbrasil, line 93: Kyo's naming typically use `init` for this kind of method. We can review to start using `make` but can you rename to `init` here for consistency for now?

kyo-core/shared/src/main/scala/kyo/internal/LayerMacros.scala
```diff
@@ -0,0 +1,238 @@
...
+                                '{ $left.to($right) }
+                    },
+                    _.value, {
+                        // TODO: MAke nIcEr PlZEaz
+                        val debugFold =
+                            targetLayer.fold[String]("(" + _ + " and " + _ + ")", "(" + _ + " to " + _ + ")", _.value.show, "Empty")
+
+                        report.errorAndAbort(
+                            s"""
+                    Empty layer found as input to Layer with non-zero requirements. Did you fully resolve dependencies?
```
fwbrasil, line 76: Indentation is off here.

kyo-core/shared/src/main/scala/kyo/internal/LayerMacros.scala
```diff
@@ -0,0 +1,238 @@
...
+
+object LayerLike:
+    def apply[A](value: A): LayerLike[A]            = Value(value)
+    given [A]: CanEqual[LayerLike[A], LayerLike[A]] = CanEqual.derived
+
+final case class Node[Key, Value](inputs: Set[Key], outputs: Set[Key], value: Value)
+object Node:
+    given [Key, Value]: CanEqual[Node[Key, Value], Node[Key, Value]] = CanEqual.derived
+
+enum Validated[+E, +A]:
```
fwbrasil, line 151: I should have a proper public impl for this but I think it's ok to keep it internal for now!

kyo-core/shared/src/test/scala/kyoTest/layersTest.scala
```diff
@@ -0,0 +1,257 @@
...
+            case class B(a: A)
+            case class C(b: B)
+
+            val a = Layers(IOs(new A))
+            val b = Layers.from((a: A) => Aborts.when(false)("").andThen(B(a)))
+            val c = Layers.from((b: B) => Vars.get[Unit].andThen(C(b)))
+
+            assertCompiles("""Layers.make[C](a, b, c)""")
+        }
+        "pruneable inputs" in pendingUntilFixed {
```
fwbrasil, line 222: Why does this fail? Is it because `TypeMap` is covariant?

kyo-core/shared/src/test/scala/kyoTest/layersTest.scala
```diff
@@ -0,0 +1,257 @@
...
+            combinedLayer.run.map { env =>
+                assert(sourceInitializationCount == 1)
+                assert(env.get[Wizard].name == "Gandalf")
+                assert(env.get[Spell].powerLevel == 100)
+                assert(env.get[Spell].source.name == "Fireball")
+                assert(env.get[Wizard].source.name == "Fireball")
+            }
+        }
+    }
+    "effects!" - {
```
fwbrasil, line 115: It's great how the API plays well with other effects 👏

kyo-core/shared/src/main/scala/kyo/internal/LayerMacros.scala
```diff
@@ -0,0 +1,238 @@
...
+                    s"Missing input ${input.show} for ${parent.value.show}".red
+
+                case GraphError.MissingInput(input, None) =>
+                    s"Missing layer for target ${input.show}".red
+
+                case GraphError.AmbiguousInputs(target, found, parent) =>
+                    s"Ambigious input ${target.show} found in ${found.map(_.value.show).mkString("{", ", ", "}")}".red
+
+                case GraphError.CircularDependency(node) =>
+                    s"MAKE CIRCULAR DEPENDENYC ERROR PRETTIESY $node".red
```
fwbrasil, line 36: Can we have a better placeholder message for now? :)

kyo-core/shared/src/main/scala/kyo/internal/LayerMacros.scala
fwbrasil: Agree! We could move impls to a companion object to avoid polluting the `kyo.internal` namespace as well.

kyo-core/shared/src/test/scala/kyoTest/layersTest.scala
hearnadam: The goal is that we should fail compilation when too many layers are provided. Currently, we happily only use some of the layers to construct a new layer.

kyo-core/shared/src/main/scala/kyo/layers.scala
hearnadam: Done!

kyo-core/shared/src/main/scala/kyo/layers.scala
hearnadam: I went ahead and marked it private as I don't have a good example of using `.run` on a layer, but it may need to be public eventually.

hearnadam:
@fwbrasil all feedback should be addressed. The implementation is sound, but I think we can revisit error reporting a bit.

Regarding naming, I followed your comments. I plan to revisit naming for all of Kyo once the new core is merged.

fwbrasil:
woohoo! The next release will be amazing 🎉

PR #447, opened 2024-05-30
improve toString for Kyo computations + bug error messages

fwbrasil:
The `toString` of Kyo computations currently print `<function1>`. This PR introduces a custom `toString` showing the details of the computation, including the new `Trace` information, to help aid in debugging and error reporting. I've also improved the bug error messages to use the new `toString`.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/internal/Trace.scala
+++ b/kyo-core/shared/src/main/scala/kyo/internal/Trace.scala
@@ -10,10 +10,10 @@ object Trace:
     private val maxSnippetLines = 3
 
     extension (t: Trace)
-        def show: String       = t
-        def position: Position = Position(t.takeWhile(_ != '\n').drop(1))
-        def method: String     = t.drop(position.show.length + 2).takeWhile(isIdentifierPart)
-        def snippet: String    = t.drop(position.show.length + 2)
+        def show: String       = s"Trace(${position},method=${method},snippet=${snippet})"
+        def position: Position = Position(t.takeWhile(_ != '\n'))
+        def method: String     = t.drop(position.show.length + 1).takeWhile(isIdentifierPart)
+        def snippet: String    = t.drop(position.show.length + 1)
     end extension
 
     implicit inline def derive: Trace =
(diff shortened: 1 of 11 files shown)


kyo-core/shared/src/main/scala/kyo/internal/Trace.scala
```diff
@@ -10,10 +10,10 @@ object Trace:
     private val maxSnippetLines = 3
 
     extension (t: Trace)
-        def show: String       = t
-        def position: Position = Position(t.takeWhile(_ != '\n').drop(1))
```
fwbrasil, line 14: there was a bug here removing the first char of the position

kyo-core/shared/src/main/scala/kyo/package.scala
```diff
@@ -68,30 +68,24 @@ package object kyo:
...
-                failTag(actual, expected)
-
-        def when(cond: Boolean)(msg: String): Unit =
-            if cond then bug(msg)
+            if kyo.tag =!= expected then
+                failTag(kyo, expected)
 
         def apply(msg: String): Nothing =
-            throw KyoBugException(msg + " Please open an issue 🥹 https://github.com/getkyo/kyo/issues")
+            throw KyoBugException(msg + " Please open an issue 🥹  https://github.com/getkyo/kyo/issues")
```
fwbrasil, line 89: added a space here because, at least in iTerm2, it renders without a space between the emoji and the url. I imagine due to emoji modifiers.

kyo-core/shared/src/main/scala/kyo/internal/Trace.scala
```diff
@@ -10,10 +10,10 @@ object Trace:
...
-        def method: String     = t.drop(position.show.length + 2).takeWhile(isIdentifierPart)
-        def snippet: String    = t.drop(position.show.length + 2)
+        def show: String       = s"Trace(${position},method=${method},snippet=${snippet})"
```
hearnadam, line 13: Do we need the `Trace` prefix?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/internal/Trace.scala
fwbrasil: I put it to avoid having to label the values in `Suspend.toString`.

PR #449, opened 2024-05-30
Have the singular form of Duration extension methods

carlosedp:
This would look nicer with cases like `1.second` or `1.milli` instead of the plural form.


fwbrasil:
Thanks! What do you think about using a singleton type to restrict the methods only to literal `1` values:

```
extension (value: 1)
    inline def nano: Duration   = value.asNanos
    ....
```

carlosedp:
Neat, didn't knew it would work... it's even cleaner. It would be neat if the plural form was not allowed on values == 1 ... but I don't think this is possible.

fwbrasil:
It seems possible:

```
extension (value: 1)
    inline def millis: Duration = compiletime.error("use milli")
```

Isn't Scala amazing? :)

carlosedp:
Oh!
image

Holy shit, Scala is gorgeous....

carlosedp:
I know it's too picky... but do you want me add both to both extensions?

fwbrasil:
I think it makes sense if you want to make the change. If you could improve the error message as well.

carlosedp:
Done.. let me know if that's ok or if I should cleanup and remove the extra methods...
Thanks for the tip!

fwbrasil:
LGTM! Could you add tests for it? I know scalatest has `assertDoesNotCompile` to check that code doesn't compile but I'm not sure about zio-test cc/ @hearnadam

hearnadam:
zio-test has `typecheck` which is similar.

README.md
```diff
@@ -18,7 +18,7 @@ Drawing inspiration from [ZIO](https://zio.dev/)'s [effect rotation](https://deg
 Kyo is available on Maven Central in multiple modules:
 
 | Module         | Scala 3 | Scala JS | Description                      |
-|----------------|---------|----------|----------------------------------|
```
fwbrasil, line 21: thanks :)

README.md
```diff
@@ -29,7 +29,7 @@ Kyo is available on Maven Central in multiple modules:
 
 For Scala 3:
 
-```scala
```
fwbrasil, line 32: I think we need these spaces here. There's this hack to allow code examples that aren't compiled: https://github.com/getkyo/kyo/blob/d41f50e8a41f524cf7aaf1cbbc3aeca54300772e/build.sbt#L388. We should address this hack, though.

carlosedp: Ah, got it... sorry. Will revert.

kyo-core/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -51,6 +51,39 @@ extension (value: Long)
...
+    inline def second: Duration  = value.as(Seconds)
+    inline def minute: Duration  = value.as(Minutes)
+    inline def hour: Duration    = value.as(Hours)
+    inline def day: Duration     = value.as(Days)
+    inline def week: Duration    = value.as(Weeks)
+    inline def month: Duration   = value.as(Months)
+    inline def year: Duration    = value.as(Years)
+    inline def nanos: Duration   = compiletime.error("please use `.nano`")
+    inline def micros: Duration  = compiletime.error("please use `.micro`")
+    inline def millis: Duration  = compiletime.error("please use `.milli`")
```
hearnadam, line 79: I would prefer just not to populate these, I think the plural version should work for all values

carlosedp: Ok! Agreed it's too much.

hearnadam (approved):

fwbrasil:
Thank you @carlosedp! Congrats on you first contribution 👏🏼

Issue #342, opened 2024-05-09
Speed up build


fwbrasil:
I've noticed Kyo's codebase is taking longer to compile for some time now. It seems the linting compiler flags added some time ago are the main culprit. Removing these flags produced a significant speedup in an initial test:

```
"-Wvalue-discard",
"-Wunused:all",
"-language:strictEquality"
```

fwbrasil:
I've tried adding the new option `ThisBuild / usePipelining := true` that was [recommended on reddit](https://www.reddit.com/r/scala/comments/1d0ga7k/scala_3_usepipelining_build_flag_is_a_game_changer/) but it seems to infer a wrong compilation order and fails:

```
[error] -- [E008] Not Found Error: /Users/fwbrasil/workspace/kyo/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/package.scala:6:11
[error] 6 |import kyo.stats.internal.StatsRegistry
[error]   |       ^^^^^^^^^
[error]   |       value stats is not a member of kyo
```

Removing the flag makes the build work again.

regiskuckaertz:
Maybe try https://scans.gradle.com/?_gl=1*ut6ed*_ga*MjA0NjI5MzM3NC4xNzE2NzEwNzM4*_ga_7W7NC6YNPT*MTcxNjcxMDczNy4zLjAuMTcxNjcxMDczNy42MC4wLjA.#sbt

kyri-petrou:
I know this doesn't solve the issue, but the build should be considerably faster starting with Scala 3.5.x since they fixed unused imports linting performance. Note that 3.5.0-RC1 is already out already in case you want to test it

PR #450, opened 2024-05-30
improve build times

fwbrasil:
- Split root project into `kyoJVM` and `kyoJS`. The JVM project is selected by default, this way the regular development cycle is faster but there's the downside that we might not catch issues in JS before CI. It seems a reasonable tradeoff since it's possible to run the tests manually by selecting the `kyoJS` project.
- Avoid the dependency on `kyo-core` tests in other modules. All other modules end up stuck waiting for the core tests to compile because of the dependency. I had to duplicate some of the code in `KyoTest` into the other modules to achieve that.
- Move `scalafmt` from the build to a commit hook.
- Remove `-Wunused:all` compiler option. It significantly increases the compilation times for tests. We can attempt to add it back once we're on Scala 3.5, which seems to [improve build times](https://github.com/getkyo/kyo/issues/342#issuecomment-2135255841) with the option enabled.

Results with a cold JVM:

**Main**
```
$ sbt ";clean;compile"
[success] Total time: 24 s, completed May 30, 2024, 1:58:13 PM
$ sbt ";clean;test:compile"
[success] Total time: 58 s, completed May 30, 2024, 1:59:58 PM
$ sbt ";clean;test"
[success] Total time: 88 s (01:28), completed May 30, 2024, 2:01:45 PM
```

**This PR (default `kyoJVM` only)**
```
$ sbt ";clean;compile"
[success] Total time: 14 s, completed May 30, 2024, 1:54:29 PM
$ sbt ";clean;test:compile"
[success] Total time: 27 s, completed May 30, 2024, 1:55:51 PM
$ sbt ";clean;test"
[success] Total time: 39 s, completed May 30, 2024, 1:56:59 PM
```

I won't close https://github.com/getkyo/kyo/issues/342 since there's always opportunity to optimize the build.


kyo-caliban/src/test/scala/kyo/resolversTest.scala
```diff
@@ -21,9 +21,9 @@ class resolversTest extends KyoTest:
             zio.Runtime.default.unsafe.run(v).getOrThrow()
         )
 
-    def testServer() =
+    def testServer(port: Int) =
```
fwbrasil, line 24: I'm not sure these tests are failing in CI because the port isn't available. I imagine it could be the short shutdown timeout. I've different ports for each test as a workaround.

project/plugins.sbt
```diff
@@ -9,4 +9,6 @@ addSbtPlugin("org.scalameta" % "sbt-mdoc" % "2.5.2")
 
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "2.0.12")
 
+// addSbtPlugin("com.gradle" % "sbt-develocity" % "1.0.1")
```
fwbrasil, line 12: I used develocity to debug the build and find optimization opportunities. We should eventually enable it in CI.

hearnadam (approved):
LGTM. Perhaps we can migrate to kyo-test for non kyo-core modules...

fwbrasil:
> LGTM. Perhaps we can migrate to kyo-test for non kyo-core modules...

Good point! Something I noticed is that the tests using zio-test seem to take significantly longer than the ones in scalatest, though.

Issue #451, opened 2024-05-30
Incorporate zikyo into the main repo


fwbrasil:
I've been noticing a high interest in zikyo's ZIO-like fluent API. Given how important the API will likely become, it'd be ideal if it's under the `kyo` package and built with the other modules so we don't have issues with syncing versions over time. I spoke to @jiveshungerford in private and he's open to moving zikyo to `getkyo/kyo`.

John raised a couple aspects for us to consider with the move:

1. Is it a separate module? If so, what do we call it? keep it zikyo? kyo-combinators?
2. Can we have it in the same package? So import kyo.* would import the KYO object and all extensions? Or would we have a new subpackage kyo.zikyo or kyo.combinators?

I think it should be a separate module. For the name, `kyo-combinators` seems to work well. Another option is `kyo-syntax` but I think it's confusing. For example, `kyo-direct` is also about syntax.

I'd say keeping the APIs under `kyo` would be ideal. If the user has the module in the classpath, the extensions would be made automatically available.

cc/ @hearnadam

johnhungerford:
@fwbrasil let me know if there's anything we need to add for it to publish properly to Maven. I literally just copied over the zio-core project definition and changed the names. Seems to work ok.

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/452

PR #452, opened 2024-05-31
Added zikyo as kyo-combinators module

johnhungerford:
Address #451 

* Added kyo-combinators module
* Copied sbt config from kyo-core
* Copied zikyo code into kyo-combinators
* Replaced ZiKyoTest with KyoTest
* Minor fixes for scala 3.4.2 and changes to kyo API (CanEqual has problems!!)


johnhungerford:
@fwbrasil let me know if there's anything we need to add for it to publish properly to Maven. I literally just copied over the zio-core project definition and changed the names. Seems to work ok.

kyo-combinators/shared/src/main/scala/kyo/KYO.scala
```diff
@@ -0,0 +1,209 @@
...
+    )(
+        effect: A < (SA & Envs[E])
+    )(
+        using
+        he: HasEnvs[D, E] { type Remainder = SR },
+        t: Tag[D],
+        fl: Flat[A]
+    ): A < (SA & SD & SR) =
+        dependency.map(d => Envs.run[D, A, SA, E, SR](d)(effect))
+        // V >: Nothing: Tag, T: Flat, S, VS, VR
```
hearnadam, line 163: ?

kyo-combinators/shared/src/main/scala/kyo/KYO.scala
```diff
@@ -0,0 +1,209 @@
...
+import kyo.Envs.HasEnvs
+import scala.annotation.implicitNotFound
+import scala.annotation.tailrec
+import scala.annotation.targetName
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+import scala.util.NotGiven
+
+
+object KYO:
```
hearnadam, line 12: Would prefer `Kyo`.

kyo-combinators/shared/src/main/scala/kyo/KYO.scala
johnhungerford: That was just to help me work out what type parameters to use. Didn't mean to leave it there.

kyo-combinators/shared/src/main/scala/kyo/KYO.scala
johnhungerford: I had made it KYO after ZIO, but I'm not sure we need to keep it that way. Is there a `Kyo` type anywhere this would clash with?

hearnadam:
I think so! I wonder if the tool generating the website can copy from multiple files.

johnhungerford:
@fwbrasil do you have a preferred way to add the documentation for this?

kyo-combinators/shared/src/main/scala/kyo/KYO.scala
fwbrasil: There isn't a user-facing Kyo type yet but I'd prefer to keep it reserved for kyo-core. In a sense, the pending type is an expression of the Kyo monad :) 

We don't have an IO type yet. Perhaps we could define one in kyo-core and then provide extensions for both the class and companion object in kyo-combinators?

Something that would be interesting to try is defining IO as a case class. Considering the target user base of these extensions, allowing arbitrary nesting of computations better matches the mental model of how this kind of API is traditionally used. There should be a performance cost but it might not be significant since the core might still be inlinable by the JIT compiler and this kind of temporary wrapper is easy to virtualize.

I'm thinking something like:

case class IO[T, S](kyo: T < S)

object IO:
    // not sure if we need methods here, maybe only extensions in kyo-combinators?

// keep IOs as the effect (I know we need to talk about plurals 😅)
object IOs:
    def get[T, S](io: IO[T, S]) = ...
    // current IOs methods

fwbrasil: Actually, I think `IO` wouldn't make sense unless we want to have the `IOs` effect by default. Maybe it could be `Kyo` but in `kyo-core` with extensions in `kyo-combinators`. We could still provide a `map` for it but taking a function `T => Kyo[U, S2] | U < S2`.

johnhungerford: Agreed that `IO` isn't the ideal object for this method.

I just added a `Kyo` object to the `kyo` package in `kyo-core` and turned the `KYO` object definition into extension methods on it.

kyo-combinators/shared/src/main/scala/kyo/combinators.scala
```diff
@@ -0,0 +1,429 @@
...
+
+    @targetName("zipLeft")
+    def <*[A1, S1](next: => A1 < S1): A < (S & S1) =
+        effect.map(e => next.as(e))
+
+    @targetName("zip")
+    def <*>[A1, S1](next: => A1 < S1): (A, A1) < (S & S1) =
+        effect.map(e => next.map(n => (e, n)))
+
+    def as[A1, S1](value: => A1 < S1): A1 < (S & S1) =
```
hearnadam, line 26: Can you inline the trivial methods? I think this will cause the trace to be captured in this file instead of in the caller's file.

kyo-combinators/shared/src/main/scala/kyo/combinators.scala
```diff
@@ -0,0 +1,429 @@
...
+
+    def optionsToUnit(using Flat[A]): A < (S & Aborts[Unit]) =
+        effect.optionsToAborts(())
+
+    def optionsToChoices(using Flat[A]): A < (S & Choices) =
+        Options.run(effect).map(aOpt => Choices.get(aOpt.toSeq))
+end extension
+
+extension [A, S, E](effect: A < (S & Envs[E]))
+    def provide[E1, S1, SR](dependency: E1 < S1)(
```
hearnadam, line 305: Unless you are already using these methods, we will need to implement a stronger version of Provides for the layer implementation. We can later just use this extension as syntax over the `Layer.provide`

kyo-combinators/shared/src/main/scala/kyo/constructors.scala
```diff
@@ -0,0 +1,208 @@
...
+        promise.map(p => fromFuture(p.future))
+
+    def fromOption[A, S](option: => Option[A] < S): A < (S & Options) =
+        Options.get(option)
+
+    def fromSeq[A, S](sequence: => Seq[A] < S): A < (S & Choices) =
+        sequence.flatMap(seq => Choices.get(seq))
+
+    def fromTry[A, S](_try: => scala.util.Try[A] < S): A < (S & Aborts[Throwable]) =
+        _try.map(t => Aborts.get(t.toEither))
```
hearnadam, line 98: https://github.com/getkyo/kyo/pull/461

kyo-combinators/shared/src/main/scala/kyo/combinators.scala
johnhungerford: Good idea

kyo-combinators/shared/src/main/scala/kyo/combinators.scala
johnhungerford: Yeah provides should definitely be for layers. That's not finished though, yet, is it?

kyo-combinators/shared/src/main/scala/kyo/constructors.scala
johnhungerford: Nice, since 461 isn't merged, for now I just improved the implementation to avoid the additional allocation

README.md
```diff
@@ -1367,13 +1367,13 @@ import kyo._
 
 // A computation that sleeps for 1s
 val a: Unit < Fibers =
-  Fibers.sleep(1.seconds)
+  Fibers.sleep(1.second)
```
fwbrasil, line 1370: bad merge?

kyo-combinators/shared/src/main/scala/kyo/KYO.scala
fwbrasil: thanks! I'm exploring [another way](https://github.com/getkyo/kyo/issues/462) to support nesting for any computation so I think we can go without a `case class Kyo` for now.

fwbrasil:
> @fwbrasil do you have a preferred way to add the documentation for this?

I'm looking into migrating to https://readthedocs.io/ and splitting the documentation into multiple files but can you add it to the readme for now?

fwbrasil:
LGTM besides the merge issue

fwbrasil (approved):

johnhungerford:
I fixed the weird readme issues, and I've moved documentation over

fwbrasil:
Thank you so much @jiveshungerford! 🙏

README.md
```diff
@@ -2373,6 +2376,96 @@ val a: Int < Options =
...
+    saluee      <- Consoles.readln                // Uses Consoles effect
+    _           <- Kyo.sleep(2.seconds)           // Uses Fibers (semantic blocking)
+    _           <- nameService.sayHelloTo(saluee) // Adds Aborts[Throwable] effect
+} yield ()
+
+// There are no combinators for handling IOs or blocking Fibers, since this should
+// be done at the edge of the program
+IOs.run {                                                 // Handles IOs
+    Fibers.runAndBlock(Duration.Inf) {                    // Handles Fibers
+		Kyo.scoped {                                      // Handles Resources
```
nightscape, line 2432: There seems to be a mix of tabs and spaces here...

PR #453, opened 2024-05-31
rename IOs.attempt to IOs.toTry

fwbrasil:
I'm noticing people used to ZIO sometimes try to use `IOs.attempt` as they'd use `ZIO.attempt` but the semantics are different.


hearnadam (approved):
I think this may make more sense as an extension method... 

`IOs.run(IOs(42).map(_.toTry))`?

edit: I guess it would just be `IOs(...).toTry` - `.map` would give the result

Issue #390, opened 2024-05-20
gRPC Support


ghostdogpr:
Support for protobuf and gRPC would drive adoption (at least for me).

It's definitely not easy as it involves code generation but maybe https://scalapb.github.io/docs/writing-plugins could be used like it is for zio-grpc and fs2-grpc.

There is currently the option of using zio-grpc with the zio interop but its performance is not great to start with (https://github.com/LesnyRumcajs/grpc_bench/discussions/441) so there might be an opportunity to shine.

He-Pin:
Yes, looking forward to it, grpc_bench is a great place to show .

steinybot:
This is still a WIP. I made some good progress on the code gen side of things. I haven't had much time in the last 2 weeks. I hope to get back to this in a week or so.

steinybot:
Not sure what I changed but after implementing it as a library it is now significantly slower:
```
-----------------------------------------------------------------------------------------------------------------------------------------
| name                        |   req/s |   avg. latency |        90 % in |        95 % in |        99 % in | avg. cpu |   avg. memory |
-----------------------------------------------------------------------------------------------------------------------------------------
| scala_zio                   |   31949 |       29.93 ms |       66.57 ms |       76.40 ms |      132.59 ms |   100.5% |    315.89 MiB |
| scala_kyo                   |    7056 |      140.20 ms |      214.61 ms |      502.63 ms |         1.39 s |  101.77% |    333.34 MiB |
-----------------------------------------------------------------------------------------------------------------------------------------
```

DamianReeves:
Any updates on this front?

steinybot:
I haven't managed to dig into the performance issues. I created some benchmarks but that is as far as I got. I'm in the middle of merging the new core changes in.

Unfortunately I'm unexpectedly out of a job so I actually have less free time while I apply for jobs and do interviews etc. Happy to share what I have done and split the bounty if you want to help out.

steinybot:
I'm back onto this now. I hope to have a draft PR up in a day or two.

fwbrasil:
/bounty 500

Updating since this is much more work than I expected

debaa98:
hey @fwbrasil is it like 700 or 500 ?

fwbrasil:
[...]

PR #458, opened 2024-06-02
Use Typelevel scalac options

steinybot:
I was working on https://github.com/getkyo/kyo/issues/390 which adds a few different projects for the code gen and configuring the different options for the different scala versions was a pain. This makes it a fair bit easier and removes some of the existing workarounds.

I didn't go as far as adding https://github.com/typelevel/sbt-tpolecat although that would be my preference.

I "fixed" (ignored) a few warnings which possibly started showing up as I didn't remove as many warnings for the projects with Scala 2.12.

There are still 2 left:
```
[warn] -- Warning: /Users/jason/src/kyo2/kyo-stats-registry/js/src/main/scala/kyo/stats/internal/UnsafeHistogram.scala:3:22
[warn] 3 |class UnsafeHistogram(numberOfSignificantValueDigits: Int, highestToLowestValueRatio: Long) {
[warn]   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[warn]   |                      unused explicit parameter
[warn] -- Warning: /Users/jason/src/kyo2/kyo-stats-registry/js/src/main/scala/kyo/stats/internal/UnsafeHistogram.scala:3:59
[warn] 3 |class UnsafeHistogram(numberOfSignificantValueDigits: Int, highestToLowestValueRatio: Long) {
[warn]   |                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^
[warn]   |                                                 unused explicit parameter
```


kyo-scheduler/jvm/src/main/scala/kyo/scheduler/InternalClock.scala
```diff
@@ -32,6 +32,8 @@ final private[kyo] case class InternalClock(executor: Executor) {
     def stop(): Unit =
         _stop = true
 
-    private val gauge =
-        statsScope.scope("clock").gauge("skew")((System.currentTimeMillis() - millis).toDouble)
+    locally {
+        val _ =
+            statsScope.scope("clock").gauge("skew")((System.currentTimeMillis() - millis).toDouble)
```
steinybot, line 37: I couldn't use the `@unused` annotation because it isn't available in Scala 2.12. You are supposed to be able to annotate with `: Unit` to avoid discarded non-unit value warnings but I have very little success using that approach. Perhaps `UnsafeGuage` should have a `void` method or there should be another `gauge` method that returns `Unit`?

fwbrasil: Metrics are held via weak references in `kyo-stats-registry` so we need the class field holding the `gauge` reference to avoid it getting garbage collected. Maybe we could move [`kyo.discard`](https://github.com/getkyo/kyo/blob/6e152e77f780fe5e9af5c03a1716b878d9229666/kyo-core/shared/src/main/scala/kyo/package.scala#L62) to `kyo-scheduler`?

steinybot: I moved `discard` to `kyo-scheduler` but it is using Scala 3 top level function definitions and so isn't going to work. I'll have to come up with something else.

steinybot: Ok scratch that. You already removed the `-Wunused:all` option so this is now a non-issue.

PR #463, opened 2024-06-03
provide trace origin for fiber interruptions

### LINKED ISSUE #448: Throw a specific exception in case of timeout
Maybe it would be nice to have `Fibers.timeout` throw a specific exception like `scala.concurrent.TimeoutException` or `java.lang.Throwable` instead of `kyo.fibers$package$Fibers$Interrupted$: Fiber Interrupted`.

fwbrasil:
Fixes #448

Kyo currently doesn't show any information regarding the origin of fiber interrupts. For example, when a timeout happens we can only see that the fiber got interrupted. This PR addresses this limitation by storing the origin `Trace` in `Fibers.Interrupted` and adding it to the error message.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/fibers.scala
+++ b/kyo-core/shared/src/main/scala/kyo/fibers.scala
@@ -124,13 +124,11 @@ object Fibers extends Joins[Fibers] with fibersPlatformSpecific:
 
     type Effects = FiberGets & IOs
 
-    case object Interrupted
-        extends RuntimeException("Fiber Interrupted")
+    case class Interrupted(origin: Trace)
+        extends RuntimeException(s"Fiber Interrupted at ${origin}")
         with NoStackTrace:
         override def getCause() = null
     end Interrupted
-
-    private[kyo] val interrupted = IOs.fail(Interrupted)
 
     def run[T: Flat](v: T < Fibers)(using Trace): Fiber[T] < IOs =
         FiberGets.run(v)
--- a/kyo-core/shared/src/test/scala/kyoTest/fibersTest.scala
+++ b/kyo-core/shared/src/test/scala/kyoTest/fibersTest.scala
@@ -183,7 +183,7 @@ class fibersTest extends KyoTest:
         }
 
         "failure" in run {
-            IOs.attempt(Fibers.timeout(1.millis)(Fibers.sleep(100.millis))).map {
+            IOs.toTry(Fibers.timeout(1.millis)(Fibers.sleep(100.millis))).map {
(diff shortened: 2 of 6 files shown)


kyo-core/shared/src/test/scala/kyoTest/fibersTest.scala
```diff
@@ -150,26 +150,44 @@ class fibersTest extends KyoTest:
...
+        "success" in run {
+            for
+                result <- Fibers.timeout(100.millis)(Fibers.init(42).map(_.get))
+            yield assert(result == 42)
+        }
+
+        "failure" in run {
+            IOs.attempt(Fibers.timeout(1.millis)(Fibers.sleep(100.millis))).map {
+                case Failure(Fibers.Interrupted(trace)) =>
+                    assert(trace.snippet == "timeout(1.millis)(Fibers.sleep(100.millis))")
```
fwbrasil, line 188: Checking only the snippet so the test doesn't start failing in case the line number in the `Trace` changes due to the test scenario changing position in the file.

kyo-core/shared/src/main/scala/kyo/fibers.scala
```diff
@@ -124,14 +124,12 @@ object Fibers extends Joins[Fibers] with fibersPlatformSpecific:
 
     type Effects = FiberGets & IOs
 
-    case object Interrupted
-        extends RuntimeException("Fiber Interrupted")
+    case class Interrupted(origin: Trace)
+        extends RuntimeException(s"Fiber Interrupted at ${origin}")
         with NoStackTrace:
         override def getCause() = null
```
hearnadam, line 130: For later: Can we make a `KyoNoStackTrace(message: String)` abstract class? that way you won't need to extend multiple items or override a member.

hearnadam (approved):
Nice!

fwbrasil:
There's an issue with the interruption integration with sttp. It can't inspect if the completion is an interruption anymore with this change: https://github.com/getkyo/kyo/blob/e8c68604dfcdc03fe0c82ad6f4b1f2e3b12f5a53/kyo-sttp/shared/src/main/scala/kyo/internal/KyoSttpMonad.scala#L58

I can't see a good solution for it so I'll close this for now and work on a change to avoid `T < IOs` as the completion value of fibers. It was a quick hack to avoid allocations for fiber completions but we can use a proper union type instead.

kyo-core/shared/src/main/scala/kyo/fibers.scala
fwbrasil: yeah, maybe we should always have the `Trace` as well to make debugging easier.

PR #464, opened 2024-06-04
Maybe: allocationless Option (with nesting support!)

fwbrasil:
The implementation matches the behavior of Scala's `Option`, even when there's nesting! The key insight is that nesting is only a problem when there are nested `Some(None)`, which the implementation tracks via cached internal `SomeNone` instances. The only scenarios where this implementation allocates objects is when the nesting of `Some(None)` exceeds `100` (quite unlikely) and when `Some.unapply` is used. I had to use Scala's `Option` for pattern matching because the compiler isn't able to locate `isDefined`/`get` as extension methods to consider `Some.unapply` an extractor.

I'm a bit uneasy about naming this new impl `Option` since it can be confusing to users but we have a precedence with `Duration`.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Maybe.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Maybe.scala
@@ -74,7 +74,7 @@ object Maybe:
         inline def getOrElse[B >: T](inline default: => B): B =
             if isEmpty then default else get
 
-        inline def fold[U](inline ifEmpty: => U, inline ifDefined: T => U): U =
+        inline def fold[U](inline ifEmpty: => U)(inline ifDefined: T => U): U =
             if isEmpty then ifEmpty else ifDefined(get)
 
         inline def map[U](inline f: T => U): Maybe[U] =
--- a/kyo-core/shared/src/main/scala/kyo/Maybe.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Maybe.scala
@@ -1,0 +1,159 @@
+package kyo
+
+import Maybe.*
+import Maybe.internal.*
+
+opaque type Maybe[+T] >: (Empty | Defined[T]) = Empty | Defined[T]
+
+object Maybe:
+    given [T, U](using CanEqual[T, U]): CanEqual[Maybe[T], Maybe[U]] = CanEqual.derived
+    given [T]: Conversion[Maybe[T], IterableOnce[T]]                 = _.iterator
+    given [T: Flat]: Flat[Maybe[T]]                                  = Flat.unsafe.bypass
+
+    def apply[T](v: T): Maybe[T] =
+        if isNull(v) then
+            Empty
+        else
+            v match
+                case v: DefinedEmpty => v.nest
+                case v               => v
+
+    def fromScala[T](opt: Option[T]): Maybe[T] =
+        opt match
+            case Some(v) => Defined(v)
+            case None    => Empty
+
+    def empty[T]: Maybe[T] = Empty
+
+    def when[T](cond: Boolean)(v: => T): Maybe[T] =
+        if cond then v else Empty
+
+    opaque type Defined[+T] = T | DefinedEmpty
+
+    object Defined:
+
+        def apply[T](v: T): Defined[T] =
+            v match
+                case v: DefinedEmpty => v.nest
+                case v: Empty        => DefinedEmpty()
+                case v               => v
+
+        // TODO avoid allocation
+        def unapply[T](opt: Maybe[T]): Option[T] =
+            opt.toScala
+    end Defined
+
+    sealed abstract class Empty
+    case object Empty extends Empty
+
+    extension [T](self: Maybe[T])
+
+        inline def toScala: Option[T] =
+            if isEmpty then None
+            else Some(get)
+
+        inline def isEmpty: Boolean =
+            self match
+                case _: Empty => true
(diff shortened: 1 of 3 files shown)


kyo-core/shared/src/main/scala/kyo/Option.scala
```diff
@@ -0,0 +1,209 @@
...
+        v match
+            case v: SomeNone => v.nest
+            case v: None     => SomeNone()
+            case v           => v
+
+    // TODO avoid allocation
+    def unapply[T](opt: Option[T]): scala.Option[T] =
+        opt.toScala
+
+    extension [T](self: Some[T])
```
fwbrasil, line 23: I had to add methods to `Some` and `None` because Scala isn't able to locate the `Option` extensions for them in some cases.

guizmaii:
Did you see Jon Pretty lib providing allocationless Option implementation? 

https://x.com/propensive/status/1796574146316366077?s=46

kyo-core/shared/src/main/scala/kyo/Option.scala
```diff
@@ -0,0 +1,209 @@
+package kyo
+
+import Option.internal.*
+
+opaque type Option[+T] >: (None | Some[T]) = None | Some[T]
```
hearnadam, line 5: How about `Optional`? `Maybe`? `Opt`?

fwbrasil:
Renamed to `Maybe`

kyo-core/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -0,0 +1,159 @@
...
+                    DefinedEmpty(depth - 1)
+                else
+                    Empty
+            def nest =
+                DefinedEmpty(depth + 1)
+        end DefinedEmpty
+
+        object DefinedEmpty:
+            val cache = (0 until 100).map(new DefinedEmpty(_)).toArray
+            def apply(depth: Int = 1): DefinedEmpty =
```
hearnadam, line 152: I would prefer a `DefinedEmpty.One` vs the default argument...

kyo-core/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -0,0 +1,159 @@
...
+    end Defined
+
+    sealed abstract class Empty
+    case object Empty extends Empty
+
+    extension [T](self: Maybe[T])
+
+        inline def toScala: Option[T] =
+            if isEmpty then None
+            else Some(get)
```
hearnadam, line 53: Just noting that this is only shallow unpacking... it may be annoying in the case of non-trivial nesting. Though I suspect there is rarely a usecase where you want to unpack many nested Maybes to nested Options. The conversion would be quite expensive.

kyo-core/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -0,0 +1,159 @@
...
+
+        inline def filter(inline f: T => Boolean): Maybe[T] =
+            if isEmpty || f(get) then self else Empty
+
+        inline def filterNot(inline f: T => Boolean): Maybe[T] =
+            if isEmpty || !f(get) then self else Empty
+
+        // TODO compilation failure if inlined
+        def contains[U](elem: U)(using CanEqual[T, U]): Boolean =
+            !isEmpty && self.get == elem
```
hearnadam, line 97: Does compilation fail if you remove the `self.` prefix from `get`?

kyo-core/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -0,0 +1,159 @@
...
+        def unapply[T](opt: Maybe[T]): Option[T] =
+            opt.toScala
+    end Defined
+
+    sealed abstract class Empty
+    case object Empty extends Empty
+
+    extension [T](self: Maybe[T])
+
+        inline def toScala: Option[T] =
```
hearnadam, line 51: How about `toOption` now that the class has a unique name?

kyo-core/shared/src/main/scala/kyo/Maybe.scala
fwbrasil: I think we'd need a macro to transform nested types

kyo-core/shared/src/main/scala/kyo/Maybe.scala
fwbrasil: It does. I've removed `self.`. I had added it because the compiler has trouble when I inline this method.

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/Maybe.scala
hearnadam: Couldn't it be recursive?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/Maybe.scala
fwbrasil: The issue is that we'd need to change the type of the nested values and the method operates only at "one level". `Maybe(Maybe(1)).toOption == Option(Maybe(1))`. It becomes even more complicated if it's in a generic scope `def test[T](v: Maybe[T]): Option[T] = v.toOption; test(Maybe(Maybe(1))`

PR #466, opened 2024-06-04
Result (lightweight Try)

fwbrasil:
Following a similar approach as https://github.com/getkyo/kyo/pull/464. The implementation can avoid allocations in case of success but has to allocate in case of failures.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Result.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Result.scala
@@ -95,9 +95,13 @@ object Result:
                 case Success(value) => Result(f(value))
                 case failure        => failure.asInstanceOf[Result[U]]
 
-        inline def fold[U](inline ifFailure: Throwable => U, inline ifSuccess: T => U): U =
+        inline def fold[U](inline ifFailure: Throwable => U)(inline ifSuccess: T => U): U =
             (self: @unchecked) match
-                case Success(value)     => ifSuccess(value)
+                case Success(value) =>
+                    try ifSuccess(value)
+                    catch
+                        case ex if NonFatal(ex) =>
+                            ifFailure(ex)
                 case Failure(exception) => ifFailure(exception)
 
         inline def filter(inline p: T => Boolean): Result[T] =
--- a/kyo-core/shared/src/main/scala/kyo/Result.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Result.scala
@@ -1,0 +1,140 @@
+package kyo
+
+import Result.*
+import kyo.Flat
+import scala.util.control.NonFatal
+
+opaque type Result[+T] >: (Success[T] | Failure[T]) = Success[T] | Failure[T]
+
+object Result:
+    import internal.*
+
+    given [T: Flat]: Flat[Result[T]]                                = Flat.unsafe.bypass
+    given [T](using CanEqual[T, T]): CanEqual[Result[T], Result[T]] = CanEqual.derived
+
+    def apply[T](expr: => T): Result[T] =
+        try
+            Success(expr)
+        catch
+            case ex: Throwable if NonFatal(ex) => Failure(ex)
+
+    def success[T](value: T): Result[T] = Success(value)
+
+    def failure[T](exception: Throwable): Result[T] = Failure(exception)
+
+    def fromEither[T](either: Either[Throwable, T]): Result[T] = either match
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/Result.scala
```diff
@@ -0,0 +1,140 @@
...
+        def unapply[T](self: Result[T]): Option[T] =
+            (self: @unchecked) match
+                case _: Failure[?] => None
+                case s: SuccessFailure[?] =>
+                    Some(s.unnest.asInstanceOf[T])
+                case s: T @unchecked => Some(s)
+
+    end Success
+
+    case class Failure[+T](exception: Throwable)
```
hearnadam, line 49: Assuming this will be the backing for `Fiber`, do we want to store other information? Trace? Do we need a `Cause` type like ZIO?

fwbrasil: Yes, I'm planning to use it to avoid `T < IOs` in fiber completions. The `Option`/`Maybe` was a little detour to explore how to support nesting without allocations :)

I think we could keep the trace information in the exception itself since we can decide which exceptions benefit from a origin `Trace` and which don't.

kyo-core/shared/src/main/scala/kyo/Result.scala
```diff
@@ -0,0 +1,140 @@
...
+            (self: @unchecked) match
+                case Success(value) => ev(value)
+                case failure        => failure.asInstanceOf[Result[U]]
+
+        inline def map[U](inline f: T => U): Result[U] =
+            (self: @unchecked) match
+                case Success(value) => Result(f(value))
+                case failure        => failure.asInstanceOf[Result[U]]
+
+        inline def fold[U](inline ifFailure: Throwable => U, inline ifSuccess: T => U): U =
```
hearnadam, line 98: Would it be beneficial to use multiple parameter lists to improve type inference?

kyo-core/shared/src/test/scala/kyoTest/ResultTest.scala
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+
+    "edge cases" - {
+        "should handle exceptions during fold" in {
+            val tryy      = Success(1)
+            val exception = new RuntimeException("exception")
+            val result =
+                try
+                    tryy.fold(_ => throw exception, _ => throw exception)
```
hearnadam, line 253: https://github.com/scala/scala/blob/2.13.x/src/library/scala/util/Try.scala#L288-L289

I don't think this is the behavior most users of `Try` would expect.

fwbrasil: good catch! fixed

kyo-core/shared/src/test/scala/kyoTest/ResultTest.scala
```diff
@@ -0,0 +1,301 @@
...
+    "getOrElse" - {
+        "returns the value for Success" in {
+            assert(Success(1).getOrElse(0) == 1)
+        }
+        "returns the default value for Failure" in {
+            assert(Failure[Int](new Exception("error")).getOrElse(0) == 0)
+        }
+    }
+
+    "orElse" - {
```
hearnadam, line 55: Can you add a type inference test:

https://github.com/zio/zio/issues/8890

ZIO.succeed(Option.empty[List[String]]).someOrElse(List.empty)
```
Result(Option.empty[List[String]).orElse(Result(List.empty))
```

hearnadam (approved):

kyo-core/shared/src/test/scala/kyoTest/ResultTest.scala
fwbrasil: It doesn't compile. What return type should it have? `Result[IterableOnce]`?

hearnadam: Ah no, sorry that is a bad example. I forgot the `someOrElse` operator will squash the Option.

Perhaps just `Result(List.empty[String]).orElse(List.empty)`

fwbrasil: I've added a couple of inference tests. It seems to be working correctly. I think inference here is simpler than in ZIO's case.

PR #469, opened 2024-06-05
add: TypeMap.show, improving fatal error message

hearnadam:


kyo-core/shared/src/test/scala/kyoTest/typeMapTest.scala
```diff
@@ -168,4 +168,21 @@ class typeMapTest extends KyoTest:
         }
     }
 
+    ".show" - {
+        "many" in {
+            val t = TypeMap("str", true, 42, 'c').add(None).add(List[Any]()).add(Map('k' -> 'v'))
+            val expected =
+                "TypeMap(" +
+                    "Tag[java.lang.String] -> str, " +
```
hearnadam, line 176: @fwbrasil maybe we should add `.showTpe` or equivalent to `Tag` so that the `Tag` is not wrapping each item. Though I don't like permuting the number of methods.

kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -38,6 +38,8 @@ extension [A](self: TypeMap[A])
     inline def size: Int        = self.size
     inline def isEmpty: Boolean = self.isEmpty
 
+    def show: String = self.map { case (tag, value) => s"${tag.show} -> $value" }.toList.sorted.mkString("TypeMap(", ", ", ")")
```
fwbrasil, line 41: Cool! I didn't know this `mkString` with multiple params

kyo-core/shared/src/test/scala/kyoTest/typeMapTest.scala
fwbrasil: It seems useful

fwbrasil (approved):

PR #470, opened 2024-06-05
tags: fix perf regression with int encoding

fwbrasil:
I noticed a performance regression introduced by https://github.com/getkyo/kyo/pull/432. The mechanism to decode `Int`s has to scan an array of chars, which introduces overhead and even allocations in some cases because `indexOf` is an unoptimized Scala implementation.

This PR fixes the regression by moving the search to compile time and producing a table that can be used with a simple offset lookup.

Benchmark results:

image
https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/872121366422a6cd4d1d19b9df1e95b9/raw/c4ccc79d42f61514c7fcdbdfc9ca12f4178caf67/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/872121366422a6cd4d1d19b9df1e95b9/raw/c4ccc79d42f61514c7fcdbdfc9ca12f4178caf67/jmh-result-candidate.json

### DIFF
--- a/kyo-tag/shared/src/main/scala/kyo/tags.scala
+++ b/kyo-tag/shared/src/main/scala/kyo/tags.scala
@@ -1,5 +1,6 @@
 package kyo
 
+import kyo.internal.IntEncoder
 import scala.annotation.switch
 import scala.annotation.tailrec
 import scala.collection.immutable
@@ -151,11 +152,11 @@ object Tag:
             checkSegment(subTag, superTag, subIdx, superIdx) && checkParams(subTag, superTag, subIdx, superIdx)
 
         def checkSegment(subTag: String, superTag: String, subIdx: Int, superIdx: Int): Boolean =
-            val superSize = decodeInt(superTag.charAt(superIdx))
+            val superSize = IntEncoder.decode(superTag.charAt(superIdx))
             @tailrec def loop(subIdx: Int): Boolean =
                 if subIdx >= subTag.length() || superIdx >= superTag.length() then false
                 else
-                    val subSize = decodeInt(subTag.charAt(subIdx))
+                    val subSize = IntEncoder.decode(subTag.charAt(subIdx))
                     (subSize == superSize && subTag.regionMatches(subIdx + 1, superTag, superIdx + 1, subSize)) ||
                     loop(subIdx + subSize + 3)
                 end if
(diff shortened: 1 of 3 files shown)


kyo-tag/shared/src/test/scala/kyoTest/internal/IntEncoderTest.scala
```diff
@@ -0,0 +1,26 @@
...
+
+    "fails out of range" in {
+        intercept[Exception](IntEncoder.encode(-1))
+        intercept[Exception](IntEncoder.encode(89))
+        intercept[Exception](IntEncoder.encode(Int.MaxValue))
+        intercept[Exception](IntEncoder.encode(Int.MinValue))
+    }
+
+    "encodes hashes" in {
+        for _ <- 0 until 100 do
```
hearnadam, line 22: What's the loop for?

kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -307,8 +294,8 @@ object Tag:
                     val sym = tpe.typeSymbol
                     val path = tpe.dealias.baseClasses.map { sym =>
                         val name = toCompact.getOrElse(sym.fullName, sym.fullName)
-                        val size = encodeInt(name.length())
-                        val hash = encodeInt(Math.abs(name.hashCode()) % charsPrintable.length)
+                        val size = IntEncoder.encode(name.length())
+                        val hash = IntEncoder.encodeHash(name.hashCode())
```
hearnadam, line 298: Will this be a problem for custom hashcodes?

hearnadam (approved):

kyo-tag/shared/src/test/scala/kyoTest/internal/IntEncoderTest.scala
fwbrasil: just to generate a few hashcodes

kyo-tag/shared/src/main/scala/kyo/tags.scala
fwbrasil: Not sure I understood. Do you mean using the encoder in other places besides tags?

Issue #317, opened 2024-05-04
Unify/Refine Public APIs


hearnadam:
Unify public APIs:

- Define commonly used names for type parameters
  - eg, `S` is often used for 'pending effects'
- Define commonly used names for parameters (used in the same context)
- Define commonly used names for methods, as well as the semantic meaning of each
  - `get`/`run`

- Document all of these in a shared section to quickly ramp users to Kyo's patterns

hearnadam:
Over the past few weeks I have been thinking about potential encodings. I will first cover type parameters/type names, then we can followup with function naming.

### Pending Effects
- Current: `S`

When describing the pending set of effects to users, I have thought of 3 reasonable conclusions:
- pending **S**et -> `S` type parameter 
  - Pros: consistent with existing code + not hard to understand
  - Con: without an explicit explanation, it's hard to tell where the `S` came from.
- **P**ending set -> `P` type parameter
  - Pros: users can (hopefully) infer from the language around pending type (`<`) where this comes from
- **U**sing Effects -> `U` 
  - Pros: more in line with the caprese alternative decisions
  - Cons: using is very overload, eg the Scala 3 `using` keyword

- My preference:  `S`/`P`

### Errors (Aborts)
- Current: `V`

I don't think there is much question here from me. Without a doubt, we should reuse what other effect systems use (ZIO, Cats).

- `E` -> Error

### Results (All effects)
- Current: `T`

I have 3 ideas here:
- `A` -> historically used for result type in Scala.
  - Pros: `f: A => B` is extremely common, can easily 'enumerate' type parameters
- `T` -> more in line with Java developers/libraries, not my choice, but pre-existing
- `R` -> **R**esult 


### Dependency Injection (Envs)
- Current: `V`

- `R` -> similar to `ZIO[R, _, _]`
- `V` -> not sure the reason behind the name, but I don't mind it
- `D` -> Dependency

fwbrasil:
> pending Set -> S type parameter

I think `S` has been working well. It indicates that the type is meant as a type-level representation of a set, which I think helps understanding. Also, `<` is already named "pending" so we'd use the same meaning for two different things if we adopt `P`.

> E -> Error
> R -> similar to ZIO[R, _, _]

ZIO needs different type parameter names because they're used in the same scope since short-circuiting and dependency injection are encoded in the same monad. Kyo is different given that definition of both effects are in completely isolated scopes. I wouldn't be opposed to changing it, though.

> R -> Result

[...]

PR #471, opened 2024-06-06
Remove separate os lib module + fixes + readme

fwbrasil:
The code was in a separate module to isolate the dependency on `os-lib` but, given that the dependency is not needed anymore, I think keeping it in `kyo-core` is ideal so users don't need to import the separate module.

I've also made a few refactorings to avoid polluting the `kyo` namespace and added a misssing IOs suspension.


kyo-core/jvm/src/main/scala/kyo/Process.scala
```diff
@@ -0,0 +1,235 @@
+package kyo
```
fwbrasil, line 1: I tried to keep the file rename in the git history but I couldn't make it work :(

kyo-core/jvm/src/main/scala/kyo/files.scala
```diff
@@ -127,7 +127,7 @@ class Files(val path: List[String]):
       */
     def readLinesStream(charset: Charset = java.nio.charset.StandardCharsets.UTF_8)(using Trace): Stream[Unit, String, Fibers] < Resources =
         readLoop[String, String, BufferedReader](
-            JFiles.newBufferedReader(toJava, Charset.defaultCharset()),
+            IOs(JFiles.newBufferedReader(toJava, Charset.defaultCharset())),
```
fwbrasil, line 130: Fix to include a missing `IOs` suspension.

fwbrasil:
cc/ @pablf

kyo-core/jvm/src/main/scala/kyo/Process.scala
fwbrasil: It seems it actually worked but github doesn't show it: https://github.com/getkyo/kyo/pull/471/commits/d2b781bca6d72aca0d3276c8a0fd2b6c7b35007b

hearnadam (approved):

fwbrasil:
> @fwbrasil LGTM, it seems better organized this way! Maybe it's more coherent with the rest of the library to use Files.scala instead of files.scala

Yeah, our naming patterns are a mess at the moment. I'll mention reviewing the file names as well in https://github.com/getkyo/kyo/issues/317.

PR #477, opened 2024-06-10
Disable cooperative yielding in zio scheduler

ghostdogpr:
The flag landed in 2.1.2 so we can now disable the automatic yielding that ZIO does every 128 forks and 10240 operations.

Is there a good way to check the impact of this? I see there's a `bench` job but you trigger it manually?


ghostdogpr:
It doesn't really seem to show much improvement, it's even a bit worse.
Screenshot 2024-06-11 at 9 45 01 AM

fwbrasil:
@ghostdogpr yeah, there are benchmarks where preemption is a relevant factor. This change only part of what we'd need for a proper integration since Kyo's preemption signal also needs to be dispatched to ZIO's fibers.

PR #480, opened 2024-06-11
result: support null values

fwbrasil:
I'm changing the fiber completion to use `Result[T]` instead of `T < IOs` and noticed that `Result` produces `MatchError`s for successful `null` values. This PR fixes that and also introduces a new convenience `toTry` method.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Result.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Result.scala
@@ -2,6 +2,7 @@ package kyo
 
 import Result.*
 import kyo.Flat
+import scala.util.Try
 import scala.util.control.NonFatal
 
 opaque type Result[+T] >: (Success[T] | Failure[T]) = Success[T] | Failure[T]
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/Result.scala
```diff
@@ -65,69 +60,69 @@ object Result:
...
-            (self: @unchecked) match
-                case Success(value) => value
-                case Failure(_)     => default
+            fold(_ => default)(v => v)
 
-        inline def orElse[U >: T](inline alternative: => Result[U]): Result[U] =
-            (self: @unchecked) match
-                case Success(value) => self
-                case Failure(_)     => alternative
+        def orElse[U >: T](alternative: => Result[U]): Result[U] =
```
fwbrasil, line 83: Not sure why I started getting compilation errors for this method and had to remove `inline`.

kyo-core/shared/src/test/scala/kyoTest/ResultTest.scala
```diff
@@ -306,4 +307,70 @@ class ResultTest extends KyoTest:
...
+            assert(result.get == null)
+        }
+
+        "Success with null value flatMap" in {
+            val result: Result[String] = Success(null)
+            val flatMapped             = result.flatMap(str => Success(s"mapped: $str"))
+            assert(flatMapped == Success("mapped: null"))
+        }
+
+        "Failure with null exception" taggedAs jvmOnly in {
```
fwbrasil, line 355: marked these as `jvmOnly` because JS throws a different exception from `NullPointerException`

fwbrasil:
I'm merging this to unblock other work but lease feel free to provide feedback so I can address it!

PR #481, opened 2024-06-11
use Result for Fiber completions

fwbrasil:
Fiber completions use `T < IOs` to avoid allocations but we have `Result` now that provides a proper solution for this use case and allows us to avoid `IOs.run` in a few places.


fwbrasil:
Merging because I need this for the work on the new core.

PR #483, opened 2024-06-12
Add workaround for git worktree

steinybot:
Take 2.

Turns out it is actually quite easy to detect if the current workspace is a linked workspace because `.git` is a file instead of a directory.


sideeffffect:
Could this even be upstreamed to sbt-git?

fwbrasil:
LGTM! Let's remove the workaround later in case you manage to get the change upstream.

PR #485, opened 2024-06-13
Use Typelevel scalac-options

steinybot:
Take 2.

Since starting this, the `-Wunused:all` option had been removed which makes things a bit easier.

FTR If we do need to ignore unused locals in a way that is compatible with Scala 2.12, 2.13 and 3 I think we can do this:
```
    /**
     * A way to ignore the unused warning that works in Scala 2.12, 2.13 and 3.
     */
    @elidable(ASSERTION)
    private[scheduler] def unused(a: => Any): Unit = {
        val _ = a
        ()
    }
```

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -26,4 +26,7 @@ fileOverride {
   "glob:**/kyo-stats-otel/**" {
     runner.dialect = scala212source3
   }
-} 
+  "glob:**/scala-3/**" {
+    runner.dialect = scala3
+  }
+}
--- a/build.sbt
+++ b/build.sbt
@@ -29,7 +29,6 @@ lazy val `kyo-settings` = Seq(
     scalaVersion       := scala3Version,
     crossScalaVersions := List(scala3Version),
     scalacOptions     ++= scalacOptionTokens(compilerOptions).value,
-    Test / scalacOptions --= scalacOptionToken(ScalacOptions.languageStrictEquality).value,
     scalafmtOnCompile := false,
     organization      := "io.getkyo",
     homepage          := Some(url("https://getkyo.io")),
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,6 @@
+import org.scalajs.jsenv.nodejs.*
+import org.typelevel.scalacoptions.{ScalaVersion, ScalacOption, ScalacOptions}
+
 val scala3Version   = "3.4.2"
 val scala212Version = "2.12.19"
 val scala213Version = "2.13.14"
@@ -5,14 +8,14 @@ val scala213Version = "2.13.14"
 val zioVersion       = "2.1.2"
 val scalaTestVersion = "3.2.18"
 
-val compilerOptions = Seq(
-    "-encoding",
-    "utf8",
-    "-feature",
-    "-unchecked",
-    "-deprecation",
-    "-Wvalue-discard",
-    "-language:strictEquality"
+val compilerOptions = Set(
+    ScalacOptions.encoding("utf8"),
+    ScalacOptions.feature,
+    ScalacOptions.unchecked,
+    ScalacOptions.deprecation,
+    ScalacOptions.warnValueDiscard,
+    ScalacOptions.languageStrictEquality,
+    ScalacOptions.release("11")
 )
 
 ThisBuild / scalaVersion           := scala3Version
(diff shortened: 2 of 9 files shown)


build.sbt
```diff
@@ -25,7 +28,8 @@ lazy val `kyo-settings` = Seq(
     fork               := true,
     scalaVersion       := scala3Version,
     crossScalaVersions := List(scala3Version),
-    scalacOptions ++= compilerOptions,
+    scalacOptions     ++= scalacOptionTokens(compilerOptions).value,
+    Test / scalacOptions --= scalacOptionToken(ScalacOptions.languageStrictEquality).value,
```
steinybot, line 32: Formatting is a bit out of whack here. I didn't want to rock the boat too much so I did something that looks ok-ish.

fwbrasil: Doesn't this disable strict equality for tests? I'd prefer to keep it on since it's easy to make mistakes and compare kyo computations directly to pure values.

build.sbt
```diff
@@ -43,7 +47,6 @@ lazy val `kyo-settings` = Seq(
     sonatypeProfileName                := "io.getkyo",
     Test / testOptions += Tests.Argument("-oDG"),
     ThisBuild / versionScheme := Some("early-semver"),
-    scalacOptions ++= Seq("-release:11"),
```
fwbrasil, line 46: Is this set automatically? I think the default might be java 8.

fwbrasil:
LGTM besides my comments + the merge conflict.

build.sbt
steinybot: It does yeah because a bunch of them already do this. I can keep it in those specific projects if you prefer.

build.sbt
steinybot: I moved it to be with the rest of the `compilerOptions`

kyo-stats-registry/shared/src/main/scala/kyo/stats/internal/StatsRegistry.scala
```diff
@@ -57,7 +57,7 @@ object StatsRegistry {
         val counterGauges  = new Store[UnsafeCounterGauge]
         lazy val exporters = Collections.newSetFromMap(new ConcurrentHashMap[StatsExporter, java.lang.Boolean])
 
-        class Store[T] {
+        class Store[T <: AnyRef] {
```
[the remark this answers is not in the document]
steinybot: I'll have to double check. I think it was the strict equality one.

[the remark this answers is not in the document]
steinybot: Yeah it was.
```
sbt:kyoJVM> Test/compile
[info] compiling 7 Scala sources to /Users/jason/src/kyo2/kyo-stats-registry/jvm/target/scala-3.4.2/classes ...
[warn] bad option '-Xsource:3' was ignored
[info] compiling 2 Scala sources to /Users/jason/src/kyo2/kyo-tag/jvm/target/scala-3.4.2/test-classes ...
[error] -- [E172] Type Error: /Users/jason/src/kyo2/kyo-stats-registry/shared/src/main/scala/kyo/stats/internal/StatsRegistry.scala:67:20
[error] 67 |                if (value == null) {
[error]    |                    ^^^^^^^^^^^^^
[error]    |             Values of types T and Null cannot be compared with == or !=
[error] -- [E172] Type Error: /Users/jason/src/kyo2/kyo-stats-registry/shared/src/main/scala/kyo/stats/internal/StatsRegistry.scala:76:24
[error] 76 |                    if (ref._1.get() == null) {
[error]    |                        ^^^^^^^^^^^^^^^^^^^^
[error]    |             Values of types T and Null cannot be compared with == or !=
[warn] one warning found
[error] two errors found
[error] (kyo-stats-registry / Compile / compileIncremental) Compilation failed
[error] Total time: 6 s, completed 21/06/2024, 8:50:50 pm
```

build.sbt
steinybot: I put a complete list of all the options down below. This options is still set, albeit with a space instead of colon.

build.sbt
steinybot: I thought it was more prevalent. It was just the `TaskTests` that were missing a `CanEquals[Task, Task]`. I've fixed that and removed this line.

.scalafmt.conf
```diff
@@ -26,4 +26,7 @@ fileOverride {
   "glob:**/kyo-stats-otel/**" {
     runner.dialect = scala212source3
   }
-} 
+  "glob:**/scala-3/**" {
+    runner.dialect = scala3
+  }
```
steinybot, line 31: Without this I get:
```
[error] org.scalafmt.sbt.ScalafmtSbtReporter$ScalafmtSbtError: scalafmt: /home/runner/work/kyo/kyo/kyo-scheduler/shared/src/test/scala-3/kyoTest/scheduler/TaskImplicits.scala:5: error: [dialect scala212source3] type application is not allowed for postfix operators
[error] given CanEqual[Task, Task] = CanEqual.derived
[error]                            ^ [/home/runner/work/kyo/kyo/kyo-scheduler/shared/src/test/scala-3/kyoTest/scheduler/TaskImplicits.scala]
```

fwbrasil (approved):

Issue #462, opened 2024-06-03
New Encoding Again—Maybe


kitlangton:
As I was falling asleep last night, my mind wandered again—oh, the curse of programming—to implementing algebraic effect systems. When I worked on my little talk, I had a few extemporaneous practice sessions with my wife. I was trying to see just how incrementally I could make the refactoring from a basic free monad to a fully-powered, ad hoc effect handler implementation.

One eventually gets to the point where a generic `Command/Operation` has been extracted, and the handle loop must decide if it `accepts` it or not. This leads to the bit which I had the most trouble decomposing pedagogically: The necessity of the continuation which *continues* the current handler loop *only after* some other, heretofore unknown, handler clears the *unaccepted* operation blocking the path ahead. This step introduces a great deal of complexity all at once (I actually never even reached this stage in my final talk 😅—due to taking too long on the earlier steps).

Anyhow, when working with my wife on this problem, we’d reached the following impasse:

```scala
trait Handler[Operation[_]]:
  def apply[A](operation: Operation[A]): A
  def accepts(any: Any): Boolean

def run[Operation[_], A](dsl: DSL[A], handler: Handler[Operation[A]]): A =
  dsl match
    case ReadLine()         => StdIn.readLine()
    case PrintLine(message) => println(message)

    case Embed(operation) =>
      if handler.accepts(operation) then handler(operation.asInstanceOf[Operation[A]])
      else ???

    case Succeed(value) => value

    case FlatMap(dslX, f) =>
      val x = f(run(dslX, handlers*))
      run(x, handlers*)
```

Instead of simply launching headlong into the solution, I first tried to clarify the issue, saying something like: "The issue here is that this particular `handler` may not be able to handle the embedded operation. So, what can we do if our handler cannot handle this? How can we ever get the final `A` out?”.

In posing this question, it really hit me how non-incremental of a change this would need to be. Yes, we certainly *cannot* return an `A`, so this must still return a `DSL[A]`, until we've handled everything—cue our type-level tricks for tracking effects as intersection types. And then of course, we need to deal with creating a new continuation.

It then struck me then that there’s another, much simpler solution to the question: *"What can we do if our handler cannot handle this?”.* The obvious, incremental solution, which allows us to *actually* return an `A` right here in this method, is to ensure that we could indeed handle everything and anything that is thrown at us.

And how could we do that? Well, we simply gather up all of the handlers *before* we ever try to handle everything.

```scala
[...]

PR #489, opened 2024-06-19
new core design (kyo-prelude)

fwbrasil:
This PR is part of the work to introduce a new overall design for the library core abstractions as explored in https://github.com/getkyo/kyo/issues/462. Making a single large change for all effects would be too challenging so I'm splitting the work into two main changes:

1. **Introduce `kyo-prelude`**: I've been wanting to split code that doesn't rely on `IOs` into a separate module (see https://github.com/getkyo/kyo/pull/308) so I'm using this opportunity to introduce the new module as an isolated implementation with the new design under a separate package `kyo2`. Enabling both versions of the core design to be in the codebase will ease migration. For example, we can implement benchmarks for it in `kyo-bench`. The module name is inspired by ZIO :)
2. **Port `kyo-core`**: I'm planning to work on a separate PR that ports `kyo-core` to use the new design in `kyo-prelude`. For the files that this PR duplicates, I'll first delete the duplicate file, move the previous one to the new location, and then update it with the new code. This will allow us to properly keep the git history. Other modules will also need to be ported but they shouldn't require significant changes.

This new core design introduces several major improvements:

## 1. Computation nesting

Kyo's flat representation of computations doesn't seem a major limitation given that composable effects generally preclude the need for nesting but the flat representation introduces adoption and usability issues. Users can get confused by the approach diverging from other similar libraries and passing the `Flat` evidence around in user code doesn't seem ideal but is required in generic contexts that perform effect handling.

In this new design, the `<` type is an `AnyVal`. This enables efficient unboxed execution in the majority of the scenarios and, when a Kyo computation is passed to a generic scope, it's automatically boxed by the compiler, which enables safe execution of nested computations and avoids the need for the `Flat` check. 

I've also explored defining `<` as a context function: `type <[+T, -S] = Runtime ?=> Kyo[T, S]`. Although the compiler desugaring is able to avoid allocations in more scenarios than with `AnyVal`, the fact that context functions can't be aliased by an `opaque type` introduces usability issues since it exposes the internal `Kyo[T, S]` type to the user and the compiler can infer it instead of `<` in certain situations. I think we can still explore this path further later, especially as the compiler has been evolving significantly lately.

## 2. Isolated and user-friendly kernel

I've renamed `core` to `kernel` since it'd be confusing for the `core` to be in `kyo-prelude` instead of `kyo-core`. The new `kernel` implementation reorganizes APIs to be more user-friendly so we can start allowing users to create new effects.

Something that has been challenging while evolving Kyo's design is the fact that internal `core` APIs are used in a few places of the codebase like in `IOTask` and `ZIOs`. The new `kernel` API is designed to fulfill the needs of all effects, including `IOTask` and `ZIOs` via new APIs that enable handling multiple effects at once. This will allow us to iterate faster on the core design in isolation. I think the only missing feature we'll need to port `kyo-core` is how to handle `IOs.ensure`/`Resources` and preemptions, which I'm planning to work on as a follow up.

## 3. Naming improvements

Kyo's naming for effects is based on early versions of the library where most effects had a monad counterpart: `Future`/`Futures`, `Try`/`Tries`, `Option`/`Options`, etc. This hasn't been the case for some time now so there's no need for the pattern using plurals for effect names. In case we have such cases, we can also have internal effects in the companion object (see `Local` in this PR as an example).

I'm also moving away from `T`/`U`/etc for generic type names to `A`/`B`/etc

## 4. Type-safe effect handling

The current lack of type safety when handling effects is a major limitation to allow users to create new effects. In this new design, effects are defined as an abstract polymorphic arrow:

```scala
// Note how an effect has to define lambdas for its input and output types
abstract class Effect[-I[_], +O[_]]

// For example, `Sum` is defined as an arrow that takes a constant `V` and returns `Unit`
sealed trait Sum[V] extends Effect[Const[V], Const[Unit]]

// `Choice` that takes a `Seq` of an arbitrary type and returns the type itself (`Id`)
sealed trait Choice extends Effect[Seq, Id]
```

The effect handling APIs are also more streamlined without requiring `Handler` classes. This means that handling is allocation-free by default without a need to cache handler instances. The new encoding of effects as arrows also enables fully type-safe effect handling. The best example of it is `Var`:

```scala
// If I pass the wrong type to `cont` for each of the input types, the compilation fails!
Effect.handle.state(tag, state, v) {
    [C] =>
        (input, state, cont) =>
            input match
                case input: Get[?] =>
                    (state, cont(state)) 
                case input: Set[?] =>
                    (input(), cont(())) // fails to compile if I pass `state` to cont for example!
                case input: Update[?] =>
                    (input(state), cont(()))
}
```

The effect handling is also "pinned" to the result type of the computation being handled, which is something that wasn't possible with the previous design using separate handler classes. I think we don't have effects that can leverage that yet but it opens new possibilities for how to implement effects.

## 5. Stack safety

Kyo's current design doesn't provide stack safety by default and users need to rely on effect suspensions or `Loops` to ensure stack safety for recursive computations. Although this isn't a major limitation given that recursive computations without effect suspensions don't make much sense and we provide convenient alternatives, providing stack safety should help with adoption since it follows the behavior of other effect systems.

The execution of the new kernel is stack-safe for recursive computations in general but there's still a limitation with computations that dynamically accumulate a large number `map` transformations without any effect suspensions since the kernel still leverages the stack to "unfold" the nested computation when resuming from an effect suspension. I've added pending tests showing this limitation. I explored approaches to avoid this limitation as well but this is a central aspect to optimize the execution of Kyo computations via monomorphization. It seems a reasonable limitation and I'm keeping the `Loop` effect as an alternative.

## 6. Built-in Defer effect

The current core design attempts to provide the core abstractions to define effects without including any effects in the core itself. This approach has two major downsides:

- The core has no way to suspend execution by itself (no way to provide stack safety)
- Effects that only need suspensions like `Locals` have to rely on another effect (currently `IOs` but it could be `Defers`)

The new kernel has an internal `Defer` effect that is used to provide stack safety and encode the use of local values without a separate effect. This approach means that a `A < Any` is actually a `A < Defer` internally, which is why `kyo.pure` is renamed to `kyo.eval` since it can have pending internal `Defer` suspensions.

## 7. New runtime effects

There are effects like `Env` and `Local` that only require propagating values into the computation without the need for a proper effect handling with a continuation. The new kernel provides a new kind of effect for these scenarios:

```scala
// The `A` type parameter specifies the type of the local value the effect uses
abstract class RuntimeEffect[+A]

// For example, `Env` uses a `TypeMap`
sealed trait Env[+R] extends RuntimeEffect[TypeMap[R]]

// `Local` uses an untyped map (State is its internal effect type)
sealed private trait State extends RuntimeEffect[Map[Local[?], AnyRef]]
```

Suspension of runtime effects has an interesting characteristic. If the suspension provides a default value, the effect isn't added to the pending set.

```scala
// The `Env` effect suspends without a default value so it's added to the pending set
// (this is a simplification of the actual impl that requires erased tags)
def get[R](using tag: Tag[Env[R]]): R < Env[R] =
    RuntimeEffect.suspend(tag)(_.get(using tag))
    
// But `Local` provides a default value (`Map.empty`) so it doesn't get added to the pending set
def get: T < Any =
    RuntimeEffect.suspend(Tag[Local.State], Map.empty)(_.getOrElse(this, default).asInstanceOf[T]) 
```

This approach is designed so we can easily propagate runtime effects when forking fibers. The method `RuntimeEffect.boundary` saves the state of the runtime, which includes runtime effect values and traces, and provides a function to resume the computation using the original runtime state. The method is currently unused and I'll explore it further when porting `kyo-core`.

```scala
// I'm using a match type to remove runtime effects from the pending set
type WithoutRuntimeEffects[S] = S match
    case RuntimeEffect[x] => Any
    case s1 & s2          => WithoutRuntimeEffects[s1] & WithoutRuntimeEffects[s2]
    case _                => S

// Note how `f` is a function that produces a computation without pending runtime effects
// since they're already handled by the "propagated" runtime
def boundary[A, B, S, S2](v: A < S)(
    f: (() => A < WithoutRuntimeEffects[S]) => B < S2
)(using _frame: Frame): B < (S & S2) =
```

## 8. Stack traces

The new `Runtime` mechanism also collects traces of the execution and automatically injects frames into exception stack traces. The `boundary` method is also designed to propagate the trace to forked fibers. Example output:

```
java.lang.Exception: test exception
	at kyo2.TraceTest.ex(TraceTest.scala:8)
	at      x.map(_ => throw ex) @ kyo2.TraceTest.boom(TraceTest.scala:10)
	at , y).map(_ + _).map(boom) @ kyo2.TraceTest.withEffects(TraceTest.scala:17)
	at  Kyo.zip(x, y).map(_ + _) @ kyo2.TraceTest.withEffects(TraceTest.scala:17)
	at             Kyo.zip(x, y) @ kyo2.TraceTest.withEffects(TraceTest.scala:17)
	at       Env.use[Int](_ * 2) @ kyo2.TraceTest.loop(TraceTest.scala:16)
	at             Kyo.zip(x, y) @ kyo2.TraceTest.withEffects(TraceTest.scala:17)
	at       Env.use[Int](_ + 1) @ kyo2.TraceTest.loop(TraceTest.scala:15)
	at        Env.run(1)(z).eval @ kyo2.TraceTest.withEffects(TraceTest.scala:18)
	at kyo2.TraceTest.withEffects(TraceTest.scala:18)
```

Note how traces also include a short snippet of the code of the frame. This behavior is on by default but we should provide a way to disable it since I imagine it could break tooling that read stack traces. I think we could also explore providing a separate way to render traces by compacting the information into a base64 string and providing a link to a static page that reads the payload and renders the trace.

## 9. Remove Options effect

I wasn't sure what to name `Options` since we're moving away from plurals and decided to remove it instead. I think I've never used it and we can provide the same functionality via `Abort` directly. I've added a `Abort.get(someOption)` method instead. 

## Conclusion

I apologize for the large and intrusive change but I think it's necessary to evolve the library faster. This new design gets us much closer to a stable set of primitives.


build.sbt
```diff
@@ -14,7 +14,8 @@ val compilerOptions = Seq(
     "-unchecked",
     "-deprecation",
     "-Wvalue-discard",
-    "-language:strictEquality"
+    "-language:strictEquality",
+    "-Ykind-projector"
```
fwbrasil, line 18: To make it easier to define effects like `Var`.

build.sbt
```diff
@@ -162,7 +181,7 @@ lazy val `kyo-core` =
             libraryDependencies += "dev.zio"      %%% "zio-laws-laws"   % "1.0.0-RC27" % Test,
             libraryDependencies += "dev.zio"      %%% "zio-test-sbt"    % "2.1.3"      % Test,
             libraryDependencies += "ch.qos.logback" % "logback-classic" % "1.5.6"      % Test,
-            libraryDependencies += "javassist"      % "javassist"       % "3.12.1.GA"  % Test
+            libraryDependencies += "org.javassist"  % "javassist"       % "3.30.2-GA"  % Test
```
fwbrasil, line 184: I had trouble with `BytecodeTest` failing in `kyo-prelude` and noticed that we're on an old version of `javassist` . Scala steward isn't updating it because the package name changed.

kyo-prelude/jvm/src/main/scala/kyo2/kernel/Platform.scala
```diff
@@ -0,0 +1,25 @@
+package kyo2.kernel
+
+import kyo2.kernel.Mode
+import scala.concurrent.ExecutionContext
+
+object Platform:
+
+    val mode: Mode =
```
fwbrasil, line 8: I've added this `Mode` mechanism so we can vary configurations based on it. For example, if the mode is `Debugging`, we could render a fancier trace. This is unused yet, though.

kyo-prelude/jvm/src/test/scala/kyo2/MonadLawsTest.scala
```diff
@@ -0,0 +1,82 @@
...
+import zio.prelude.Equal
+import zio.prelude.coherent.CovariantDeriveEqual
+import zio.prelude.coherent.CovariantDeriveEqualIdentityFlatten
+import zio.prelude.laws.*
+import zio.test.*
+import zio.test.laws.*
+
+object MonadLawsTest extends ZIOSpecDefault:
+
+    case class Myo[+T](v: T < (Env[String] & Abort[String] & Sum[Int] & Var[Boolean]))
```
fwbrasil, line 13: I was hoping we wouldn't need this wrapper anymore but the compiler doesn't allow extending interfaces with an `AnyVal` type parameter. The encoding using a context function had the same issue with the compiler not allowing to extend interfaces with a context function type.

kyo-prelude/jvm/src/test/scala/kyo2/kernel/BytecodeTest.scala
```diff
@@ -0,0 +1,55 @@
...
+    class TestSuspend:
+        def test() = Effect.suspend[Int](Tag[TestEffect.type], 42)
+
+    class TestMap:
+        def test(v: Int < TestEffect.type) = v.map(_ + 1)
+
+    class TestHandle:
+        def test(v: Int < TestEffect.type) = Effect.handle(Tag[TestEffect.type], v)([C] => (input, cont) => cont(input))
+
+    "suspend" in {
```
fwbrasil, line 20: The new kernel generates larger trees to keep the trace information and provide stack safety. I'll follow up later to try reduce the generated bytecode size.

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -0,0 +1,93 @@
...
+import scala.reflect.ClassTag
+import scala.util.Failure
+import scala.util.Success
+import scala.util.Try
+
+sealed trait Abort[-E] extends Effect[Const[E], Const[Unit]]
+
+object Abort:
+
+    private inline def erasedTag[E]: Tag[Abort[E]] = Tag[Abort[Any]].asInstanceOf[Tag[Abort[E]]]
```
fwbrasil, line 17: We still need untyped effect handling here because it's not possible to use `<:<` to compare tags or use `Tag.Full` in the kernel due to the performance cost. I'm planning to continue the work to integrate union and intersection type tags into the tag string representation itself and avoid string scans by prefixing payloads with their length, which should enable us to avoid the erased tag here and in `Env` (see branch https://github.com/getkyo/kyo/tree/tags-take-3)

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -0,0 +1,93 @@
...
+                                case input: E => true
+                                case _        => false,
+                    handle = [C] => (input, _) => Left(input)
+                ).asInstanceOf[Either[E, A] < ER & B]
+            } {
+                case fail: E => Left(fail)
+            }
+    end RunOps
+
+    inline def run[E >: Nothing]: RunOps[E] = RunOps(())
```
fwbrasil, line 68: Improvements of the `*Dsl` classes:

1. Renamed to `*Ops`, I think `Dsl` can be confusing to users
2. The `Ops` classes are now `AnyVal` so the compiler can avoid emitting code for their allocation
3. Inlined the methods that produce `Ops` classes

We should migrate to [clause interleaving](https://docs.scala-lang.org/sips/clause-interleaving.html) once it's not experimental anymore.

kyo-prelude/shared/src/main/scala/kyo2/Aspect.scala
```diff
@@ -0,0 +1,65 @@
+package kyo2
+
+object Aspect:
```
fwbrasil, line 3: I've kept the same APIs as the original effects in `kyo-core`, only adapting the internals if needed. For example, in this case no changes were needed.

kyo-prelude/shared/src/main/scala/kyo2/Choice.scala
```diff
@@ -0,0 +1,32 @@
...
+
+    inline def get[T](seq: Seq[T]): T < Choice =
+        Effect.suspend[T](Tag[Choice], seq)
+
+    inline def eval[T, U, S](seq: Seq[T])(inline f: T => U < S): U < (Choice & S) =
+        seq match
+            case Seq(head) => f(head)
+            case seq       => Effect.suspend[T](Tag[Choice], seq, f)
+
+    inline def dropIf[S](condition: Boolean): Unit < (Choice & S) =
```
fwbrasil, line 18: A minor API change renaming this method from `filter` to `dropIf`. It seems more intuitive, `filter` made more sense when the effect was still named `Seqs`.

kyo-prelude/shared/src/main/scala/kyo2/Kyo.scala
```diff
@@ -0,0 +1,136 @@
+package kyo2
+
+import kernel.<
+import kernel.Frame
+import kernel.Loop
+import scala.annotation.tailrec
+
+object Kyo:
```
fwbrasil, line 8: I've moved some methods to the `Kyo` companion object. I'll review how this plays with `kyo-combinators` when porting it to the new design.

kyo-prelude/shared/src/main/scala/kyo2/Kyo.scala
```diff
@@ -0,0 +1,136 @@
...
+    def zip[T1, T2, S](v1: T1 < S, v2: T2 < S)(using Frame): (T1, T2) < S =
+        v1.map(t1 => v2.map(t2 => (t1, t2)))
+
+    def zip[T1, T2, T3, S](v1: T1 < S, v2: T2 < S, v3: T3 < S)(using Frame): (T1, T2, T3) < S =
+        v1.map(t1 => v2.map(t2 => v3.map(t3 => (t1, t2, t3))))
+
+    def zip[T1, T2, T3, T4, S](v1: T1 < S, v2: T2 < S, v3: T3 < S, v4: T4 < S)(using Frame): (T1, T2, T3, T4) < S =
+        v1.map(t1 => v2.map(t2 => v3.map(t3 => v4.map(t4 => (t1, t2, t3, t4)))))
+
+    object seq:
```
fwbrasil, line 19: This is the code from `Seqs` in `kyo-core`.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Frame.scala
```diff
@@ -0,0 +1,91 @@
+package kyo2.kernel
+
+import scala.annotation.tailrec
+import scala.quoted.*
+
+opaque type Frame <: AnyRef = String
+
+object Frame:
```
fwbrasil, line 8: Equivalent to `Trace` in `kyo-core` with a few improvements

kyo-prelude/shared/src/main/scala/kyo2/kernel/Runtime.scala
```diff
@@ -0,0 +1,124 @@
...
+final private class Runtime(initDepth: Int, initState: Runtime.State):
+    import Runtime.State
+
+    private val owner         = Thread.currentThread()
+    private[kernel] var depth = initDepth
+    private var traceIdx      = initState.traceIdx
+    private val trace         = Runtime.copyTrace(initState.trace, traceIdx)
+
+    private def enter(_frame: Frame, force: Boolean = false): Int =
+        assert(Thread.currentThread eq owner, "Leaked runtime! " + (Thread.currentThread, owner))
```
fwbrasil, line 18: This is something I'm not sure about: I'm using implicits as a way to propagate the current `Runtime` without having to read from `ThreadLocal` on each transformation to reduce the performance overhead. I tried to avoid it but there can be situations where a `Runtime` instance gets captured by a computation and reused later from another thread. I'll explore this further when porting fibers to the new design.

kyo-prelude/shared/src/test/scala/kyo2/ChoiceTest.scala
```diff
@@ -0,0 +1,98 @@
...
+        try
+            assert(
+                kyo2.Choice.run(kyo2.Choice.get(largeChoice)).eval == largeChoice
+            )
+        catch
+            case ex: StackOverflowError => fail()
+        end try
+    }
+
+    "large number of suspensions" in pendingUntilFixed {
```
fwbrasil, line 84: This is still stack-unsafe because the nested `Choice.run` call during effect handling doesn't go through the new stack depth tracking mechanism. We can probably fix it later.

kyo-tag/shared/src/main/scala/kyo/tags.scala
```diff
@@ -19,7 +19,7 @@ object Tag:
     inline given apply[T]: Tag[T] = ${ tagImpl[T] }
 
     extension [T](t1: Tag[T])
-        def erased: Tag[Any] = t1.asInstanceOf[Tag[Any]]
+        inline def erased: Tag[Any] = t1.asInstanceOf[Tag[Any]]
```
fwbrasil, line 22: This method not being inlined was preventing optimizations when suspending `Abort` and `Env` with erased tags.

kyo-prelude/shared/src/test/scala/kyo2/EnvTest.scala
```diff
@@ -0,0 +1,314 @@
...
+            val intTypeMap    = TypeMap(123)
+            val boolTypeMap   = TypeMap(true)
+            assert(
+                kyo2.Env.run(true)(
+                    kyo2.Env.runTypeMap(stringTypeMap)(kyo2.Env.runTypeMap(intTypeMap)(kyo2.Env.runTypeMap(boolTypeMap)(kyo)))
+                ).eval == ("Hello", 123, true)
+            )
+        }
+
+        "providing the wrong env map" in {
```
fwbrasil, line 260: Note how the scenarios where the effect getting handled isn't pending are now supported and correctly leave effects still pending. This is possible because the computation gets nested via `AnyVal` boxing when the effects have a mismatch.

kyo-prelude/js/src/main/scala/java/lang/ThreadLocal.scala
```diff
@@ -0,0 +1,11 @@
+package java.lang
+
+import java.util.function.Supplier
+
+class ThreadLocal[T](private var value: T):
+    def this() = this(null.asInstanceOf[T])
```
steinybot, line 6: Although it doesn't do a whole lot, it might be more correct to do this?
```suggestion
class ThreadLocal[T](private var value: T | Null = null):
```

fwbrasil:
I've just pushed an important improvement. The encoding with `<` defined as `opaque type <[+T, -S] >: T` didn't allow us to avoid nesting of computations by the compiler since the value was widened directly by the type system. With the new kernel, the lifting is done via an implicit conversion so it's possible to use a `NotGiven` to avoid lifting nested computations:

```scala
implicit inline def lift[A, S](v: A)(using inline ng: NotGiven[A <:< (Any < Nothing)]): A < S = <(v) 
```

This is an important behavior for usability because mismatch of pending effects trigger lifting. For example:

```scala
def test[T](v: T < IO): T < IO = ...
val a: Int < Choice = ...

// Without `NotGiven` the compiler nests the computation if there's a mismatch.
// This fails to compile after my last change
val b: Int < Choice < IO = test(a)
```

Although nesting is safe with the new encoding, this behavior when there's a mismatch in pending effects is an important usability issue. Note that nesting is still possible indirectly as demonstrated in the tests (see PendingTest.flatten as an example).

kyo-prelude/shared/src/main/scala/kyo2/kernel/Runtime.scala
fwbrasil: I've made [a change](https://github.com/getkyo/kyo/pull/489/commits/9bb294f17b837ac76f12bcfed32eab7ac24a984c) to provide graceful handling of runtime leaks. In case of mismatch, the computation suspends and the corresponding `.eval` call will re-wire the correct `Runtime` instance for the current thread.

I've also added tests for it and a good thing is that I couldn't reproduce a leak using Kyo's APIs directly, only by manually saving the Runtime instance (last two scenarios).

kyo-prelude/shared/src/main/scala/kyo2/kernel/RuntimeEffect.scala
```diff
@@ -0,0 +1,103 @@
...
+                    kyo.asInstanceOf[B]
+        handleLoop(v)
+    end handle
+
+    type WithoutRuntimeEffects[S] = S match
+        case RuntimeEffect[x] => Any
+        case s1 & s2          => WithoutRuntimeEffects[s1] & WithoutRuntimeEffects[s2]
+        case _                => S
+
+    def boundary[A, B, S, S2](v: A < S)(
```
fwbrasil, line 75: @johnhungerford I'm sorry for asking for your help again with these type-level issues 😅 but I'm having trouble using this new boundary method. The idea is to use it when forking fibers:

```scala
object Fiber:
    // I'd need to restrict `S` here only to runtime effects
    def init[A, S](v: A < (S & Fibers)): Fiber[A] < (IO & S) =
        Runtime.boundary(v) { (cont: A < Fibers) =>
            // schedule execution without `S` pending
        }
```

I've tried a few ways, including the `Has` and `Reducible` mechanisms, but had no luck. Using type constraints like `_ <: RuntimeEffect[?]` also doesn't work because `(SomeRuntimeEffect & NonRuntimeEffect) <:< RuntimeEffect[?]`.

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -0,0 +1,101 @@
...
+            frame: Frame
+        ): A < (Abort[E] & B) =
+            Effect.catching(v) {
+                case ex: E => Abort.fail(ex)
+            }
+    end CatchingOps
+
+    inline def catching[E <: Throwable]: CatchingOps[E] = CatchingOps(())
+
+    sealed trait HasAbort[E, ES]:
```
fwbrasil, line 91: I've tried to migrate to the new `Reducible` approach you suggested but I couldn't declare the `Eliminable` instances separate from the `Eliminable` companion object. I wanted to keep `Reducible`/ `Eliminable` in the `kyo.kernel` package, which shouldn't have references to concrete effects. Since the `Has` mechanism is simpler and `Reducible` still has a type cast, I don't see much benefit unless the mechanism is generic in `kyo.kernel`.

fwbrasil, line 91: @johnhungerford I've tried to migrate to the new Reducible approach you suggested but I couldn't declare the `Eliminable` instances separate from the `Eliminable` companion object. I wanted to keep `Reducible`/`Eliminable` in the `kyo.kernel` package, which shouldn't have references to concrete effects. Since the `Has` mechanism is simpler and `Reducible` still has a type cast, I don't see much benefit unless the mechanism is generic in kyo.kernel.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Runtime.scala
```diff
@@ -0,0 +1,121 @@
...
+    private def exit(depth: Int): Unit =
+        this.depth = depth - 1
+
+    private[kernel] def save(values: Values) =
+        State(Runtime.copyTrace(trace, traceIdx), traceIdx, values)
+end Runtime
+
+object Runtime:
+
+    implicit def get: Runtime = local.get()
```
hearnadam: Assuming this should not be called often (if ever) by users, should it be `private[kyo]`?

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
hearnadam: Is there any way to create low priority implicits without extension? I assume we could expose a constructor if that's the case.

kyo-prelude/shared/src/main/scala/kyo2/Env.scala
```diff
@@ -0,0 +1,62 @@
...
+            using tag: Tag[R]
+        ): A < (Env[R] & S) =
+            def loop(): A < (Env[R] & S) =
+                RuntimeEffect.suspend(erasedTag[R]) { map =>
+                    val value = map.getOrElse(null)(using tag.erased).asInstanceOf[R]
+                    if isNull(value) then loop()
+                    else f(value)
+                }
+            end loop
+            loop()
```
hearnadam, line 43: Why is the loop now necessary?

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
johnhungerford: There's no necessity that `Eliminable` be a sealed trait, in which case you should be able to define given instances of `Eliminable[X]` in the companion object of `X`. So you'd have:

```scala
object Envs:
  given Eliminable[Envs[Any]] with {}
```

kyo-prelude/shared/src/test/scala/kyo2/kernel/TraceTest.scala
```diff
@@ -105,6 +105,7 @@ class TraceTest extends Test:
                 ex.printStackTrace(printWriter)
                 printWriter.flush()
                 val trace = stringWriter.toString.linesIterator.takeWhile(!_.contains("$init$$$anonfun")).mkString("\n")
+                println(trace)
```
hearnadam, line 108: Debug println 🙂

kyo-prelude/shared/src/main/scala/kyo2/kernel/RuntimeEffect.scala
johnhungerford: I'll take a look at this tomorrow. This might be tricky though. I haven't had much luck using type class derivation or inlining to iterate over the components of an intersection type.

kyo-prelude/jvm/src/main/scala/kyo2/kernel/Platform.scala
```diff
@@ -0,0 +1,25 @@
...
+        Option(System.getProperty("kyo.kernel.Platform.mode"))
+            .map(Mode.valueOf).getOrElse {
+                if isDebugEnabled then
+                    Mode.Debugging
+                else
+                    Mode.Development
```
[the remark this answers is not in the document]
fwbrasil: The idea was to have `Development` as the default and `Debugging` as a specialization of it when the development environment has the debug port enabled. But reconsidering it, it seems a bit dangerous that `Debugging` might have behaviors that make only sense in a development environment but can be enabled in production just by setting the debug port configurations. Perhaps we could remove `Mode.Debugging` since it's still possible to check if the debug port is enabled via `Platform.isDebugEnabled`.

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
fwbrasil: > Can you change name for Unit from ign -> dummy?

Good point! I'm renaming it to `dummy`.

> Additionally, is there a benefit to inlining the def run?

It saves a few bytecodes because calling a method uses more instructions than referring to `()`. It's possible to see that in `BytecodeTest`. If I remove the `inline` from `Effect.suspend`, the test fails with:

```
[info] BytecodeTest:
[info] - suspend *** FAILED *** (34 milliseconds)
[info]   Map("test" -> 27) did not equal Map("test" -> 20) (BytecodeTest.scala:25)
[info]   Analysis:
[info]   Map("test": 27 -> 20)
[info] - suspendMap (14 milliseconds)
[info] - map (8 milliseconds)
[info] - handle (6 milliseconds)
```

kyo-prelude/shared/src/main/scala/kyo2/Choice.scala
fwbrasil: No, it's probably a leftover. Good catch!

kyo-prelude/shared/src/main/scala/kyo2/Choice.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import kyo.Tag
+import kyo2.kernel.*
+
+sealed trait Choice extends Effect[Seq, Id]
+
+object Choice:
+
+    inline def get[T](seq: Seq[T]): T < Choice =
+        Effect.suspend[T](Tag[Choice], seq)
```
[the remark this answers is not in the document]
fwbrasil: It's actually already a constant string in the class constant pool. By passing it as the parameter, the tag gets inlined into the `tag` method of the `KyoSuspend` class generated by `Effect.suspend`. If we move it to a field, it'd become a captured value and the compiler will generate a field for it in the `KyoSuspend` class, which increases the memory footprint.

Example decompiled code from `BytecodeTest.TestSuspend`:

```java
public class BytecodeTest.TestSuspend {
    public BytecodeTest.TestSuspend() {
        if (BytecodeTest.this == null) {
            throw new NullPointerException();
        }
    }

    public Object test() {
        package.internal$.KyoSuspend p$proxy1 = new package.internal$.KyoSuspend<?, ?, BytecodeTest.TestEffect$, Object, Object, BytecodeTest.TestEffect$>(){

            // Note how the frame information is also stored without a field
            public String frame() {
                // Scala's inlining typically leaves these unused values but they're trivially removed by the JIT
                Frame$package$ frame$package$ = Frame$package$.MODULE$;
                Frame.package.Frame$ Frame$_this = Frame.package.Frame$.MODULE$;
                Frame$package$ Frame$package$_this = frame$package$;
                return "kyo2.kernel.BytecodeTest.TestSuspend\u00a3test\u00a3BytecodeTest.scala\u00a312\u00a367\u00a3Tag[TestEffect.type], 42)\u00a3class TestSuspend:\n    def test() = Effect.suspend[Int](Tag[TestEffect.type], 42)\ud83d\udccd";
            }

            public String tag() {
                tags$package$ tags$package$2;
                tags$package$ tags$package$_this = tags$package$2 = tags$package$.MODULE$;
                return "H1kyo2.kernel.BytecodeTest.TestEffect$;5?kyo2.kernel.Effect;!T0;!U1;!V2;";
            }

            public int input() {
                return 42;
            }

            public Object apply(int v, Map values, Safepoint x$3) {
                return $less$.MODULE$.apply((Object)BoxesRunTime.boxToInteger((int)v));
            }
        };
        // The compiler unfortunately doesn't reduce this tree to `p$proxy1` but
        // JIT compilers can trivially elide this kind of allocation
        return new .less((Object)p$proxy1).kyo2$kernel$$less$$curr();
    }

    public final /* synthetic */ BytecodeTest kyo2$kernel$BytecodeTest$TestSuspend$$$outer() {
        return BytecodeTest.this;
    }
}
```

kyo-prelude/shared/src/main/scala/kyo2/Env.scala
```diff
@@ -0,0 +1,62 @@
...
+import scala.util.NotGiven
+
+sealed trait Env[+R] extends RuntimeEffect[TypeMap[R]]
+
+object Env:
+
+    private inline def erasedTag[R] = Tag[Env[Any]].asInstanceOf[Tag[Env[R]]]
+
+    inline def get[R](using inline tag: Tag[R]): R < Env[R] =
+        use[R](v => v)
```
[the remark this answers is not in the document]
fwbrasil: I thought the compiler wouldn't reduce the function call in `use` with `identity` but it does! I'll update it.

kyo-prelude/shared/src/main/scala/kyo2/kernel/package.scala
```diff
@@ -0,0 +1,68 @@
...
+            final override def toString =
+                val parsed = frame.parse
+                s"Kyo(${tag.show}, Input($input), ${parsed.position}, ${parsed.snippetShort})"
+        end Suspend
+
+        abstract class KyoDefer[T, S] extends Suspend[Const[Unit], Const[Unit], Defer, Any, T, S]:
+            final def tag   = Tag[Defer]
+            final def input = ()
+
+        class Values(values: immutable.Map[Tag[Any], AnyRef]) extends AnyVal:
```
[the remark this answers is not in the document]
fwbrasil: Yes, this uses only exact tag matches for performance. I'm hoping the new tag improvements will allow us to use `TypeMap` here.

kyo-prelude/shared/src/main/scala/kyo2/Local.scala
```diff
@@ -0,0 +1,39 @@
...
+    val default: T
+
+    private val _get: T < Any =
+        RuntimeEffect.suspend(Tag[Local.State], Map.empty)(_.getOrElse(this, default).asInstanceOf[T])
+
+    def get: T < Any = _get
+
+    // TODO Compilation error if inlined
+    def use[U, S](f: T => U < S)(using Frame): U < S =
+        RuntimeEffect.suspend(Tag[Local.State], Map.empty)(map => f(map.getOrElse(this, default).asInstanceOf[T]))
```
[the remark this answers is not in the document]
fwbrasil: yeah, it seems to allocate two options. One for `get` and another for the `Some` pattern matching:

```scala
  def get(key: K): Option[V]
  
  def getOrElse[V1 >: V](key: K, default: => V1): V1 = get(key) match {
    case Some(v) => v
    case None => default
  }
  ```
  
I haven't seen it in profiling sessions, though. This kind of temporary object is normally easy for the JIT compiler to optimize.

File: kyo-prelude/shared/src/main/scala/kyo2/kernel/Runtime.scala (hunk 20 above)
Comment (fwbrasil) [reply kyo-prelude/shared/src/main/scala/kyo2/kernel/Runtime.scala]: It needs to be available in user code to resolve the `Safepoint` (previously `Runtime`) when performing transformations.

File: kyo-prelude/shared/src/main/scala/kyo2/package.scala [hunk 28]
```diff
@@ -0,0 +1,31 @@
...
+private[kyo2] inline def isNull[T](v: T): Boolean =
+    v.asInstanceOf[AnyRef] eq null
+
+private[kyo2] inline def discard[T](v: T): Unit =
+    val _ = v
+    ()
+
+private[kyo2] object bug:
+
+    case class KyoBugException(msg: String) extends Exception(msg)
```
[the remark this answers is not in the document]
fwbrasil: Good idea! Scala's `NonFatal` has some problems too, we could explore a better implementation.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Pending.scala
```diff
@@ -0,0 +1,76 @@
+package kyo2.kernel
+
+import internal.*
+import kyo.*
+import scala.annotation.tailrec
+import scala.language.implicitConversions
+import scala.util.control.NonFatal
+
+case class <[+A, -S](private val curr: A | Kyo[A, S]) extends AnyVal
```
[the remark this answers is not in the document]
fwbrasil: The compiler doesn't allow it because of `AnyVal`.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Pending.scala
```diff
@@ -0,0 +1,76 @@
...
+
+object `<`:
+
+    implicit private[kyo2] inline def apply[A, S](p: Kyo[A, S]): A < S = new <(p)
```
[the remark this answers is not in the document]
fwbrasil: The compiler needs it here otherwise it infers a recursive call to `apply`.

kyo-prelude/shared/src/main/scala/kyo2/Env.scala
fwbrasil: Sorry, it's a leftover from debugging. Fixed.

fwbrasil:
`Runtime` is an overloaded term, especially considering the parallel with ZIO's `Runtime`, and gives an impression of an impure effect, which isn't the case. I've renamed `Runtime` to `Safepoint`, `RuntimeEffect` to `ContextEffect`, and `Values` to `Context`. It seems to communicate better what those types do.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
```diff
@@ -0,0 +1,121 @@
...
+import internal.*
+import java.util.Arrays
+import kyo2.isNull
+import scala.annotation.tailrec
+import scala.util.control.NonFatal
+
+final class Safepoint(initDepth: Int, initState: Safepoint.State):
+    import Safepoint.State
+
+    private val owner         = Thread.currentThread()
```
hearnadam, line 12: Is there any case where this value could be shared to avoid the method invocation?

```
final class Safepoint(initDepth: Int, initState: Safepoint.State, private val owner = Thread.currentThread())
```

Not sure if it's worth the complexity

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
```diff
@@ -0,0 +1,121 @@
...
+                if t != null then
+                    result(i) = t
+                    loop(i + 1)
+        loop(0)
+        result
+    end copyTrace
+
+    import internal.*
+
+    final class State(
```
hearnadam, line 110: I assume you don't want a case class here due to the equality issues with Arrays?

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -0,0 +1,101 @@
...
+                Effect.handle(erasedTag[E], v.map(Right(_): Either[E, A]))(
+                    accept = [C] =>
+                        (input, _) =>
+                            (input.asInstanceOf[Any]) match
+                                case input: E => true
+                                case _        => false,
+                    handle = [C] => (input, _) => Left(input)
+                ).asInstanceOf[Either[E, A] < ER & B]
+            } {
+                case fail: E => Left(fail)
```
hearnadam, line 72: Should `catching` only be used if `E` is a subtype of `Throwable`? I wouldn't think an unbounded type parameter would work in this case.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
```diff
@@ -0,0 +1,121 @@
...
+        val trace: Array[Frame],
+        val traceIdx: Int,
+        val context: Context
+    )
+
+    object State:
+        val empty = State(new Array(maxStackDepth), 0, Context.empty)
+
+    private[kernel] val local = ThreadLocal.withInitial(() => Safepoint(0, State.empty))
```
hearnadam, line 119: Do you expect any serious performance limitation of Thread locals? I wonder if `ScopedValue` could be used here - though that limits the executor to newer JDKs.

kyo-prelude/shared/src/main/scala/kyo2/kernel/ContextEffect.scala
```diff
@@ -0,0 +1,69 @@
...
+        def handleLoop(v: B < (E & S))(using Safepoint): B < S =
+            v match
+                case <(kyo: KyoSuspend[IX, OX, EX, Any, B, S] @unchecked) =>
+                    new KyoSuspend[IX, OX, EX, Any, B, S]:
+                        val tag   = kyo.tag
+                        val input = kyo.input
+                        def frame = _frame
+                        def apply(v: OX[Any], context: Context)(using Safepoint) =
+                            val tag   = _tag
+                            val value = context.getOrElse(tag, null)
```
hearnadam, line 59: can't this just be `get`? Or is it cheaper to avoid the `bug` function allocation?

kyo-prelude/shared/src/main/scala/kyo2/kernel/package.scala
```diff
@@ -0,0 +1,65 @@
...
+        class Context(context: immutable.Map[Tag[Any], AnyRef]) extends AnyVal:
+            inline def getOrElse[A, E <: ContextEffect[A], B >: A](tag: Tag[E], inline default: => B): B =
+                if !context.contains(tag.erased) then default
+                else context(tag.erased).asInstanceOf[B]
+
+            inline def get[A, E <: ContextEffect[A]](tag: Tag[E]): A =
+                getOrElse(tag, bug(s"Missing value for context effect '${tag}'. Values: $context"))
+
+            inline def set[A, E <: ContextEffect[A]](tag: Tag[E], value: A): Context =
+                Context(context.updated(tag.asInstanceOf[Tag[Any]], value.asInstanceOf[AnyRef]))
```
hearnadam, line 59: can you use `tag.erased`?

kyo-prelude/shared/src/main/scala/kyo2/kernel/package.scala
```diff
@@ -0,0 +1,65 @@
...
+        end Context
+
+        object Context:
+            inline def empty: Context = Context(immutable.Map.empty)
```
hearnadam, line 63: Can't this be a `val`?

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
fwbrasil: yeah, I think it should have a similar execution and only move `Thread.currentThread()` from the constructor to a default param.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
fwbrasil: exactly!

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
fwbrasil: I think so since it can only recover thrown exceptions. The type is bound to `Throwable` here and in `Effect.catching`.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
fwbrasil: Interesting, I thought `ScopedValue` was just a scoped wrapper for `ThreadLocal` but it's actually a more efficient mechanism. I'll play with it, if the performance is better we might be able to avoid the optimization to propagate `Safepoint` via implicits.

kyo-prelude/shared/src/main/scala/kyo2/kernel/ContextEffect.scala
fwbrasil: I was a bit lazy here, the `null` is used to check if the the value is in the context or not. I've exposed a `Context.contains` method to avoid it. About `bug`, in `Context` there's no allocation for it because `default` gets inlined in `getOrElse` directly into the `if` condition body.

kyo-prelude/shared/src/main/scala/kyo2/kernel/package.scala
fwbrasil: There shouldn't be a need. The compiler optimizes the final code to `immutable.Map.empty`, which already uses a cached value internally without allocations.

kyo-prelude/shared/src/main/scala/kyo2/kernel/package.scala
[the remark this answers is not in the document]
hearnadam: We can probably add a `private[kyo] def getExact` in `TypeMap` to avoid the duplication. Feel free to followup later.

kyo-prelude/shared/src/main/scala/kyo2/Local.scala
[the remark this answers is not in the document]
hearnadam: I think HashMap avoids this - but we can followup later. It may not be an improvement with the small sizes specializations.

kyo-prelude/shared/src/main/scala/kyo2/kernel/package.scala
hearnadam: I was talking about the `Context.empty` - not the Map. Is there a reason you used `inline def`?

kyo-prelude/shared/src/main/scala/kyo2/Local.scala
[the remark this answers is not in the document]
fwbrasil: Makes sense, I'm starting to think that we need our own low-allocation `Map` impl

kyo-prelude/shared/src/main/scala/kyo2/kernel/package.scala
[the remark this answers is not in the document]
fwbrasil: cool cool, it'll be nice to use `TypeMap` here!

kyo-prelude/shared/src/main/scala/kyo2/Env.scala
```diff
@@ -0,0 +1,55 @@
...
+        Frame
+    ): T < (S & RR) =
+        ContextEffect.handle(erasedTag[R], env, _.union(env))(value).asInstanceOf[T < (S & RR)]
+
+    class UseOps[R >: Nothing](dummy: Unit) extends AnyVal:
+        inline def apply[A, S](inline f: R => A < S)(
+            using tag: Tag[R]
+        ): A < (Env[R] & S) =
+            ContextEffect.suspend(erasedTag[R]) { map =>
+                f(map.get(using tag.erased).asInstanceOf[R])
```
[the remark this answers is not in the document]
hearnadam: IIUC, the result cast and the `.erased` are necessary due to the `TypeMap` tag being erased.

I think we can safely cast the Map instead of the tag + the result:

`f(map.asInstanceOf[TypeMap[R]].get(using tag))`

fwbrasil:
After multiple battles with the compiler, I've finally been able to make `Boundary` work to fork fibers! The match type approach wasn't working properly so I tried a few mechanisms similar to `Has` we have for `Abort` and `Env`, which didn't work, and ended landing on a macro to make the type-level transformation: https://github.com/getkyo/kyo/pull/489/commits/d631a99d1980606ea833f0a11d0f4412ea9a43c7

I think there isn't anything major pending in this PR after this change.

kyo-prelude/shared/src/main/scala/kyo2/kernel/package.scala
fwbrasil: Yeah, it doesn't need to be `inline`. I've fixed it. I'm not sure why I had used `AnyVal` here. I've changed to `opaque type` since `AnyVal` can end up allocating.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
fwbrasil: I took a look at `ScopedValue` and it doesn't seem possible to use it to store the `Safepoint` since its use isn't always scoped. For example, if a computation is executed without any suspensions, it'll get the safepoint from the thread local even if the code isn't within an `.eval` execution. We should review this later, though. We might need a more involved solution when the scheduler's Loom integration is enabled since the thread local increases the memory footprint of virtual threads, and there can be many of them.

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
fwbrasil: @johnhungerford I probably did something wrong in my first try. I was able to migrate `Abort` and `Env` to `Reducible`: https://github.com/getkyo/kyo/pull/489/commits/ff6c55c4f3bfac926ca477bc888bd9c313d06546

@hearnadam Do you mean the low-priority implicit trait? Is it possible to encode it via a constructor?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
```diff
@@ -33,8 +29,30 @@ object Boundary:
...
+            tpe match
+                case AndType(left, right) => flatten(left) ++ flatten(right)
+                case t                    => List(t)
+
+        val s = flatten(TypeRepr.of[S])
+
+        val r = flatten(TypeRepr.of[Ctx]).filter(tpe => !s.exists(_ <:< tpe))
+
+        if !r.forall(tpe => (tpe <:< TypeRepr.of[ContextEffect[?]]) || (tpe =:= TypeRepr.of[Any])) then
+            report.errorAndAbort("")
```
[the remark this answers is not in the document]
fwbrasil: I added a better error message yesterday: https://github.com/getkyo/kyo/pull/489/commits/233735240254ab9dcbaad078310e4d17076ee70b

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
fwbrasil: I see, thanks for the clarification! It's safe to do it even without the bound since the tag will simply mismatch when catching exceptions. That said, maybe we should constrain `Abort` only to exception types. It'd allow us to inject the trace into the stack trace for example

PR #490, opened 2024-06-20
Fix field nullability in Caliban schema instances

ghostdogpr:
I copied the Cats Effect integration but it turns out [there was a bug in it](https://github.com/ghostdogpr/caliban/pull/2296) 😄


kyri-petrou (approved):

fwbrasil:
thanks folks!

PR #493, opened 2024-06-24
add: TypeMap[A].prune[B]

hearnadam:
I colocated the extension methods so only `import kyo.TypeMap` is necessary for consumers.

I plan to add benchmarks comparing `ZEnvironment` next

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/typeMap.scala
+++ b/kyo-core/shared/src/main/scala/kyo/typeMap.scala
@@ -5,48 +5,51 @@ import scala.collection.immutable.HashMap
 
 opaque type TypeMap[+A] = HashMap[Tag[Any], Any]
 
-extension [A](self: TypeMap[A])
-
-    private inline def fatal[T](using t: Tag[T]): Nothing =
-        throw new RuntimeException(s"fatal: kyo.TypeMap of contents [${self.show}] missing value of type: [${t.showTpe}].")
-
-    def get[B >: A](using t: Tag[B]): B =
-        val b = self.getOrElse(t.erased, null)
-        if !isNull(b) then b.asInstanceOf[B]
-        else
-            var sub: B = null.asInstanceOf[B]
-            val it     = self.iterator
-            try // iterator should never throw given type constraint on B
-                while isNull(sub) do
-                    val (tag, item) = it.next()
-                    if tag <:< t then
-                        sub = item.asInstanceOf[B]
-                end while
-            catch
-                case _: NoSuchElementException => fatal
-            end try
-            sub
-        end if
-    end get
-
-    inline def add[B](b: B)(using inline t: Tag[B]): TypeMap[A & B] =
-        self.updated(t.erased, b)
-
-    inline def union[B](that: TypeMap[B]): TypeMap[A & B] =
-        self ++ that
-
-    inline def size: Int        = self.size
-    inline def isEmpty: Boolean = self.isEmpty
-
-    def show: String = self.map { case (tag, value) => s"${tag.showTpe} -> $value" }.toList.sorted.mkString("TypeMap(", ", ", ")")
-
-    private[kyo] inline def tag: Intersection[?] = Intersection(self.keySet.toIndexedSeq)
-
-    private[kyo] inline def <:<[T](tag: Tag[T]): Boolean =
-        self.keySet.exists(_ <:< tag)
-end extension
-
 object TypeMap:
+    extension [A](self: TypeMap[A])
+
+        private inline def fatal[T](using t: Tag[T]): Nothing =
+            throw new RuntimeException(s"fatal: kyo.TypeMap of contents [${self.show}] missing value of type: [${t.showTpe}].")
+
+        def get[B >: A](using t: Tag[B]): B =
+            val b = self.getOrElse(t.erased, null)
+            if !isNull(b) then b.asInstanceOf[B]
+            else
+                var sub: B = null.asInstanceOf[B]
+                val it     = self.iterator
+                try // iterator should never throw given type constraint on B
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -5,48 +5,51 @@ import scala.collection.immutable.HashMap
...
 
-    inline def union[B](that: TypeMap[B]): TypeMap[A & B] =
-        self ++ that
+        inline def union[B](that: TypeMap[B]): TypeMap[A & B] =
+            self ++ that
 
-    inline def size: Int        = self.size
-    inline def isEmpty: Boolean = self.isEmpty
+        def prune[B >: A](using t: Tag[B]): TypeMap[B] =
+            if t =:= Tag[Any] then empty
```
steinybot, line 40: Shouldn't this be?
```suggestion
            if t =:= Tag[Any] then self
```
`prune` is keeping everything that is < B right?

hearnadam: Technically yes - and I am happy to make that change for 'soundness'.

I intentionally took this departure from `ZEnvironment` as I think it's likely more useful.

What can you `.get` from a `TypeMap[Any]`? I suppose you could invoke `.get[Any]` to get the first element? But that does not seem useful nor practical. The widening of the `A` -> `Any` renders the resulting map practically useless.

Am I missing a usecase?

steinybot: I agree that it is not very useful but neither is an empty map so I'm not sure whether there is any benefit there. I would expect `map.get[B] eq map.prune[B].get[B]` for all `B >: A`.

steinybot: Thinking about this some more, does prune even make sense?

If it is keeping everything that is < B then you are only left with a single type and you can only retrieve that first value.

For example. This doesn't work:
```
val e = TypeMap(42, "", true)
val p = e.prune[Int & String]
```

All you can do is:
```
val p1 = e.prune[Int]
```

which I think (apart from the size) is effectively the same as:
```
val p2: TypeMap[Int] = e
```

and:
```
val p3 = TypeMap(e.get[Int])
```

It might make more sense to reverse what `prune` does. Rather than keeping everything that is < B, how about removing everything that is < B?

It would also suit the name better IMO. The metaphor of pruning sounds more like what you are taking off rather than what you are leaving behind.

I'm guessing the point of this is remove unused values so they can be GCed? Otherwise it is redundant since `TypeMap[A]` is already < `TypeMap[A & B]`.

hearnadam: If you check the tests, you can see that
1. Intersection eventually should be supported
2. Pruning to single type may not result in a single value.

kyo-core/shared/src/test/scala/kyoTest/typeMapTest.scala
```diff
@@ -157,14 +158,44 @@ class typeMapTest extends KyoTest:
...
     ".prune" - {
-        "Env[Int & String] -> Env[Int]" in pendingUntilFixed {
-            assertCompiles(
-                """
-                  | val e = TypeMap(42, "")
-                  | val p = e.prune[Int]
-                  | assert(p.size == 1)
-                  |""".stripMargin
-            )
+        "Env[Int] -> Env[Any]" in {
```
fwbrasil, line 161: Can you add type annotations to the values in the tests? It's not clear to me what's expected.

kyo-core/shared/src/main/scala/kyo/typeMap.scala
fwbrasil: I'm not sure if I understood the semantics well but would `filter` be a better name for this new method? Regarding `Any`, maybe we could fail compilation in case it's passed via `scala.compiletime.error`?

kyo-core/shared/src/test/scala/kyoTest/typeMapTest.scala
fwbrasil: If you don't expect the annotations to change the behavior of the API, it could be all tests but just a couple of them should be enough to make it easier to understand the semantics of the API.

kyo-core/shared/src/test/scala/kyoTest/typeMapTest.scala
```diff
@@ -157,14 +158,44 @@ class typeMapTest extends KyoTest:
...
+            val e =
+                TypeMap
+                    .empty
+                    .add(new Throwable)
+                    .add(new Exception)
+                    .add(new RuntimeException)
+                    .add(new NullPointerException)
+                    .add(new ClassCastException)
+            val p = e.prune[RuntimeException]
+            assert(p.size == 3)
```
steinybot, line 191: Is there a way in which it is useful to keep all 3 entries? Once you have gone to `TypeMap[RuntimeException]` you can only ever get some `B` > `RuntimeException` out and that will always be the same value no matter what you choose for `B`. So those other 2 values are sitting there unable to be accessed.

hearnadam: In this case no - these are just types which don't require custom classes that show the behaviors I expect.

In general, Kyo tags will support intersections soon and then it will be useful to retain multiple.

kyo-core/shared/src/main/scala/kyo/typeMap.scala
hearnadam: I was attempting to keep similar naming to ZEnv, though that's not necesssary. I think `prune` is likely the most accurate. I think `.filter` would to support a predicate.

PR #494, opened 2024-06-24
Distinct TypeMap

steinybot:
The discussion in https://github.com/getkyo/kyo/pull/493/files?diff=split&w=1#r1650388812 got me wondering whether it makes sense to have overlapping types and whether it would be possible to keep them distinct.

It seems like it is possible, at least when it come time to adding them, however `TypeMap[A]` is still > `TypeMap[A & B]` so you can have two distinct types `A extends C` and `B extends C` and so you can go from `TypeMap[A & B]` to `TypeMap[C]`. There would now appear to be two overlapping types but I think that is fine since that is what you get from `get[C]` anyway. At least now when you replace/merge you can overwrite the previous values. It might be worth looking at whether `get` can return the last matching type that was added rather than the order being undefined. It's not entirely clear if that is a desirable property or not.

Depends on https://github.com/getkyo/kyo/pull/493 and it'll be interesting to see how the performance compares. It'll obviously be worse but I wonder by how much. More objects should be eligible for collection. It might be a worthwhile tradeoff considering how small these are likely to be in practice.


hearnadam:
Nice experiment! There is definitely some interesting work here.

I don't know if there is value in directly implementing this as it may break common expectations of `Layers` or this data structure in general. Limiting all subtypes from being shared in the map will result in some potentially useful states being impossible to construct.

I believe this PR makes the assumption `TypeMap[Super] =:= TypeMap[Sub & Sub]`

Where do you think this would provide the most value?

hearnadam:
Can you sync with mainline and give the benchmarks a shot? I'm curious to see how this performs.

fwbrasil:
@steinybot could we close this for now? I'd prefer to avoid having PRs open for long, even if a draft.

PR #501, opened 2024-06-26
prelude: make boundary constructor private + add convenience apply

fwbrasil:
@hearnadam addressing your feedback on https://github.com/getkyo/kyo/pull/489

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
@@ -3,7 +3,7 @@ package kyo2.kernel
 import internal.*
 import scala.quoted.*
 
-final class Boundary[Ctx, S](dummy: Unit) extends AnyVal:
+final class Boundary[Ctx, S] private (dummy: Unit) extends AnyVal:
 
     inline def apply[A, B, S2, S3](inline v: A < (Ctx & S & S2))(
         inline f: A < (S & S2) => B < S3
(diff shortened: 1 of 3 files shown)


kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
```diff
@@ -39,6 +39,10 @@ object Boundary:
 
     inline given derive[Ctx, S]: Boundary[Ctx, S] = ${ boundaryImpl[Ctx, S] }
 
+    def apply[Ctx, S](using boundary: Boundary[Ctx, S]): Boundary[Ctx, S] = boundary
+
+    private def create[Ctx, S]: Boundary[Ctx, S] = new Boundary(())
```
fwbrasil, line 44: This is needed because it's not possible to use an `AnyVal` with a private constructor in an `inline` method.

kyo-prelude/shared/src/test/scala/kyo2/kernel/BoundaryTest.scala
```diff
@@ -115,7 +115,7 @@ class BoundaryTest extends Test:
     }
 
     "detects for non-context effects" in {
-        assertDoesNotCompile("summon[Boundary[NotContextEffect, Any]]")
+        assertDoesNotCompile("Boundary[NotContextEffect, Any]")
     }
```
hearnadam, line 119: Check the compiler error here - you may need to pass a value. You could pass a parameter if you don't want to define an actual effect.

hearnadam (approved):

kyo-prelude/shared/src/test/scala/kyo2/kernel/BoundaryTest.scala
fwbrasil: It looks ok:
```
[error] 118 |        Boundary[NotContextEffect, Any]
[error]     |                                       ^
[error]     | Expected context effects. Found: 'BoundaryTest.this.NotContextEffect'
```

PR #502, opened 2024-06-26
prelude: typed failures in Result

fwbrasil:
I'm working on porting `kyo-core` to `kyo-prelude` and part of the change introduces typed error handling for fibers. This PR makes `Result` a low-allocation mix of `Either` and `Try` to use it in fiber completions.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/Result.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Result.scala
@@ -1,6 +1,7 @@
 package kyo2
 
 import Result.*
+import scala.annotation.implicitNotFound
 import scala.util.Try
 import scala.util.control.NonFatal
 
@@ -24,28 +25,27 @@ object Result:
         catch
             case ex: Throwable if NonFatal(ex) => Error(ex)
 
-    def success[E, A](value: A): Result[E, A] = Success(value)
-
-    def error[E, A](error: E): Result[E, A] = Error(error)
-
+    def success[E, A](value: A): Result[E, A]           = Success(value)
+    def error[E, A](error: E): Result[E, A]             = Error(error)
     def panic[E, A](exception: Throwable): Result[E, A] = Panic(exception)
 
-    def fromEither[E, A](either: Either[E, A]): Result[E, A] = either match
-        case Right(value) => Success(value)
-        case Left(error)  => Error(error)
+    def fromEither[E, A](either: Either[E, A]): Result[E, A] =
+        either match
+            case Right(value) => Success(value)
+            case Left(error)  => Error(error)
 
-    opaque type Success[+T] = T | SuccessError[T]
+    opaque type Success[+A] = A | SuccessError[A]
 
     object Success:
 
-        def apply[T](value: T): Success[T] =
+        def apply[A](value: A): Success[A] =
             value match
-                case v: SuccessError[?]     => v.nest.asInstanceOf[Success[T]]
-                case v: Error[T] @unchecked => SuccessError(v)
+                case v: SuccessError[?]     => v.nest.asInstanceOf[Success[A]]
+                case v: Error[A] @unchecked => SuccessError(v)
                 case v                      => v
 
         // TODO avoid Option allocation
-        def unapply[E, T](self: Result[E, T]): Option[T] =
+        def unapply[E, A](self: Result[E, A]): Option[A] =
             self.fold(_ => None)(_ => None)(Some(_))
 
     end Success
@@ -63,7 +63,7 @@ object Result:
 
     case class Panic(exception: Throwable) extends Failure[Nothing]
 
-    extension [E, T](self: Result[E, T])
+    extension [E, A](self: Result[E, A])
 
         def isSuccess: Boolean =
             self match
@@ -76,7 +76,7 @@ object Result:
         def isPanic: Boolean =
             self.isInstanceOf[Panic]
 
-        inline def fold[U](inline ifError: E => U)(inline ifPanic: Throwable => U)(inline ifSuccess: T => U): U =
+        inline def fold[B](inline ifError: E => B)(inline ifPanic: Throwable => B)(inline ifSuccess: A => B): B =
             (self: @unchecked) match
                 case self: Error[E] =>
                     ifError(self.error)
@@ -85,41 +85,41 @@ object Result:
                 case _ =>
                     try
                         self match
-                            case self: SuccessError[T] @unchecked =>
-                                ifSuccess(self.unnest.asInstanceOf[T])
+                            case self: SuccessError[A] @unchecked =>
+                                ifSuccess(self.unnest.asInstanceOf[A])
                             case self =>
-                                ifSuccess(self.asInstanceOf[T])
+                                ifSuccess(self.asInstanceOf[A])
                     catch
                         case ex if NonFatal(ex) => ifPanic(ex)
 
-        def get(using E =:= Nothing): T =
+        def get(using E =:= Nothing): A =
             fold(ex => throw new NoSuchElementException(s"Failure: $ex"))(throw _)(identity)
 
-        inline def getOrElse[U >: T](inline default: => U): U =
+        inline def getOrElse[B >: A](inline default: => B): B =
             fold(_ => default)(_ => default)(identity)
 
-        def orElse[E2, U >: T](alternative: => Result[E2, U]): Result[E | E2, U] =
+        def orElse[E2, B >: A](alternative: => Result[E2, B]): Result[E | E2, B] =
             fold(_ => alternative)(_ => alternative)(Result.success)
 
-        inline def flatMap[E2, U](inline f: T => Result[E2, U]): Result[E | E2, U] =
+        inline def flatMap[E2, B](inline f: A => Result[E2, B]): Result[E | E2, B] =
             (self: @unchecked) match
                 case self: Failure[E] => self
                 case self =>
-                    try f(self.asInstanceOf[Success[T]].get)
+                    try f(self.asInstanceOf[Success[A]].get)
                     catch
                         case ex if NonFatal(ex) =>
                             Panic(ex)
 
-        def flatten[E2, U](using ev: T <:< Result[E2, U]): Result[E | E2, U] =
+        def flatten[E2, B](using ev: A <:< Result[E2, B]): Result[E | E2, B] =
             flatMap(ev)
 
-        inline def map[U](inline f: T => U): Result[E, U] =
+        inline def map[B](inline f: A => B): Result[E, B] =
             flatMap(v => Result.success(f(v)))
 
-        inline def withFilter(inline p: T => Boolean): Result[E | NoSuchElementException, T] =
+        inline def withFilter(inline p: A => Boolean): Result[E | NoSuchElementException, A] =
             filter(p)
 
-        inline def filter(inline p: T => Boolean): Result[E | NoSuchElementException, T] =
+        inline def filter(inline p: A => Boolean): Result[E | NoSuchElementException, A] =
             flatMap { v =>
                 if !p(v) then
                     Error(new NoSuchElementException("Predicate does not hold for " + v))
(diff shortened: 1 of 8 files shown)


kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -4,134 +4,159 @@ import Result.*
...
+    extension [E, T](self: Result[E, T])
+
+        def isSuccess: Boolean =
             self match
-                case _: Failure[?] => false
-                case _             => true
+                case _: Cause[?] => false
+                case _           => true
 
-        inline def isFailure: Boolean = !isSuccess
```
fwbrasil, line 59: A few the methods don't benefit much from inlining since they don't invoke functions so I removed `inline` for them.

fwbrasil:
I've also changed `Abort.run` to return `Result` and catch unexpected exceptions to `Result.panic`.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -4,134 +4,159 @@ import Result.*
...
-                case v: Failure[A] @unchecked => SuccessFailure(v)
+                case v: SuccessFailure[?]     => v.nest.asInstanceOf[Success[T]]
+                case v: Failure[T] @unchecked => SuccessFailure(v)
                 case v                        => v
 
         // TODO avoid Option allocation
-        def unapply[A](self: Result[A]): Option[A] =
-            self.fold(_ => None)(Some(_))
+        def unapply[E, T](self: Result[E, T]): Option[T] =
+            self.fold(_ => None)(_ => None)(Some(_))
```
hearnadam, line 43: I think if you add an `isDefined` method to `Result`, you can directly return `self` in the case of `Success`'s `unapply`, avoiding any allocation

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -4,134 +4,159 @@ import Result.*
...
-                                ifSuccess(self.asInstanceOf[A])
+                                ifSuccess(self.asInstanceOf[T])
                     catch
-                        case ex if NonFatal(ex) => ifFailure(ex)
+                        case ex if NonFatal(ex) => ifPanic(ex)
 
-        inline def get: A =
-            fold(ex => throw ex)(identity)
+        def get(using E =:= Nothing): T =
+            fold(ex => throw new NoSuchElementException(s"Cause: $ex"))(throw _)(identity)
```
hearnadam, line 90: Should we specialize this in the case that `E =:= Throwable`, such that the original exception is re-thrown?

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -4,134 +4,159 @@ import Result.*
...
 
-        inline def getOrElse[B >: A](inline default: => B): B =
-            fold(_ => default)(identity)
+        inline def getOrElse[U >: T](inline default: => U): U =
+            fold(_ => default)(_ => default)(identity)
 
-        def orElse[B >: A](alternative: => Result[B]): Result[B] =
-            fold(_ => alternative)(v => Result.success(v))
+        def orElse[E2, U >: T](alternative: => Result[E2, U]): Result[E | E2, U] =
+            fold(_ => alternative)(_ => alternative)(v => Result.success(v))
```
hearnadam, line 96: minor: I prefer function syntax where possible - hopefully it would desugar the same anyways:
`fold(_ => alternative)(_ => alternative)(Result.success)`

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -4,134 +4,159 @@ import Result.*
...
 
-        inline def map[B](inline f: A => B): Result[B] =
+        inline def map[U](inline f: T => U): Result[E, U] =
             flatMap(v => Result.success(f(v)))
 
-        inline def filter(inline p: A => Boolean): Result[A] =
+        inline def withFilter(inline p: T => Boolean): Result[E | NoSuchElementException, T] =
+            filter(p)
+
+        inline def filter(inline p: T => Boolean): Result[E | NoSuchElementException, T] =
```
hearnadam, line 116: I don't think the `E` is correct here - the Panic will not be visible in the result type.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -4,134 +4,159 @@ import Result.*
...
         // TODO avoid Option allocation
-        def unapply[A](self: Result[A]): Option[A] =
-            self.fold(_ => None)(Some(_))
+        def unapply[E, T](self: Result[E, T]): Option[T] =
+            self.fold(_ => None)(_ => None)(Some(_))
 
     end Success
 
-    case class Failure[+A](exception: Throwable)
+    sealed abstract class Cause[+E]
```
hearnadam, line 47: Perhaps we can keep the `Failure` naming for parity with Success, then have

`Error` and `Panic` as the underlying types extending this.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -4,134 +4,159 @@ import Result.*
...
+                    case _ => self
+            catch
+                case ex: Throwable if NonFatal(ex) =>
+                    Panic(ex)
+
+        def toEither: Either[E | Throwable, T] =
+            fold(Left(_))(Left(_))(Right(_))
+
+        def toTry: Try[T] =
+            fold(e => scala.util.Failure(new NoSuchElementException(s"Cause: $e")))(scala.util.Failure(_))(scala.util.Success(_))
```
hearnadam, line 148: I think we should use Either's implementation for inspiration and require evidence of `E <: Throwable`

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -4,134 +4,159 @@ import Result.*
...
 
 object Result:
     import internal.*
 
-    given [A](using CanEqual[A, A]): CanEqual[Result[A], Result[A]] = CanEqual.derived
+    given [E, A](using CanEqual[A, A]): CanEqual[Result[E, A], Result[E, A]] = CanEqual.derived
+    given [E, A]: CanEqual[Result[E, A], Panic]                              = CanEqual.derived
 
-    def apply[A](expr: => A): Result[A] =
+    def apply[A](expr: => A): Result[Nothing, A] =
```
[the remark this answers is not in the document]
fwbrasil: The idea is that `Throwable` is always possible via `Panic`, similarly to `Try`. I'm not sure tracking it also here helps much.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -4,134 +4,159 @@ import Result.*
...
-    def success[A](value: A): Result[A] = Success(value)
+    def success[E, A](value: A): Result[E, A] = Success(value)
 
-    def failure[A](exception: Throwable): Result[A] = Failure(exception)
+    def failure[E, A](error: E): Result[E, A] = Failure(error)
 
-    def fromEither[A](either: Either[Throwable, A]): Result[A] = either match
+    def panic[E, A](exception: Throwable): Result[E, A] = Panic(exception)
+
+    def fromEither[E, T](either: Either[E, T]): Result[E, T] = either match
```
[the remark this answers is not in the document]
fwbrasil: I sometimes write an entire class and notice I'm using the old naming only before committing 😂

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
fwbrasil: The compiler can't find `isDefined`/`get` defined as extension methods :( We might be able to work it around by returning an `AnyVal` wrapper only to satisfy pattern matching.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
fwbrasil: My reasoning was that `E =:= Nothing` is essentially that: if there's no typed error, `Throwable` is still present via `Panic`. A behavior similar to `Try.get`.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
fwbrasil: I meant to return `Failure`, not `Panic`. I've fixed it

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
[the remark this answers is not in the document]
hearnadam: How about adding an `Result.attempt(_): Result[Throwable, A]` and leave `Result.apply(_): Result[E, A]`?

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -4,134 +4,159 @@ import Result.*
...
+                    case _ => self
+            catch
+                case ex: Throwable if NonFatal(ex) =>
+                    Panic(ex)
+
+        def toEither: Either[E | Throwable, T] =
+            fold(Left(_))(Left(_))(Right(_))
+
+        def toTry: Try[T] =
+            fold(e => scala.util.Failure(new NoSuchElementException(s"Failure: $e")))(scala.util.Failure(_))(scala.util.Success(_))
```
hearnadam, line 148: I would still prefer to require `toTry` to be restricted to `Result[Throwable, A]` to avoid serializing the `e`, but it's not the end of the world.

kyo-prelude/shared/src/test/scala/kyo2/ResultTest.scala
```diff
@@ -1,55 +1,79 @@
...
         }
     }
 
     "get" - {
         "returns the value for Success" in {
-            assert(Success(1).get == 1)
+            assert(Result.success(1).get == 1)
+        }
+        "can't be called for Error" in {
+            assertDoesNotCompile("Result.failure(ex).get")
```
hearnadam, line 61: Function needs updating

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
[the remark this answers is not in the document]
fwbrasil: Good idea! I've added it.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
fwbrasil: That's a good point, I've added an implicit to restrict it.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -1,137 +1,172 @@
...
+            catch
+                case ex: Throwable if NonFatal(ex) =>
+                    Panic(ex)
+
+        def toEither: Either[E | Throwable, A] =
+            fold(Left(_))(Left(_))(Right(_))
+
+        def toTry(using
+            @implicitNotFound("Error type must be a 'Throwable' to invoke 'toTry'. Found: '${E}'")
+            ev: E <:< Throwable
```
hearnadam, line 155: I assume it's friendlier to consumers to use the implicit vs an extension method limited to `Result[Throwable, A]`?

fwbrasil: yeah, I don't think it's possible to use `implicitNotFound` if we restrict it via the extension method declaration

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
hearnadam: https://github.com/zio/zio/blob/c7b4074d1b40727e4cca53d5b8ab3eafd6b9e549/core/shared/src/main/scala/zio/ZIO.scala#L6352-L6361

My suggestion was more so a form of specialization that requires users to be explicit with the behaviors they expect.

`getOrThrow` -> `get` - requiring `E <: Throwable`
`getOrThrowFiberFailure` -> `getOrThrowPanic` - where `E` is anything

PR #503, opened 2024-06-26
chunks: avoid ClassTag cache lookup

fwbrasil:
I've been doing some performance testing as part of the work on the new core design and noticed thread contention when using chunks due to `ClassTag` lookups, which are synchronized by Scala's `ClassTag` cache. We don't really need the `ClassTag` lookup since the goal in chunks is to always use `Any` for internal arrays. I've tried to introduce a benchmark for it but I couldn't reproduce the issue in isolation.

I'm updating both versions in `kyo-core` and `kyo-prelude`.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/chunks.scala
+++ b/kyo-core/shared/src/main/scala/kyo/chunks.scala
@@ -9,6 +9,8 @@ sealed abstract class Chunk[T] derives CanEqual:
 
     import Chunks.internal.*
 
+    private inline given ClassTag[T] = ClassTag.Any.asInstanceOf[ClassTag[T]]
+
     //////////////////
     // O(1) methods //
     //////////////////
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/chunks.scala
```diff
@@ -256,7 +258,7 @@ sealed abstract class Chunk[T] derives CanEqual:
                 else
                     acc
 
-            val unnested = new Array[U](totalSize())(using ClassTag(classOf[Any]))
+            val unnested = new Array[U](totalSize())(using ClassTag.Any.asInstanceOf[ClassTag[U]])
```
hearnadam, line 261: Is there a reason the given doesn't work here? Perhaps it could be named?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/chunks.scala
fwbrasil: Do you mean the local implicit? It's for `A` but we need `U` here. By the way let me fix the generic name 😅

fwbrasil: ah, the prelude version already has it as `B`

PR #504, opened 2024-06-26
prelude: safepoint interceptors

fwbrasil:
I'm working on providing a way to customize `Safepoint` with additional `enter` and `exit` implementations. The mechanism will be used for preemption in fibers but it can be easily extended to provide other features like profiling and debugging. This PR propagates the current value of the computation alongside its frame to `Safepoint.enter` as preparation. This will allow a custom `enter` to also inspect the current value. The main use case would be a `Debug` effect that also presents values that appear during execution.

**Edit**: I've pushed a commit with the new interceptor mechanism. `Safepoint.immediate` will be used by fibers to preempt execution in evaluation loop (suspension is managed by the fiber itself) and `Safepoint.propagating` provides similar functionality but ensuring propagation across suspensions.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
@@ -11,8 +11,11 @@ object Effect:
     def defer[A, S](f: Safepoint ?=> A < S): A < S =
         new KyoDefer[A, S]:
             def frame = summon[Frame]
-            def apply(dummy: Unit, context: Context)(using Safepoint) =
-                f
+            def apply(v: Unit, context: Context)(using Safepoint) =
+                Safepoint.handle(v)(
+                    suspend = this,
+                    continue = f
+                )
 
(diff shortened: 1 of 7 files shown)


kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
```diff
@@ -11,8 +11,11 @@ object Effect:
     def defer[A, S](f: Safepoint ?=> A < S): A < S =
         new KyoDefer[A, S]:
             def frame = summon[Frame]
-            def apply(dummy: Unit, context: Context)(using Safepoint) =
-                f
+            def apply(v: Unit, context: Context)(using Safepoint) =
+                Safepoint.handle(v)(
```
fwbrasil, line 15: `Safepoint.handle` was missing here.

hearnadam (approved):
Very clever mechanism to drop in behavior changes as needed. I like how simple it is as well!

PR #505, opened 2024-06-27
prelude: Coroutine effect

fwbrasil:
I've been trying to find an effect that can leverage the fact that the new kernel makes effect handling aware of the return type of the computation. My first try was a `Shift` effect that encodes delimited continuations (`shift`/`reset`) since it requires the type of the computation being handled to be the same as the shift being handled. Although the effect is [remarkably safer and cleaner](https://gist.github.com/fwbrasil/8f9e7e6204ed88ca1d90014a0ce0f884) with the new core design, I think it's too low-level to introduce in the codebase.

This PR introduces a purely functional `Coroutine` implementation. It provides cooperative multitasking within a single thread, allowing for deterministic and more controlled concurrency. The effect uses suspensions to provide two operations:

1. `Coroutine.pause`: Yields the execution, moving the current computation to the end of the queue.
2. `Coroutine.spawn`: Submits a sub-task to the queue and immediately resumes execution.

The return type of the effect handling is used to maintain the final result of the computation via a `Maybe`, even when other sub-tasks are still being processed. An important limitation of this initial version is that spawn returns `Unit`, so it's not possible to directly use the result of a spawned computation. But, as the tests demonstrate, it's possible to use other effects to propagate values in and out of spawned computations. The use of `Queue` is also not optimized yet.

This new effect **is not meant** as a replacement for `Fiber` but it can be useful to provide computation scheduling behavior without `IO`. Something I worry about is users getting confused with Kyo providing both `Coroutine` and `Fiber`, maybe we could use a more opaque name for this new effect like `Coop` to avoid that?

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/Coroutine.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Coroutine.scala
@@ -1,0 +1,48 @@
+package kyo2
+
+import kernel.*
+import kyo.Tag
+import scala.collection.immutable.Queue
+
+sealed trait Coroutine[+S] extends Effect[Const[Coroutine.Op[S]], Const[Unit]]
+
+object Coroutine:
+
+    sealed trait Op[-S] derives CanEqual
+    object Op:
+        case object Pause extends Op[Any]
+        abstract class Spawn[-S] extends Op[S]:
+            def apply(): Unit < (Coroutine[S] & S)
+    end Op
+
+    private def erasedTag[S] = Tag[Coroutine[Any]].asInstanceOf[Tag[Coroutine[S]]]
+
+    inline def pause: Unit < Coroutine[Any] =
+        Effect.suspend[Any](erasedTag[Any], Op.Pause)
+
+    inline def spawn[A, S](inline v: => Unit < (Coroutine[S] & S)): Unit < Coroutine[S] =
+        Effect.suspend[Any](erasedTag[S], (() => v): Op.Spawn[S])
+
+    def run[A, S](v: A < (Coroutine[S] & S)): A < S =
+        Loop.transform(Queue(v.map(Maybe(_))), Maybe.empty[A]) { (state, result) =>
+            if state.isEmpty then Loop.done(result.get)
+            else
+                val (task, rest) = state.dequeue
+                Effect.handle.state(erasedTag[S], rest, task)(
+                    handle = [C] =>
+                        (input, state, cont) =>
+                            input match
+                                case Op.Pause =>
+                                    (state.enqueue(cont(())), Maybe.empty)
+                                case input: Op.Spawn[S] =>
+                                    (state.enqueue(input().andThen(Maybe.empty[A])), cont(()))
+                            end match
+                    ,
+                    done = (state, result2) =>
+                        Loop.continue(state, result.orElse(result2))
+                )
+            end if
+        }
+    end run
+
+end Coroutine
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo2/Coroutine.scala
```diff
@@ -0,0 +1,48 @@
...
+object Coroutine:
+
+    sealed trait Op[-S] derives CanEqual
+    object Op:
+        case object Pause extends Op[Any]
+        abstract class Spawn[-S] extends Op[S]:
+            def apply(): Unit < (Coroutine[S] & S)
+    end Op
+
+    private def erasedTag[S] = Tag[Coroutine[Any]].asInstanceOf[Tag[Coroutine[S]]]
```
fwbrasil, line 18: This is the only part of the effect that isn't fully type-safe. We need it because the kernel still uses `=:=` due to performance concerns with `<:<` to compare tags. Since the `S` is naturally bound to the pending effects of the computation being handled, it's safe to use an erased tag.

kyo-prelude/shared/src/test/scala/kyo2/CoroutineTest.scala
```diff
@@ -0,0 +1,255 @@
...
+                    yield ()
+                }
+            yield ()
+
+        val result =
+            Coroutine.run(effect)
+                .andThen(Var.get[List[String]])
+                .pipe(Var.run(List.empty[String]))
+                .pipe(Var.run(0))
+                .pipe(Env.run("env"))
```
fwbrasil, line 212: I forgot to mention in the initial prelude PR that I've introduced this `pipe` method. It seems convenient to make effect handling more fluent.

kyo-prelude/shared/src/test/scala/kyo2/CoroutineTest.scala
```diff
@@ -0,0 +1,255 @@
...
+
+        val result = Coroutine.run(effect)
+        assert(result.eval == ())
+        assert(executed == List(5, 0, 1, 2, 3, 1))
+    }
+
+    "error propagation" in run {
+        val effect =
+            for
+                _ <- Coroutine.spawn { throw new RuntimeException("Spawn error") }
```
hearnadam, line 158: I think it would make sense for errors to be captured by `spawn`. Perhaps we should nest `Result/Maybe`?

hearnadam: Thinking about this more, it might make sense to have `spawn`  and `spawnDaemon` - a variant which can proceed when other tasks fail.

fwbrasil: We're limited by `spawn` returning `Unit` in this initial version of the effect. Would `spawnDaemon` swallow exceptions? I explored this approach:

```scala
inline def spawnDaemon[A, S](inline v: => Unit < (Coroutine[S] & S)): Unit < Coroutine[Abort[Throwable] & S] =
    spawn(Abort.catching[Throwable](v))
```

but `run` would still stop the execution on the first `Abort`. I think we'd need a `run` method that collects multiple failures and aggregates them. I'd prefer to keep this initial version of the effect without explicit error handling and review it once we evolve it to allow joining spawned computations since we can do something similar to the new `Fiber` design.

fwbrasil:
I'm closing this for now. I was hoping it'd be useful for something I'm working on but the lack of a way to compose the result of sub-tasks is too limiting. I'm exploring a new `Join` effect with a similar deterministic behavior but with batching like in ZQuery.

PR #509, opened 2024-06-29
prelude: Maybe and Result improvements

fwbrasil:
This PR introduces a few improvements. The main one is allocation-free pattern matching for both `Maybe` and `Result`. The new implicit class `Maybe.Ops` wraps a `Maybe` into an `AnyVal`, which satisfies the compiler to consider the method an extractor.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
@@ -127,10 +127,10 @@ object Maybe:
         def toList: List[A] =
             if isEmpty then List.empty else get :: Nil
 
-        def toRight[X](left: => X): Either[X, A] =
+        inline def toRight[X](inline left: => X): Either[X, A] =
             if isEmpty then Left(left) else Right(get)
 
-        def toLeft[X](right: => X): Either[A, X] =
+        inline def toLeft[X](inline right: => X): Either[A, X] =
             if isEmpty then Right(right) else Left(get)
 
     end extension
--- a/kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
@@ -115,7 +115,7 @@ object Maybe:
                 end if
             else Empty
 
-        def orElse[B >: A](alternative: => Maybe[B]): Maybe[B] =
+        inline def orElse[B >: A](inline alternative: => Maybe[B]): Maybe[B] =
             if isEmpty then alternative else self
 
         def zip[B](that: Maybe[B]): Maybe[(A, B)] =
--- a/kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
@@ -41,7 +41,7 @@ object Maybe:
 
     end Defined
 
-    implicit class Ops[A](maybe: Maybe[A]) extends AnyVal:
+    implicit final class Ops[A](maybe: Maybe[A]) extends AnyVal:
         def isEmpty: Boolean = maybe.isEmpty
         def get: A           = maybe.get
 
@@ -55,9 +55,7 @@ object Maybe:
             else Some(get)
 
         def isEmpty: Boolean =
-            self match
-                case _: Empty => true
-                case _        => false
+            self.isInstanceOf[Empty]
 
         inline def isDefined: Boolean = !isEmpty
 
--- a/kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
@@ -6,8 +6,8 @@ import Maybe.internal.*
 opaque type Maybe[+A] >: (Empty | Defined[A]) = Empty | Defined[A]
 
 object Maybe:
-    given [A, B](using CanEqual[A, B]): CanEqual[Maybe[A], Maybe[B]] = CanEqual.derived
-    given [A]: Conversion[Maybe[A], IterableOnce[A]]                 = _.iterator
+    inline given [A, B](using inline ce: CanEqual[A, B]): CanEqual[Maybe[A], Maybe[B]] = CanEqual.derived
+    given [A]: Conversion[Maybe[A], IterableOnce[A]]                                   = _.iterator
 
     def apply[A](v: A): Maybe[A] =
         if isNull(v) then
@@ -24,7 +24,7 @@ object Maybe:
 
     def empty[A]: Maybe[A] = Empty
 
-    def when[A](cond: Boolean)(v: => A): Maybe[A] =
+    inline def when[A](cond: Boolean)(inline v: => A): Maybe[A] =
         if cond then v else Empty
 
     opaque type Defined[+A] = A | DefinedEmpty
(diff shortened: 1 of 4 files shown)


kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
```diff
@@ -6,8 +6,8 @@ import Maybe.internal.*
 opaque type Maybe[+A] >: (Empty | Defined[A]) = Empty | Defined[A]
 
 object Maybe:
-    given [A, B](using CanEqual[A, B]): CanEqual[Maybe[A], Maybe[B]] = CanEqual.derived
-    given [A]: Conversion[Maybe[A], IterableOnce[A]]                 = _.iterator
+    inline given [A, B](using inline ce: CanEqual[A, B]): CanEqual[Maybe[A], Maybe[B]] = CanEqual.derived
```
fwbrasil, line 9: Inlining to avoid Scala's lazy initialization of `given` values. It's also a "free" inlining since it doesn't increase bytecode size (a method call getting replaced by another)

kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
```diff
@@ -24,7 +24,7 @@ object Maybe:
 
     def empty[A]: Maybe[A] = Empty
 
-    def when[A](cond: Boolean)(v: => A): Maybe[A] =
+    inline def when[A](cond: Boolean)(inline v: => A): Maybe[A] =
```
fwbrasil, line 27: `inline` to avoid the `v` thunk allocation

kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
```diff
@@ -37,21 +37,24 @@ object Maybe:
...
+    implicit class Ops[A](maybe: Maybe[A]) extends AnyVal:
+        def isEmpty: Boolean = maybe.isEmpty
+        def get: A           = maybe.get
+
     sealed abstract class Empty
     case object Empty extends Empty
 
     extension [A](self: Maybe[A])
 
-        inline def toOption: Option[A] =
```
fwbrasil, line 50: I've removed inline from methods that don't take function/by-name parameters.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -10,29 +10,30 @@ opaque type Result[+E, +A] >: (Success[A] | Failure[E]) = Success[A] | Failure[E
...
+    inline given [E, A](using inline ce: CanEqual[A, A]): CanEqual[Result[E, A], Result[E, A]] = CanEqual.derived
+    given [E, A]: CanEqual[Result[E, A], Panic]                                                = CanEqual.derived
 
-    def apply[A](expr: => A): Result[Nothing, A] =
+    inline def apply[A](expr: => A): Result[Nothing, A] =
         try
             Success(expr)
         catch
-            case ex: Throwable if NonFatal(ex) => Panic(ex)
+            case ex => Panic(ex)
```
fwbrasil, line 20: The `NonFatal` check is now done by `Panic.apply`. This reduces the size of the inlined bytecode.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -77,23 +86,26 @@ object Result:
             self.isInstanceOf[Panic]
 
         inline def fold[B](inline ifError: E => B)(inline ifPanic: Throwable => B)(inline ifSuccess: A => B): B =
```
fwbrasil, line 88: Reduces the inlined bytecode size by moving some of the logic to `get`. It'll now throw in case of `Panic` but, since it won't produce a new stack trace, there isn't a performance penalty.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -10,29 +10,30 @@ opaque type Result[+E, +A] >: (Success[A] | Failure[E]) = Success[A] | Failure[E
...
+    inline def panic[E, A](inline exception: Throwable): Result[E, A] = Panic(exception)
 
     def fromEither[E, A](either: Either[E, A]): Result[E, A] =
-        either match
-            case Right(value) => Success(value)
-            case Left(error)  => Error(error)
+        either.fold(error, success)
+
+    def fromTry[A](t: Try[A]): Result[Throwable, A] =
+        t.fold(error, success)
```
fwbrasil, line 36: New `Result.fromTry` method.

kyo-prelude/shared/src/test/scala/kyo2/ResultTest.scala
```diff
@@ -27,6 +27,43 @@ class ResultTest extends Test:
...
+
+        "should return Error for failed Try" in {
+            val exception = new RuntimeException("Test exception")
+            val tryValue  = scala.util.Try(throw exception)
+            val result    = Result.fromTry(tryValue)
+            assert(result == Result.error(exception))
+        }
+    }
+
+    "fromEither" - {
```
fwbrasil, line 45: These tests were missing.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -76,39 +88,40 @@ object Result:
         def isPanic: Boolean =
             self.isInstanceOf[Panic]
 
-        inline def fold[B](inline ifError: E => B)(inline ifPanic: Throwable => B)(inline ifSuccess: A => B): B =
-            (self: @unchecked) match
-                case self: Error[E] =>
-                    ifError(self.error)
-                case self: Panic =>
-                    ifPanic(self.exception)
+        inline def fold[B](inline ifFailure: Failure[E] => B)(inline ifSuccess: A => B): B =
```
fwbrasil, line 91: I've just pushed an additional commit simplifying `fold` to take two functions. Other methods like `recover` use a similar approach.

kyo-prelude/shared/src/main/scala/kyo2/Maybe.scala
```diff
@@ -37,21 +37,24 @@ object Maybe:
...
                 case v               => v
 
-        // TODO avoid allocation
-        def unapply[A](opt: Maybe[A]): Option[A] =
-            opt.toOption
+        def unapply[A](opt: Maybe[A]): Maybe.Ops[A] = opt
+
     end Defined
 
+    implicit class Ops[A](maybe: Maybe[A]) extends AnyVal:
```
hearnadam, line 44: Can be `final`

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -10,29 +10,30 @@ opaque type Result[+E, +A] >: (Success[A] | Failure[E]) = Success[A] | Failure[E
...
         catch
-            case ex: Throwable if NonFatal(ex) => Error(ex)
+            case ex if NonFatal(ex) => Error(ex)
 
-    def success[E, A](value: A): Result[E, A]           = Success(value)
-    def error[E, A](error: E): Result[E, A]             = Error(error)
-    def panic[E, A](exception: Throwable): Result[E, A] = Panic(exception)
+    inline def success[E, A](inline value: A): Result[E, A]           = Success(value)
+    inline def error[E, A](inline error: E): Result[E, A]             = Error(error)
+    inline def panic[E, A](inline exception: Throwable): Result[E, A] = Panic(exception)
```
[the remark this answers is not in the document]
fwbrasil: These are meant to help Scala's type inference. If I return `Failure`, this kind of use fails to compile:

```scala
[error] -- [E008] Not Found Error: /Users/fwbrasil/workspace/kyo/kyo-prelude/shared/src/test/scala/kyo2/ResultTest.scala:148:77
[error] 148 |            assert(Result.error[String, Int]("error").flatMap(x => Success(x + 1)) == Error("error"))
[error]     |                                                                           ^^^
[error]     |value + is not a member of Any, but could be made available as an extension method.
[error]     |
[error]     |The following import might fix the problem:
[error]     |
[error]     |  import org.scalactic.Prettifier.default
```

hearnadam (approved):

PR #510, opened 2024-06-29
prelude: Var.run improvements

fwbrasil:
Simplify the `Var.run` method to avoid the `Ops` class. The explicit type parameter isn't necessary since the `state` param already bounds the `V` type. I've also added a `Var.runTuple` method, which returns the value of the var with the result of the computation.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/Var.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Var.scala
@@ -26,22 +26,27 @@ object Var:
     inline def update[V](inline f: V => V)(using inline tag: Tag[Var[V]]): Unit < Var[V] =
         Effect.suspend[Unit](tag, (v => f(v)): Update[V])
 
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo2/Var.scala
```diff
@@ -26,22 +26,27 @@ object Var:
...
-                                (state, cont(state))
-                            case input: Set[?] =>
-                                (input(), cont(()))
-                            case input: Update[?] =>
-                                (input(state), cont(()))
-            }
-    end RunOps
-
-    inline def run[V >: Nothing]: RunOps[V] = RunOps(())
+    private inline def runWith[V, A, S, B, S2](state: V)(v: A < (Var[V] & S))(
```
fwbrasil, line 29: I had initially left this method `public` but it doesn't seem very useful, so I made it `private` to reduce the API surface.

hearnadam (approved):

PR #512, opened 2024-07-01
prelude: Abort/Maybe fixes and improvements

fwbrasil:
I noticed a few issues while porting `kyo-core` to `kyo-prelude`. More background in comments.

### DIFF
--- a/kyo-prelude/jvm/src/test/scala/kyo2/kernel/BytecodeTest.scala
+++ b/kyo-prelude/jvm/src/test/scala/kyo2/kernel/BytecodeTest.scala
@@ -37,7 +37,7 @@ class BytecodeTest extends Test:
 
     "handle" in {
         val map = methodBytecodeSize[TestHandle]
-        assert(map == Map("test" -> 50, "anonfun" -> 17, "handleLoop" -> 273))
+        assert(map == Map("test" -> 30, "anonfun" -> 15, "handleLoop" -> 266))
     }
 
     def methodBytecodeSize[T](using ct: ClassTag[T]): Map[String, Int] =
--- a/kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
@@ -5,51 +5,50 @@ import kernel.Const
 import kernel.Effect
 import kernel.Frame
 import kyo.Tag
+import kyo2.Result.*
 import kyo2.kernel.Reducible
 import scala.annotation.targetName
 import scala.reflect.ClassTag
-import scala.util.Failure
-import scala.util.Success
-import scala.util.Try
 
-sealed trait Abort[-E] extends Effect[Const[E], Const[Unit]]
+sealed trait Abort[-E] extends Effect[Const[Failure[E]], Const[Unit]]
 
 object Abort:
 
     given eliminateAbort: Reducible.Eliminable[Abort[Nothing]] with {}
     private inline def erasedTag[E]: Tag[Abort[E]] = Tag[Abort[Any]].asInstanceOf[Tag[Abort[E]]]
 
-    inline def fail[E](inline value: E): Nothing < Abort[E] =
-        Effect.suspendMap[Any](erasedTag[E], value)(_ => ???)
+    inline def error[E](inline value: E): Nothing < Abort[E] =
+        Effect.suspendMap[Any](erasedTag[E], Error(value))(_ => ???)
 
-    inline def when[E](b: Boolean)(inline value: E)(
-        using inline tag: Tag[Abort[E]]
-    ): Unit < Abort[E] =
-        if b then fail(value)
+    inline def panic[E](ex: Throwable): Nothing < Abort[E] =
+        Effect.suspendMap[Any](erasedTag[E], Panic(ex))(_ => ???)
+
+    inline def when[E](b: Boolean)(inline value: => E): Unit < Abort[E] =
+        if b then error(value)
         else ()
 
     final class GetOps[E >: Nothing](dummy: Unit) extends AnyVal:
         inline def apply[A](either: Either[E, A]): A < Abort[E] =
             either match
                 case Right(value) => value
-                case Left(value)  => fail(value)
+                case Left(value)  => error(value)
 
-        inline def apply[A](opt: Option[A]): A < Abort[None.type] =
+        inline def apply[A](opt: Option[A]): A < Abort[Maybe.Empty] =
             opt match
-                case None    => fail(None)
+                case None    => error(Maybe.Empty)
                 case Some(v) => v
 
-        inline def apply[A](e: Try[A]): A < Abort[Throwable] =
+        inline def apply[A](e: scala.util.Try[A]): A < Abort[Throwable] =
             e match
-                case Success(t) => t
-                case Failure(v) => fail(v)
+                case scala.util.Success(t) => t
+                case scala.util.Failure(v) => error(v)
 
         inline def apply[E, A](r: Result[E, A]): A < Abort[E | Throwable] =
-            r.fold(e => fail(e.getFailure))(identity)
+            r.fold(e => error(e.getFailure))(identity)
 
         @targetName("maybe")
         inline def apply[A](m: Maybe[A]): A < Abort[Maybe.Empty] =
-            m.fold(fail(Maybe.Empty))(identity)
+            m.fold(error(Maybe.Empty))(identity)
     end GetOps
 
     inline def get[E >: Nothing]: GetOps[E] = GetOps(())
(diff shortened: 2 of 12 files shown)


kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -5,51 +5,50 @@ import kernel.Const
...
+sealed trait Abort[-E] extends Effect[Const[Failure[E]], Const[Unit]]
 
 object Abort:
 
     given eliminateAbort: Reducible.Eliminable[Abort[Nothing]] with {}
     private inline def erasedTag[E]: Tag[Abort[E]] = Tag[Abort[Any]].asInstanceOf[Tag[Abort[E]]]
 
-    inline def fail[E](inline value: E): Nothing < Abort[E] =
-        Effect.suspendMap[Any](erasedTag[E], value)(_ => ???)
+    inline def error[E](inline value: E): Nothing < Abort[E] =
```
fwbrasil, line 20: I'm renaming this method to follow the naming in `Result`. `Fiber` will also have an `error` method.

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -5,51 +5,50 @@ import kernel.Const
...
+        Effect.suspendMap[Any](erasedTag[E], Error(value))(_ => ???)
 
-    inline def when[E](b: Boolean)(inline value: E)(
-        using inline tag: Tag[Abort[E]]
-    ): Unit < Abort[E] =
-        if b then fail(value)
+    inline def panic[E](ex: Throwable): Nothing < Abort[E] =
+        Effect.suspendMap[Any](erasedTag[E], Panic(ex))(_ => ???)
+
+    inline def when[E](b: Boolean)(inline value: => E): Unit < Abort[E] =
```
fwbrasil, line 26: This method was missing the `=>` for `value`. It doesn't matter because the method is inlined and `value` gets spliced where it's used but it could be a problem if we remove `inline`.

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -5,51 +5,50 @@ import kernel.Const
...
 
     final class GetOps[E >: Nothing](dummy: Unit) extends AnyVal:
         inline def apply[A](either: Either[E, A]): A < Abort[E] =
             either match
                 case Right(value) => value
-                case Left(value)  => fail(value)
+                case Left(value)  => error(value)
 
-        inline def apply[A](opt: Option[A]): A < Abort[None.type] =
+        inline def apply[A](opt: Option[A]): A < Abort[Maybe.Empty] =
```
fwbrasil, line 36: Even if the value is an `Option`, this method new returns an `Abort[Maybe.Empty]` so we can avoid having both `Abort[None.type]` and `Abort[Maybe.Empty]` in the pending set.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
```diff
@@ -88,6 +89,23 @@ object Result:
         def isPanic: Boolean =
             self.isInstanceOf[Panic]
 
+        def value: Maybe[A] =
```
fwbrasil, line 92: New convenience methods. It's great that we can provide this kind of API without worrying about allocations!

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
```diff
@@ -60,11 +60,11 @@ object Effect:
         )(
             inline handle: Safepoint ?=> [C] => (I[C], O[C] => A < (E & S & S2)) => A < (E & S & S2),
             inline done: A => B < S3 = (v: A) => v,
-            accept: [C] => (I[C], Tag[Any]) => Boolean = [C] => (v: I[C], tag2: Tag[Any]) => tag =:= tag2
+            inline accept: [C] => I[C] => Boolean = [C] => (v: I[C]) => true
```
fwbrasil, line 63: There was a bug in `Result.run` because the provided `accept` function was ignoring the `Tag`, which was matching inputs of other effects. I've changed `Effect.handle` to always check the tag and allow `accept` to check only the input.

kyo-prelude/shared/src/test/scala/kyo2/AbortTest.scala
```diff
@@ -67,10 +69,8 @@ class AbortsTest extends Test:
             }
             "in handle" in {
                 val effect: Int < Abort[String | Boolean] =
-                    kyo2.Abort.fail("failure")
-                val handled: Result[String | Boolean, Int] =
-                    kyo2.Abort.run[String | Boolean](effect).eval
-                assert(handled == Result.error("failure"))
+                    kyo2.Abort.error("failure")
+                assertDoesNotCompile("kyo2.Abort.run[String | Boolean](effect)")
```
fwbrasil, line 73: This scenario wasn't correct and I've added a `Tag` to make it fail at compile time. The `ClassTag` was inferred to `Any`, effectively handling any `Abort` suspensions.

kyo-prelude/shared/src/test/scala/kyo2/AbortTest.scala
```diff
@@ -316,8 +316,8 @@ class AbortsTest extends Test:
                     val a = kyo2.Abort.run[Distinct1](distinct)
                     val b = kyo2.Abort.run[Distinct2](a).eval
                     assert(b == Result.success(Result.error(d1)))
-
-                    val c = kyo2.Abort.run(distinct).eval
+                    assertDoesNotCompile("kyo2.Abort.run[Distinct1 | Distinct2](distinct)")
```
fwbrasil, line 319: Another scenario that was incorrect. The `ClassTag` was getting widened to `Throwable`, which would catch aborts not specified in the type union.

kyo-prelude/jvm/src/test/scala/kyo2/kernel/BytecodeTest.scala
```diff
@@ -37,7 +37,7 @@ class BytecodeTest extends Test:
 
     "handle" in {
         val map = methodBytecodeSize[TestHandle]
-        assert(map == Map("test" -> 50, "anonfun" -> 17, "handleLoop" -> 273))
+        assert(map == Map("test" -> 30, "anonfun" -> 15, "handleLoop" -> 266))
```
fwbrasil, line 40: Reduced size because `accept` in `Effect.handle` is now `inline`. I'm not sure why I didn't do this initially, it might be an oversight.

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -58,21 +57,25 @@ object Abort:
         def apply[A, S, ES, ER](v: => A < (Abort[E | ER] & S))(
             using
             ct: ClassTag[E],
+            tag: Tag[E], // TODO Used only to ensure E isn't a type union. There should be a more lightweight solution for this.
```
fwbrasil, line 60: @hearnadam I'm using `Tag` here only as a way to deny the use of type unions for `E`. We should have a proper solution for this as we discussed in private.

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
hearnadam: How would you feel about renaming the method `error` -> `fail` in `Result`? the underlying `Error` type can stay the same.

using the 'verb' for the constructor method:
- `success` -> `succeed`
- `error` -> `fail`
- `panic` -> `panic`

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -5,51 +5,50 @@ import kernel.Const
...
-        Effect.suspendMap[Any](erasedTag[E], value)(_ => ???)
+    inline def error[E](inline value: E): Nothing < Abort[E] =
+        Effect.suspendMap[Any](erasedTag[E], Error(value))(_ => ???)
 
-    inline def when[E](b: Boolean)(inline value: E)(
-        using inline tag: Tag[Abort[E]]
-    ): Unit < Abort[E] =
-        if b then fail(value)
+    inline def panic[E](ex: Throwable): Nothing < Abort[E] =
+        Effect.suspendMap[Any](erasedTag[E], Panic(ex))(_ => ???)
```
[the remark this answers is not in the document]
hearnadam: Nice!

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
hearnadam: We can explore a more unified solution for this sort of problem once intersection/union types are supported by a single String tag.

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
hearnadam: Can these be supported by `unapply`? Or is there a limitation?

I don't see an `unapply` for `Panic`

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
fwbrasil: I think having different names for the type `Error` and the correspondent constructor `fail` is less intuitive. Another suggestion:

```scala
opaque type Result[+E, +A] >: (Success[A] | Error[E]) = Success[A] | Error[E]

sealed abstract class Error[+E]
case class Fail[+E](error: E) extends Error[E] // or Failure?
case class Panic(exception: Throwable) extends Error[Nothing]

// Other types like Abort and Fiber would also provide methods with the same name:
Result.success(...)
Result.fail(...)
Result.panic(...)
```

kyo-prelude/shared/src/main/scala/kyo2/Result.scala
fwbrasil: Yes, it's still possible to use the extractors. `Panic` is already has an extractor since it's a `case class`. We have a custom extractor for `Error` to help the compiler infer the `E` type correctly but `Panic` doesn't need it since it extracts a `Throwable` instead of a parametrized value like in `Error`.

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
fwbrasil: Btw, what do you think about renaming `Success` to `Value`? It's shorter and matches `result.value: Maybe[A]`. We could also rename `result.value` to `result.success` but I think it's less intuitive.

hearnadam: I like your first suggestion. In my opinion, `Value` loses the semantic meaning of a computation with the potential for failure. I think it is also more likely to feel familiar to users of 
`Try`

fwbrasil: Naming updated

hearnadam (approved):

PR #513, opened 2024-07-01
prelude: Boundary for sequences

fwbrasil:
I'm porting fibers to the new kernel. Methods like `parallel` have to use `Boundary` but calling `Boundary.apply` for each element would introduce overhead. This PR introduces an overloaded `Boundary.apply` that handles a sequence of computations at once.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
@@ -5,34 +5,42 @@ import scala.quoted.*
 
 final class Boundary[Ctx, S] private (dummy: Unit) extends AnyVal:
 
-    inline def apply[A, B, S2, S3](inline v: A < (Ctx & S & S2))(
+    inline def apply[A, B, S2, S3](inline v: => A < (Ctx & S & S2))(
         inline f: A < (S & S2) => B < S3
     )(using inline _frame: Frame): B < (Ctx & S3) =
         new KyoDefer[B, Ctx & S3]:
             def frame = _frame
             def apply(dummy: Unit, context: Context)(using safepoint: Safepoint) =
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
```diff
@@ -5,34 +5,42 @@ import scala.quoted.*
 
 final class Boundary[Ctx, S] private (dummy: Unit) extends AnyVal:
 
-    inline def apply[A, B, S2, S3](inline v: A < (Ctx & S & S2))(
+    inline def apply[A, B, S2, S3](inline v: => A < (Ctx & S & S2))(
```
fwbrasil, line 8: I noticed `=>` was missing here.

kyo-prelude/shared/src/test/scala/kyo2/kernel/BoundaryTest.scala
```diff
@@ -221,4 +221,78 @@ class BoundaryTest extends Test:
...
+            }
+            val handled = ContextEffect.handle(Tag[TestEffect], 0, _ => 0)(result)
+            assert(handled.eval == 42)
+        }
+
+        "sequence of pure values" in {
+            val seq = Seq(1, 2, 3)
+            val result = Boundary[TestEffect, Any](seq.map(x => x: Int < TestEffect)) { isolatedSeq =>
+                assert(isolatedSeq.size == 3)
+                isolatedSeq.map(_.eval).sum
```
hearnadam, line 239: Should we add a test for maintaining the order of the seq?

fwbrasil: added!

hearnadam (approved):

PR #514, opened 2024-07-01
prelude: introduce Abort.error + test improvements

fwbrasil:


kyo-prelude/shared/src/test/scala/kyo2/AbortTest.scala
```diff
@@ -11,71 +11,71 @@ class AbortsTest extends Test:
     "pure" - {
         "handle" in {
             assert(
-                kyo2.Abort.run[Ex1](kyo2.Abort.get[Ex1](Right(1))).eval ==
+                Abort.run[Ex1](Abort.get[Ex1](Right(1))).eval ==
```
fwbrasil, line 14: I think automated refactoring had unnecessarily introduced the package name.

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -17,11 +17,12 @@ object Abort:
...
 
-    inline def fail[E](inline value: E): Nothing < Abort[E] =
-        Effect.suspendMap[Any](erasedTag[E], Fail(value))(_ => ???)
+    inline def fail[E](inline value: E): Nothing < Abort[E] = error(Fail(value))
 
-    inline def panic[E](ex: Throwable): Nothing < Abort[E] =
-        Effect.suspendMap[Any](erasedTag[E], Panic(ex))(_ => ???)
+    inline def panic[E](ex: Throwable): Nothing < Abort[E] = error(Panic(ex))
+
+    inline def error[E](e: Error[E]): Nothing < Abort[E] =
```
hearnadam, line 24: Shouldn't `e` be inlined here?

Do we want to make this package private? It should be stable, but I prefer to keep the public API small.

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
fwbrasil: good catch! I've added the missing inline in `panic` as well

fwbrasil: > Do we want to make this package private? It should be stable, but I prefer to keep the public API small.

Makes sense, I've made it `private[kyo]`.

hearnadam (approved):

PR #515, opened 2024-07-02
prelude: tracing fixes and improvements

fwbrasil:
I had thought that we wouldn't need to pass `Frame` explicit in methods marked as `inline` but that's not the the case. Instead of using the frame from the outer call, the compiler summons a new one. This PR introduces the `Frame` implicit to inline methods, reduces the size of the frame payload by inferring the short snippet from the expanded one, and introduces a few other improvements (some unrelated to tracing)

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/Env.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Env.scala
@@ -12,7 +12,7 @@ object Env:
     given eliminateEnv: Reducible.Eliminable[Env[Any]] with {}
     private inline def erasedTag[R] = Tag[Env[Any]].asInstanceOf[Tag[Env[R]]]
 
-    inline def get[R](using inline tag: Tag[R]): R < Env[R] =
+    inline def get[R](using inline tag: Tag[R])(using inline frame: Frame): R < Env[R] =
         use[R](identity)
 
     def run[R >: Nothing: Tag, T, S, VR](env: R)(v: T < (Env[R & VR] & S))(
@@ -33,7 +33,7 @@ object Env:
         inline def apply[A, S](inline f: R => A < S)(
             using tag: Tag[R]
         ): A < (Env[R] & S) =
-            ContextEffect.suspend(erasedTag[R]) { map =>
+            ContextEffect.suspendMap(erasedTag[R]) { map =>
                 f(map.asInstanceOf[TypeMap[R]].get(using tag))
             }
     end UseOps
--- a/kyo-prelude/shared/src/main/scala/kyo2/Local.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Local.scala
@@ -9,14 +9,12 @@ abstract class Local[T]:
 
     val default: T
 
-    private val _get: T < Any =
-        ContextEffect.suspend(Tag[Local.State], Map.empty)(_.getOrElse(this, default).asInstanceOf[T])
-
-    def get: T < Any = _get
+    def get(using Frame): T < Any =
+        ContextEffect.suspendMap(Tag[Local.State], Map.empty)(_.getOrElse(this, default).asInstanceOf[T])
 
     // TODO Compilation error if inlined
     def use[U, S](f: T => U < S)(using Frame): U < S =
-        ContextEffect.suspend(Tag[Local.State], Map.empty)(map => f(map.getOrElse(this, default).asInstanceOf[T]))
+        ContextEffect.suspendMap(Tag[Local.State], Map.empty)(map => f(map.getOrElse(this, default).asInstanceOf[T]))
 
     def let[U, S](value: T)(v: U < S)(using Frame): U < S =
         ContextEffect.handle(Tag[Local.State], Map(this -> value), _.updated(this, value.asInstanceOf[AnyRef]))(v)
--- a/kyo-prelude/shared/src/main/scala/kyo2/internal/BaseKyoTest.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/internal/BaseKyoTest.scala
@@ -39,6 +39,15 @@ trait BaseKyoTest[S]:
     given eitherCanEqual[T, U]: CanEqual[Either[T, U], Either[T, U]] = CanEqual.derived
     given throwableCanEqual: CanEqual[Throwable, Throwable]          = CanEqual.derived
 
+    def retry[S](f: => Boolean < S): Boolean < S =
+        def loop(): Boolean < S =
+            f.map {
+                case true  => true
+                case false => loop()
+            }
+        loop()
+    end retry
+
     def timeout =
         if Platform.isDebugEnabled then
             Duration.Infinity
--- a/kyo-prelude/shared/src/main/scala/kyo2/internal/BaseKyoTest.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/internal/BaseKyoTest.scala
@@ -39,14 +39,14 @@ trait BaseKyoTest[S]:
     given eitherCanEqual[T, U]: CanEqual[Either[T, U], Either[T, U]] = CanEqual.derived
     given throwableCanEqual: CanEqual[Throwable, Throwable]          = CanEqual.derived
 
-    def retry[S](f: => Boolean < S): Boolean < S =
+    def untilTrue[S](f: => Boolean < S): Boolean < S =
         def loop(): Boolean < S =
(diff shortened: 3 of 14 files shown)


kyo-prelude/shared/src/main/scala/kyo2/Env.scala
```diff
@@ -33,7 +33,7 @@ object Env:
         inline def apply[A, S](inline f: R => A < S)(
             using tag: Tag[R]
         ): A < (Env[R] & S) =
-            ContextEffect.suspend(erasedTag[R]) { map =>
+            ContextEffect.suspendMap(erasedTag[R]) { map =>
```
fwbrasil, line 36: A minor unrelated refactoring: I've changed `ContextEffect` to use `suspendMap` as the name for the methods that suspend with a continuation. It follows the method naming in `Effect`.

kyo-prelude/shared/src/main/scala/kyo2/Local.scala
```diff
@@ -9,14 +9,12 @@ abstract class Local[T]:
 
     val default: T
 
-    private val _get: T < Any =
```
fwbrasil, line 12: The idea here was to cache the `get` operation to avoid allocations but I think it's best to ensure the tracing produces correct information. We can revisit later in case we optimize this code.

kyo-prelude/shared/src/main/scala/kyo2/internal/BaseKyoTest.scala
```diff
@@ -39,6 +39,15 @@ trait BaseKyoTest[S]:
     given eitherCanEqual[T, U]: CanEqual[Either[T, U], Either[T, U]] = CanEqual.derived
     given throwableCanEqual: CanEqual[Throwable, Throwable]          = CanEqual.derived
 
+    def retry[S](f: => Boolean < S): Boolean < S =
```
fwbrasil, line 42: minor: ported this method from `KyoTest` in `kyo-core`

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
```diff
@@ -209,10 +209,11 @@ object Effect:
                         val tag   = kyo.tag
                         val input = kyo.input
                         def frame = _frame
-                        def apply(v: OX[Any], context: Context)(using Safepoint) =
+                        def apply(v: OX[Any], context: Context)(using safepoint: Safepoint) =
                             try catchingLoop(kyo(v, context))
                             catch
                                 case ex: Throwable if (NonFatal(ex) && pf.isDefinedAt(ex)) =>
+                                    Safepoint.insertTrace(ex)
```
fwbrasil, line 216: It isn't enough to handle exceptions only in `Safepoint.eval` since exceptions can be caught before reaching `eval`. All APIs that provide exception handling have to use `Effect.catching` (including `Abort`) so we can introduce the trace here as well.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Frame.scala
```diff
@@ -7,7 +7,7 @@ opaque type Frame <: AnyRef = String
 
 object Frame:
 
-    private val snippetShortMaxChars = 25
```
fwbrasil, line 10: This change makes the frame payload have only the long-form version of the snippet and then the short version is inferred when parsing a frame.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
```diff
@@ -120,25 +121,32 @@ object Safepoint:
...
+                }.reverse.map { frame =>
+                    StackTraceElement(
+                        frame.snippetShort.reverse.padTo(toPad, ' ').reverse + " @ " + frame.declaringClass,
+                        frame.methodName,
+                        frame.position.fileName,
+                        frame.position.lineNumber
+                    )
+                }.reverse
+            val prefix = cause.getStackTrace.takeWhile(e =>
+                e.getFileName() != elements(0).getFileName() || e.getLineNumber != elements(0).getLineNumber()
```
fwbrasil, line 144: I noticed this edge case with fibers: when a fiber starts executing and fails before reaching a transformation that pushes a frame, `elements` is empty. I've added a guard `size > 0` to fix it.

kyo-prelude/shared/src/main/scala/kyo2/internal/BaseKyoTest.scala
hearnadam, line 42: I would probably rename this method `untilTrue` or `retryBoolean`, but doesn't matter too much as it's for testing.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
hearnadam: Doesn't it make more sense to capture the trace only if the partial function is not defined?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
fwbrasil: When the partial function is defined, the exception "escapes" the scope of the kernel evaluation (typically into a `Result` via `Abort`) so it's essentially the last opportunity to inject the trace. In case the partial function is undefined, the exception still gets the trace included either by an outer `Effect.catching` call or `<.eval`.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
```diff
@@ -120,25 +121,33 @@ object Safepoint:
...
-        val trace = copyTrace(self.trace, self.traceIdx).filter(_ != null).map(_.parse)
-        val toPad = trace.map(_.snippetShort.size).max + 1
-        val elements = trace.map { frame =>
-            StackTraceElement(
-                frame.snippetShort.reverse.padTo(toPad, ' ').reverse + " @ " + frame.declaringClass,
-                frame.methodName,
-                frame.position.fileName,
-                frame.position.lineNumber
+    // TODO compiler crash if private[kyo2]
+    def insertTrace(cause: Throwable)(using self: Safepoint): Unit =
```
fwbrasil, line 125: The compiler ends up generating invalid bytecode with this method marked as `private[kyo]` 😱 https://github.com/getkyo/kyo/actions/runs/9764789434/job/26955864828?pr=515#step:4:1988

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
hearnadam: Ah, that makes more sense.

kyo-prelude/shared/src/main/scala/kyo2/internal/BaseKyoTest.scala
```diff
@@ -39,6 +39,15 @@ trait BaseKyoTest[S]:
     given eitherCanEqual[T, U]: CanEqual[Either[T, U], Either[T, U]] = CanEqual.derived
     given throwableCanEqual: CanEqual[Throwable, Throwable]          = CanEqual.derived
 
+    def untilTrue[S](f: => Boolean < S): Boolean < S =
```
hearnadam, line 42: Are you sure this should be by name?

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
hearnadam: @fwbrasil perhaps we can mark all methods that are non-private (where they could/should be, but the compiler won't allow it) `@deprecated` with the reason `kyo.internal`. Then we should be able to silence that warning in our own code.

kyo-prelude/shared/src/main/scala/kyo2/internal/BaseKyoTest.scala
fwbrasil: Yes, some tests perform side effects in the thunk

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
fwbrasil: Interesting! We have some other APIs that should also be private. For example, commands in effects like `Var` are public only because the effect type is public.

PR #516, opened 2024-07-02
prelude: add `evalNow` and `evalPartial` methods to the pending type

fwbrasil:
I'm using both methods to implement fibers based on `kyo-prelude`. 

The `evalNow` method inspects if the computation has no pending effects to return its value if possible (via `Maybe`). In fibers, this method is used to avoid forking computations that are already complete. I'm wasn't sure about keeping this method public since it's low-level but it seems useful to allow users and library authors to implement logic that avoids overhead in case the computation is already complete.

The `evalPartial` is used from the `IOTask` effect handling to install the interceptor and stop evaluation once the interceptor signals to stop (preemption in fibers).

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/kernel/Pending.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/kernel/Pending.scala
@@ -2,6 +2,7 @@ package kyo2.kernel
 
 import internal.*
 import kyo.*
+import kyo2.Maybe
 import scala.annotation.tailrec
 import scala.language.implicitConversions
 import scala.util.NotGiven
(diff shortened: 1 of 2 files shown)


fwbrasil:
FYI: I had pushed this change to `main` by mistake but I've just reverted it.

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo2/kernel/Pending.scala
```diff
@@ -50,6 +51,26 @@ object `<`:
                         )
             mapLoop(v)
         end map
+
+        inline def evalNow: Maybe[A] =
```
hearnadam, line 55: Could this be the new `pure`?

fwbrasil: I had named it `pure` initially but thought it could be confusing given that we used to provide the method with a different behavior. `evalNow` seems to better communicate what the API does as well.

PR #517, opened 2024-07-02
prelude: add Safepoint.ensure

fwbrasil:
This PR introduces `Safepoint.ensure` to be used from `IO.ensure`. It ensures the execution of a thunk at the end of the computation even if the computation fails. In case an `Interceptor` is used to preempt execution (like in `IOTask`), the interceptor can also collect the thunks of the `ensure` calls to execute them later. It's a mechanism [similar](https://github.com/getkyo/kyo/blob/14c1d921441bfd23279434559e7da9e5987e2c5f/kyo-core/shared/src/main/scala/kyo/ios.scala#L89) to the current core.

This seems to be the last feature we need to port `kyo-core` to the new design! \o/

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
@@ -53,8 +53,8 @@ object Safepoint:
     implicit def get: Safepoint = local.get()
 
     abstract private[kyo2] class Interceptor:
-        def addEnsure(f: () => Unit): Unit    = {}
-        def removeEnsure(f: () => Unit): Unit = {}
+        def addEnsure(f: () => Unit): Unit
+        def removeEnsure(f: () => Unit): Unit
         def enter(frame: Frame, value: Any): Boolean
         def exit(): Unit
     end Interceptor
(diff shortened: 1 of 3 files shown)


kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
```diff
@@ -92,6 +96,38 @@ object Safepoint:
         immediate(p)(loop(v))
     end propagating
 
+    private[kyo2] def ensure[A, S](f: => Unit)(v: => A < S)(using safepoint: Safepoint, _frame: Frame): A < S =
+        // ensures the function is called once even if an
+        // interceptor executes it multiple times
```
fwbrasil, line 101: This behavior is necessary because `f` can be executed multiple times depending on the execution path. For example the function can be evaluated at the end of the computation normally but then an exception is thrown and the catch in `eval` attempts to execute it again. `IOTask` also has some performance optimizations that can end up invoking the function multiple times.

hearnadam: Very clever! Perhaps we should name `def eval` distinctly from the function which is a few lines below `ensure`?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
```diff
@@ -53,6 +53,8 @@ object Safepoint:
     implicit def get: Safepoint = local.get()
 
     abstract private[kyo2] class Interceptor:
+        def addEnsure(f: () => Unit): Unit    = {}
+        def removeEnsure(f: () => Unit): Unit = {}
```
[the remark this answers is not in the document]
fwbrasil: makes sense, updated!

PR #520, opened 2024-07-02
make Effect.handle* methods partially tail-recursive

fwbrasil:
I'm doing some performance testing and noticed a regression due to the `Effect.handle*` methods of the new kernel not being tail-recursive when there are multiple suspensions of the same effect. I hadn't noticed this issue yet because the new safepoint mechanism prevents stack overflows.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
@@ -157,6 +157,22 @@ object Safepoint:
                 finally self.exit(depth)
     end handle
 
+    private[kernel] inline def handle[A, B, S](value: Any)(
+        inline eval: => A,
+        inline continue: A => B < S,
+        inline suspend: Safepoint ?=> B < S
+    )(using inline frame: Frame, self: Safepoint): B < S =
+        self.enter(frame, value) match
+            case -1 =>
+                Effect.defer(suspend)
+            case depth =>
+                val a =
+                    try eval
+                    finally self.exit(depth)
+                continue(a)
+        end match
+    end handle
+
     // TODO compiler crash if private[kyo2]
     def insertTrace(cause: Throwable)(using self: Safepoint): Unit =
         val size = Math.min(self.traceIdx, maxTraceFrames)
--- a/kyo-prelude/jvm/src/test/scala/kyo2/kernel/BytecodeTest.scala
+++ b/kyo-prelude/jvm/src/test/scala/kyo2/kernel/BytecodeTest.scala
@@ -37,7 +37,7 @@ class BytecodeTest extends Test:
 
     "handle" in {
         val map = methodBytecodeSize[TestHandle]
-        assert(map == Map("test" -> 30, "anonfun" -> 15, "handleLoop" -> 266))
+        assert(map == Map("test" -> 30, "anonfun" -> 15, "handleLoop" -> 267))
     }
 
     def methodBytecodeSize[A](using ct: ClassTag[A]): Map[String, Int] =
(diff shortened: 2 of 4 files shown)


kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
```diff
@@ -157,6 +157,22 @@ object Safepoint:
...
+        inline suspend: Safepoint ?=> B < S
+    )(using inline frame: Frame, self: Safepoint): B < S =
+        self.enter(frame, value) match
+            case -1 =>
+                Effect.defer(suspend)
+            case depth =>
+                val a =
+                    try eval
+                    finally self.exit(depth)
+                continue(a)
```
fwbrasil, line 172: The compiler couldn't make the `handleLoop` methods tail-recursive when the computation isn't suspended because `self.exit` made the recursive method call not be in the tail position. By the way, isn't it amazing that we compiler can apply tail call elimination even when the recursive call is in an inlined method 😻

fwbrasil: Example decompiled handle loop (not including https://github.com/getkyo/kyo/pull/519):

```java
public final Object kyo2$IO$$$_$handleLoop$1(String x$2$1, Object v, Map context, Safepoint safepoint) {
        Object kyo;
        Object object;
        Object object2;
        Object object3;
        while ((object3 = $less$.MODULE$._1$extension(object2 = $less$.MODULE$.unapply(object = v))) instanceof package.internal.KyoSuspend) {
            Object a;
            Object object4;
            tags$package$ tags$package$2;
            package.internal.KyoSuspend kyo2 = (package.internal.KyoSuspend)object3;
            tags$package$ tags$package$_this = tags$package$2 = tags$package$.MODULE$;
            if (!tags.package.Tag$.MODULE$.$eq$colon$eq((Object)"'Qkyo2.IO;5?kyo2.kernel.Effect;!T0;!U1;!V2;", (Object)kyo2.tag())) break;
            BoxedUnit boxedUnit = (BoxedUnit)kyo2.input();
            if (!true) break;
            int n = safepoint.enter(x$2$1, (Object)kyo2);
            if (-1 == n) {
                return Effect$.MODULE$.defer((Function1 & Serializable)contextual$2 -> new .less(this.handleLoop$1$$anonfun$1(kyo2, context, x$2$1, (Safepoint)contextual$2)));
            }
            int depth = n;
            try {
                BoxedUnit boxedUnit2 = (BoxedUnit)kyo2.input();
                Function1 & Serializable intersect = (Function1 & Serializable)_$9 -> new .less(kyo2.apply((Object)_$9, context, safepoint));
                Object object5 = intersect.apply((Object)BoxedUnit.UNIT);
                object4 = object5 == null ? null : ((.less)object5).kyo2$kernel$$less$$curr();
            }
            finally {
                safepoint.exit(depth);
            }
            v = a = object4;
        }
        if (object3 instanceof package.internal.KyoSuspend) {
            package.internal.KyoSuspend kyo3 = (package.internal.KyoSuspend)object3;
            package.internal.KyoSuspend p$proxy1 = new package.internal.KyoSuspend<Object, Object, Effect, Object, T, S>(kyo3, x$2$1){
                private final String x$2$4;
                private final package.internal.KyoSuspend kyo$4;
                private final String tag;
                private final Object input;
                {
                    this.x$2$4 = x$2$3;
                    this.kyo$4 = kyo$3;
                    this.tag = kyo$3.tag();
                    this.input = kyo$3.input();
                }

                public String tag() {
                    return this.tag;
                }

                public Object input() {
                    return this.input;
                }

                public String frame() {
                    return this.x$2$4;
                }

                public Object apply(Object v, Map context, Safepoint x$3) {
                    return IO$.MODULE$.kyo2$IO$$$_$handleLoop$1(this.x$2$4, this.kyo$4.apply(v, context, x$3), context, x$3);
                }
            };
            return p$proxy1;
        }
        Object object6 = kyo = object3;
        return $less$.MODULE$.apply(object6);
    }
 ```

hearnadam (approved):

PR #522, opened 2024-07-03
fix(stats): Handle delta overflows in Counter and Countergauge 

### LINKED ISSUE #410: Stats: handle overflows
Kyo now has its own implementation of metrics introduced by https://github.com/getkyo/kyo/pull/409. As [pointed out](https://github.com/getkyo/kyo/pull/409#discussion_r1610859626) in the review, the implementation doesn't handle overflows when reporting deltas to exporters.

A solution for this issue must address this issue for both `Counter` and `CounterGauge`. I haven't implemented a solution for this problem yet but it seems we could detect the overflow and still infer the correct delta.

mobley-trent:
Closes #410 
/claim #410 

- Handles overflow while calculating the delta in `Counter` and `CounterGauge`

### DIFF
--- a/kyo-core/js/src/main/scala/kyo/addersStubs.scala
+++ b/kyo-core/js/src/main/scala/kyo/addersStubs.scala
@@ -14,4 +14,9 @@ class LongAdder():
     def increment()  = curr += 1
     def sum(): Long  = curr
     def reset()      = curr = 0
+    def sumThenReset(): Long =
+        val res = curr
+        curr = 0
+        res
+    end sumThenReset
 end LongAdder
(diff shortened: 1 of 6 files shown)


mobley-trent:
@fwbrasil I think this takes care of the rollover

hearnadam:
I think the issue was looking for a more holistic solution, but regardless tests will be necessary for this bounty claim.

mobley-trent:
Working on a better solution + tests

mobley-trent:
Hi @fwbrasil @hearnadam this solution addresses the overflow on `last` by rolling over the remainder of the value to start from zero. This would mitigate the inaccuracies when dealing with negative numbers. Kindly review and advise if this was not the intended behaviour

mobley-trent:
@fwbrasil @hearnadam I would appreciate your feedback

fwbrasil:
Thank you! Sorry for the delay to review @mobley-trent

Issue #368, opened 2024-05-15
streams take does not terminate the stream early


lukestephenson:
Consider the following application which emits 10,000 stream elements, but a downstream `take(5)` limits that to the first 5 operations. I'd expect this to signal early termination up the stream and stop emitting elements.
```
package kyo.example

import kyo.{KyoApp, Streams}
import kyo._
import scala.concurrent.duration._

object KyoBufferExample {
  val seq = (1 to 10000)
  val streamApp = Streams.initSeq(seq)
    .transform { i =>
      for {
        _ <- Consoles.println(s"pre take got $i")
      } yield i
    }
    .take(5)
    .transform { i =>
      for {
        _ <- Consoles.println(s"post take got $i")
        _ <- Fibers.sleep(1.seconds)
      } yield i
    }
    .runFold(0)(_ + _)

  def main(args: Array[String]): Unit = {
    KyoApp.run(streamApp)
  }

}
```
However, when this is run, after the first 5 elements are consumed, the remaining elements are still printed to the console as "pre take"

The equivalent zio example completes early:
```
    ZStream.range(1, 20, 1)
      .tap(i => Console.printLine(s"got $i"))
      .take(2)
      .tap(i => Console.printLine(s"take got $i").delay(1.second))
      .runCollect
```

hearnadam:
This seems to have been introduced by stream chunking: https://github.com/getkyo/kyo/pull/270

@fwbrasil, correct me if I am wrong, but I think we need a `Chunk[V].map` which produces `Chunk[U < S]` instead of `Chunk[U] < S`
https://github.com/getkyo/kyo/blob/main/kyo-core/shared/src/main/scala/kyo/chunks.scala#L84-L95

lukestephenson:
@hearnadam It doesn't specifically look related to chunking to me. Check out the top level definition of `take`:
https://github.com/getkyo/kyo/blob/3ebea0ccefe666487ea9e24c84af763257ddf0ba/kyo-core/shared/src/main/scala/kyo/streams.scala#L22-L26

The way I read that is, when invoked as `take(0)` that it still tries to process (`runDiscard`) the stream, just not passing it through any further downstream handlers (but anything upstream will still run to stream completion).

fwbrasil:
[...]

PR #523, opened 2024-07-03
prelude: Stream and Emit effects

fwbrasil:
This PR introduces a new `Emit` effect to be used as the underlying effect of the new `Stream` implementation. The effect replaces `Sum` since it has a very similar functionality.

The `Stream` implementation introduces two major changes:

1. The type now has only two parameters `Stream[S, V]` and the underlying computation now produces a `Unit`. It's a simplification that reduces the expressivity of the effect but I think usability is more important. We can later add a separate implementation like `Flow[S, V, A]` with a functionality like the previous Stream implementation if needed. I placed the `S` parameter as the first one. It's similar to the new `Fiber[E, A]` design.

2. The implementation now properly handles early termination (https://github.com/getkyo/kyo/issues/368). I won't close the ticket for now since I still need to add the `IO`-based operations as extension methods when porting `kyo-core`.

I've also made a minor improvement to `Var.update` so it returns the updated value.


kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
```diff
@@ -159,26 +159,24 @@ object Effect:
...
 
-        inline def state[I[_], O[_], E <: Effect[I, O], State, A, B, S, S2](
+        inline def state[I[_], O[_], E <: Effect[I, O], State, A, B, S, S2, S3](
             inline tag: Tag[E],
             inline state: State,
             v: A < (E & S)
         )(
-            inline handle: Safepoint ?=> [C] => (I[C], State, O[C] => A < (E & S & S2)) => (State, A < (E & S & S2)),
-            inline done: (State, A) => B < (S & S2) = (_: State, v: A) => v,
+            inline handle: Safepoint ?=> [C] => (I[C], State, O[C] => A < (E & S & S2)) => (State, A < (E & S & S2)) < S3,
```
fwbrasil, line 167: This is something the previous core used to support and was missing from the new kernel. It allows `handle` to perform other effects before returning the tuple, which is something necessary for `Stream` since methods like `take` have to emit values of the same type that is getting handled.

kyo-prelude/shared/src/test/scala/kyo2/EmitTest.scala
```diff
@@ -0,0 +1,70 @@
+package kyo2
+
+class EmitTest extends Test:
```
fwbrasil, line 3: Ported from `SumTest`.

kyo-prelude/shared/src/test/scala/kyo2/StreamTest.scala
```diff
@@ -0,0 +1,650 @@
+package kyo2
+
+class StreamTest extends Test:
```
fwbrasil, line 3: Ported from `kyo-core` without the `IO`-based operations + new test scenarios.

kyo-prelude/shared/src/main/scala/kyo2/Stream.scala
```diff
@@ -0,0 +1,202 @@
...
+        Stream[S & S2, V2](Effect.handle.state(tagV, (), v)(
+            [C] =>
+                (input, _, cont) =>
+                    if input.isEmpty then
+                        Emit.andMap(Chunk.empty[V2])(ack => ((), cont(ack)))
+                    else
+                        f(input).map(c => Emit.andMap(c)(ack => ((), cont(ack))))
+        ))
+
+    def flatMap[S2, V2, S3](f: V => Stream[S2, V2] < S3)(
```
fwbrasil, line 39: `flatMap` and `flatMapChunk` are new methods

kyo-prelude/shared/src/test/scala/kyo2/StreamTest.scala
```diff
@@ -0,0 +1,650 @@
...
+                    Seq.empty
+            )
+        }
+
+        "with effects" in {
+            val stream = Stream.init(Seq(1, 2, 3, 4, 5))
+            val taken = stream.takeWhile { v =>
+                Var.update[Int](_ + 1).map(_ < 4)
+            }.runSeq
+            assert(Var.run(0)(taken).eval == Seq(1, 2, 3))
```
hearnadam, line 133: Can you also validate the result of the Var? This way we can check that take actually stops the stream when receiving the ack.

kyo-prelude/shared/src/main/scala/kyo2/Emit.scala
```diff
@@ -0,0 +1,78 @@
...
+        object Continue:
+            def apply(): Continue              = Int.MaxValue
+            def apply(maxItems: Int): Continue = Math.max(0, maxItems)
+            def unapply(ack: Ack): Maybe.Ops[Int] =
+                if ack < 0 then Maybe.empty
+                else Maybe(ack)
+        end Continue
+
+        opaque type Stop <: Ack = Int
+        val Stop: Ack = -1
```
hearnadam, line 30: Is the `Stop` type actually used?

hearnadam (approved):

PR #524, opened 2024-07-04
prelude: Memo and Layer

fwbrasil:
This PR ports layers to the new design in `kyo-prelude`. I've introduced the `Memo` effect to replace the use of `IOs`. As described in https://github.com/getkyo/kyo/pull/523, I'll make sure to restore the git history when porting `kyo-core` to the new design. cc/ @kitlangton


kyo-prelude/shared/src/main/scala/kyo2/Layer.scala
```diff
@@ -0,0 +1,105 @@
...
+            Kyo.zip(Env.get[A], Env.get[B], Env.get[C])
+                .map { case (a, b, c) => f(a, b, c) }
+        }
+
+    def from[A: Tag, B: Tag, C: Tag, D: Tag, E: Tag, S](f: (A, B, C, D) => E < S): Layer[E, Env[A & B & C & D] & S] =
+        apply {
+            Kyo.zip(Env.get[A], Env.get[B], Env.get[C], Env.get[D]).map { case (a, b, c, d) => f(a, b, c, d) }
+        }
+
+    @targetName("fromPure1")
```
fwbrasil, line 52: I had to introduce these methods because the compiler can't infer the result type as `<` from a pure function anymore since the lifting of pure values into `<` now happens via an implicit conversion.

hearnadam: Hmmm. I would like to avoid this if possible. I wonder if a private implicit in this scope might change things...

kyo-prelude/shared/src/main/scala/kyo2/Memo.scala
```diff
@@ -0,0 +1,32 @@
+package kyo2
+
+import kyo.Tag
+import kyo2.kernel.*
+
+opaque type Memo = Var[Memo.Cache]
+
+object Memo:
+
+    private[kyo2] case class Cache(map: Map[(Any, Any), Any])
```
hearnadam, line 10: Can this be an AnyVal? Or will it be boxed anyways?

kyo-prelude/shared/src/main/scala/kyo2/Memo.scala
```diff
@@ -0,0 +1,32 @@
...
+    def apply[A, B, S](f: A => B < S)(using Frame): A => B < (S & Memo) =
+        val token = new Object
+        a =>
+            val key = (a, token)
+            Var.use[Cache] { cache =>
+                cache.map.get(key) match
+                    case Some(cached) => cached.asInstanceOf[B]
+                    case None =>
+                        f(a).map { result =>
+                            Var.update[Cache](c => c.copy(c.map + (key -> result)))
```
hearnadam, line 23: How would you feel about a `get` and `updated` method on Cache type?

kyo-prelude/shared/src/main/scala/kyo2/Memo.scala
```diff
@@ -0,0 +1,32 @@
...
+opaque type Memo = Var[Memo.Cache]
+
+object Memo:
+
+    private[kyo2] case class Cache(map: Map[(Any, Any), Any])
+
+    private val empty = Cache(Map.empty)
+
+    def apply[A, B, S](f: A => B < S)(using Frame): A => B < (S & Memo) =
+        val token = new Object
```
hearnadam, line 15: What's the object for? I'm surprised this creates the behavior shown in the tests. Is the hashcode stable for a given A?

kyo-prelude/shared/src/main/scala/kyo2/Layer.scala
```diff
@@ -0,0 +1,105 @@
...
+    final infix def and[Out2, S2](that: Layer[Out2, S2]): Layer[Out & Out2, S & S2]                   = And(self, that)
+    final infix def using[Out2, S2, In2](that: Layer[Out2, Env[In2] & S2]): Layer[Out & Out2, S & S2] = self and (self to that)
+
+end Layer
+
+object Layer:
+
+    extension [In, Out, S](layer: Layer[Out, Env[In] & S])
+        def run[R](using reduce: Reducible[Env[In]]): TypeMap[Out] < (S & reduce.SReduced & Memo) =
+            reduce(doRun(layer))
```
hearnadam, line 22: Nice!

kyo-prelude/shared/src/main/scala/kyo2/Memo.scala
fwbrasil: Yeah, it's used as a generic in `Var` so it'd box when handling the effect

kyo-prelude/shared/src/main/scala/kyo2/Layer.scala
fwbrasil: I can't see a way to avoid it, the trouble is on the callers of `Layer.from`. The compiler isn't able to infer the implicit conversion to lift the result type of the provided functions. If you removed these methods, `LayerTest` shows the issue.

kyo-prelude/shared/src/main/scala/kyo2/Memo.scala
fwbrasil: It's used to ensure each memoized function has a different key space in the cache, otherwise values from one function could be returned by another. The `hashCode` is the identity of the object itself created with each function, unrelated to `A`. I've done some refactoring to make it clearer.

kyo-prelude/shared/src/main/scala/kyo2/Memo.scala
```diff
@@ -0,0 +1,45 @@
...
+
+    // Used to ensure each memoized function
+    // has a different key space
+    private[kyo2] class MemoIdentity
+
+    private[kyo2] class Cache(map: Map[(Any, Any), Any]):
+        def get[A](input: A, id: MemoIdentity): Maybe[Any] =
+            val key = (input, id)
+            if map.contains(key) then
+                Maybe(map(key))
```
hearnadam, line 18: this restricts to non-null values, but should be okay.

hearnadam (approved):

PR #528, opened 2024-07-04
prelude: error handling improvements

fwbrasil:


kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -83,23 +83,25 @@ object Abort:
                     )
                 }
             } {
-                case fail: E => Result.fail(fail)
-                case fail    => Result.panic(fail)
+                case fail: E if classOf[Throwable].isAssignableFrom(ct.runtimeClass) =>
```
fwbrasil, line 86: @hearnadam I think you had warned about this 🙈 I'm using `Abort` for the new `Fiber` implementation and it's incorrectly translating thrown exceptions that should be a `Panic` to `Fail` instead. I could work it around in `Fiber` but this seems a safer behavior.

hearnadam: Can you try using `if tag <:< Tag[Throwable]` to check if the result should be a `fail`?

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
```diff
@@ -83,23 +83,25 @@ object Abort:
...
+                    Result.fail(fail)
+                case fail => Result.panic(fail)
             }
     end RunOps
 
     inline def run[E >: Nothing]: RunOps[E] = RunOps(())
 
     final class CatchingOps[E <: Throwable](dummy: Unit) extends AnyVal:
-        def apply[A, B](v: => A < B)(
```
hearnadam, line 94: Do we need `catching` if `run` also catches?

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
```diff
@@ -176,7 +177,7 @@ object Safepoint:
     // TODO compiler crash if private[kyo2]
     def insertTrace(cause: Throwable)(using self: Safepoint): Unit =
         val size = Math.min(self.traceIdx, maxTraceFrames)
-        if size > 0 then
+        if size > 0 && !cause.isInstanceOf[NoStackTrace] then
```
hearnadam, line 180: Do we want to have an `else` to capture a single Frame with the Throwable's name (even if it doesn't have a stack trace)? Not sure if that's already captured somewhere else...

kyo-prelude/shared/src/test/scala/kyo2/ResultTest.scala
```diff
@@ -175,6 +175,21 @@ class ResultTest extends Test:
...
 
+    "getOrThrow" - {
+        "returns the value for Success" in {
+            assert(Result.success(1).getOrThrow == 1)
+        }
+        "doesn't compile for non-Throwable Fail" in {
+            assertDoesNotCompile("Result.fail(1).getOrThrow")
+        }
+        "throws for Throwable Fail" in {
+            assert(Try(Result.fail(ex).getOrThrow) == scala.util.Failure(ex))
```
hearnadam, line 186: `Result.attempt` 😜

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
fwbrasil: Yep, tests pass with it as well! It's hard to tell if it'd have better performance than `isAssignableFrom`, though. I'd prefer to keep it as is since the `Tag` in this method is more of a workaround.

kyo-prelude/shared/src/main/scala/kyo2/Abort.scala
fwbrasil: `catching` is provides exception handling without handling the effect itself (return type is `A`, not `Result`). It's useful to isolate code that might throw exceptions and lift them into `Abort` suspensions.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Safepoint.scala
fwbrasil: The `Throwable` instance already has its type/name. This condition only avoids mutating the stack trace to insert Kyo's frames.

hearnadam (approved):

PR #530, opened 2024-07-06
Lazily initialize `default` in `Local`

hearnadam:
When working with KyoApp locally, I noticed that SLF4J was being initialized even when I was providing a Local value to override it from the very start. This should prevent that behavior at some cost to the first access.

@fwbrasil if you prefer, I can create an `initLazy`.


hearnadam:
We can also add an extension:
```
    extension [A](local: Local[A])
        def strict: Local[A] =
            val _ = local.default
            local
```

fwbrasil:
LGTM! I'm not sure how useful `strict` would be but it seems ok to add as well.

fwbrasil (approved):

PR #532, opened 2024-07-09
prelude: lift pure functions

fwbrasil:
I'm noticing more scenarios where pure functions fail compile when passed to methods that take effectful functions while porting `kyo-core`. This PR introduces implicit conversions for pure functions with up to four parameters.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/kernel/Pending.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/kernel/Pending.scala
@@ -14,7 +14,28 @@ object `<`:
 
     implicit private[kyo2] inline def apply[A, S](p: Kyo[A, S]): A < S = new <(p)
 
-    implicit inline def lift[A, S](v: A)(using inline ng: NotGiven[A <:< (Any < Nothing)]): A < S = <(v)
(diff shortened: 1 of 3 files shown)


kyo-prelude/shared/src/main/scala/kyo2/kernel/Pending.scala
```diff
@@ -14,7 +14,28 @@ object `<`:
 
     implicit private[kyo2] inline def apply[A, S](p: Kyo[A, S]): A < S = new <(p)
 
-    implicit inline def lift[A, S](v: A)(using inline ng: NotGiven[A <:< (Any < Nothing)]): A < S = <(v)
+    implicit inline def lift[A, S](v: A)(using inline ng: NotGiven[A <:< (Any < Nothing)]): A < S =
+        <(v)
+
+    implicit inline def liftPureFunction1[A1, B](inline f: A1 => B)(
+        using inline ng: NotGiven[B <:< (Any < Nothing)]
```
hearnadam, line 21: Does this need a name?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo2/kernel/Pending.scala
fwbrasil: yeah, the compiler requires it because of `inline`

Issue #531, opened 2024-07-07
optimize `map` call after a large number of effect suspensions


marcinzh:
Hello! I've been refreshing my microbenchmark suite and ran into this problem.

Step 0:

```scala
  def program: Int < Vars[Int] =
    Vars.get[Int].flatMap: n =>
      if n <= 0
      then n
      else Vars.set(n - 1).andThen(program)

  Vars.run(100000)(program).pure
```

This works OK.

Step 1: Rewrite to use `for` comprehension.

```scala
  def program: Int < Vars[Int] =
    for
      n <- Vars.get[Int]
      x <-
        if n <= 0
        then n
        else Vars.set(n - 1).andThen(program)
    yield x

  Vars.run(100000)(program).pure
```

This results in SO.

Step 2: Add `IOs` to prevent SO.

```scala
  def program: Int < (Vars[Int] & IOs) =
    for
      n <- Vars.get[Int]
      x <-
        if n <= 0
        then n
        else Vars.set(n - 1).andThen(IOs(program))
    yield x

  IOs.run:
    Vars.run(100000)(program)
```

Now the program runs **very** slow. I had to reduce the initial counter to 1000 to make it ever finish.

Step 3:

The problem must be related to how Scala desugars `for`. We can replicate it by adding a `map` after the recursion:

```scala
  def program: Int < (Vars[Int] & IOs) =
    Vars.get[Int].flatMap: n =>
      if n <= 0
      then n
      else Vars.set(n - 1).andThen(IOs(program)).map(x => x)

  IOs.run:
    Vars.run(100000)(program)
```

Now it behaves as in Step 2.

fwbrasil:
Thank you for the report @marcinzh! These are essentially different computations, which is why we see a different outcome. The additional `IOs` suspension and `map` transformation naturally introduces more overhead, mainly due to their allocations. Have you had a chance to compare the performance of a similar computation in another library?

The need to introduce suspensions like `IOs` for recursion is something that is going away with the next release of the library (see https://github.com/getkyo/kyo/pull/489 for more background). You can also optimize `Vars.get` + `flatMap` to `Vars.use`. `Loops` would be more idiomatic as well and might provide better performance.

marcinzh:
@fwbrasil Adding `IOs` itself isn't a big deal. In a loop that uses 4 effects, it only incurs ~10% performance penalty.

Using `for` though, slows down the loop ~500 times. That's with iteration count reduced to 1000 to get a measurable result.

[...]

PR #534, opened 2024-07-10
prelude: Var improvements

fwbrasil:
I'm taking a look at the profiling session for https://github.com/getkyo/kyo/issues/531 and noticed that we can avoid the allocation of `Input.Set`. This PR updates the effect to use a union type for its input, introduces new APIs, and includes scaladocs (you read that right 😅)

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/Var.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Var.scala
@@ -4,29 +4,41 @@ import Var.internal.*
 import kyo.Tag
 import kyo2.kernel.*
 
-sealed trait Var[V] extends Effect[Input[V, *], Id]
+sealed trait Var[V] extends Effect[Const[Op[V]], Const[V]]
 
 object Var:
 
+    /** Obtains the current value of the 'Var'. */
     inline def get[V](using inline tag: Tag[Var[V]], inline frame: Frame): V < Var[V] =
         use[V](identity)
 
     final class UseOps[V](dummy: Unit) extends AnyVal:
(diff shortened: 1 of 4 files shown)


kyo-prelude/shared/src/main/scala/kyo2/Var.scala
```diff
@@ -4,62 +4,73 @@ import Var.internal.*
...
     inline def use[V]: UseOps[V] = UseOps(())
 
-    inline def set[V](inline value: V)(using inline tag: Tag[Var[V]], inline frame: Frame): Unit < Var[V] =
-        Effect.suspend[Unit](tag, (() => value): Set[V])
+    /** Sets a new value and returns the previous one. */
+    inline def set[V](inline value: V)(using inline tag: Tag[Var[V]], inline frame: Frame): V < Var[V] =
+        Effect.suspend[Unit](tag, value: Op[V])
 
+    /** Sets a new value and returns `Unit`. */
+    inline def setUnit[V](inline value: V)(using inline tag: Tag[Var[V]], inline frame: Frame): Unit < Var[V] =
```
hearnadam, line 32: I don't love how it sounds like you are setting the value to `Unit` - perhaps `setDiscard`?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo2/Var.scala
fwbrasil: Makes sense, I'll start using this nomenclature. Should we keep `pending.unit` or rename it to `pending.discard` for consistency?

PR #535, opened 2024-07-10
prelude: Loop improvements

fwbrasil:
This PR introduces an optimization to specialize `Continue*` instances. They're currently generic, which requires boxing of primitive types in the `Continue` instance fields. After this change, the fields in the final bytecode are specialized. Example from a decompiled loop:

```java
new Loop.Continue2<String, Object>(i1, i2){
    private final String i1$tailLocal1$1;
    private final int i2$tailLocal1$1;
```

There's still boxing when the value is recovered in the generic scope of the `Loop` body but that's something the JIT can easily optimize.

I'm also using abbreviated generic names. It makes it easier for me to read the code but I'd be ok to moving back to more descriptive names. The `Loop.transform*` methods are renamed to `Loop.apply*` to improve readability since `transform` doesn't seem to communicate much meaning in this context.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/kernel/Loop.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/kernel/Loop.scala
@@ -7,77 +7,77 @@ import scala.util.NotGiven
 
 object Loop:
 
-    abstract class Continue[V1]:
-        private[Loop] def _v1: V1
-    abstract class Continue2[V1, V2]:
-        private[Loop] def _v1: V1
-        private[Loop] def _v2: V2
-    abstract class Continue3[V1, V2, V3]:
-        private[Loop] def _v1: V1
-        private[Loop] def _v2: V2
-        private[Loop] def _v3: V3
+    abstract class Continue[A]:
+        private[Loop] def _1: A
+    abstract class Continue2[A, B]:
+        private[Loop] def _1: A
+        private[Loop] def _2: B
+    abstract class Continue3[A, B, C]:
+        private[Loop] def _1: A
+        private[Loop] def _2: B
+        private[Loop] def _3: C
     end Continue3
-    abstract class Continue4[V1, V2, V3, V4]:
-        private[Loop] def _v1: V1
-        private[Loop] def _v2: V2
-        private[Loop] def _v3: V3
-        private[Loop] def _v4: V4
+    abstract class Continue4[A, B, C, D]:
+        private[Loop] def _1: A
+        private[Loop] def _2: B
+        private[Loop] def _3: C
+        private[Loop] def _4: D
     end Continue4
 
-    opaque type Result[I, O]               = O | Continue[I]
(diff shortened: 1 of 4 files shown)


kyo-prelude/shared/src/main/scala/kyo2/kernel/Loop.scala
```diff
@@ -7,212 +7,239 @@ import scala.util.NotGiven
...
+        private[Loop] def _v1: V1
+    abstract class Continue2[V1, V2]:
+        private[Loop] def _v1: V1
+        private[Loop] def _v2: V2
+    abstract class Continue3[V1, V2, V3]:
+        private[Loop] def _v1: V1
+        private[Loop] def _v2: V2
+        private[Loop] def _v3: V3
+    end Continue3
+    abstract class Continue4[V1, V2, V3, V4]:
```
fwbrasil, line 20: I've also added support for loops with four parameters.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Loop.scala
```diff
@@ -7,212 +7,239 @@ import scala.util.NotGiven
 
 object Loop:
 
-    case class Continue[Input](input: Input)
-    private case class Continue2[Input1, Input2](input1: Input1, input2: Input2)
-    private case class Continue3[Input1, Input2, Input3](input1: Input1, input2: Input2, input3: Input3)
+    abstract class Continue[V1]:
+        private[Loop] def _v1: V1
```
hearnadam, line 11: How about using `_1` just like tuple?

kyo-prelude/shared/src/main/scala/kyo2/kernel/Loop.scala
```diff
@@ -7,212 +7,239 @@ import scala.util.NotGiven
 
 object Loop:
 
-    case class Continue[Input](input: Input)
-    private case class Continue2[Input1, Input2](input1: Input1, input2: Input2)
-    private case class Continue3[Input1, Input2, Input3](input1: Input1, input2: Input2, input3: Input3)
+    abstract class Continue[V1]:
```
hearnadam, line 10: How about `A, B, C`, etc for type parameters? You can keep `O` distinct if you want.

kyo-prelude/shared/src/test/scala/kyo2/kernel/LoopTest.scala
```diff
@@ -4,27 +4,27 @@ import kyo2.*
 
 class LoopTest extends Test:
 
-    "transform" - {
+    "apply" - {
         "with a single iteration" in {
             assert(
-                Loop.transform(1)(i => if i < 5 then Loop.continue(i + 1) else Loop.done(i)).eval == 5
+                Loop(1)(i => if i < 5 then Loop.continue(i + 1) else Loop.done(i)).eval == 5
```
hearnadam, line 10: Definitely improves readability!

hearnadam (approved):

PR #536, opened 2024-07-10
prelude: avoid allocations for tracing

fwbrasil:
The new kernel allocates new traces on each `Boundary` use. This is proving too expensive when porting fibers to the new design in `kyo-prelude`. I've been avoiding caching state in the library since it can become problematic depending on the characteristics of specific workloads. For example, a cache might show great performance in a benchmark but become a major bottleneck in real applications if it doesn't have enough space and the hit ratio drops.

That said, this seems to be the first case where there's no alternative to avoid allocations in such critical piece of code since fiber execution is a key aspect of the library's performance. With this change, traces are cached and reused as needed, which avoids the vast majority of allocations. To reduce contention, each thread maintains a local cache and replenishes from the global cache as necessary.


kyo-prelude/shared/src/main/scala/kyo2/kernel/TracePool.scala
```diff
@@ -0,0 +1,99 @@
+package kyo2.kernel
+
+import internal.*
+import java.util.Arrays
+import java.util.concurrent.atomic.AtomicInteger
+import scala.annotation.tailrec
+
+private[kernel] object TracePool:
+    inline def globalCapacity: Int = 1024
+    inline def localCapacity: Int  = 32
```
fwbrasil, line 10: These defaults should work well for most use cases but it could introduce a significant increase in memory footprint for systems that rely on blocking of large numbers of threads with Loom. We're not advertising the Loom integration yet but we might need to provide an option to disable the local caches if it becomes a first-class feature.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Trace.scala
```diff
@@ -0,0 +1,103 @@
...
+class Trace(
+    private[kernel] var frames: Array[Frame],
+    private[kernel] var index: Int
+)
+
+object Trace:
+
+    private[kernel] def init: Trace = Trace(new Array[Frame](maxTraceFrames), 0)
+
+    abstract class Owner extends TracePool.Local:
```
fwbrasil, line 18: This is used as the super class of `Safepoint` to "flatten" the memory layout and avoid pointer chasing given how hot this code is.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Trace.scala
```diff
@@ -0,0 +1,103 @@
...
+
+        final private[kyo2] def copyTrace(trace: Trace): Trace =
+            val newTrace = borrow()
+            System.arraycopy(trace.frames, 0, newTrace.frames, 0, Math.min(trace.index, maxTraceFrames - 1))
+            newTrace.index = trace.index
+            newTrace
+        end copyTrace
+
+        final private[kyo2] def releaseTrace(trace: Trace): Unit =
+            release(trace)
```
fwbrasil, line 53: This will be called from a point in the fiber implementation where it's guaranteed that the trace won't be used anymore because the fiber that holds it has completed.

kyo-prelude/shared/src/test/scala/kyo2/kernel/TraceTest.scala
```diff
@@ -120,7 +120,10 @@ class TraceTest extends Test:
                 val printWriter  = new PrintWriter(stringWriter)
                 ex.printStackTrace(printWriter)
                 printWriter.flush()
-                val trace = stringWriter.toString.linesIterator.takeWhile(!_.contains("anonfun")).mkString("\n")
+                val full   = stringWriter.toString
```
fwbrasil, line 123: This test was failing again because the execution flow changed. I've updated the logic to verify only frames introduce by Kyo.

fwbrasil:
I've replaced the global pool with a JCTools queue for better performance. There was also an issue with the local queues not replenishing correctly.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Trace.scala
```diff
@@ -0,0 +1,109 @@
+package kyo2.kernel
+
+import internal.*
+import java.util.Arrays
+import java.util.concurrent.atomic.AtomicInteger
+import scala.annotation.tailrec
+import scala.util.control.NonFatal
+import scala.util.control.NoStackTrace
+
+class Trace(
```
hearnadam, line 10: minor: this should be final

kyo-prelude/shared/src/main/scala/kyo2/kernel/Trace.scala
```diff
@@ -0,0 +1,109 @@
...
+                withTrace(trace)(f)
+            finally
+                release(trace)
+            end try
+        end withNewTrace
+
+        final private[kernel] def enrich(ex: Throwable): Unit =
+            val size = frames.size
+            if size > 0 && !ex.isInstanceOf[NoStackTrace] then
+                val trace = frames.filter(_ != null).map(_.parse)
```
hearnadam, line 85: `.withFilter`?

kyo-prelude/jvm/src/main/scala/kyo2/kernel/TracePool.scala
```diff
@@ -0,0 +1,58 @@
...
+                global.offer(trace)
+                ()
+            end if
+        end release
+
+        final private def clear(trace: Trace) =
+            val arr = trace.frames
+            var i   = 0
+            while i < maxTraceFrames && arr(i) != null do
+                arr(i) = null.asInstanceOf[Frame]
```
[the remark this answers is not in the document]
fwbrasil: Interesting, I didn't know about `uninitialized`. It seems available only for mutable fields:

```
[error] -- Error: /Users/fwbrasil/workspace/kyo/kyo-prelude/jvm/src/main/scala/kyo2/kernel/TracePool.scala:52:25
[error] 52 |                arr(i) = uninitialized
[error]    |                         ^^^^^^^^^^^^^
[error]    |`uninitialized` can only be used as the right hand side of a mutable field definition
```

kyo-prelude/shared/src/main/scala/kyo2/kernel/Trace.scala
fwbrasil: Updated. I forgot to optimize this method. It's an important one because the overhead to produce stack traces when a system is in a failure state can become a source of major degradation. I'll prepare a follow-up PR.

fwbrasil:
I'm merging to post a follow-up but please feel free to provide feedback!

PR #542, opened 2024-07-11
scheduler: avoid draining workers multiple times

fwbrasil:
I noticed some performance overhead while working on the new fiber implementation because workers keep draining over and over. This PR replaces the `running` flag in `Worker` by a proper `State` object. After this change, workers get drained only once when they transitioning from `Running` to `Stalled`.

### DIFF
--- a/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
+++ b/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
@@ -70,7 +70,7 @@ final class Scheduler(
         var worker: Worker = null
         if (submitter eq null) {
             worker = Worker.current()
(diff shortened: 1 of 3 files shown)


kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -70,7 +70,7 @@ final class Scheduler(
         var worker: Worker = null
         if (submitter eq null) {
             worker = Worker.current()
-            if ((worker ne null) && !worker.checkAvailability(nowMs))
+            if ((worker ne null) && ((worker eq submitter) || !worker.checkAvailability(nowMs)))
```
fwbrasil, line 73: This check was missing here, making the worker try to enqueue to itself while draining.

hearnadam (approved):

PR #546, opened 2024-07-14
Typemap deterministic get

### LINKED ISSUE #545: TypeMap retrieval is non-deterministic
This fails:
```scala
        "deterministic" in {
            trait A
            trait B extends A
            val b = new B {}
            trait C extends A
            val c = new C {}
            trait D extends A
            val d = new D {}
            trait E extends A
            val e = new E {}
            trait F extends A
            val f = new F {}
            trait G extends A
            val g = new G {}
            trait H extends A
            val h = new H {}
            trait I extends A
            val i = new I {}
            trait J extends A
            val j = new J {}
            trait K extends A
            val k = new K {}

            var map: TypeMap[A] = TypeMap(b)
            assert(map.get[A] eq b)
            map = map.add(c)
            assert(map.get[A] eq b)
            map = map.add(d)
            assert(map.get[A] eq b)
            map = map.add(e)
            assert(map.get[A] eq b)
            map = map.add(f)
            assert(map.get[A] eq b)
            map = map.add(g)
            assert(map.get[A] eq b)
            map = map.add(h)
            assert(map.get[A] eq b)
            map = map.add(i)
            assert(map.get[A] eq b)
            map = map.add(j)
            assert(map.get[A] eq b)
            map = map.add(k)
            assert(map.get[A] eq b)
        }
```

After adding `b` and `c` it will return `b` but after adding `d` it will now return `d`.

The full order is: `b` `b` `d` `e` `e` `e` `h` `h` `j` `j`

I realised this possibility while tinkering with #494.

steinybot:
Fixes #545

I would probably expect it to retrieve the last modified first but I couldn't think of an easy way to do this efficiently. First modified first still seems better than a seemingly random order.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/typeMap.scala
+++ b/kyo-core/shared/src/main/scala/kyo/typeMap.scala
@@ -12,13 +12,16 @@ object TypeMap:
             throw new RuntimeException(s"fatal: kyo.TypeMap of contents [${self.show}] missing value of type: [${t.showTpe}].")
 
         def get[B >: A](using t: Tag[B]): B =
-            val it = self.iterator
-            while it.hasNext do
-                val (tag, item) = it.next()
-                if tag <:< t then
-                    return item.asInstanceOf[B]
-            end while
-            fatal
+            def search: Any =
+                val it = self.iterator
+                while it.hasNext do
+                    val (tag, item) = it.next()
+                    if tag <:< t then
+                        return item
+                end while
+                fatal
+            end search
+            self.getOrElse(t.erased, search).asInstanceOf[B]
         end get
 
         inline def add[B](b: B)(using inline t: Tag[B]): TypeMap[A & B] =
--- a/kyo-core/shared/src/main/scala/kyo/typeMap.scala
+++ b/kyo-core/shared/src/main/scala/kyo/typeMap.scala
@@ -1,9 +1,9 @@
 package kyo
 
 import kyo.Tag.Intersection
-import scala.collection.immutable.HashMap
+import scala.collection.immutable.TreeSeqMap
 
-opaque type TypeMap[+A] = HashMap[Tag[Any], Any]
+opaque type TypeMap[+A] = TreeSeqMap[Tag[Any], Any]
 
 object TypeMap:
     extension [A](self: TypeMap[A])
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -12,22 +12,13 @@ object TypeMap:
...
-                        val (tag, item) = it.next()
-                        if tag <:< t then
-                            sub = item.asInstanceOf[B]
-                    end while
-                catch
-                    case _: NoSuchElementException => fatal
-                end try
-                sub
-            end if
+            val it = self.iterator
```
fwbrasil, line 15: would it be possible to keep the exact lookup first? The linear search on every `get` seems expensive.

steinybot: Woops. Yeah that was an oversight. Fixed.

hearnadam:
Thanks for the PR - can you run the benchmarks before and after?

I think this is likely quite valuable, but may require a custom implementation for performance reasons...

fwbrasil:
Agree on the need for a dedicated data structure but fixing the indeterminism seems already valuable. A prefix tree based on the hierarchy of the tags might be a good option for a more optimized representation. It's possible to recover the type hierarchy from the tag payload.

hearnadam:
The dark horse option is to sort the iterator of the standard map. This would maintain the exact same cost for all operations aside from `get` - when `get` exact fails. My fear is that it could be significant overhead where users are using subtypes as the elements in their map...

fwbrasil:
@hearnadam how do you feel about merging this fix for now? We can follow up with optimization later

kyo-core/shared/src/main/scala/kyo/typeMap.scala
```diff
@@ -12,22 +12,16 @@ object TypeMap:
...
-                    case _: NoSuchElementException => fatal
-                end try
-                sub
-            end if
+            def search: Any =
+                val it = self.iterator
+                while it.hasNext do
+                    val (tag, item) = it.next()
+                    if tag <:< t then
+                        return item
```
hearnadam, line 20: Generally prefer not using premature returns, but seems okay in this case.

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/typeMap.scala
steinybot: Same here actually but the trailrec loop didn't look as good.

PR #548, opened 2024-07-16
prelude: optimizations for IOTask

fwbrasil:
This seems to be the last optimizations needed to port `kyo-core` to the new design. I had designed a generic `Boundary` mechanism that could be used even by user-defined effects but the abstraction requires a few allocations in the critical path of the the fiber execution (`IOTask`). I'm kind of throwing the towel for now and providing more low-level APIs designed specifically to ensure peak performance for fibers.

### DIFF
--- a/kyo-prelude/jvm/src/test/scala/kyo2/kernel/SafepointTest.scala
+++ b/kyo-prelude/jvm/src/test/scala/kyo2/kernel/SafepointTest.scala
@@ -244,7 +244,7 @@ class SafepointTest extends Test:
                     yield 42
                 }.eval
 
-                assert(count == 6)
+                assert(count == 4)
             }
 
             "restores previous interceptor after completion" in {
@@ -278,7 +278,7 @@ class SafepointTest extends Test:
                 }.eval
 
                 assert(outerCount == 0)
-                assert(innerCount == 6)
+                assert(innerCount == 4)
             }
         }
 
@@ -319,19 +319,11 @@ class SafepointTest extends Test:
             assert(result == 26)
 
             val expectedLogs = Seq(
-                "Entering apply with value: ()",
-                "Exiting method",
                 "Entering computation with value: 6",
-                "Exiting method",
-                "Entering apply with value: ()",
                 "Exiting method",
                 "Entering Unknown with value: 12",
                 "Exiting method",
-                "Entering apply with value: ()",
-                "Exiting method",
                 "Entering computation with value: 13",
-                "Exiting method",
-                "Entering apply with value: ()",
                 "Exiting method",
                 "Entering $anonfun with value: 26",
                 "Exiting method"
@@ -383,7 +375,7 @@ class SafepointTest extends Test:
                 assert(duration < 3000000)
             }
 
-            assert(interceptor.log.size == 8)
+            assert(interceptor.log.size == 4)
             assert(interceptor.log.exists(_._1.parse.methodName == "computation"))
         }
     }
--- a/kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
@@ -1,51 +1,29 @@
 package kyo2.kernel
 
 import internal.*
+import kyo.Tag
+import scala.annotation.tailrec
 import scala.quoted.*
 
-final class Boundary[Ctx, S] private (dummy: Unit) extends AnyVal:
-
-    inline def apply[A, B, S2, S3](inline v: => A < (Ctx & S & S2))(
-        inline f: A < (S & S2) => B < S3
-    )(using inline _frame: Frame): B < (Ctx & S3) =
-        new KyoDefer[B, Ctx & S3]:
-            def frame = _frame
-            def apply(dummy: Unit, context: Context)(using safepoint: Safepoint) =
-                val state = safepoint.save(context)
-                f(Effect.defer(boundaryLoop(v, state)))
-
-    inline def apply[A, B, S2, S3](seq: Seq[A < (Ctx & S & S2)])(
-        inline f: Seq[A < (S & S2)] => B < S3
-    )(using inline _frame: Frame): B < (Ctx & S3) =
-        new KyoDefer[B, Ctx & S3]:
-            def frame = _frame
-            def apply(dummy: Unit, context: Context)(using safepoint: Safepoint) =
-                val state = safepoint.save(context)
-                f(seq.map(item => boundaryLoop(item, state)))
-
-    private def boundaryLoop[A, S2](v: A < (Ctx & S & S2), state: Safepoint.State)(using Frame): A < (S & S2) =
-        v match
-            case <(kyo: KyoSuspend[IX, OX, EX, Any, A, S] @unchecked) =>
-                new KyoContinue[IX, OX, EX, Any, A, S & S2](kyo):
-                    def frame = summon[Frame]
-                    def apply(v: OX[Any], context: Context)(using Safepoint) =
-                        val parent = Safepoint.local.get()
-                        Safepoint.local.set(Safepoint(parent.depth, parent.interceptor, state))
-                        val r =
-                            try kyo(v, state.context)
-                            finally Safepoint.local.set(parent)
-                        boundaryLoop(r, state)
-                    end apply
-            case <(kyo) =>
-                kyo.asInstanceOf[A]
+final class Boundary[Ctx, +S] private (dummy: Unit) extends AnyVal:
 
 end Boundary
 
-object Boundary:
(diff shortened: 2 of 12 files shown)


kyo-prelude/jvm/src/test/scala/kyo2/kernel/SafepointTest.scala
```diff
@@ -244,7 +244,7 @@ class SafepointTest extends Test:
                     yield 42
                 }.eval
 
-                assert(count == 6)
```
fwbrasil, line 247: This is a related fix. The `Effect.defer` method doesn't need to call `Safepoint.handle` since it's executed right after the evaluation decides that the computation can continue. These `Effect.defer` frames are also synthetic, not generated from user code, so I think it makes sense not to have them in the generated traces.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
```diff
@@ -1,51 +1,29 @@
...
-                kyo.asInstanceOf[A]
+final class Boundary[Ctx, +S] private (dummy: Unit) extends AnyVal:
 
 end Boundary
 
-object Boundary:
+private[kyo2] object Boundary:
+
+    extension [Ctx, S](boundary: Boundary[Ctx, S])
+        private[kyo2] inline def apply[A, S2](inline f: (Trace, Context) => A < S2)(using inline _frame: Frame): A < (S & S2) =
```
fwbrasil, line 15: The main optimization here is avoiding the `Boundary.State` allocation and passing the trace and context objects directly to the fiber implementation.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Context.scala
```diff
@@ -0,0 +1,27 @@
+package kyo2.kernel
+
+import kyo.Tag
+import kyo2.bug
+
+opaque type Context = Map[Tag[Any], AnyRef]
```
fwbrasil, line 6: Moved from `kernel.internal` since the type is now exposed in `Boundary`.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
```diff
@@ -13,10 +13,7 @@ object Effect:
         new KyoDefer[A, S]:
             def frame = summon[Frame]
             def apply(v: Unit, context: Context)(using Safepoint) =
-                Safepoint.handle(v)(
```
fwbrasil, line 16: No need to call `Safepoint.handle` here as elaborated previously.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
```diff
@@ -165,6 +162,51 @@ object Effect:
             handle3Loop(v, Context.empty)
         end apply
 
+        private[kyo2] inline def partial[I1[_], O1[_], E1 <: Effect[I1, O1], I2[_], O2[_], E2 <: Effect[I2, O2], I3[_], O3[_], E3 <: Effect[
```
fwbrasil, line 165: This is designed specifically for the needs of `IOTask`. You can see its use here: https://github.com/getkyo/kyo/blob/82890d80159f366e80fe4a3f81b8c551129c9b61/kyo-core2/shared/src/main/scala/kyo2/scheduler/IOTask.scala#L39

kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
```diff
@@ -1,51 +1,29 @@
...
+            new KyoDefer[A, S & S2]:
+                def frame = _frame
+                def apply(v: Unit, context: Context)(using safepoint: Safepoint) =
+                    f(safepoint.saveTrace(), context)
 
     inline given derive[Ctx, S]: Boundary[Ctx, S] = ${ boundaryImpl[Ctx, S] }
 
-    def apply[Ctx, S](using boundary: Boundary[Ctx, S]): Boundary[Ctx, S] = boundary
```
fwbrasil, line 48: This `apply` method was conflicting with the new extension method. Since it was used only for testing, I'm using `Boundary.derive` directly instead in tests.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala
fwbrasil: `apply` is also an extension method now because I made the `S` type parameter covariant to improve type inference.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
```diff
@@ -13,10 +13,7 @@ object Effect:
         new KyoDefer[A, S]:
             def frame = summon[Frame]
             def apply(v: Unit, context: Context)(using Safepoint) =
```
hearnadam, line 15: Can you remind me why defer needs `Unit`?

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
```diff
@@ -165,6 +162,51 @@ object Effect:
...
+                else
+                    v match
+                        case <(kyo: KyoSuspend[?, ?, ?, ?, ?, ?]) =>
+                            if kyo.tag =:= Tag[Defer] then
+                                val k = kyo.asInstanceOf[KyoSuspend[Const[Unit], Const[Unit], Defer, Any, A, Any]]
+                                partialLoop(k((), context), context)
+                            else
+                                safepoint.pushFrame(kyo.frame)
+                                if tag1 =:= kyo.tag then
+                                    val k = kyo.asInstanceOf[KyoSuspend[I1, O1, E1, Any, A, E1 & E2 & E3 & S & S2]]
```
hearnadam, line 191: Perhaps helpful to make a `type Suspend[I, O, E] = KyoSuspend[I, O, E, Any, A, E1 & E2 & E3 & S & S2]` on line 189?

kyo-prelude/shared/src/main/scala/kyo2/kernel/Context.scala
```diff
@@ -0,0 +1,27 @@
...
+
+object Context:
+    val empty: Context = Map.empty
+
+    extension (context: Context)
+        inline def isEmpty = context eq Map.empty
```
hearnadam, line 12: Can this just be `context eq empty`?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
fwbrasil: All effects need an input value to suspend with. In case of `Defer`, it's a `Unit` placeholder since the effect only provides suspension and doesn't use its input.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
fwbrasil: nice cleanup! I also changed the `Defer` branch to use it.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Effect.scala
```diff
@@ -165,6 +162,52 @@ object Effect:
...
+                else
+                    v match
+                        case <(kyo: KyoSuspend[?, ?, ?, ?, ?, ?]) =>
+                            type Suspend[I[_], O[_], E <: Effect[I, O]] = KyoSuspend[I, O, E, Any, A, E1 & E2 & E3 & S & S2]
+                            if kyo.tag =:= Tag[Defer] then
+                                val k = kyo.asInstanceOf[Suspend[Const[Unit], Const[Unit], Defer]]
+                                partialLoop(k((), context), context)
+                            else
+                                safepoint.pushFrame(kyo.frame)
+                                if tag1 =:= kyo.tag then
```
hearnadam, line 191: What would you think about a `def unapply(other: Tag[Any]): Boolean` for `Tag` to check `=:=`?

This could become a `match` expression + should be possible to use `@switch` annotation

fwbrasil: I think the tag string comparison has to be done for each tag even if we use an extractor. Given the low number of branches, a linear search is typically efficient given that the JIT can leverage branch profiling to optimize the execution. A more efficient `TypeMap` might be a good way to avoid the linear search in case we allow effect handling of a larger number of effects at once.

hearnadam: This comment was more about aesthetics, but I agree

PR #555, opened 2024-07-22
ansi frame rendering

fwbrasil:
This PR introduces a basic syntax highlight mechanism to improve the rendering of frames. Example output of `frame.show`:

```scala
object main extends App:
    def myMethod(i: Int)(using f: Frame) = f
    val f =
        myMethod(Math.pow(1, 0).toInt)
    println(f.show)
end main
```

image

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/Ansi.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/Ansi.scala
@@ -1,0 +1,111 @@
+package kyo2
+
+object Ansi:
+
+    extension (str: String)
+        def black: String   = s"\u001b[30m$str\u001b[0m"
+        def red: String     = s"\u001b[31m$str\u001b[0m"
+        def green: String   = s"\u001b[32m$str\u001b[0m"
+        def yellow: String  = s"\u001b[33m$str\u001b[0m"
+        def blue: String    = s"\u001b[34m$str\u001b[0m"
+        def magenta: String = s"\u001b[35m$str\u001b[0m"
+        def cyan: String    = s"\u001b[36m$str\u001b[0m"
+        def white: String   = s"\u001b[37m$str\u001b[0m"
+        def grey: String    = s"\u001b[90m$str\u001b[0m"
+
+        def bold: String      = s"\u001b[1m$str\u001b[0m"
+        def dim: String       = s"\u001b[2m$str\u001b[0m"
+        def italic: String    = s"\u001b[3m$str\u001b[0m"
+        def underline: String = s"\u001b[4m$str\u001b[0m"
+
+        def stripAnsi: String = str.replaceAll("\u001b\\[[0-9;]*[a-zA-Z]", "")
+    end extension
+
+    object highlight:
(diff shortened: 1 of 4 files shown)


kyo-prelude/shared/src/main/scala/kyo2/Ansi.scala
```diff
@@ -0,0 +1,111 @@
+package kyo2
+
+object Ansi:
+
+    extension (str: String)
+        def black: String   = s"\u001b[30m$str\u001b[0m"
```
hearnadam, line 6: Is there any notable performance difference with string concatenation vs interpolation?

kyo-prelude/shared/src/main/scala/kyo2/Ansi.scala
```diff
@@ -0,0 +1,111 @@
...
+            val headerLines  = if header.nonEmpty then header.split("\n") else Array.empty[String]
+            val codeLines    = code.split("\n").dropWhile(_.trim.isEmpty).reverse.dropWhile(_.trim.isEmpty).reverse
+            val trailerLines = if trailer.nonEmpty then trailer.split("\n") else Array.empty[String]
+
+            val allLines        = headerLines ++ codeLines ++ trailerLines
+            val toDrop          = codeLines.filter(_.trim.nonEmpty).map(_.takeWhile(_ == ' ').length).minOption.getOrElse(0)
+            val lineNumberWidth = (startLine + codeLines.length - 1).toString.length
+            val separator       = "│".dim
+
+            val processedLines = allLines.zipWithIndex.map { case (line, index) =>
```
hearnadam, line 35: At some point we should impl optimized iterable extensions, eg: `indexedMap`

kyo-prelude/shared/src/main/scala/kyo2/Ansi.scala
```diff
@@ -0,0 +1,111 @@
...
+
+                s"${lineNumber.dim} $separator $highlightedLine"
+            }
+
+            processedLines.mkString("\n")
+        end apply
+
+        def apply(code: String): String = apply("", code, "", 1)
+
+        private def highlightLine(line: String): String =
```
hearnadam, line 58: it would be nice if this was more configurable, eg for passing `String => String` functions, but not sure how that would impact performance

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo2/Ansi.scala
fwbrasil: agree!

kyo-prelude/shared/src/main/scala/kyo2/Ansi.scala
fwbrasil: I'm not sure I understood it, can you elaborate?

hearnadam: It should be possible to pass in color functions as parameters. This would enable you to add any color scheme you wanted.

ie,
```
comments: String => String = _.green,
literals: String => String = ???
```

PR #561, opened 2024-08-05
prelude: fix use of the Trace circular buffer in Trace.Owner.enrich

fwbrasil:
Traces use a circular buffer internally to store frames, which can make the first frame to land in any index of the array, not necessarily at index `0`. This PR fixes `Trace.Owner.enrich` to consider this scenario correctly.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo2/kernel/Trace.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo2/kernel/Trace.scala
@@ -39,7 +39,7 @@ object Trace:
                 System.arraycopy(frames, splitIndex, newFrames, 0, firstPartLength)
                 System.arraycopy(frames, 0, newFrames, firstPartLength, splitIndex)
             end if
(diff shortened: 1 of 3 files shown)


kyo-prelude/shared/src/main/scala/kyo2/kernel/Trace.scala
```diff
@@ -80,24 +80,38 @@ object Trace:
...
+                val toPad = parse(0, 0)
+
                 val elements =
-                    trace.foldLeft(List.empty[Frame.Parsed]) {
+                    parsed.foldLeft(List.empty[Frame.Parsed]) {
                         case (acc, curr) =>
                             acc match
                                 case `curr` :: tail => acc
                                 case _              => curr :: acc
-                    }.reverse.map { frame =>
```
fwbrasil, line 93: Not sure why these `reverse` calls were here.

fwbrasil:
merging to unblock other work

PR #562, opened 2024-08-05
prelude: stream improvements

fwbrasil:
`Stream` currently places pending effects (`S`) as the first generic while `Layer` places it as the last one. I'm also working on a `Join` effect with an `S` parameter and wanted to introduce this convention to make all classes that take `S` place it at the end. I'd be ok with it as the first parameter as well but it'd be important to have a strong convention.

I've also introduced `stream.runForeach` and `Stream.runForeachChunk`.


fwbrasil:
merging to unblock other work

PR #563, opened 2024-08-08
prelude: revert pending type encoding to an opaque type

fwbrasil:
I'm porting the other modules to the new design in `kyo-prelude` but there's a performance regression that I can't see how to avoid with the new encoding in the current version of the Scala 3 compiler:

- Given the lack of specialization, the generated bytecode contains a significant amount of unexpected `<` boxing. Fortunately, the JIT is able to avoid most of these allocations but profiling sessions show `<` allocations in several benchmarks. It isn't a bottleneck in the benchmarks I analyzed but introduces some overhead.

- The compiler doesn't allow implementing interfaces with `AnyVal` generic type parameters. As a workaround, I was planning to use a wrapper class like in `MonadLawsTest` but the approach is too expensive in integrations like in `kyo-sttp`. The wrapper class essentially introduces a new allocation to all methods in [KyoSttpMonad](https://github.com/getkyo/kyo/blob/0279f30163759d834b1793cbba5bcf7d08f9f0c3/kyo-sttp/shared/src/main/scala/kyo/internal/KyoSttpMonad.scala#L9), which are heavily used by sttp's transformations. Given how critical the integrations with sttp and tapir are, this regression doesn't seem ideal.

As described in https://github.com/getkyo/kyo/issues/531, the purpose of the new `AnyVal` encoding was to allow arbitrary nesting of Kyo computations. Although it's a nice usability improvement, the current limitations of the Scala 3 compiler doesn't seem to make it worth it and finalizing the migration to the new design seems a much higher priority. I can create a ticket so we can try again later.

Related compiler tickets: https://github.com/scala/scala3/issues/11264 https://github.com/scala/scala3/issues/15532

### DIFF
--- a/kyo-prelude/jvm/src/test/scala/kyo2/MonadLawsTest.scala
+++ b/kyo-prelude/jvm/src/test/scala/kyo2/MonadLawsTest.scala
@@ -1,5 +1,6 @@
 package kyo2
 
+import Flat.unsafe.bypass
 import kyo2.Result
 import zio.Trace
 import zio.prelude.Equal
@@ -11,7 +12,7 @@ import zio.test.laws.*
 
 object MonadLawsTest extends ZIOSpecDefault:
 
-    case class Myo[+A](v: A < (Env[String] & Abort[String] & Emit[Int] & Var[Boolean]))
+    type Myo[+A] = A < (Env[String] & Abort[String] & Emit[Int] & Var[Boolean])
 
     val listGenF: GenF[Any, Myo] =
         new GenF[Any, Myo]:
@@ -48,17 +49,17 @@ object MonadLawsTest extends ZIOSpecDefault:
                             _ <- Abort.when(s.length() > 10)("length exceeded")
                         yield v
                     )
-                ).map(Myo(_))
+                )
             end apply
 
     given CovariantDeriveEqualIdentityFlatten[Myo] =
         new CovariantDeriveEqualIdentityFlatten[Myo]:
             override def flatten[A](ffa: Myo[Myo[A]]): Myo[A] =
-                Myo(ffa.v.flatMap(_.v))
+                ffa.flatten
             override def any: Myo[Any] =
-                Myo(())
+                ()
             override def map[A, B](f: A => B): Myo[A] => Myo[B] =
-                m => Myo[B](m.v.map(f(_)))
+                _.map(f(_))
             override def derive[A: Equal]: Equal[Myo[A]] =
                 new Equal[Myo[A]]:
                     protected def checkEqual(l: Myo[A], r: Myo[A]): Boolean =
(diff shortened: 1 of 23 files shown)


kyo-prelude/jvm/src/test/scala/kyo2/MonadLawsTest.scala
```diff
@@ -1,5 +1,6 @@
 package kyo2
 
+import Flat.unsafe.bypass
```
fwbrasil, line 3: No need for a wrapper class anymore but we need to bypass the `Flat` check. It's safe since `zio-laws` can only compare two monad instances by calling the provided `Equal` and should never rely on the regular `==`, even if the monad instance is nested into another. In other words, the `A` provided to the `Equal` instance can not be nested.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Flat.scala
```diff
@@ -0,0 +1,88 @@
+package kyo2.kernel
+
+import kyo.Tag
+import scala.quoted.*
+
+opaque type Flat[T] = Null
```
fwbrasil, line 6: copied from `kyo-core`. I'll make sure to restore the git history in next PRs.

kyo-prelude/shared/src/main/scala/kyo2/kernel/Pending.scala
```diff
@@ -7,15 +7,12 @@ import scala.annotation.tailrec
 import scala.language.implicitConversions
 import scala.util.NotGiven
 
-// TODO Constructor should be private but there's an issue with inlining
-case class <[+A, -S](private val curr: A | Kyo[A, S]) extends AnyVal
+opaque type <[+A, -S] = A | Kyo[A, S]
```
fwbrasil, line 10: This is declared as `opaque type <[+A, -S] >: A = A | Kyo[A, S]` in `kyo-core`. Note that I didn't add the `>: A` back, which provides automatic lifting via the type system. The approach using implicit conversion has better usability because it can avoid unexpected nesting of computations (see `NotGiven` in `lift` bellow).

kyo-prelude/shared/src/main/scala/kyo2/package.scala
```diff
@@ -6,8 +6,12 @@ import kyo2.kernel.Safepoint
 private[kyo2] type Frame = kernel.Frame
 private[kyo2] inline def Frame = kernel.Frame
 
-export kernel.<
```
fwbrasil, line 9: I had to deal with a few cryptic compiler crashes while working on this change that went away once I avoided `export`s here.

kyo-prelude/jvm/src/test/scala/kyo2/MonadLawsTest.scala
```diff
@@ -48,30 +49,33 @@ object MonadLawsTest extends ZIOSpecDefault:
...
                             Var.run(true)(
                                 Emit.run(
                                     Abort.run(
-                                        Env.run("test")(m.v)
+                                        Env.run("test")(m)
                                     )
                                 )
                             ).eval._2
-
-                        run(l).equals(run(r))
```
fwbrasil, line 74: This was incorrect since it wasn't using the provided `Equal[A]`.

hearnadam (approved):

Issue #429, opened 2024-05-28
Rename this project to Kyos


hearnadam:
Given the implications of pending effects being plural, @kitlangton and I think it's only fitting that we rename the project.

Otherwise we plan to commit mutiny.

carlosedp:
To give @fwbrasil more work... I'd prefer to have all effects singular... like `IO`, `Fiber`, `Console` and so on... :)

fwbrasil:
For background on the current design: the initial versions of Kyo where focused on providing effects for existing monads like `Future`/`Futures`, `Try`/`Tries`, etc. Using plurals was a good way to avoid having to have effects named as `FutureEffect`, `TryEffect`, etc.

The design changed over time and the parity with monads is not as common. For example, even `Fibers` use a `FiberGets` effect under the hood. Given that, I think we could move away from plurals for effects.

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/572

Issue #340, opened 2024-05-08
Improve UX around streams


calvinlfer:
Capturing our convo on Discord
https://discord.com/channels/1087005439859904574/1087005440623259681/1237487763235016724

I started off by trying write the following
```scala
def select(query: Statement[?]): Stream[Unit, Row, Fibers] < Fibers =
    def go(rs: AsyncResultSet, channel: Channel[Chunk[Row] | Stream.Done]): Channel[Chunk[Row] | Stream.Done] < Fibers =
      // fetch the next result-set if need be
      val next =
        if rs.hasMorePages() then Fibers.fromCompletionStage(rs.fetchNextPage()).map(go(_, channel))
        else channel.put(Stream.Done).map(_ => channel)

      // emit the current result-set
      if rs.remaining() > 0 then channel.put(Chunks.initSeq(rs.currentPage().asScala.toSeq)).map(_ => next)
      else next

    Channels
      .init[Chunk[Row] | Stream.Done](1024, Access.Spmc)
      .map: channel =>
        val firstRs = Fibers.fromCompletionStage(session.executeAsync(query))
        val out     = firstRs.map(go(_, channel))
        out.flatMap(_ => Streams.initChannel(channel)) // is this how you lift the channel into the stream?
```

When you showed me that Stream was a wrapper on T < Streams[V] & S
```scala
def example[T, V, S, S2](s: Stream[T, V, S] < S2): Stream[T, V, S & S2] = 
  Streams.initSource(s.map(_.get))
```

I was wondering if the opaque type around Stream buys us anything

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/572

PR #572, opened 2024-08-14
port codebase to the new design in kyo-prelude

### LINKED ISSUE #462: New Encoding Again—Maybe

fwbrasil:
Fixes #462, #448, #429, #368, #340, #317, #229, #213

Following up on https://github.com/getkyo/kyo/pull/489, this PR migrates the rest of the codebase to the new design. Main highlights:

1. **Async optimizations**: The new `Async` effect, previously `Fibers`, has reduced allocations (no `Fiber` and `Promise` wrapper classes anymore) and `race` and `parallel` can now avoid forking if computations are already complete.
2. **`Fiber` error tracking**: Fibers now take an `E` type parameter that is automatically inferred from/to the `Abort` effect by `fiber.get` and `Async.run`. 
3. **Forking with effects**: The new `Boundary` mechanism in `kyo-prelude` is used to allow `Async`, `Abort`, `IO`, and any other context effects (currently `Env`, `Local`, `Resource`, and other effects based on those like `Console`).
4. **Simplified `IO` effect**: In this new design, the `IO` effect isn't responsible for handling unexpected failures since they're covered by `Abort` now. 

To preserve the git history, I first renamed all files to their new locations and then updated them with their new contents. Github can't show that in the default PR view, see https://github.com/getkyo/kyo/commit/620a593894b8067999213bbf34072eb9e3aa17e0 for a cleaner diff. For the cases where the file moved from `kyo-core` to `kyo-prelude`, I've kept the history of the original in `kyo-core` instead of the new one.

This change leaves pending tasks I'll create issues for:

1. Update the documentation. I've disabled the `readme` build for now.
2. Reintroduce `Stream.buffer`. The new mechanism for stream early termination makes it harder to implement it efficiently so I've removed it for now alongside its benchmark.

Most [benchmarks](https://gist.githubusercontent.com/fwbrasil/1a94d231024613ab57b66ec159f629f2/raw/f5df8ae9a6a76dfcf047589c271522719b187914/jmh-result-candidate.json) have a significant regression with the new design. The main reason is the new bookkeeping for tracing and to provide stack safety in `Safepoint`. It's not ideal but the other improvements seem worth it. It'll also give a more realistic opportunity for other effect systems to catch up and I'll definitely have some fun exploring new optimizations :) 

image

For perspective, I've collected benchmark [results](https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/f5bb216c54622e29e9eeb82267bc2093/raw/376255acbffad825932111bb039fb4c3c40f3c23/jmh-result.json) for all effect systems with this branch.


build.sbt
```diff
@@ -77,6 +77,7 @@ lazy val kyoJVM = project
         `kyo-bench`.jvm,
         `kyo-test`.jvm,
         `kyo-zio`.jvm,
+        `kyo-combinators`.jvm,
```
fwbrasil, line 80: `kyo-combinators` was missing in the aggregate project

build.sbt
```diff
@@ -318,6 +321,13 @@ lazy val `kyo-examples` =
         .dependsOn(`kyo-core`)
         .settings(
             `kyo-settings`,
+            fork := true,
```
fwbrasil, line 324: configs to make `sbt run` work out-of-the box to run the example server

kyo-bench/src/main/scala/kyo/bench/Bench.scala
```diff
@@ -1,17 +1,11 @@
 package kyo.bench
 
-import cats.effect.IO
```
fwbrasil, line 3: Imports reorganized to avoid conflicts now that both kyo and cats-effect have an `IO` type.

kyo-bench/src/main/scala/kyo/bench/Bench.scala
```diff
@@ -1,17 +1,11 @@
...
-import kyo.*
 import org.openjdk.jmh.annotations.*
-import zio.UIO
-import zio.ZLayer
 
 @State(Scope.Benchmark)
 @Fork(
     value = 1,
     jvmArgs = Array(
-        "-Dcats.effect.tracing.mode=DISABLED",
```
fwbrasil, line 14: Tracing was disabled for cats-effect to make the results more fair since Kyo didn't have tracing. The default `CACHED` config also isn't very fair because the equivalent of Kyo's tracing would be `FULL` but cats-effect has a prohibitive performance cost with it.

kyo-bench/src/main/scala/kyo/bench/BroadFlatMapBench.scala
```diff
@@ -20,24 +20,13 @@ class BroadFlatMapBench extends Bench.SyncAndFork(BigInt(610)):
...
 
-        def kyoFib(n: Int): BigInt < IOs =
+        def kyoFib(n: Int): BigInt < IO =
             if n <= 1 then BigInt(n)
             else kyoFib(n - 1).flatMap(a => kyoFib(n - 2).flatMap(b => a + b))
 
         kyoFib(depth)
     end kyoBench
 
-    @Benchmark
```
fwbrasil, line 30: These `Kyo2` benchmarks were temporary to access the performance of the new design in `kyo-prelude`.

kyo-bench/src/main/scala/kyo/bench/DeepBindBench.scala
```diff
@@ -7,8 +7,8 @@ class DeepBindBench extends Bench.SyncAndFork(()):
     def kyoBench() =
         import kyo.*
 
-        def loop(i: Int): Unit < IOs =
-            IOs {
```
fwbrasil, line 11: `IOs` was used to ensure stack safety via suspension. Since Kyo's execution is stack-safe by default now, the suspension isn't necessary anymore, which makes Kyo's impl of the benchmark closer to the others.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -0,0 +1,392 @@
...
+        using
+        ev: E => Throwable | ER,
+        f: Flat[A],
+        reduce: Reducible[Abort[ER]]
+    ): A < (S & reduce.SReduced) =
+        Abort.run[Throwable](effect.asInstanceOf[A < (Abort[Throwable | ER] & S)])
+            .map(_.fold(e => throw e.getFailure)(identity))
+end extension
+
+class SomeAbortToChoiceOps[A, S, E, E1 <: E](effect: A < (Abort[E] & S)) extends AnyVal:
```
fwbrasil, line 211: @jiveshungerford The new `Reducible` mechanism worked well for other APIs but it got quite complicated here. I had to add these `Ops` classes + make the methods require an additional `()`.

fwbrasil: I've also removed the `Options` APIs since the effect got removed.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -0,0 +1,368 @@
...
+import scala.util.control.NonFatal
+import scala.util.control.NoStackTrace
+
+opaque type Async <: (IO & Async.Join) = Async.Join & IO
+
+object Async:
+
+    sealed trait Join extends Effect[IOPromise[?, *], Result[Nothing, *]]
+
+    inline def run[E, A: Flat, Ctx](inline v: => A < (Abort[E] & Async & Ctx))(
```
fwbrasil, line 30: `Fibers` used to have `run` and `init` with a similar role returning `Fiber` instances. I've kept only `run` in this new design.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -0,0 +1,368 @@
...
+        using
+        boundary: Boundary[Ctx, IO],
+        frame: Frame
+    ): A < (Abort[E | Timeout] & IO & Ctx) =
+        run(v).map { fiber =>
+            IO(Abort.get(fiber.block(deadline(timeout))))
+        }
+    end runAndBlock
+
+    opaque type Promise[E, A] <: Fiber[E, A] = IOPromise[E, A]
```
fwbrasil, line 48: No wrapper class allocations for `Fiber` and `Promise` anymore.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -0,0 +1,368 @@
...
+    ): A < (Abort[E | Timeout] & Async & Ctx) =
+        boundary { (trace, context) =>
+            val task = IOTask(v, trace, context)
+            Timer.schedule(d)(task.completeUnit(Result.fail(Timeout(frame)))).map { t =>
+                IO.ensure(t.cancel.unit)(Async.get(task))
+            }
+        }
+    end timeout
+
+    def race[E, A: Flat, Ctx](seq: Seq[A < (Abort[E] & Async & Ctx)])(
```
fwbrasil, line 252: I've moved these methods from the `Joins` class and removed it

kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -0,0 +1,123 @@
+package kyo
+
+import kyo.internal.LogPlatformSpecific
+
+object Log extends LogPlatformSpecific:
```
fwbrasil, line 5: I've removed the `debug` methods here. I'm planning to introduce a new `Debug` effect based on `Safepoint.Interceptor`.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -1,151 +1,181 @@
...
+import kyo.scheduler.Scheduler
 import scala.annotation.tailrec
 import scala.util.control.NonFatal
+import scala.util.control.NoStackTrace
 
-private[kyo] class IOPromise[T](state: State[T])
-    extends AtomicReference(state):
+private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Interceptor:
+
+    @volatile private var state: State[E, A] = init
```
fwbrasil, line 20: I've changed the code to use `state` via `VarHandle` so it's possible to extend from the `Safepoint.Interceptor` class and avoid interface dispatches. Regular `IOPromise`s have the super class but as a no-op and `IOTask` overrides the no-op impls.

kyo-sttp/shared/src/main/scala/kyo/Request.scala
```diff
@@ -0,0 +1,39 @@
...
+        local.let(b)(v)
+
+    type BasicRequest = RequestT[Empty, Either[FailedRequest, String], Any]
+
+    val basicRequest: BasicRequest = sttp.client3.basicRequest.mapResponse {
+        case Left(value)  => Left(FailedRequest(value))
+        case Right(value) => Right(value)
+    }
+
+    def apply[E, A](f: BasicRequest => Request[Either[E, A], Any])(using Frame): A < (Async & Abort[E]) =
```
fwbrasil, line 32: These methods used to ignore the error type provided by sttp

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -0,0 +1,53 @@
...
+import kyo.kernel.*
+import scala.util.control.NonFatal
+import scala.util.control.NoStackTrace
+import zio.Cause
+import zio.Task
+import zio.ZIO
+
+opaque type ZIOs <: Async = GetZIO & Async
+
+object ZIOs:
```
fwbrasil, line 13: This new impl propagate typed errors between Kyo and ZIO.

fwbrasil:
I've just pushed an important improvement restricting the APIs of the pending type to only non-nested computations, allowing only `flatten`. Nesting typically happens in Kyo when there's a user mistake, and providing a compilation error closer to where the nesting happens with a specific error message via `@implicitNotFound` should provide better usability. 

I introduced a new `Flat.Weak` evidence based on `NotGiven` for this change. While `Flat` guarantees that a type isn't nested even in generic contexts, `Flat.Weak` can be bypassed in generic contexts where a type isn't statically known to be nested. `Flat.Weak` is strictly for usability given that it doesn't provide guarantees.

Commit: https://github.com/getkyo/kyo/pull/572/commits/685eb0268672adfd39b6911b88324c12778a663f

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -0,0 +1,53 @@
...
+        Effect.suspendMap(Tag[GetZIO], task)(Abort.get(_))
+
+    def get[A](v: ZIO[Any, Nothing, A])(using Frame): A < ZIOs =
+        Effect.suspend(Tag[GetZIO], v)
+
+    inline def get[R: zio.Tag, E, A](v: ZIO[R, E, A])(using Tag[Env[R]], Frame): A < (Env[R] & ZIOs) =
+        compiletime.error("ZIO environments are not supported yet. Please handle them before calling this method.")
+
+    def run[E, A](v: A < (Abort[E] & ZIOs))(using Frame): ZIO[Any, E, A] =
+        ZIO.suspendSucceed {
```
hearnadam, line 26: I think this can just become `suspend`:
https://github.com/zio/zio/blob/series/2.x/core/shared/src/main/scala/zio/ZIO.scala#L4803-L4813

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -0,0 +1,53 @@
...
+import scala.util.control.NoStackTrace
+import zio.Cause
+import zio.Task
+import zio.ZIO
+
+opaque type ZIOs <: Async = GetZIO & Async
+
+object ZIOs:
+
+    def get[E >: Nothing: Tag, A](v: ZIO[Any, E, A])(using Frame): A < (Abort[E] & ZIOs) =
```
hearnadam, line 15: Should params be by name throughout?

kyo-test/shared/src/main/scala/kyo/test/KyoSpecDefault.scala
```diff
@@ -1,13 +1,12 @@
...
 
 import kyo.*
-import kyo.test.interop.*
 import zio.ZIO
 import zio.test.Spec
 
 abstract class KyoSpecDefault extends KyoSpecAbstract[KyoApp.Effects]:
     final override def run[In: Flat](v: => In < KyoApp.Effects): ZIO[Environment, Throwable, In] =
-        ZIO.fromKyoFiber(KyoApp.runFiber(timeout)(v)).flatMap(ZIO.fromTry)
+        ZIOs.run(Resource.run(v))
```
hearnadam, line 9: Nice cleanup!

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -0,0 +1,53 @@
...
+    def run[E, A](v: A < (Abort[E] & ZIOs))(using Frame): ZIO[Any, E, A] =
+        ZIO.suspendSucceed {
+            try
+                Effect.handle(Tag[GetZIO], v.map(r => ZIO.succeed(r): ZIO[Any, E, A]))(
+                    [C] => (input, cont) => input.flatMap(r => run(cont(r)).flatten)
+                ).pipe(Async.run).map { fiber =>
+                    ZIO.asyncInterrupt[Any, E, A] { cb =>
+                        fiber.unsafe.onComplete {
+                            case Result.Fail(ex)   => cb(ZIO.fail(ex))
+                            case Result.Panic(ex)  => cb(ZIO.die(ex))
```
hearnadam, line 34: These should be `Exit.fail` and `Exit.die` to avoid suspension.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -1,151 +1,181 @@
...
+                case p: Pending[E, A] @unchecked =>
                     false
-                case l: Linked[T] @unchecked =>
-                    loop(l.p)
+                case l: Linked[E, A] @unchecked =>
+                    isDoneLoop(l.p, i + 1)
                 case _ =>
                     true
-        loop(this)
+        isDoneLoop(this, 0)
```
hearnadam, line 47: What's the `i` for?

kyo-prelude/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -73,17 +90,26 @@ object `<`:
 
     // TODO Compiler crash if inlined
     extension [A, S](v: A < S)
```
hearnadam, line 92: Can the evidence be in the extension?

kyo-core/shared/src/main/scala/kyo/Async.scala
fwbrasil: I've followed the naming in [`Boundary`](https://github.com/getkyo/kyo/blob/719d67a411562aa540dc7f21d9e44b03d73e161a/kyo-prelude/shared/src/main/scala/kyo2/kernel/Boundary.scala#L8). It makes it easier for me to reason about the code because `S` typically represents an unconstrained set of pending effects. Since the set in this case is limited to `ContextEffect`s, naming it `Ctx` better indicates what the set contains. For example, someone looking at `IOTask` and seeing an `S` type param might think that it can contain arbitrary effects. Using `Ctx` there as well I think makes the code clearer.

kyo-core/shared/src/main/scala/kyo/Adder.scala
```diff
@@ -0,0 +1,25 @@
...
+    inline def add(v: Long)(using Frame): Unit < IO = IO(ref.add(v))
+    inline def decrement(using Frame): Unit < IO    = IO(ref.decrement())
+    inline def increment(using Frame): Unit < IO    = IO(ref.increment())
+    inline def get(using Frame): Long < IO          = IO(ref.sum())
+    inline def reset(using Frame): Unit < IO        = IO(ref.reset())
+    inline def sumThenReset(using Frame): Long < IO = IO(ref.sumThenReset())
+end LongAdder
+
+object LongAdder:
+    def init(using Frame): LongAdder < IO = IO(LongAdder(new j.LongAdder))
```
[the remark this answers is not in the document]
fwbrasil: I'm not sure I understood, can you elaborate more?

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -0,0 +1,368 @@
...
+
+    opaque type Fiber[E, A] = IOPromise[E, A]
+
+    object Fiber extends FiberPlatformSpecific:
+
+        inline given [E, A]: Flat[Fiber[E, A]] = Flat.unsafe.bypass
+
+        private val _unit = success(())
+
+        def unit[E]: Fiber[E, Unit] = _unit.asInstanceOf[Fiber[E, Unit]]
```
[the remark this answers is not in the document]
fwbrasil: hmm, I think the explicit cast is easier to follow

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -0,0 +1,368 @@
...
+            import scala.util.*
+            IO {
+                val p = new IOPromise[Nothing, A]()
+                f.onComplete {
+                    case Success(v) =>
+                        p.complete(Result.success(v))
+                    case Failure(ex) =>
+                        p.complete(Result.panic(ex))
+                }(ExecutionContext.parasitic)
+                p
```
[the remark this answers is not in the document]
fwbrasil: It does but then the result would be a `Fiber[Throwable, A]` instead of `Fiber[Nothing, A]`. I'm honestly not sure what error scenarios we should track but tracking `Throwable` doesn't seem very valuable since `Panic` is essentially that.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -0,0 +1,368 @@
...
+                IO(self.interrupt(error))
+
+            def interruptUnit(error: Panic)(using Frame): Unit < IO =
+                IO(discard(self.interrupt(error)))
+
+            private[kyo] inline def unsafe: IOPromise[E, A] = self
+
+        end extension
+
+        case class Interrupted(at: Frame) extends NoStackTrace
```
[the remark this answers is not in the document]
fwbrasil: Added. I've just noticed that I had forgotten to include this optimization:  https://github.com/getkyo/kyo/commit/0765846add87d245108701e83e6a46c5336910f0. Fixed as well.

kyo-core/shared/src/main/scala/kyo/Atomic.scala
```diff
@@ -0,0 +1,65 @@
+package kyo
+
+import java.util.concurrent.atomic as j
+
+class AtomicInt private[kyo] (ref: j.AtomicInteger) extends AnyVal:
```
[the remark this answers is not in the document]
fwbrasil: That'd be nice and we can probably ensure specialization via inlining. It seem a good follow-up!

kyo-core/shared/src/main/scala/kyo/Meter.scala
```diff
@@ -0,0 +1,139 @@
...
+    def run[A, S](v: => A < S)(using Frame): A < (S & Async)
+
+    def tryRun[A, S](v: => A < S)(using Frame): Maybe[A] < (IO & S)
+
+    def close(using Frame): Boolean < IO
+end Meter
+
+object Meter:
+
+    def initNoop: Meter =
```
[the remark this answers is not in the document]
fwbrasil: good point!

kyo-core/shared/src/main/scala/kyo/Adder.scala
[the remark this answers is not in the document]
hearnadam: The comment is generally irrelevant to the line.

It's preferable to avoid expanding `Frame`s in Kyo's codebase, as the goal is to find where the methods are called in user code. For this reason, it may be useful to disable expanding the `Frame` to a value in Kyo's code.

ZIO does so with an implicit 'marker' class:
https://github.com/zio/zio/blob/series/2.x/stacktracer/shared/src/main/scala/zio/stacktracer/TracingImplicits.scala
https://github.com/zio/zio/blob/series/2.x/stacktracer/shared/src/main/scala-3/zio/internal/stacktracer/Macros.scala#L51-L67

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -1,131 +1,121 @@
...
-                        end match
-                    else
-                        IOs(bug.failTag(kyo, Tag.Intersection[FiberGets & IOs]))
-                case _ =>
-                    complete(Result.success(curr.asInstanceOf[T]))
-                    finalize()
-                    nullIO
-        end if
-    end eval
+    final override def addEnsure(f: () => Unit) =
```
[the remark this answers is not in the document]
fwbrasil: I think in this case making `()` explicit helps given the importance of avoiding evaluating the thunk

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
fwbrasil: I had tried it initially but I get this error:

```
Found:    zio.ZIO[Any, E, A]
Required: zio.RIO[Nothing, Any]
```

It seems `ZIO.suspend` doesn't support typed errors in the produced computation.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
fwbrasil: I've added it to `run` but `get` doesn't seem possible because we need the concrete `ZIO` value to suspend the computation with

kyo-core/shared/src/main/scala/kyo/Async.scala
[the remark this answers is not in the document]
hearnadam: Given that `Future` tracks Throwables, I think it makes sense that should be tracked in the Fiber as well.

Panics should usually be reserved for unrecoverable errors, for which I don't think we expect from `Future`.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
hearnadam: Ah right, this is because `RIO` is an alias for `ZIO[R, Throwable, A]`. In this case we need to support arbitrary `E` types. I think it would make sense to use `Exit.die` to propagate unexpected errors to ZIO's error channel.

Can you use `ZIO.fatalWith`?

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -1,131 +1,121 @@
...
-        st: Locals.State,
-        ensures: Ensures,
-        runtime: Int
-    ): IOTask[T] =
-        val f =
-            if st eq Locals.State.empty then
-                new IOTask[T](v, ensures)
+object IOTask:
+
+    private val _frame                = Frame.derive
```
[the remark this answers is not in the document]
hearnadam: How would you feel about a `Frame.empty`/`Frame.omitted`?

kyo-core/shared/src/main/scala/kyo/Adder.scala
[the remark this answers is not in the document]
fwbrasil: Thanks for the clarification! I agree it'd be good to have this restriction mechanism. Maybe we could have some more strict in the macro to always fail to compile in case the frame is in a symbol under the `kyo` package?

kyo-core/shared/src/main/scala/kyo/Async.scala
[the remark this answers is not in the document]
fwbrasil: There are a few other places where I didn't track exceptions as well, for example when `Closed` can be thrown in `Channel`. Could we create a "review tracked errors" ticket and address this separately?

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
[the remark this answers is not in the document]
fwbrasil: The cost of the string itself is low and having at least some information of the frame can help with debugging as well. For example, if you're in a debugger and there's a transformation from `IOTask`, it's possible to see where it came from.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
fwbrasil: I imagine you mean `isFatalWith` to handle the exception? I've just updated it.

kyo-core/shared/src/main/scala/kyo/Async.scala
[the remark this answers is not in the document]
hearnadam: Sure that's fine

kyo-core/shared/src/main/scala/kyo/Atomic.scala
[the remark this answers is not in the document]
hearnadam: Sounds good

kyo-core/shared/src/main/scala/kyo/Adder.scala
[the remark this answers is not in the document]
hearnadam: That would be great, but we may also need the flag for library authors at some point.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -0,0 +1,57 @@
...
+        ZIO.suspendSucceed {
+            try
+                Effect.handle(Tag[GetZIO], v.map(r => ZIO.succeed(r): ZIO[Any, E, A]))(
+                    [C] => (input, cont) => input.flatMap(r => run(cont(r)).flatten)
+                ).pipe(Async.run).map { fiber =>
+                    ZIO.asyncInterrupt[Any, E, A] { cb =>
+                        fiber.unsafe.onComplete {
+                            case Result.Fail(ex)   => cb(Exit.fail(ex))
+                            case Result.Panic(ex)  => cb(Exit.die(ex))
+                            case Result.Success(v) => cb(v)
```
hearnadam, line 36: Does this need `Exit.succeed`?

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -0,0 +1,57 @@
...
+                        }
+                        Left(ZIO.succeed {
+                            fiber.unsafe.interrupt(interrupt)
+                        })
+                    }
+                }.pipe(IO.run).eval
+            catch
+                case ex =>
+                    ZIO.isFatalWith { isFatal =>
+                        if !isFatal(ex) then Exit.Failure(Cause.die(ex))
```
hearnadam, line 46: minor: `Exit.die`?

Would it be possible to use `Exit.fail` when `E <:< Throwable`? Or is it not worth the effort?

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -0,0 +1,57 @@
...
+                        else throw ex
+                    }
+        }
+    end run
+
+    private[kyo] object internal:
+        class ZIOsInterrupt extends NoStackTrace
```
hearnadam, line 53: Minor: a message would be nice here. Maybe we make our own

`abstract class NoStackTrace(message: String) extends Exception(message, null, true, false)`

kyo-prelude/shared/src/main/scala/kyo/Kyo.scala
hearnadam, line 1: As discussed offline, it would be good to remove the `seq` nesting + potentially the `Seq` specialization of the related methods, though that could be done in a followup.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
fwbrasil: No, the result is already a `ZIO[Any, E, A]`

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
fwbrasil: I wouldn't expect a failure here. Any exceptions in `v` would already be handled by `Async.run` so this should catch only bugs

hearnadam (approved):

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
fwbrasil: Actually, we don't need the custom exception here. I've changed it to use `Fiber.Interrupted`. I'll follow up on a custom `NoStackTrace`.

PR #578, opened 2024-08-20
make Chunk a Seq

fwbrasil:
This PR makes `Chunk` extend from `Seq` so we can avoid additional allocations to transform between the types and improve usability. It's not ideal since Scala's `Seq` methods aren't well-optimized but it seems a good tradeoff.

The methods that used to take effectful functions like `map` and `filter` were moved to `Kyo.*` and the methods from `Kyo.seq.*` are now available directly in `Kyo.*`.

### DIFF
--- a/kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
+++ b/kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
@@ -40,27 +40,13 @@ extension (kyoObject: Kyo.type)
     )(
         useElement: PartialFunction[A, A1 < S1]
     ): Seq[A1] < (S & S1) =
-        sequence.flatMap((seq: Seq[A]) => Kyo.seq.collect(seq.collect(useElement)))
+        sequence.flatMap((seq: Seq[A]) => Kyo.collect(seq.collect(useElement)))
 
     def debugln[S](message: => String < S): Unit < (S & IO) =
         message.map(m => Console.println(m))
 
     def fail[E, S](error: => E < S): Nothing < (S & Abort[E]) =
         error.map(e => Abort.fail(e))
-
-    def foreach[A, S, A1, S1](
-        sequence: => Seq[A] < S
-    )(
-        useElement: A => A1 < S1
-    ): Seq[A1] < (S & S1) =
-        collect(sequence)(a => useElement(a))
-
-    def foreachDiscard[A, S, S1](
-        sequence: => Seq[A] < S
-    )(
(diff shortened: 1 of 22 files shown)


kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
```diff
@@ -181,13 +167,13 @@ extension (kyoObject: Kyo.type)
...
     ): Seq[A] < (S & S1) =
-        sequence.flatMap((seq: Seq[A < S]) => Kyo.seq.collect(seq))
+        sequence.flatMap((seq: Seq[A < S]) => Kyo.collect(seq))
 
     def traverseDiscard[A, S, S1](
         sequence: => Seq[A < S] < S1
     ): Unit < (S & S1) =
         sequence.flatMap { (seq: Seq[A < S]) =>
-            Kyo.seq.collect(seq.map(_.map(_ => ()))).unit
+            Kyo.collect(seq.map(_.map(_ => ()))).unit
```
hearnadam, line 176: is there a reason not to use `_.unit`?

kyo-prelude/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -15,121 +15,202 @@ object Kyo:
...
+                    case seq =>
+                        val indexed = toIndexed(seq)
+                        Loop.indexed(Chunk.empty[B]) { (idx, acc) =>
+                            if idx == size then Loop.done(acc)
+                            else f(indexed(idx)).map(u => Loop.continue(acc.append(u)))
+                        }
+                end match
+
+    def foreachDiscard[A, B, S](seq: Seq[A])(f: A => Unit < S)(using Frame): Unit < S =
+        seq.size match
```
hearnadam, line 39: I think you want to use `knownSize` to avoid iterating `List`s.

hearnadam (approved):
LGTM aside from `size` issue.

kyo-prelude/shared/src/main/scala/kyo/Kyo.scala
fwbrasil: thanks! I didn't know `knownSize`

kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
fwbrasil: yeah, I've changed it to use `Kyo.collectDiscard`

PR #579, opened 2024-08-20
debug effect

fwbrasil:
This new effect uses the `Safepoint.Interceptor` mechanism introduced by https://github.com/getkyo/kyo/pull/489 and the code snippets captured by `Frame` to output debugging information. The `Debug.apply` method only logs the result of the computation with its frame while `Debug.trace` outputs all transformations of a computation with their values. 

There's an important limitation: the `Safepoint.Interceptor` currently isn't propagated to forked fibers so the debugging won't output the trace of computations in forked fibers. I'll follow up on this limitation but it's a bit tricky because we need to avoid increasing the memory footprint of `IOTask`.

Example output of `Debug.trace`:

image

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -166,6 +166,7 @@ lazy val `kyo-prelude` =
         .in(file("kyo-prelude"))
         .settings(
             `kyo-settings`,
+            libraryDependencies += "com.lihaoyi" %%% "pprint"        % "0.9.0",
             libraryDependencies += "org.jctools"   % "jctools-core"  % "4.0.5",
             libraryDependencies += "dev.zio"     %%% "zio-laws-laws" % "1.0.0-RC27" % Test,
             libraryDependencies += "dev.zio"     %%% "zio-test-sbt"  % "2.1.2"      % Test,
@@ -182,7 +183,6 @@ lazy val `kyo-core` =
         .in(file("kyo-core"))
         .settings(
             `kyo-settings`,
-            libraryDependencies += "com.lihaoyi"  %%% "pprint"          % "0.9.0",
             libraryDependencies += "org.jctools"    % "jctools-core"    % "4.0.5",
             libraryDependencies += "org.slf4j"      % "slf4j-api"       % "2.0.16",
             libraryDependencies += "dev.zio"      %%% "zio-laws-laws"   % "1.0.0-RC27" % Test,
--- a/kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
+++ b/kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
@@ -195,6 +195,15 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
                     v.asInstanceOf[Result[E | Timeout, A]]
         blockLoop(this)
     end block
+
+    override def toString =
+        val stateString =
+            state match
+                case p: Pending[?, ?] => s"Pending(waiters = ${p.waiters})"
+                case l: Linked[?, ?]  => s"Linked(promise = ${l.p})"
+                case r                => s"Done(result = ${r.asInstanceOf[Result[Any, Any]].show})"
+        s"IOPromise(state = ${stateString})"
+    end toString
 end IOPromise
 
 private[kyo] object IOPromise extends IOPromisePlatformSpecific:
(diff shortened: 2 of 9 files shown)


build.sbt
```diff
@@ -152,6 +152,7 @@ lazy val `kyo-prelude` =
         .in(file("kyo-prelude"))
         .settings(
             `kyo-settings`,
+            libraryDependencies += "com.lihaoyi" %%% "pprint"        % "0.9.0",
```
fwbrasil, line 155: I wasn't sure if I should add the effect to `kyo-prelude` or `kyo-core`. Although it isn't a pure effect, it seems important to make it available for codebases that have only `kyo-prelude`. I moved the `pprint` to `kyo-prelude` to use it when outputting the debugging info.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -194,6 +194,15 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
                     v.asInstanceOf[Result[E | Timeout, A]]
         blockLoop(this)
     end block
+
```
fwbrasil, line 197: Just improving `toString` to show in the debug output.

kyo-prelude/shared/src/main/scala/kyo/debug/Debug.scala
```diff
@@ -0,0 +1,51 @@
+package kyo
+
+import Ansi.*
+import kyo.kernel.Effect
+import kyo.kernel.Safepoint
+
+object Debug:
+
+    def apply[A, S](v: => A < S)(using frame: Frame): A < S =
```
fwbrasil, line 9: I decided to use `println` directly and avoid introducing suspensions so it's easy to use the effect anywhere. For example, if the method had `IO` suspensions, adding `Debug` to a computation would require the user to add the `IO` effect if it's not already present. I think it's ok since it's meant only for debugging.

fwbrasil: I also considered using `Log` but it's common for users to have trouble with logs not appearing due to missing configs.

kyo-prelude/shared/src/main/scala/kyo/kernel/Frame.scala
```diff
@@ -21,7 +21,7 @@ object Frame:
     ) derives CanEqual:
 
         def show: String =
-            Ansi.highlight(s"// $declaringClass $methodName", snippetLong, s"// $position", position.lineNumber)
+            Ansi.highlight(s"// $position $declaringClass $methodName", snippetLong, s"", position.lineNumber)
```
fwbrasil, line 24: I've moved the file position to the header and left the footer empty to make the rendering easier to follow.

kyo-prelude/shared/src/main/scala/kyo/debug/Debug.scala
```diff
@@ -0,0 +1,51 @@
...
+                    value
+                }
+            } { ex =>
+                printValue(ex)
+                throw ex
+            }
+        }
+    end trace
+
+    private def printValue(value: Any) =
```
hearnadam, line 46: there seems like an opportunity to print the frame as well, but perhaps not all the usages need that.

build.sbt
hearnadam: Can you me why we need pprint?

hearnadam (approved):

build.sbt
fwbrasil: It's used here to print the values in the debug log. It inspects the types and outputs their structure with ansi colors instead of calling `toString`. It uses reflection for that. Example:

image

kyo-prelude/shared/src/main/scala/kyo/debug/Debug.scala
fwbrasil: Both callers of this method print the frame but in different order. `Debug.apply` logs the frame and then the value since it's the result of the transformation but `Debug.trace` prints the value before the frame because it's the input.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -194,6 +194,15 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
...
+    override def toString =
```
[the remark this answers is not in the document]
fwbrasil: I've changed `Debug` to truncate values with more than 500 chars

fwbrasil:
I've added a new `Debug.values` method that brings the functionality we had in `Log` before. Example output:

image

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/debug/Debug.scala
sideeffffect: Alternatively, you could also consider using logging via java.util.logging. It's available everywhere, doesn't bring any dependency and is interoperable.

PR #582, opened 2024-08-21
Update kyo-combinators for kyo-prelude refactor

johnhungerford:
- Replaces old `Options` methods (`Kyo.fromOption`, `handleOption`, `optionToAborts`) with `Maybe`-based methods (`Kyo.fromMaybe`, `handleEmptyAbort`, `emptyAbortToFailure`) that use `Abort[Maybe.Empty]` to represent short-circuiting with an empty result
- Cleaned up references to `Seqs` changing everything to `Choice` (and other formerly plural effects, like `Aborts`)
- Support handling of all abort types by using `Abort.run[Any]`
- Added `provideLayer` method to support layers
- Updated docs to reflect change to `Seqs` and `Options`


kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -136,44 +136,43 @@ end extension
...
 extension [A, S, E](effect: A < (Abort[E] & S))
     def handleAbort(
-        using
-        ct: ClassTag[E],
-        tag: Tag[E],
-        flat: Flat[A]
+        using flat: Flat[A]
     ): Result[E, A] < S =
-        Abort.run[E](effect)
+        Abort.run[Any](effect).asInstanceOf[Result[E, A] < S]
```
johnhungerford, line 141: @fwbrasil what do you think of catching all errors? I see that `Tag` doesn't support union types, but there should be some way to handle all composed error types at one time.

The danger in this approach is that it will end up handling "hidden" `Abort` effect (e.g., opaque type alias on `Abort[X]` for some failure `X`).

fwbrasil: The previous behavior allowing type unions was unsound because the `ClassTag` gets inferred as `Any` in those cases, essentially handling any `Abort`, not only the ones in the type union. The `Tag` was added to `Abort.run` as a way to prevent this scenario. We might be able to eventually lift this limitation with the new tag implementation I'm planning to work on.  

I suggest keeping it restricted here as well to avoid issues.

kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
```diff
@@ -65,11 +65,17 @@ extension (kyoObject: Kyo.type)
...
+    def fromEither[E, A, S](either: => Either[E, A] < S): A < (S & Abort[E]) =
+        either.map(Abort.get(_))
 
-    def fromResult[E, A](result: => Result[E, A]): A < Abort[E] =
-        Abort.get(result)
+    def fromOption[A, S](option: => Option[A] < S): A < (S & Abort[Maybe.Empty]) =
+        option.map(o => Abort.get(o.toRight[Maybe.Empty](Maybe.Empty)))
+
+    def fromMaybe[A, S](maybe: => Maybe[A] < S): A < (S & Abort[Maybe.Empty]) =
+        maybe.map(m => Abort.get(m.toRight[Maybe.Empty](Maybe.Empty)))
```
hearnadam, line 75: You can directly invoke `Abort.get`:
https://github.com/getkyo/kyo/blob/main/kyo-prelude/shared/src/main/scala/kyo/Abort.scala#L52-L54

johnhungerford: Missed that -- thanks!

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -208,6 +205,27 @@ extension [A, S, E](effect: A < (Abort[E] & S))
             .map(_.fold(e => throw e.getFailure)(identity))
 end extension
 
+extension [A, S, E](effect: A < (Abort[Maybe.Empty] & S))
+    def handleEmptyAbort(using f: Flat[A]): Maybe[A] < S =
```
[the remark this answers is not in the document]
johnhungerford: I've tried to be consistent with the method names for handling and converting error types. Also, this method doesn't really "unmaybe" anything, does it? It produces a `Maybe`.

fwbrasil (approved):
Thank you @johnhungerford! Sorry that I left a bit a mess 🫣

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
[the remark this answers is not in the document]
hearnadam: I was thinking that combinators would follow ZIO's general naming conventions, eg https://github.com/zio/zio/blob/series/2.x/core/shared/src/main/scala/zio/ZIO.scala#L2316-L2322

I guess this would be `unmaybe`

hearnadam (approved):

fwbrasil (commented):
add missing frames

PR #583, opened 2024-08-21
introduce kyo-data

fwbrasil:
This PR moves data types from `kyo-prelude` to a new isolated module `kyo-data`: `TypeMap`, `Duration`, `Result`, and `Maybe`. The main goal is enabling adoption without the effect system given how generic the data types are. The isolation should be helpful for long-term maintenance and the separate module also improves build times by allowing more pipelining.


kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -8,7 +8,6 @@ opaque type Maybe[+A] >: (Empty | Defined[A]) = Empty | Defined[A]
 object Maybe:
     inline given [A, B](using inline ce: CanEqual[A, B]): CanEqual[Maybe[A], Maybe[B]] = CanEqual.derived
     given [A]: Conversion[Maybe[A], IterableOnce[A]]                                   = _.iterator
-    inline given [A]: Flat[Maybe[A]]                                                   = Flat.unsafe.bypass
```
fwbrasil, line 11: `Flat` is in `kyo-prelude` so it can't be used in `kyo-data`

kyo-prelude/shared/src/main/scala/kyo/kernel/Flat.scala
```diff
@@ -66,6 +70,15 @@ private object FlatMacro:
         def isConcrete(t: TypeRepr) =
             t.typeSymbol.isClassDef
 
+        def isKyoData(t: TypeRepr): Boolean =
```
fwbrasil, line 73: My initial attempt was adding these as evidences in the `Flat` companion object but I had trouble to make it work well with type inference and the presence of the additional implicits hide the macro error message because it fails due to ambiguous implicits instead.

kyo-data/shared/src/test/scala/kyo/ResultTest.scala
```diff
@@ -372,18 +371,18 @@ class ResultTest extends Test:
             assert(flatMapped == Success("mapped: null"))
         }
 
-        "Fail with null error" taggedAs jvmOnly in {
```
fwbrasil, line 375: not sure why this was here

kyo-data/shared/src/main/scala/kyo/package.scala
```diff
@@ -0,0 +1,8 @@
+package kyo
+
+private[kyo] inline def isNull[A](v: A): Boolean =
+    v.asInstanceOf[AnyRef] eq null
+
+private[kyo] inline def discard[A](v: A): Unit =
```
hearnadam, line 6: Should this be by-name? I know the `inline` should take care of that, but it's still a bit odd to see.

hearnadam (approved):

kyo-data/shared/src/main/scala/kyo/package.scala
fwbrasil: good catch!

PR #584, opened 2024-08-22
move Chunk to kyo-data

fwbrasil:
I forgot to move `Chunk` to `kyo-data` in https://github.com/getkyo/kyo/pull/583.


kyo-data/shared/src/test/scala/kyo/ChunkTest.scala
```diff
@@ -203,34 +203,6 @@ class ChunkTest extends Test:
             val result = chunk.map(_ * 2)
             assert(result == Chunk(2, 4, 6, 8, 10))
         }
-        "with a function using Env and Abort" in {
```
fwbrasil, line 206: Not necessary anymore since `Chunk` doesn't handle effectful functions anymore. The `Kyo.*` methods are already covered in `KyoTest`

hearnadam (approved):

PR #585, opened 2024-08-22
ContextEffect.handle for constant values

fwbrasil:
A minor API improvement providing an overloaded method `ContextEffect.handle` that takes a constant value instead of `ifDefined`/`ifUndefined`.


kyo-prelude/shared/src/main/scala/kyo/kernel/ContextEffect.scala
```diff
@@ -39,7 +39,17 @@ object ContextEffect:
                     continue = f(context.getOrElse(_tag, default).asInstanceOf[A])
                 )
 
-    inline def handle[A: Flat, E <: ContextEffect[A], B, S](
+    inline def handle[A, E <: ContextEffect[A], B, S](
```
fwbrasil, line 42: The `Flat` evidence here was redundant

hearnadam (approved):

PR #586, opened 2024-08-22
Refactor effect types

fwbrasil:
I'm working on documenting custom effects and the current naming using `Effect` and `ContextEffect` is a bit confusing. The `Effect` name doesn't communicate much meaning and I have to explain in the documentation that it's an abstract arrow.

This PR refactors the effect types with an abstract `Effect` and two sub-classes: `ArrowEffect` and `ContextEffect`.


kyo-prelude/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -0,0 +1,289 @@
+package kyo.kernel
+
+import internal.*
+import kyo.Tag
+import scala.annotation.tailrec
+import scala.util.control.NonFatal
+
+abstract class ArrowEffect[-I[_], +O[_]] extends Effect
```
fwbrasil, line 8: The files are properly renamed in the git history but github doesn't render the rename. The individual commits should be easier to review.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -46,5 +46,5 @@ object ZIOs:
         }
     end run
 
-    sealed private[kyo] trait GetZIO extends Effect[ZIO[Any, Nothing, *], Id]
+    sealed private[kyo] trait GetZIO extends ArrowEffect[ZIO[Any, Nothing, *], Id]
```
hearnadam, line 49: Initially I was thinking this was a problem, but actually the context effects will be shifted into `Env` when we get line 20 working.

kyo-prelude/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
hearnadam: I think your rebase may have broke this.

hearnadam (commented):
Naming seems good to me. I'll check the diff closely once you fix the commits.

kyo-prelude/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
fwbrasil: sorry, fixed now!

hearnadam (approved):

PR #588, opened 2024-08-22
rename Loop.Result to Loop.Outcome

fwbrasil:
Given that we have `kyo.Result` now

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/kernel/Loop.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/kernel/Loop.scala
@@ -24,56 +24,56 @@ object Loop:
         private[Loop] def _4: D
     end Continue4
 
-    opaque type Result[A, O]           = O | Continue[A]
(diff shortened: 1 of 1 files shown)


kyo-prelude/shared/src/main/scala/kyo/kernel/Loop.scala
```diff
@@ -24,56 +24,56 @@ object Loop:
         private[Loop] def _4: D
     end Continue4
 
-    opaque type Result[A, O]           = O | Continue[A]
-    opaque type Result2[A, B, O]       = O | Continue2[A, B]
-    opaque type Result3[A, B, C, O]    = O | Continue3[A, B, C]
-    opaque type Result4[A, B, C, D, O] = O | Continue4[A, B, C, D]
+    opaque type Outcome[A, O]           = O | Continue[A]
```
hearnadam, line 27: Could also be called `Done`

hearnadam (approved):

PR #589, opened 2024-08-22
avoid generating internal frames

fwbrasil:
This PR makes the `Frame` macro fail to compile in case the frame is in the `kyo` package and fixes the methods in the codebase that were missing the implicit.


kyo-bench/src/main/scala/kyo/bench/Bench.scala
```diff
@@ -42,6 +42,8 @@ end Bench
 
 object Bench:
 
+    inline given kyo.kernel.Frame = kyo.kernel.Frame.internal
```
fwbrasil, line 45: There are some situations where an internal frame is necessary. `Frame.internal` is an `private[kyo]` API that allows generating a frame within the `kyo` package.

kyo-examples/jvm/src/main/scala/examples/ledger/Models.scala
```diff
@@ -1,4 +1,4 @@
-package kyo.examples.ledger
+package examples.ledger
```
fwbrasil, line 1: Move the examples out of the `kyo` package so they can generate frames.

kyo-caliban/src/main/scala/kyo/Schema.scala
```diff
@@ -17,6 +17,7 @@ given zioSchema[R, A: Flat, S](using ev: Schema[R, A], ev2: (A < S) <:< (A < (Ab
             ev.toType_(isInput, isSubscription)
 
         override def resolve(value: A < S): Step[R] =
+            given Frame = Frame.internal
```
fwbrasil, line 20: There are some situations where an internal frame is necessary. Frame.internal is an private[kyo] API that allows generating a frame within the kyo package.

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/kernel/Frame.scala
```diff
@@ -49,15 +50,37 @@ object Frame:
...
+        if !internal then
+            findEnclosing { sym =>
+                val fileName = pos.sourceFile.name
+                sym.fullName.startsWith("kyo") && !allowKyoFileSuffixes.exists(fileName.endsWith)
+            }.foreach { sym =>
+                report.errorAndAbort(
+                    s"""Frame cannot be derived within the kyo package: ${sym.owner.fullName}
+                       |
+                       |To resolve this issue:
+                       |1. Propagate the Frame from user code using an implicit parameter.
```
hearnadam, line 75: `implicit`?? What is this, Scala 2.13?

hearnadam (commented):
Looks like there is an issue with kyo-sttp files under JS

PR #596, opened 2024-08-22
readme: Duration

fwbrasil:


kyo-data/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -92,7 +92,7 @@ object Duration:
                 case Duration.Zero => JavaDuration.ZERO
                 case n             => JavaDuration.of(n.toNanos, NANOS)
 
-        inline def render: String = s"Duration($self ns)"
+        inline def show: String = s"Duration($self ns)"
```
fwbrasil, line 95: to make it more consistent with other APIs

hearnadam (approved):

PR #597, opened 2024-08-22
readme: Ansi

fwbrasil:


README.md
```diff
@@ -2096,6 +2096,55 @@ val q: Try[Int] = a.toTry
...
+
+// Strip ANSI codes from a string
+val plainText: String = "\u001b[31mColored\u001b[0m".stripAnsi
+
+// Highlight code snippets
+val code = """
+def hello(name: String): Unit =
+println(s"Hello, $name!")
+"""
+lazy val highlightedCode: String = Ansi.highlight(code)
```
fwbrasil, line 2128: `mdoc` uses `fansi` and it fails with an exception in case the string already has ansi codes. I've used `lazy val` to avoid the code being evaluated

hearnadam (approved):

PR #604, opened 2024-08-24
Env.runLayer can handle effects introduced by layers

johnhungerford:
Also separated `provideValue` and` provideLayer` from provide in kyo-combinators. `provide` now provides all layers, just like in ZIO

### DIFF
--- a/kyo-combinators/shared/src/test/scala/kyo/EnvCombinatorTest.scala
+++ b/kyo-combinators/shared/src/test/scala/kyo/EnvCombinatorTest.scala
@@ -21,7 +21,7 @@ class EnvCombinatorTest extends Test:
         "handle" - {
             "should provide" in {
                 val effect: Int < Env[String] = Env.get[String].map(_.length)
-                assert(effect.provide("value").eval == 5)
+                assert(effect.provideValue("value").eval == 5)
             }
 
             "should provide incrementally" in {
@@ -29,10 +29,10 @@ class EnvCombinatorTest extends Test:
                     Env.get[String] *> Env.get[Int] *> Env.get[Boolean] *> Env.get[Char].as(23)
                 val handled =
                     effect
-                        .provide('c')
-                        .provide("value")
-                        .provide(1)
-                        .provide(false)
+                        .provideValue('c')
+                        .provideValue("value")
+                        .provideValue(1)
+                        .provideValue(false)
                 assert(handled.eval == 23)
             }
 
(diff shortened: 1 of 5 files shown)


kyo-prelude/shared/src/main/scala/kyo/Env.scala
```diff
@@ -29,10 +29,12 @@ object Env:
     ): A < (S & reduce.SReduced) =
         reduce(ContextEffect.handle(erasedTag[R], env, _.union(env))(v): A < (Env[VR] & S))
 
-    transparent inline def runLayer[A, S, V](inline layers: Layer[?, ?]*)(value: A < (Env[V] & S)): A < (S & Memo) =
-        Layer.init[V](layers*).run.map { env =>
-            runTypeMap(env)(value)
-        }.asInstanceOf[A < (S & Memo)]
+    transparent inline def runLayer[A, S, V](inline layers: Layer[?, ?]*)(value: A < (Env[V] & S)): A < (S & Memo & Nothing) =
```
fwbrasil, line 32: Is the `Nothing` required?

fwbrasil: I imagine there's some limitation with the macro but wouldn't `runLayer[A, S1, S2, V](inline layers: Layer[?, S1]*)(value: A < (Env[V] & S2))` work?

johnhungerford: Doesn't work without including `Nothing`: `A < (S & Memo & Nothing)` provides an upper bound that the `transparent inline` can make more specific.

I also wasn't able to make it work with the type parameter `S1` in `Layer[?, S1]`. I'm not really sure I understand why, however. It may be that it lifts all the `In` parameters in `Layer[Out, Int]` to the same type, which makes the macro not work properly.

kyo-combinators/shared/src/test/scala/kyo/EnvCombinatorTest.scala
```diff
@@ -57,6 +57,24 @@ class EnvCombinatorTest extends Test:
                         .provideLayer(layerBool)
                 assert(Memo.run(handled).eval == 23)
             }
+
+            "should provide all layers" in {
```
fwbrasil, line 61: Can you add a tests checking that `provide` and `Env.runLayer` infers the expected pending effects? I normally do:

```scala
val x = ...
val y: ExpectedType = x
```

johnhungerford: I added them. Do you want them in separate tests?

fwbrasil: that looks good, thanks!

PR #605, opened 2024-08-25
Scala 3.5

fwbrasil:
The compiler introduced a new compilation message when an inlined method has an anonymous class warning that it'll generate a new class on the caller. That's the intended behavior in Kyo so I've added `@nowarn` to suppress it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ import org.typelevel.scalacoptions.ScalacOption
 import org.typelevel.scalacoptions.ScalacOptions
 import org.typelevel.scalacoptions.ScalaVersion
 
-val scala3Version   = "3.4.3"
+val scala3Version   = "3.5.0"
 val scala212Version = "2.12.19"
 val scala213Version = "2.13.14"
 
@@ -19,7 +19,7 @@ val compilerOptions = Set(
     ScalacOptions.warnValueDiscard,
     ScalacOptions.languageStrictEquality,
     ScalacOptions.release("11"),
-    ScalacOptions.privateKindProjector
+    ScalacOptions.advancedKindProjector
 )
 
 ThisBuild / scalaVersion := scala3Version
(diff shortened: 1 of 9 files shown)


build.sbt
```diff
@@ -19,7 +19,7 @@ val compilerOptions = Set(
     ScalacOptions.warnValueDiscard,
     ScalacOptions.languageStrictEquality,
     ScalacOptions.release("11"),
-    ScalacOptions.privateKindProjector
+    ScalacOptions.advancedKindProjector
```
fwbrasil, line 22: This option changed in 3.5

guizmaii: Have you tried the new `usePipelining` option? Did you see any compiling speed improvement?

fwbrasil: Yep, but got compiler crashes with macros :( I tried disabling pipelining for specific projects as a workaround but no luck

```
[error] -- Error: /Users/fwbrasil/workspace/kyo/kyo-prelude/shared/src/main/scala/kyo/internal/BaseKyoTest.scala:18:82
[error] 18 |    def run(v: Assertion < S): Future[Assertion] = run(v.map(Future.successful(_)))
[error]    |                                                                                  ^
[error]    |Macro code depends on the top-level definitions in package kyo in package kyo found on the classpath, but could not be loaded while evaluating the macro.
[error]    |  This is likely because class files could not be found in the classpath entry for the symbol.
[error]    |
[error]    |  A possible cause is if the origin of this symbol was built with pipelined compilation;
[error]    |  in which case, this problem may go away by disabling pipelining for that origin.
[error]    |
[error]    |  the top-level definitions in package kyo is defined in file /Users/fwbrasil/workspace/kyo/kyo-data/jvm/target/scala-3.5.0/early/kyo-data_3-0.11.0+12-7b4939f5+20240824-1922-SNAPSHOT.jar(kyo/Maybe$package.tasty)
[error]    |----------------------------------------------------------------------------
[error]    |Inline stack trace
[error]    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]    |This location contains code that was inlined from Frame.scala:53
[error] 53 |    implicit inline def derive: Frame = ${ frameImpl(false) }
[error]    |                                        ^^^^^^^^^^^^^^^^^^^^^
[error]     ----------------------------------------------------------------------------
[error] one error found
```

hearnadam (approved):

PR #610, opened 2024-08-26
fix release of JS artifacts

fwbrasil:
The split of the project into `kyoJVM` and `kyoJS` by https://github.com/getkyo/kyo/pull/450 made the release workflow not publish JS artifacts since only `kyoJVM` is selected by default.


fwbrasil:
I'm merging to test since the workflow runs only in `main`.

PR #612, opened 2024-08-27
fix typo in Async section

vkurdin:
*Async.init* is a non-existent function. *Async.run* is the real one.


fwbrasil:
thank you @vkurdin!

PR #613, opened 2024-08-28
Return Unit for `TypeMap#get[Any]` on empty TypeMap

### LINKED ISSUE #608: TypeMap.empty.get[Any] throws fatal
```
import kyo.*

TypeMap.empty.get[Any]
```
```
Caused by: java.lang.RuntimeException: fatal: kyo.TypeMap of contents [TypeMap()] missing value of type: [scala.Any].
```

Either we need to make `empty` `Map(Tag[Any] -> ())` OR we need to special case.

hearnadam:
Fixes #608

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/TypeMap.scala
+++ b/kyo-data/shared/src/main/scala/kyo/TypeMap.scala
@@ -21,7 +21,8 @@ object TypeMap:
                 end while
                 fatal
             end search
-            self.getOrElse(t.erased, search).asInstanceOf[B]
+            if isEmpty && t =:= Tag[Any] then ().asInstanceOf[B]
+            else self.getOrElse(t.erased, search).asInstanceOf[B]
         end get
 
         inline def add[B](b: B)(using inline t: Tag[B]): TypeMap[A & B] =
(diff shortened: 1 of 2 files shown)


kyo-data/shared/src/main/scala/kyo/TypeMap.scala
```diff
@@ -21,7 +21,8 @@ object TypeMap:
                 end while
                 fatal
             end search
-            self.getOrElse(t.erased, search).asInstanceOf[B]
+            if isEmpty && t =:= Tag[Any] then ().asInstanceOf[B]
```
fwbrasil, line 24: Could we have issues with other super types like `AnyRef`, `AnyVal`, `Matchable`?

hearnadam: No because those don't conform to the type bound `Any`. I can add a test for it if you want.

`TypeMap.empty.get[AnyRef]` fails to compile.

fwbrasil: I see, makes sense. I think tests for that can be useful to avoid issues with future changes.

PR #614, opened 2024-08-29
rename top-level files package.scala to avoid assembly conflicts

scalavision:
I get these errors from sbt-assembly plugin:

```bash
[error] (server / assembly)
[error] Deduplicate found different file contents in the following:
[error]   Jar name = kyo-data_3-0.11.0.jar, jar org = io.getkyo, entry target = kyo/package$package.class
[error]   Jar name = kyo-prelude_3-0.11.0.jar, jar org = io.getkyo, entry target = kyo/package$package.class
[error] Deduplicate found different file contents in the following:
[error]   Jar name = kyo-data_3-0.11.0.jar, jar org = io.getkyo, entry target = kyo/package$package$.class
[error]   Jar name = kyo-prelude_3-0.11.0.jar, jar org = io.getkyo, entry target = kyo/package$package$.class
[error] Total time: 3 s, completed Aug 29, 2024, 9:11:12 AM
```

when trying to create a fat jar having dependencies on kyo, this fixes the issue. I don't know if these names makes sense though, feel free to suggest otherwise.

I notice there are also two other files named package.scala:

```
kyo-prelude/shared/src/main/scala/kyo/kernel/package.scala
kyo-scheduler/jvm/src/main/scala/kyo/scheduler/package.scala
```

but they did not cause a problem for me, so keeping them as is.

Thank you for creating this awesome project!


hearnadam:
Thanks for the PR. Merging now to ensure this is included in upcoming release. I may revisit the naming of the scheduler `package`

PR #619, opened 2024-09-02
kernel: EffectTS-inspired 'pipe' methods

fwbrasil:
This PR introduces overloaded `pipe` methods for up to six functions similarly to EffectTS but it's a method on the pending type, not in a companion object https://effect.website/docs/guides/essentials/pipeline

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Clock.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Clock.scala
@@ -13,7 +13,7 @@ object Clock:
 
     private val local = Local.init(live)
 
-    def let[A, S](c: Clock)(f: => A < (IO & S))(using Frame): A < (IO & S) =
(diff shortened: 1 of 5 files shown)


kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -13,7 +13,7 @@ object Clock:
 
     private val local = Local.init(live)
 
-    def let[A, S](c: Clock)(f: => A < (IO & S))(using Frame): A < (IO & S) =
```
fwbrasil, line 16: Minor unrelated cleanup. This used to require `IO` because `Local` was implemented based on it but it's not the case anymore.

hearnadam (approved):
LGTM, though this is not my favorite style 🙂 

It may be worth adding a combinator alias `|>`

PR #620, opened 2024-09-02
scheduler: optimize worker cycling

fwbrasil:
I'm doing some testing where a large number of workers are created and the mechanism that periodically cycles workers introduces significant overhead.

### DIFF
--- a/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
+++ b/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
@@ -1,12 +1,14 @@
 package kyo.scheduler
 
 import Scheduler.*
+import java.util.concurrent.Callable
 import java.util.concurrent.Executor
 import java.util.concurrent.Executors
 import java.util.concurrent.ScheduledExecutorService
 import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.LongAdder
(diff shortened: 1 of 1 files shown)


kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -176,22 +178,25 @@ final class Scheduler(
         }
 
     private val cycleTask =
-        timerExecutor.scheduleAtFixedRate(
-            () => cycleWorkers(),
-            cycleNs,
-            cycleNs,
-            TimeUnit.NANOSECONDS
+        timerExecutor.submit(
```
fwbrasil, line 181: Instead of relying on Java's scheduled executor, which has additional overhead, the cycling now steals a thread and runs in a while loop with a fixed delay interval.

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -176,22 +178,25 @@ final class Scheduler(
...
+                }
+            ): Callable[Unit]
         )
 
     private def cycleWorkers(): Unit = {
         try {
             val nowMs    = clock.currentMillis()
             var position = 0
-            while (position < allocatedWorkers) {
+            while (position < currentWorkers) {
```
fwbrasil, line 197: There was no need to cycle idle workers since they can't receive new workload and exit as soon as possible.

hearnadam (approved):

PR #621, opened 2024-09-02
bench: introduce JIT profile warmup mechanism

fwbrasil:
A major downside of microbenchmarks in a JIT-compiled runtime like the JVM is how the execution can be extremely different from how the code would behave in a real application. This happens because benchmarks typically use a very narrow subset of the code and the profiling information collected by the JVM to generate optimized native code doesn't represent well how the code is actually used.

This PR introduces a warmup mechanism that executes several of the existing benchmarks to populate the JIT profile with more realistic information. I did just an initial analysis of Kyo's results and created https://github.com/getkyo/kyo/pull/620 but I'm planning to continue working on it. The mechanism is disabled by default for now.

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/Bench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/Bench.scala
@@ -1,6 +1,10 @@
 package kyo.bench
 
+import WarmupJITProfile.*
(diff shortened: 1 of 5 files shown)


fwbrasil:
merging to reduce the overhead of maintaining multiple branches

PR #622, opened 2024-09-02
prelude: scaladocs for effects

fwbrasil:


kyo-prelude/shared/src/main/scala/kyo/Aspect.scala
```diff
@@ -1,35 +1,32 @@
...
-    def init[A, B, S](default: Cut[A, B, S])(using Frame): Aspect[A, B, S] =
-        new Aspect[A, B, S](default)
-
-    def chain[A, B, S](head: Cut[A, B, S], tail: Seq[Cut[A, B, S]])(using Frame) =
-        tail.foldLeft(head)(_.andThen(_))
-end Aspect
-
 import Aspect.*
 
-abstract class Cut[A, B, S]:
```
fwbrasil, line 22: I've moved `Cut` to the companion object to avoid polluting the `kyo` package namespace. The companion object is also moved to after the class for consistency.

fwbrasil:
merging to reduce the overhead of maintaining multiple branches

PR #625, opened 2024-09-03
kernel: inline <.unit

fwbrasil:
Addressing a TODO. I couldn't see a significant change in benchmark results, though.


hearnadam (approved):
Seems fine, but my preference would be to avoid the added complexity.

fwbrasil:
I think it's worth ensuring the methods in the pending type are optimized as much as possible since they're the building blocks.

PR #628, opened 2024-09-03
core: introduce System

### LINKED ISSUE #627: Add ZIO style System core effect.
I'm coming from ZIO and was looking to the equivalent of what I would do in ZIO with kyo, and noticed that kyo currently doesn't have a core System effect/service.

I would expect this to provide the same functionality as [ZIO's](https://zio.dev/reference/services/system).

In summary:

- [ ] Access the environment using `env`
- [ ] Access properties via `properties`
- [ ] Access `lineSeparator`

I may also suggest providing support for providing the User's home directory in a platform agnostic manner, but this is admittedly beyond the scope of what ZIO's `System` service provides.

fwbrasil:
fixes #627

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/System.scala
+++ b/kyo-core/shared/src/main/scala/kyo/System.scala
@@ -1,0 +1,140 @@
+package kyo
+
+import System.Parser
+import java.lang.System as JSystem
+import kyo.kernel.Reducible
+
+abstract class System:
+    def env[E, A](name: String)(using Parser[E, A], Frame): Maybe[A] < (Abort[E] & IO)
+    def property[E, A](name: String)(using Parser[E, A], Frame): Maybe[A] < (Abort[E] & IO)
+    def lineSeparator(using Frame): String < IO
+    def userName(using Frame): String < IO
+    def userHome(using Frame): String < IO
+
+end System
+
+object System:
+
+    val local = Local.init(live)
+
+    val live: System =
+        new System:
+            def env[E, A](name: String)(using p: Parser[E, A], frame: Frame): Maybe[A] < (Abort[E] & IO) =
+                val value = JSystem.getenv(name)
+                if value == null then Maybe.empty
+                else p(value).map(Maybe(_))
+            end env
+
+            def property[E, A](name: String)(using p: Parser[E, A], frame: Frame): Maybe[A] < (Abort[E] & IO) =
+                val value = JSystem.getProperty(name)
+                if value == null then Maybe.empty
+                else p(value).map(Maybe(_))
+            end property
+
+            def lineSeparator(using Frame): String < IO = IO(JSystem.lineSeparator())
+
+            def userName(using Frame): String < IO = IO(JSystem.getProperty("user.name"))
+
+            def userHome(using Frame): String < IO = IO {
+                Option(JSystem.getProperty("user.home"))
+                    .orElse(Option(JSystem.getenv("HOME")))
+                    .getOrElse {
+                        val osName = JSystem.getProperty("os.name").toLowerCase
+                        if osName.contains("win") then
+                            JSystem.getenv("USERPROFILE")
+                        else
+                            "/home/" + JSystem.getProperty("user.name")
+                        end if
+                    }
+            }
+
+    def let[A, S](system: System)(f: => A < S)(using Frame): A < S =
+        local.let(system)(f)
+
+    class EnvOps[A](dummy: Unit) extends AnyVal:
+
+        def apply[E](name: String)(
+            using
+            parser: Parser[E, A],
+            frame: Frame,
+            reduce: Reducible[Abort[E]]
+        ): Maybe[A] < (reduce.SReduced & IO) =
+            reduce(local.use(_.env[E, A](name)))
+
+        def apply[E](name: String, default: => A)(
+            using
+            parser: Parser[E, A],
+            frame: Frame,
+            reduce: Reducible[Abort[E]]
+        ): A < (reduce.SReduced & IO) =
+            reduce(local.use(_.env[E, A](name).map(_.getOrElse(default))))
+
+    end EnvOps
+
+    def env[A]: EnvOps[A] = EnvOps(())
+
+    class PropertyOps[A](dummy: Unit) extends AnyVal:
+        def apply[E](name: String)(
+            using
+            parser: Parser[E, A],
+            frame: Frame,
+            reduce: Reducible[Abort[E]]
+        ): Maybe[A] < (reduce.SReduced & IO) =
+            reduce(local.use(_.property[E, A](name)))
+
+        def apply[E](name: String, default: => A)(
+            using
+            parser: Parser[E, A],
+            frame: Frame,
+            reduce: Reducible[Abort[E]]
+        ): A < (reduce.SReduced & IO) =
+            reduce(local.use(_.property[E, A](name).map(_.getOrElse(default))))
+    end PropertyOps
+
+    def property[A]: PropertyOps[A] = PropertyOps(())
+
+    def lineSeparator(using Frame): String < IO = local.use(_.lineSeparator)
+    def userName(using Frame): String < IO      = local.use(_.userName)
+    def userHome(using Frame): String < IO      = local.use(_.userHome)
+
+    abstract class Parser[E, A]:
+        def apply(s: String): A < Abort[E]
+
+    object Parser:
+        given (using Frame): Parser[Nothing, String]                    = v => v
+        given (using Frame): Parser[NumberFormatException, Int]         = v => Abort.catching(v.toInt)
+        given (using Frame): Parser[NumberFormatException, Long]        = v => Abort.catching(v.toLong)
+        given (using Frame): Parser[NumberFormatException, Float]       = v => Abort.catching(v.toFloat)
+        given (using Frame): Parser[NumberFormatException, Double]      = v => Abort.catching(v.toDouble)
+        given (using Frame): Parser[IllegalArgumentException, Boolean]  = v => Abort.catching(v.toBoolean)
+        given (using Frame): Parser[NumberFormatException, Byte]        = v => Abort.catching(v.toByte)
+        given (using Frame): Parser[NumberFormatException, Short]       = v => Abort.catching(v.toShort)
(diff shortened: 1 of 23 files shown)


kyo-core/shared/src/main/scala/kyo/System.scala
```diff
@@ -0,0 +1,140 @@
...
+object System:
+
+    val local = Local.init(live)
+
+    val live: System =
+        new System:
+            def env[E, A](name: String)(using p: Parser[E, A], frame: Frame): Maybe[A] < (Abort[E] & IO) =
+                val value = JSystem.getenv(name)
+                if value == null then Maybe.empty
+                else p(value).map(Maybe(_))
```
hearnadam, line 25: Can't this just be `Maybe(value).map(p)`?

kyo-core/shared/src/main/scala/kyo/System.scala
```diff
@@ -0,0 +1,140 @@
...
+    def userName(using Frame): String < IO
+    def userHome(using Frame): String < IO
+
+end System
+
+object System:
+
+    val local = Local.init(live)
+
+    val live: System =
```
hearnadam, line 20: Should this be private?

kyo-core/shared/src/main/scala/kyo/System.scala
```diff
@@ -0,0 +1,140 @@
...
+                else p(value).map(Maybe(_))
+            end property
+
+            def lineSeparator(using Frame): String < IO = IO(JSystem.lineSeparator())
+
+            def userName(using Frame): String < IO = IO(JSystem.getProperty("user.name"))
+
+            def userHome(using Frame): String < IO = IO {
+                Option(JSystem.getProperty("user.home"))
+                    .orElse(Option(JSystem.getenv("HOME")))
```
hearnadam, line 40: `Maybe`?

kyo-core/shared/src/main/scala/kyo/System.scala
```diff
@@ -0,0 +1,140 @@
...
+            v => Abort.catching(java.time.LocalDateTime.parse(v))
+
+        given (using Frame): Parser[java.net.URISyntaxException, java.net.URI] =
+            v => Abort.catching(new java.net.URI(v))
+
+        given (using Frame): Parser[java.net.MalformedURLException, java.net.URL] =
+            v => Abort.catching(new java.net.URL(v))
+
+        given [E, A](using p: Parser[E, A], frame: Frame): Parser[E, Seq[A]] =
+            v => Kyo.foreach(v.split(",").toIndexedSeq)(s => p(s.trim()))
```
hearnadam, line 133: How about `Chunk.fromArray`?

kyo-core/shared/src/main/scala/kyo/System.scala
```diff
@@ -0,0 +1,140 @@
...
+
+        given (using Frame): Parser[IllegalArgumentException, Char] =
+            v => if v.length == 1 then v.charAt(0) else Abort.fail(new IllegalArgumentException("String must have exactly one character"))
```
hearnadam, line 136: `Abort.when(p(v))(e).map(_ => v)` seems like a common pattern. Perhaps we should make a function.

kyo-core/shared/src/main/scala/kyo/System.scala
```diff
@@ -0,0 +1,140 @@
...
+
+end System
+
+object System:
+
+    val local = Local.init(live)
+
+    val live: System =
+        new System:
+            def env[E, A](name: String)(using p: Parser[E, A], frame: Frame): Maybe[A] < (Abort[E] & IO) =
```
hearnadam, line 22: Do we need an explicit IO suspension?

kyo-data/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -14,6 +14,26 @@ object Duration:
 
     given CanEqual[Duration, Duration] = CanEqual.derived
 
+    case class InvalidDuration(message: String) extends Exception(message)
+
+    def parse(s: String): Result[InvalidDuration, Duration] =
```
hearnadam, line 19: Nice! This will be useful.

kyo-core/shared/src/main/scala/kyo/System.scala
fwbrasil: It follows the pattern of `Clock`, `Console`, and `Timer`. I think having it public can use useful. For example, it allows implementing a `System` wrapper class with custom behaviors that dispatches to the live instance.

kyo-core/shared/src/main/scala/kyo/System.scala
fwbrasil: good idea! I've added `Chunk.from`.

kyo-core/shared/src/main/scala/kyo/System.scala
fwbrasil: I've added a new `Abort.ensuring` method as a more generic version of `Abort.when`: https://github.com/getkyo/kyo/pull/628/commits/62f2bd1b333808f084576394d74da903ad4a49d3

fwbrasil: I'm not very happy with the name, let me know if you have suggestions!

kyo-core/shared/src/main/scala/kyo/System.scala
fwbrasil: ouch! good catch

kyo-core/shared/src/main/scala/kyo/System.scala
fwbrasil: No, `Maybe.map` doesn't support effects so it produces a `kyo.Maybe[A < kyo.Abort[E]]`.

kyo-core/shared/src/main/scala/kyo/System.scala
```diff
@@ -0,0 +1,142 @@
...
+                    .orElse(Maybe(JSystem.getenv("HOME")))
+                    .getOrElse {
+                        val osName = JSystem.getProperty("os.name").toLowerCase
+                        if osName.contains("win") then
+                            JSystem.getenv("USERPROFILE")
+                        else
+                            "/home/" + JSystem.getProperty("user.name")
+                        end if
+                    }
+            }
```
sideeffffect, line 51: We could provide this and perhaps more directories, based on https://github.com/dirs-dev/directories-jvm

fwbrasil: Good call! I've just realized that `Path` is a better place to put these methods. I'll remove them here and create a separate PR.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -21,12 +21,12 @@ object Abort:
...
     inline def panic[E](inline ex: Throwable)(using inline frame: Frame): Nothing < Abort[E] = error(Panic(ex))
 
-    private[kyo] inline def error[E](inline e: Error[E])(using inline frame: Frame): Nothing < Abort[E] =
+    inline def error[E](inline e: Error[E])(using inline frame: Frame): Nothing < Abort[E] =
         ArrowEffect.suspendMap[Any](erasedTag[E], e)(_ => ???)
 
-    inline def when[E](b: Boolean)(inline value: => E)(using inline frame: Frame): Unit < Abort[E] =
-        if b then fail(value)
-        else ()
+    inline def ensuring[A, E](cond: Boolean, result: => A = ())(inline value: => E)(using inline frame: Frame): A < Abort[E] =
```
hearnadam, line 27: Not sure I like this change (at my own suggestion). I think it may be confusing to have default unit.

kyo-core/shared/src/main/scala/kyo/System.scala
```diff
@@ -0,0 +1,127 @@
...
+            reduce(local.use(_.property[E, A](name).map(_.getOrElse(default))))
+    end PropertyOps
+
+    def property[A]: PropertyOps[A] = PropertyOps(())
+
+    def lineSeparator(using Frame): String < IO = local.use(_.lineSeparator)
+    def userName(using Frame): String < IO      = local.use(_.userName)
+
+    abstract class Parser[E, A]:
+        def apply(s: String): A < Abort[E]
```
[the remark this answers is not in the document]
fwbrasil: good point, and it'd avoid having to pass the `Frame`. I'll update it.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
fwbrasil: maybe we could keep both `ensuring` and `when` to avoid the unit default? The new `ensuring` method seems quite useful

kyo-core/shared/src/main/scala/kyo/System.scala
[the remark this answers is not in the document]
fwbrasil: https://github.com/getkyo/kyo/pull/628/commits/dce2214ea58f3fc85c4361536be6529d24633f8b

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -167,6 +183,15 @@ object Result:
...
             flatMap(v => Result.success(f(v)))
 
+        inline def mapFail[E2](inline f: E => E2): Result[E2, A] =
+            self match
+                case Fail(e) =>
+                    try Fail(f(e))
+                    catch
+                        case ex => Panic(ex)
+                case Panic(ex) => Panic(ex)
+                case _         => self.asInstanceOf[Result[E2, A]]
```
hearnadam, line 193: Do we need the Panic case?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
hearnadam: That's fine with me!

hearnadam (approved):
LGTM (once you add back Abort.when 🙂)

kyo-data/shared/src/main/scala/kyo/Result.scala
fwbrasil: good catch! 🙏

PR #629, opened 2024-09-04
prelude: Check effect

fwbrasil:


kyo-prelude/shared/src/main/scala/kyo/Check.scala
```diff
@@ -0,0 +1,47 @@
+package kyo
+
+import Ansi.*
+import kyo.kernel.*
+
+class CheckFailed(val message: String, val frame: Frame) extends Exception(message):
```
hearnadam, line 6: extends `AssertionError`?

hearnadam (approved):

PR #633, opened 2024-09-05
prelude: Batch effect

fwbrasil:
An effect that provides functionality similar to ZIO Query.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Chunk.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Chunk.scala
@@ -258,7 +258,7 @@ sealed abstract class Chunk[A] extends Seq[A] derives CanEqual:
       * @return
       *   a flattened Chunk
       */
-    final def flatten[B](using ev: A =:= Chunk[B]): Chunk[B] =
+    final def flattenChunk[B](using ev: A =:= Chunk[B]): Chunk[B] =
         if isEmpty then Chunk.empty
         else
             val nested = this.toArrayInternal
@@ -280,7 +280,7 @@ sealed abstract class Chunk[A] extends Seq[A] derives CanEqual:
             copy()
 
             Compact(unnested)
-    end flatten
+    end flattenChunk
 
     /** Copies the elements of this Chunk to an array.
       *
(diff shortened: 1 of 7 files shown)


kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -135,7 +135,7 @@ sealed abstract class Chunk[A] extends Seq[A] derives CanEqual:
                 case c: Indexed[A] => c
                 case _             => Compact(toArrayInternal)
 
-    final def flatten[B](using ev: A =:= Chunk[B]): Chunk[B] =
+    final def flattenChunk[B](using ev: A =:= Chunk[B]): Chunk[B] =
```
fwbrasil, line 138: @hearnadam I also hit that issue with `flatten` not working if the nested collection isn't a `Chunk`.

kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -0,0 +1,93 @@
...
+        (v: A) => ArrowEffect.suspend[B](erasedTag[S], Op.Call(v, f, frame))
+
+    inline def eval[A](seq: Seq[A])(using inline frame: Frame): A < Batch[Any] =
+        ArrowEffect.suspend[A](erasedTag[Any], Op.Eval(seq))
+
+    def run[A: Flat, S, S2](v: A < (Batch[S] & S2))(using Frame): Seq[A] < (S & S2) =
+
+        case class Cont(op: Op[?, S], cont: Any => (Cont | A) < (Batch[S] & S2))
+        def runCont(v: (Cont | A) < (Batch[S] & S2)): (Cont | A) < S2 =
+            // TODO workaround, Flat macro isn't inferring correctly with nested classes
```
fwbrasil, line 33: I'll follow up on this. I couldn't find a simple fix.

kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -0,0 +1,93 @@
...
+            import Flat.unsafe.bypass
+            ArrowEffect.handle(erasedTag[S], v) {
+                [C] => (input, cont) => Cont(input, v => cont(v.asInstanceOf[C]))
+            }
+        end runCont
+
+        case class Expanded(v: Any, source: Seq[Any] => Seq[Any] < S, cont: Any => (Cont | A) < (Batch[S] & S2), frame: Frame)
```
fwbrasil, line 40: This code is not optimized since it uses several Scala collection operations. I think it's ok for an initial version so we can validate the effect and its functionality.

hearnadam (commented):
Seems fine overall - though I have no experience with zio-query.

PR #637, opened 2024-09-06
core: Clock stopwatch, deadline, and unsafe APIs

fwbrasil:
I'm planning to use the deadline API to fix `Async.block` that should use `Clock` instead of Java's `System` directly.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Clock.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Clock.scala
@@ -4,12 +4,60 @@ import java.time.Instant
 
 abstract class Clock:
     def now(using Frame): Instant < IO
+    def stopwatch(using Frame): Clock.Stopwatch < IO
+    def deadline(duration: Duration)(using Frame): Clock.Deadline < IO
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -4,12 +4,60 @@ import java.time.Instant
...
+
+    object Unsafe:
+        class Stopwatch(start: Instant, clock: Clock.Unsafe):
+            def elapsed: Duration =
+                Duration.fromJava(java.time.Duration.between(start, clock.now))
+
+        class Deadline(end: Instant, clock: Clock.Unsafe):
+            def timeLeft: Duration =
+                val remaining = java.time.Duration.between(clock.now, end)
+                if remaining.isNegative then Duration.Zero else Duration.fromJava(remaining)
```
fwbrasil, line 28: I think we should eventually have an optimized version of `Instance` that integrates well with `Duration`

hearnadam (approved):

PR #638, opened 2024-09-06
core: Path integration with directories-jvm

fwbrasil:
As suggested by @sideeffffect https://github.com/getkyo/kyo/pull/628#discussion_r1743429668


kyo-core/jvm/src/main/scala/kyo/Path.scala
```diff
@@ -339,22 +354,55 @@ object Path:
...
-                case _: Nil.type =>
-                    acc.reverse
-                case h :: t =>
-                    h match
-                        case h: String =>
-                            loop(t, h.split('/').filter(_.nonEmpty).toList.reverse ::: acc)
-                        case h: Path =>
-                            loop(h.path ::: t, acc)
-        new Path(loop(path, Nil))
+    def apply(parts: List[Part]): Path =
```
fwbrasil, line 357: I had issues with this method returning paths without the initial `/` and not handling `null` parts.

kyo-core/jvm/src/main/scala/kyo/Path.scala
```diff
@@ -331,6 +334,18 @@ class Path(val path: List[String]):
...
+    override def hashCode(): Int =
+        val prime  = 31
+        var result = 1
+        result = prime * result + (if path == null then 0 else path.hashCode)
+        result
+    end hashCode
+
+    override def equals(obj: Any): Boolean = obj match
+        case that: Path =>
+            (this eq that) || (that.isInstanceOf[Path] && this.path == that.path)
```
hearnadam, line 346: `that.isInstanceOf[Path]` should be unnecessary due to the match.

Why is the generated one not good enough?

kyo-core/jvm/src/main/scala/kyo/Path.scala
```diff
@@ -16,7 +19,7 @@ import scala.io.*
 import scala.jdk.CollectionConverters.*
 import scala.jdk.StreamConverters.*
 
-class Path(val path: List[String]):
+class Path(val path: List[String]) derives CanEqual:
```
hearnadam, line 22: Should we make the constructor private if we have a custom apply?

kyo-core/jvm/src/main/scala/kyo/Path.scala
```diff
@@ -339,22 +354,55 @@ object Path:
...
-                        case h: Path =>
-                            loop(h.path ::: t, acc)
-        new Path(loop(path, Nil))
+    def apply(parts: List[Part]): Path =
+        val flattened = parts.flatMap {
+            case p if isNull(p) => Nil
+            case s: String      => List(s)
+            case p: Path        => p.path
+        }
+        val javaPath       = if flattened.isEmpty then Paths.get("") else Paths.get(flattened.head, flattened.tail*)
```
hearnadam, line 363: This is now side effecting... not sure if we can avoid that however.

kyo-core/jvm/src/main/scala/kyo/Path.scala
```diff
@@ -339,22 +354,55 @@ object Path:
...
+        def download(using Frame): Path < IO = IO(Path(UserDirectories.get().downloadDir))
+        def font(using Frame): Path < IO     = IO(Path(UserDirectories.get().fontDir))
+        def picture(using Frame): Path < IO  = IO(Path(UserDirectories.get().pictureDir))
+        def public(using Frame): Path < IO   = IO(Path(UserDirectories.get().publicDir))
+        def template(using Frame): Path < IO = IO(Path(UserDirectories.get().templateDir))
+        def video(using Frame): Path < IO    = IO(Path(UserDirectories.get().videoDir))
+    end User
+
+    case class Project(qualifier: String, organization: String, application: String):
+        private def dirs                       = ProjectDirectories.from(qualifier, organization, application)
```
hearnadam, line 396: lazy val?

kyo-core/jvm/src/main/scala/kyo/Path.scala
fwbrasil: yep, it's unnecessary

kyo-core/jvm/src/main/scala/kyo/Path.scala
fwbrasil: hmm is Java's `Path` side effecting?

kyo-core/jvm/src/main/scala/kyo/Path.scala
```diff
@@ -339,22 +354,55 @@ object Path:
...
+
+    object Base:
+        def cache(using Frame): Path < IO      = IO(Path(BaseDirectories.get().cacheDir))
+        def config(using Frame): Path < IO     = IO(Path(BaseDirectories.get().configDir))
+        def data(using Frame): Path < IO       = IO(Path(BaseDirectories.get().dataDir))
+        def dataLocal(using Frame): Path < IO  = IO(Path(BaseDirectories.get().dataLocalDir))
+        def executable(using Frame): Path < IO = IO(Path(BaseDirectories.get().executableDir))
+        def preference(using Frame): Path < IO = IO(Path(BaseDirectories.get().preferenceDir))
+        def runtime(using Frame): Path < IO    = IO(Path(BaseDirectories.get().runtimeDir))
+    end Base
```
sideeffffect, line 380: Wouldn't it be better to create a Kyo's own `case class BaseDirectories(cache: Path, config: Path, ...)`.
And on `BaseDirectories` companion object have method
```scala
lazy val get: BaseDirectories < IO = ...
```

And the same for the other 2 cases.

kyo-core/jvm/src/main/scala/kyo/Path.scala
hearnadam: To my surprise, I don't think it is!

kyo-core/jvm/src/main/scala/kyo/Path.scala
fwbrasil: Makes sense, I've updated it. I can't be a `lazy val` because of `Frame`, though.

sideeffffect: Great, now it's perfect, thanks :pray:

PR #639, opened 2024-09-06
data: scaladocs

fwbrasil:


kyo-data/shared/src/main/scala/kyo/Ansi.scala
```diff
@@ -1,58 +1,112 @@
...
-                    if isHeader || isTrailer then
-                        line.green
-                    else
-                        highlightLine(line.drop(toDrop))
-
-                s"${lineNumber.dim} $separator $highlightedLine"
-            }
-
-            processedLines.mkString("\n")
+            try
```
fwbrasil, line 69: An unrelated improvement: this method is now used in several places to report errors. If it throws an exception, we'd lose visibility of the underlying failure so I've added a fallback to return the code without highlighting if that happens.

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -7,21 +7,47 @@ import scala.reflect.ClassTag
...
+      * @tparam A
+      *   The type of the successful result
+      * @return
+      *   A Result containing either the successful value or a Panic with the caught exception
+      */
     inline def apply[A](inline expr: => A): Result[Nothing, A] =
         try
             Success(expr)
         catch
             case ex => Panic(ex)
```
hearnadam, line 39: `NonFatal`? Line 56 as well?

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -32,10 +58,56 @@ object Result:
...
+      * @param value
+      *   The successful value
+      * @tparam E
+      *   The type of the error (not used in this case)
+      * @tparam A
+      *   The type of the successful value
+      * @return
+      *   A successful Result
+      */
+    inline def success[E, A](inline value: A): Result[E, A] = Success(value)
```
hearnadam, line 72: style: I think this should be renamed to `succeed` (perhaps I am too used to ZIO)

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -220,24 +533,51 @@ object Result:
...
+        /** Swaps the success and failure cases of the Result.
+          *
+          * @return
+          *   A new Result with success and failure swapped
+          */
         def swap: Result[A, E] =
             self match
                 case Fail(e)    => Result.success(e)
                 case Success(v) => Result.fail(v)
                 case _          => self.asInstanceOf[Result[A, E]]
```
hearnadam, line 574: Should we have a toString impl of `SuccessError`? `"Success(" * (depth - 1) + ...`

hearnadam (approved):
LGTM - mostly pre-existing comments.

kyo-data/shared/src/main/scala/kyo/Result.scala
fwbrasil: I put the `NonFatal` check in `Panic.apply` to reduce the size of the inlined bytecode.

hearnadam: Ah clever, I must have forgotten.

kyo-data/shared/src/main/scala/kyo/Result.scala
fwbrasil: Makes sense, I've added a `toString` to `DefinedEmpty` in `Maybe` as well: https://github.com/getkyo/kyo/pull/639/commits/852c90755c1761588cb88c235477e50d18cd3623

kyo-data/shared/src/main/scala/kyo/Result.scala
fwbrasil: let's follow up separately

kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -142,6 +403,9 @@ object Maybe:
                     Empty
             def nest =
                 DefinedEmpty(depth + 1)
+
+            override def toString: String =
+                "Maybe(" * depth + "Empty" + ")" * depth
```
hearnadam, line 408: I think this should be `Defined(`?

PR #640, opened 2024-09-06
core: scaladocs

fwbrasil:
Clock is covered by https://github.com/getkyo/kyo/pull/637

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Access.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Access.scala
@@ -1,10 +1,17 @@
 package kyo
 
-sealed abstract class Access derives CanEqual
+/** Represents different access patterns for concurrent queues.
+  */
+enum Access derives CanEqual:
+    /** Multiple Producer Multiple Consumer access pattern */
+    case Mpmc
 
-object Access:
-    case object Mpmc extends Access
-    case object Mpsc extends Access
-    case object Spmc extends Access
-    case object Spsc extends Access
+    /** Multiple Producer Single Consumer access pattern */
+    case Mpsc
+
+    /** Single Producer Multiple Consumer access pattern */
+    case Spmc
+
+    /** Single Producer Single Consumer access pattern */
+    case Spsc
 end Access
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -17,12 +17,35 @@ import scala.util.NotGiven
 import scala.util.control.NonFatal
 import scala.util.control.NoStackTrace
 
+/** Represents an asynchronous computation effect.
+  *
+  * This effect provides methods for running asynchronous computations, creating and managing fibers, handling promises, and performing
+  * parallel and race operations.
+  *
+  * @see
+  *   [[Async.run]] for running asynchronous computations
+  * @see
+  *   [[Async.Fiber]] for managing asynchronous tasks
+  * @see
+  *   [[Async.Promise]] for creating and managing promises
+  * @see
+  *   [[Async.parallel]] for running computations in parallel
+  * @see
+  *   [[Async.race]] for racing multiple computations
+  */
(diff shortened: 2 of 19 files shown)


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -92,18 +196,68 @@ object Async:
...
-            def isDone(using Frame): Boolean < IO                                       = IO(self.isDone())
-            def onComplete(f: Result[E, A] => Unit < IO)(using Frame): Unit < IO        = IO(self.onComplete(r => IO.run(f(r)).eval))
-            def block(timeout: Duration)(using Frame): Result[E | Timeout, A] < IO      = IO(self.block(deadline(timeout)))
 
+            /** Checks if the Fiber is done.
+              *
+              * @return
+              *   Whether the Fiber is done
+              */
+            def isDone(using Frame): Boolean < IO = IO(self.isDone())
```
hearnadam, line 238: Should we rename this to `done`?

kyo-core/shared/src/main/scala/kyo/Queue.scala
```diff
@@ -5,17 +5,86 @@ import java.util.concurrent.atomic.AtomicReference
...
-    def capacity(using Frame): Int             = unsafe.capacity
-    def size(using Frame): Int < IO            = op(unsafe.size())
-    def isEmpty(using Frame): Boolean < IO     = op(unsafe.isEmpty())
-    def isFull(using Frame): Boolean < IO      = op(unsafe.isFull())
+    /** Returns the capacity of the queue.
+      *
+      * @return
+      *   the capacity of the queue
+      */
+    def capacity(using Frame): Int = unsafe.capacity
```
hearnadam, line 23: pre-existing: can we mark all implemented methods on extendable classes as final? I know we don't have any explicit guarantees about binary compatibility or breaking changes, but better to do so now pre-1.0.

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -24,14 +40,43 @@ object Resource:
...
+      *
+      * @param resource
+      *   The effect to acquire the Closeable resource.
+      * @param frame
+      *   The implicit Frame for context.
+      * @return
+      *   The acquired Closeable resource wrapped in Resource, IO, and S effects.
+      */
     def acquire[A <: Closeable, S](resource: A < S)(using Frame): A < (Resource & IO & S) =
         acquireRelease(resource)(r => IO(r.close()))
```
hearnadam, line 69: I think we need to ensure that `r` is not null: https://github.com/zio/zio/pull/9008/files

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/Access.scala
```diff
@@ -1,10 +1,17 @@
...
+    /** Multiple Producer Multiple Consumer access pattern */
+    case Mpmc
 
-object Access:
-    case object Mpmc extends Access
-    case object Mpsc extends Access
-    case object Spmc extends Access
-    case object Spsc extends Access
+    /** Multiple Producer Single Consumer access pattern */
+    case Mpsc
```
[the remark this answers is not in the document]
fwbrasil: It was just because of the length but I think making it clearer can help newcomers. I'll update it!

kyo-core/shared/src/main/scala/kyo/Async.scala
fwbrasil: ah, that's an oversight. The refactoring that removed the `Done` fiber type should have renamed this as well. I'll follow up.

kyo-core/shared/src/main/scala/kyo/Queue.scala
fwbrasil: yep! It also helps JIT optimizations. I wonder if it's possible to have a linter for that.

PR #644, opened 2024-09-07
combinators: scaladocs, fixes, and tests

fwbrasil:
I noticed a couple of bugs and missing test coverage while adding the scaladocs.

### DIFF
--- a/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
+++ b/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
@@ -9,57 +9,130 @@ import scala.reflect.ClassTag
 
 extension [A, S](effect: A < S)
 
+    /** Performs this computation and then the next one, discarding the result of this computation.
+      *
+      * @param next
+      *   The computation to perform after this one
+      * @return
+      *   A computation that produces the result of `next`
+      */
     @targetName("zipRight")
     def *>[A1, S1](next: => A1 < S1)(using Frame): A1 < (S & S1) =
         effect.map(_ => next)
 
+    /** Performs this computation and then the next one, discarding the result of the next computation.
+      *
+      * @param next
+      *   The computation to perform after this one
+      * @return
+      *   A computation that produces the result of this computation
+      */
     @targetName("zipLeft")
     def <*[A1, S1](next: => A1 < S1)(using Frame): A < (S & S1) =
         effect.map(e => next.as(e))
 
+    /** Performs this computation and then the next one, returning both results as a tuple.
+      *
+      * @param next
+      *   The computation to perform after this one
+      * @return
+      *   A computation that produces a tuple of both results
+      */
     @targetName("zip")
     def <*>[A1, S1](next: => A1 < S1)(using Frame): (A, A1) < (S & S1) =
         effect.map(e => next.map(n => (e, n)))
 
+    /** Performs this computation, discards its result, and returns the given value.
+      *
+      * @param value
+      *   The value to return
+      * @return
+      *   A computation that produces the given value
+      */
     inline def as[A1, S1](value: => A1 < S1)(using inline frame: Frame): A1 < (S & S1) =
         effect.map(_ => value)
 
+    /** Performs this computation and prints its result to the console.
+      *
+      * @return
+      *   A computation that produces the result of this computation
+      */
     def debugValue(using Frame): A < S = Debug(effect)
 
+    /** Performs this computation and prints its result to the console with a detailed execution trace.
+      *
+      * @return
+      *   A computation that produces the result of this computation
+      */
     def debugTrace(using Frame): A < S = Debug.trace(effect)
 
+    /** Performs this computation after a delay.
+      *
+      * @param duration
+      *   The duration to delay for
+      * @return
+      *   A computation that produces the result of this computation with Async effect
+      */
     def delayed[S1](duration: Duration < S1)(using Frame): A < (S & S1 & Async) =
-        Kyo.sleep(duration) *> effect
+        duration.map(d => Async.delay(d)(effect))
 
-    def repeat(policy: Retry.Policy)(using Flat[A < S], Frame): A < (S & Async) =
-        def loop(i: Int): A < (S & Async) =
-            if i <= 0 then effect
-            else loop(i - 1).delayed(policy.backoff(i))
+    /** Performs this computation repeatedly with a backoff policy.
+      *
+      * @param policy
+      *   The backoff policy to use
+      * @return
+      *   A computation that produces the result of the last execution
+      */
+    def repeat(policy: Retry.Policy)(using Flat[A], Frame): A < (S & Async) =
+        Loop.indexed { i =>
+            if i >= policy.limit then effect.map(Loop.done)
+            else effect.delayed(policy.backoff(i)).as(Loop.continue)
+        }
 
-        loop(policy.limit)
-    end repeat
+    /** Performs this computation repeatedly with a limit.
+      *
+      * @param limit
+      *   The maximum number of times to repeat the computation
+      * @return
+      *   A computation that produces the result of the last execution
+      */
+    def repeat[S1](limit: Int < S1)(using Flat[A], Frame): A < (S & S1) =
+        limit.map { limit =>
+            Loop.indexed { i =>
+                if i >= limit then effect.map(Loop.done)
+                else effect.as(Loop.continue)
+            }
+        }
 
-    def repeat[S1](limit: => Int < S1)(using Flat[A < S], Frame): A < (S & S1) =
-        @tailrec
-        def loop(i: Int): A < (S & S1) =
-            if i <= 0 then effect
-            else loop(i - 1)
-
-        limit.map(l => loop(l))
-    end repeat
-
+    /** Performs this computation repeatedly with a backoff policy and a limit.
+      *
+      * @param backoff
+      *   The backoff policy to use
+      * @param limit
+      *   The limit to use
+      * @return
+      *   A computation that produces the result of this computation with Async effect
+      */
     def repeat[S1](backoff: Int => Duration, limit: => Int < S1)(using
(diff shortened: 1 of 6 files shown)


kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -9,57 +9,130 @@ import scala.reflect.ClassTag
...
+      *
+      * @param duration
+      *   The duration to delay for
+      * @return
+      *   A computation that produces the result of this computation with Async effect
+      */
     def delayed[S1](duration: Duration < S1)(using Frame): A < (S & S1 & Async) =
-        Kyo.sleep(duration) *> effect
-
-    def repeat(policy: Retry.Policy)(using Flat[A < S], Frame): A < (S & Async) =
```
fwbrasil, line 34: These `repeat` methods were incorrect, only executing the effect once.

fwbrasil:
cc/ @johnhungerford

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -69,31 +142,54 @@ extension [A, S](effect: A < S)
...
+    /** Performs this computation repeatedly while the given condition holds.
+      *
+      * @param fn
+      *   The condition to check after each iteration taking the current value and the number of iterations so far, and returning a tuple
+      *   with the condition and the duration to sleep between iterations
+      * @return
+      *   A computation that produces the result of the last successful execution before the condition becomes false
+      */
     def repeatWhile[S1](fn: (A, Int) => (Boolean, Duration) < S1)(using
-        Flat[A < S],
```
fwbrasil, line 73: This was still using an old encoding of `Flat`, making calls to the method not compile.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -9,57 +9,130 @@ import scala.reflect.ClassTag
...
         effect.map(e => next.map(n => (e, n)))
 
+    /** Performs this computation, discards its result, and returns the given value.
+      *
+      * @param value
+      *   The value to return
+      * @return
+      *   A computation that produces the given value
+      */
     inline def as[A1, S1](value: => A1 < S1)(using inline frame: Frame): A1 < (S & S1) =
```
hearnadam, line 52: I wonder if this should just accept pure values.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -105,36 +201,96 @@ extension [A, S](effect: A < S)
...
+      */
     def when[S1](condition: => Boolean < S1)(using Frame): A < (S & S1 & Abort[Maybe.Empty]) =
         condition.map(c => if c then effect else Abort.fail(Maybe.Empty))
 
+    /** Performs this computation catching any Throwable in an Abort[Throwable] effect.
+      *
+      * @return
+      *   A computation that produces the result of this computation with Abort[Throwable] effect
+      */
     def explicitThrowable(using Flat[A], Frame): A < (S & Abort[Throwable]) =
```
hearnadam, line 264: I prefer ZIO's naming `resurrect` - what do you think? 
It makes sense in the context of suppressing Throwables is an `orDie`. I'm open to other ideas.

kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
```diff
@@ -32,112 +56,317 @@ extension (kyoObject: Kyo.type)
...
+      *   A function to apply to each element of the sequence
+      * @return
+      *   Discards the results of the function application and returns Unit
+      */
     def foreachParDiscard[A, S, Any](
         sequence: => Seq[A] < S
     )(
         useElement: A => Any < Async
     )(using Flat[Any], Frame): Unit < (S & Async) =
         sequence.map(seq => Async.parallel(seq.map(v => useElement(v).unit))).unit
```
hearnadam, line 135: Do we need the `.unit` on each effect?

kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
```diff
@@ -32,112 +56,317 @@ extension (kyoObject: Kyo.type)
...
         message.map(m => err.map(e => Log.debug(m, e)))
 
+    /** Logs an error message to the console.
+      *
+      * @param message
+      *   The message to log
+      * @return
+      *   An effect that logs the message to the console
+      */
     inline def logError[S](message: => String < S): Unit < (S & IO) =
```
hearnadam, line 309: I can't tell what object these are under (extension), but I think they may be confusing names.

If I run

```
val io: A < (Abort[E] & IO) = ???
io.logError(...)
```

I would expect it to tap into the error `E`. Related: we may need to adjust `Log` to handle `Result.Error[E]` or `Throwable`

kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
```diff
@@ -32,112 +56,317 @@ extension (kyoObject: Kyo.type)
...
     ): Unit < (S & S1 & IO) =
         message.map(m => err.map(e => Log.trace(m, e)))
 
+    /** Creates an effect that never completes using Async.
+      *
+      * @return
+      *   An effect that never completes
+      */
     def never(using Frame): Nothing < Async =
         Promise.init[Nothing, Nothing].join
```
hearnadam, line 358: `Fiber.never.join`?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
fwbrasil: Agreed. I didn't change the methods taking effectful inputs but that's the pattern in the rest of the codebase. Making the places where `map` happens more explicit I think makes the code easier to reason about, even if it introduces more syntax overhead. It'd be nice to hear @johnhungerford's thoughts on this.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
fwbrasil: Good point. I'm also not very happy with the new success/fail/panic naming, it doesn't feel natural to me. Maybe we should just go with ZIO's naming for error handling. It's clear and it'd be easier for people coming from ZIO.

kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
fwbrasil: In this case it'd be a companion object method `Kyo.logError(...)`. It seems ok to me. Good point about logging `Result.Error`!

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
johnhungerford: @hearnadam I agree that `explicitThrowable` is no good. I wasn't aware of `resurrect`, but it would make sense to use that. We should also then use `orDie` instead of `implicitThrowable`. The alternative I had been mulling over was `throwAborts` (for `orDie`) and `abortThrows` (for `resurrect`). Happy to stick to ZIO convention, though.

@fwbrasil I don't see anything wrong with success/fail/panic except that success is a noun whereas fail is a verb. succeed/fail/panic or success/failure/panic would be marginally more graceful.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
johnhungerford: Did I not have tests for the repeat methods?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
johnhungerford: I do think it should be consistent with the rest of the codebase on this point. I'm kind of torn however: it's a really cool thing about Kyo that effects are perfectly composable and there's no cost to lifting a value to a "pure" kyo (A < Any). If there's a performance cost to the map, though, that would be a reason not to. But is that the case? Or are you concerned that the method signatures are more confusing?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
fwbrasil: Yeah, I think so. They were missing the effect on each loop iteration.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
fwbrasil: There is a performance cost but it's minimal, especially since it would not allocate if pure values are passed. I know I had highlighted "fluent APIs" as an advantage of Kyo's encoding but I've reconsidered this view. Making the monadic bindings more explicit helps to indicate where effects are coming from and simplifies method signatures. Although it introduces more syntax overhead, I believe it be beneficial for newcomers. It can be especially helpful when multiple effects are involved since it makes the flow of effects more transparent and clear to reason about.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
hearnadam: @johnhungerford I am inclined to approve this CR to get the scaladocs published. Can you followup on some alignment with ZIO's naming?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
hearnadam: Seems reasonable. For contaminators, I think we will need per method decision making. `as` should probably accept non-kyo, but `tap` or other operators can accept effects.

hearnadam (approved):

fwbrasil:
I've created two tickets for the follow ups: https://github.com/getkyo/kyo/issues/650 https://github.com/getkyo/kyo/issues/651

PR #645, opened 2024-09-08
Scala Native support for kyo-tag, kyo-data, kyo-prelude

fwbrasil:
I was surprised by how easy it was to add Scala Native support!

### DIFF
--- a/kyo-data/shared/src/test/scala/kyo/DurationSpec.scala
+++ b/kyo-data/shared/src/test/scala/kyo/DurationSpec.scala
@@ -157,7 +157,7 @@ object DurationSpec extends ZIOSpecDefault:
                 )
             }
         )
-    )
+    ) @@ TestAspect.exceptNative
 end DurationSpec
 
 object ZDurations:
--- a/kyo-prelude/native/src/main/scala/kyo/kernel/TracePool.scala
+++ b/kyo-prelude/native/src/main/scala/kyo/kernel/TracePool.scala
@@ -1,0 +1,49 @@
+package kyo.kernel
+
+import internal.*
+import java.util.Arrays
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.atomic.AtomicInteger
+import kyo.discard
+import scala.annotation.tailrec
+
+private[kernel] object TracePool:
+    inline def globalCapacity: Int = 10240
+    inline def localCapacity: Int  = 32
+
+    private val global = new ConcurrentLinkedQueue[Trace]()
+
+    abstract class Local:
+        final private val pool = new Array[Trace](localCapacity)
+        final private var size = 0
+
+        final def borrow(): Trace =
+            if size == 0 then
+                val trace = global.poll()
+                if trace != null then trace else Trace.init
+            else
+                size -= 1
+                val buffer = pool(size)
+                pool(size) = null
+                buffer
+        end borrow
+
+        final def release(trace: Trace): Unit =
+            clear(trace)
+            if size < localCapacity then
+                pool(size) = trace
+                size += 1
+            else if global.size() < globalCapacity then
+                discard(global.offer(trace))
+            end if
+        end release
+
+        final private def clear(trace: Trace) =
+            val arr = trace.frames
+            val len = math.min(trace.index, maxTraceFrames)
+            Arrays.fill(arr.asInstanceOf[Array[AnyRef]], 0, len, null)
+            trace.index = 0
+        end clear
+    end Local
+
+end TracePool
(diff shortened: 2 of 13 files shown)


kyo-data/shared/src/test/scala/kyo/DurationSpec.scala
```diff
@@ -157,7 +157,7 @@ object DurationSpec extends ZIOSpecDefault:
                 )
             }
         )
-    )
+    ) @@ TestAspect.exceptNative
```
fwbrasil, line 160: The tests actually pass but I think there's some incompatibility in the test runner that makes it report a failure.

kyo-prelude/native/src/main/scala/kyo/kernel/TracePool.scala
```diff
@@ -0,0 +1,49 @@
...
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.atomic.AtomicInteger
+import kyo.discard
+import scala.annotation.tailrec
+
+private[kernel] object TracePool:
+    inline def globalCapacity: Int = 10240
+    inline def localCapacity: Int  = 32
+
+    private val global = new ConcurrentLinkedQueue[Trace]()
```
fwbrasil, line 14: Not as optimized as the JVM version since it allocates.

kyo-prelude/native/src/main/scala/kyo/kernel/TracePool.scala
```diff
@@ -0,0 +1,49 @@
...
+
+    abstract class Local:
+        final private val pool = new Array[Trace](localCapacity)
+        final private var size = 0
+
+        final def borrow(): Trace =
+            if size == 0 then
+                val trace = global.poll()
+                if trace != null then trace else Trace.init
```
fwbrasil, line 23: Also less efficient than the JVM version because it doesn't replenish from the global pool in batches.

kyo-prelude/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -162,7 +162,11 @@ object `<`:
     // TODO Compiler crash if inlined
     extension [A, S](v: A < S)
         def repeat(i: Int)(using ev: A => Unit, frame: Frame, flat: Flat.Weak[A]): Unit < S =
-            if i <= 0 then () else v.andThen(repeat(i - 1))
+            if i <= 0 then ()
+            else
+                // TODO Test failures in Scala Native without `return`
+                return v.andThen(repeat(i - 1))
```
fwbrasil, line 168: This was the most confusing issue I had. It seems Scala Native ends up returning `()` instead of the `Unit < S` computation without the `return` clause. I wonder if this issue could appear in other places.

kyo-prelude/shared/src/test/scala/kyo/ChoiceTest.scala
```diff
@@ -81,7 +83,7 @@ class ChoiceTest extends Test:
         end try
     }
 
-    "large number of suspensions" in pendingUntilFixed {
+    "large number of suspensions" taggedAs notNative in pendingUntilFixed {
```
fwbrasil, line 86: Scala Native crashes and reports a failures due to the stack overflow even with `pendingUntilFixed`.

kyo-prelude/shared/src/main/scala/kyo/kernel/Pending.scala
hearnadam: Is this just a type inference issue? Very strange.

kyo-prelude/native/src/main/scala/kyo/kernel/TracePool.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import internal.*
+import java.util.Arrays
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.atomic.AtomicInteger
+import kyo.discard
+import scala.annotation.tailrec
+
+private[kernel] object TracePool:
+    inline def globalCapacity: Int = 10240
```
sideeffffect, line 11: 1024? 8192? 16384?

fwbrasil: ouch 🤦 good catch, thanks!

kyo-prelude/shared/src/main/scala/kyo/kernel/Pending.scala
fwbrasil: I'm not sure since I don't know how to inspect the code the compiler is generating. In this case, the `then` branch returns a `()` but the `else` returns a computation so the `()` should be lifted to `Unit < Any` via the `<.lift` implicit conversion. I've tried to reproduce it insolation but didn't have success. It could be a bad interaction between the multiple features Kyo uses: opaque types, extension methods, implicit conversions, inlining, etc. I don't think we should block on it since the tests are passing but we should have a ticket to follow up.

PR #646, opened 2024-09-08
cats-effect integration

fwbrasil:
I wanted to avoid this integration because of past issues with Typelevel, but I think it's best if we get ahead of a possible integration and ensure it's under our control in the repository.

### DIFF
--- a/kyo-cats/shared/src/main/scala/kyo/Cats.scala
+++ b/kyo-cats/shared/src/main/scala/kyo/Cats.scala
@@ -16,11 +16,8 @@ object Cats:
       * @return
       *   A Kyo effect that, when run, will execute the cats.effect.IO
       */
-    def get[A](io: CatsIO[A])(using Frame): A < Cats =
-        ArrowEffect.suspendMap(Tag[GetCatsIO], io) {
-            case Left(ex) => throw ex
-            case Right(v) => v
-        }
+    def get[A](io: CatsIO[A])(using Frame): A < (Abort[Throwable] & Cats) =
+        ArrowEffect.suspendMap(Tag[GetCatsIO], io)(Abort.get(_))
 
     /** Runs a Kyo effect that uses Cats and converts it to a cats.effect.IO.
       *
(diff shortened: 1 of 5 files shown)


kyo-cats/shared/src/main/scala/kyo/Cats.scala
```diff
@@ -0,0 +1,46 @@
...
+object Cats:
+
+    /** Lifts a cats.effect.IO into a Kyo effect.
+      *
+      * @param io
+      *   The cats.effect.IO to lift
+      * @return
+      *   A Kyo effect that, when run, will execute the cats.effect.IO
+      */
+    def get[A](io: CatsIO[A])(using Frame): A < (Abort[Throwable] & Cats) =
```
hearnadam, line 19: Should `io` be by name?
Does the return type need the `Abort`? I guess that is more clear for users.

kyo-cats/shared/src/main/scala/kyo/Cats.scala
```diff
@@ -0,0 +1,46 @@
+package kyo
+
+import Cats.GetCatsIO
+import cats.effect.IO as CatsIO
+import kyo.kernel.*
+import scala.util.control.NonFatal
+
+opaque type Cats <: (Async & Abort[Throwable]) = GetCatsIO & Async & Abort[Throwable]
```
hearnadam, line 8: Amazing how simple the integration is!

kyo-cats/shared/src/main/scala/kyo/Cats.scala
```diff
@@ -0,0 +1,46 @@
+package kyo
```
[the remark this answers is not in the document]
fwbrasil: I like the convenience of a single import. Do you worry about name conflicts? We shouldn't have another `Cats` type in the codebase.

kyo-cats/shared/src/main/scala/kyo/Cats.scala
fwbrasil: It doesn't need the `Abort` since it's in the opaque type. We don't strictly need a by-name param but that's the pattern in `ZIOs` and it'd make error handling better in case the `io` construction fails. I'll update it.

kyo-cats/shared/src/main/scala/kyo/Cats.scala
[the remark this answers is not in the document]
hearnadam: I think users with larger applications may want more granular control of what they import. Seems okay either way.

hearnadam (approved):

PR #661, opened 2024-09-14
fix readme

mio-19:
code was updated to dropIf in https://github.com/getkyo/kyo/pull/577/files but comment was not updated


fwbrasil:
thank you @mio-19!

PR #666, opened 2024-09-17
core: translate exceptions in Fiber.fromFuture

fwbrasil:


kyo-combinators/shared/src/test/scala/kyo/FiberCombinatorTest.scala
```diff
@@ -19,7 +19,7 @@ class FiberCombinatorTest extends Test:
 
             "should construct from Future" in {
                 val future        = scala.concurrent.Future(100)
-                val effect        = Kyo.fromFuture(future)
+                val effect        = Kyo.fromFuture[Throwable, Int, Any](future)
```
fwbrasil, line 22: This is a good example of how the additional type param to allow effectful inputs introduces complexity.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -170,19 +172,34 @@ object Async:
...
+          *   The Future to convert into a Fiber
+          * @tparam E
+          *   The expected error type that the Future might fail with. Use Throwable if you don't need to catch specific exceptions.
+          * @tparam A
+          *   The type of the successful result
           * @return
           *   A Fiber that completes with the result of the Future
           */
-        def fromFuture[A: Flat](f: Future[A])(using Frame): Fiber[Nothing, A] < IO =
+        def fromFuture[E <: Throwable, A: Flat](f: Future[A])(using
```
fwbrasil, line 188: The API isn't great since users will typically need to specify both type parameters. It might be ok since it seems a low-level API. Alternatively, we could have a `FromFutureOps` class to isolate the `E` type or have a separate `Fiber.fromFutureCatching` method. Suggestions are welcome

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
```diff
@@ -475,6 +475,35 @@ class AsyncTest extends Test:
         }
     }
 
+    "Fiber.fromFuture" - {
```
fwbrasil, line 478: The method was missing coverage 🫣

kyo-combinators/shared/src/test/scala/kyo/FiberCombinatorTest.scala
steinybot: Yeah this is not ideal. At least if `fromFuture` was `Fiber[Throwable, A] < IO` then we need to come up with some way of refining the `E`.

kyo-core/shared/src/main/scala/kyo/Async.scala
steinybot: At the very least it should be `Fiber[Throwable, A] < IO`. Then we would need to come up with some way of refining the `E` which could be useful in other cases too. Something orthogonal to `map` which changes the `A`.

kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
```diff
@@ -191,7 +191,7 @@ extension (kyoObject: Kyo.type)
       * @return
       *   An effect that attempts to run the given effect and handles the Future to Async.
       */
-    def fromFuture[A: Flat, S](future: => Future[A] < S)(using Frame): A < (S & Async) =
+    def fromFuture[E <: Throwable: ClassTag, A: Flat, S](future: => Future[A] < S)(using Frame): A < (S & Async & Abort[E]) =
```
hearnadam, line 194: I think this is a good example of a parameter that should not accept Kyos. By name can remain.

kyo-core/shared/src/main/scala/kyo/Async.scala
hearnadam: I think it's reasonable that this returns `Fiber[Throwable, A] < IO`, not `Fiber[E, A]`. Refinement of Aborts is another, but related issue.

fwbrasil: My initial intuition was that we should avoid replicating error handling logic that should be in `Abort` and not introduce a `fiber.mapFail` method but I think @steinybot made a good point since we already offer `fiber.map/flatMap`. Providing the method can also avoid some of the performance overhead since converting the `Fiber` to a computation requires additional allocations. I think we could always produce a `Fiber[Throwable, A]` here and provide a `fiber.mapFail` method.  @hearnadam wdyt?

steinybot: Given that a fiber is `Fiber[E, A]` I'd expect there to be an API on the container type to go to any `Fiber[F, B]`. So I'd probably cover all cases with `map` (`Fiber[E, A]` to `Fiber[E, B]`), `mapFail` (`Fiber[E, A]` to `Fiber[F, A]`) and `mapResult` (`Fiber[E, A]` to `Fiber[F, B]`).

Ideally fiber would be like how it was and not track `E`. So `fromFuture` would be `Fiber[A] < (IO & Abort[Throwable])` then it would be a lot more intuitive to operate at the `Abort` level and you wouldn't need to do the `.map(_.get)` dance.

Now I'm starting to question my why we even have a fiber container type. Shouldn't it be `Async.fromFuture` returning `A < (Abort[Throwable] & Async)`? Once again using `Abort` would make more sense.

fwbrasil: Thanks for the suggestions folks! I've update the PR to make `Fiber.fromFuture` return `Fiber[Throwable, A]`, introduced `fiber.mapResult` as a more generic method that allows mapping the error, and added a convenience `Async.fromFuture` shortcut.

steinybot (approved):

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -299,6 +308,24 @@ object Async:
...
+              *
+              * This method allows you to transform both the error and success types of the Fiber's result. It's useful when you need to
+              * modify the error type or perform a more complex transformation on the success value that may also produce a new error type.
+              *
+              * @param f
+              *   The function to apply to the Fiber's Result. It should take a Result[E, A] and return a Result[E2, B].
+              * @return
+              *   A new Fiber with the mapped Result
+              */
+            def mapResult[E2, B](f: Result[E, A] => Result[E2, B])(using Frame): Fiber[E2, B] < IO =
```
hearnadam, line 321: Another option is `sandbox` function (`Fiber[E, A] => Fiber[Nothing, Result[E, A]`) + `flatten` which could flatten one of Promise/Fiber/Result?

Otherwise seems okay to me.

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/Async.scala
fwbrasil: I think the complexity of the API might not be worth it

PR #667, opened 2024-09-18
core: async optimizations

fwbrasil:
image

The regressions seem to be just noise.

https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/1449324e996db52fc3632d0b5d115933/raw/f02a69655e1ca4f3fcd710cced12e982a619f09f/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/1449324e996db52fc3632d0b5d115933/raw/f02a69655e1ca4f3fcd710cced12e982a619f09f/jmh-result-candidate.json

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -179,13 +179,15 @@ object Async:
         def fromFuture[A: Flat](f: Future[A])(using Frame): Fiber[Nothing, A] < IO =
             import scala.util.*
             IO {
-                val p = new IOPromise[Nothing, A]()
-                f.onComplete {
-                    case Success(v) =>
-                        p.complete(Result.success(v))
-                    case Failure(ex) =>
-                        p.complete(Result.panic(ex))
-                }(ExecutionContext.parasitic)
+                val p = new IOPromise[Nothing, A] with (Try[A] => Unit):
+                    def apply(result: Try[A]) =
+                        result match
+                            case Success(v) =>
+                                completeUnit(Result.success(v))
+                            case Failure(ex) =>
+                                completeUnit(Result.panic(ex))
+
+                f.onComplete(p)(ExecutionContext.parasitic)
                 p
             }
         end fromFuture
@@ -276,8 +278,9 @@ object Async:
               */
             def map[B](f: A => B)(using Frame): Fiber[E, B] < IO =
                 IO {
-                    val p = IOPromise[E, B](interrupts = self)
-                    self.onComplete(v => p.completeUnit(v.map(f)))
+                    val p = new IOPromise[E, B](interrupts = self) with (Result[E, A] => Unit):
+                        def apply(v: Result[E, A]) = completeUnit(v.map(f))
+                    self.onComplete(p)
                     p
                 }
 
@@ -290,8 +293,9 @@ object Async:
               */
             def flatMap[E2, B](f: A => Fiber[E2, B])(using Frame): Fiber[E | E2, B] < IO =
                 IO {
-                    val p = IOPromise[E | E2, B](interrupts = self)
-                    self.onComplete(_.fold(p.completeUnit)(v => p.becomeUnit(f(v))))
+                    val p = new IOPromise[E | E2, B](interrupts = self) with (Result[E, A] => Unit < IO):
+                        def apply(r: Result[E, A]) = r.fold(completeUnit)(v => becomeUnit(f(v)))
+                    self.onComplete(p)
                     p
                 }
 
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -179,13 +179,15 @@ object Async:
...
             import scala.util.*
             IO {
-                val p = new IOPromise[Nothing, A]()
-                f.onComplete {
-                    case Success(v) =>
-                        p.complete(Result.success(v))
-                    case Failure(ex) =>
-                        p.complete(Result.panic(ex))
-                }(ExecutionContext.parasitic)
+                val p = new IOPromise[Nothing, A] with (Try[A] => Unit):
```
fwbrasil, line 182: A few optimizations are just merging function interfaces with `IOPromise`. It avoids the function allocation in addition to the promise.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -440,16 +444,18 @@ object Async:
...
-            val p = IOPromise[Nothing, Unit]()
-            if d.isFinite then
-                Timer.schedule(d)(p.completeUnit(Result.success(()))).map { t =>
-                    IO.ensure(t.cancel.unit)(get(p))
-                }
-            else
-                get(p)
-            end if
-        }
+        if d == Duration.Zero then ()
```
fwbrasil, line 447: I noticed that `Retry` ends up calling `Async.sleep` with a zero duration for the first retry with the default policy.

kyo-core/shared/src/main/scala/kyo/scheduler/Ensures.scala
```diff
@@ -13,11 +14,9 @@ private[kyo] object Ensures:
     val empty: Ensures = Empty
 
     private val bufferCache = new MpmcArrayQueue[ArrayDeque[() => Unit]](1000)
+
     private def buffer(): ArrayDeque[() => Unit] =
-        val b = bufferCache.poll()
-        if b == null then new ArrayDeque()
-        else b
-    end buffer
+        Maybe(bufferCache.poll()).getOrElse(new ArrayDeque())
```
fwbrasil, line 19: Minor refactoring.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -211,7 +212,8 @@ private[kyo] object IOPromise extends IOPromisePlatformSpecific:
         def waiters: Int
         def run(v: Result[E, A]): Pending[E, A]
 
-        def add(f: Result[E, A] => Unit): Pending[E, A] =
+        @nowarn("msg=anonymous")
+        inline def add(inline f: Result[E, A] => Unit): Pending[E, A] =
```
fwbrasil, line 216: Merges two allocations for `f` and `Pending.add` into a single one.

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -7,7 +7,7 @@ import kyo.kernel.ArrowEffect
 import kyo.scheduler.IOTask.*
 import scala.util.control.NonFatal
 
-private[kyo] class IOTask[Ctx, E, A] private (
+sealed private[kyo] class IOTask[Ctx, E, A] private (
```
fwbrasil, line 10: minor refactoring since there shouldn't be external extensions of the class.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -112,7 +113,7 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
         becomeLoop(other.compress())
     end become
 
-    final def onComplete(f: Result[E, A] => Unit): Unit =
+    inline def onComplete(inline f: Result[E, A] => Unit): Unit =
```
fwbrasil, line 116: Merges two allocations for f and Pending.add into a single one.

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -55,7 +55,7 @@ private[kyo] class IOTask[Ctx, E, A] private (
                                 val trace = this.trace
                                 this.trace = null.asInstanceOf[Trace]
                                 input.onComplete { r =>
-                                    val task = IOTask(IO(cont(r.asInstanceOf[Result[Nothing, C]])), trace, context, ensures, runtime)
```
fwbrasil, line 58: The `IO` suspension here isn't strictly necessary since this is already unsafe execution. It might execute more pure transformations during flushing but it should be ok. For reference, Twitter's `Future` always executes continuations in the thread that flushes the continuations.

fwbrasil: I'm reverting this change. There's actually a good reason for the `IO` suspension here: error handling. If the continuation throws when applied, the failure won't be piped correctly without `IO`.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -179,13 +179,15 @@ object Async:
...
-                }(ExecutionContext.parasitic)
+                val p = new IOPromise[Nothing, A] with (Try[A] => Unit):
+                    def apply(result: Try[A]) =
+                        result match
+                            case Success(v) =>
+                                completeUnit(Result.success(v))
+                            case Failure(ex) =>
+                                completeUnit(Result.panic(ex))
+
+                f.onComplete(p)(ExecutionContext.parasitic)
```
hearnadam, line 190: Do we want this to be done on Kyo's threads?

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -440,16 +444,18 @@ object Async:
...
-            else
-                get(p)
-            end if
-        }
+        if d == Duration.Zero then ()
+        else
+            IO {
+                val p = IOPromise[Nothing, Unit]()
+                if d.isFinite then
+                    Timer.schedule(d)(p.completeUnit(Result.success(()))).map { t =>
```
hearnadam, line 452: Should be 'free', but: `Result.unit`?

hearnadam (approved):

PR #673, opened 2024-09-20
Fix Stream.runDiscard not processing any stream elements

### LINKED ISSUE #672: Stream.runDiscard doesn't process stream elements
Calling `Stream.runDiscard` doesn't process any stream elements.  This conflicts with the docs and I doubt is what is intended.

image

Cause:
https://github.com/getkyo/kyo/blob/631c35f4d2973af6b18373aa43955f2ff2ce78ec/kyo-prelude/shared/src/main/scala/kyo/Stream.scala#L261-L269

One potential solution is to implement `forDiscard` as `runForeach(_ => ())`

lukestephenson:
Fixes #672

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
@@ -264,9 +264,7 @@ final case class Stream[V, -S](v: Ack < (Emit[Chunk[V]] & S)):
       *   A unit effect that runs the stream without collecting results
       */
     def runDiscard(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Unit < S =
-        ArrowEffect.handle(tag, v.unit)(
-            [C] => (input, cont) => cont(Stop)
-        )
+        runForeach(_ => ())
 
     /** Runs the stream and applies the given function to each emitted value.
       *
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -264,9 +264,7 @@ final case class Stream[V, -S](v: Ack < (Emit[Chunk[V]] & S)):
       *   A unit effect that runs the stream without collecting results
       */
     def runDiscard(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Unit < S =
-        ArrowEffect.handle(tag, v.unit)(
-            [C] => (input, cont) => cont(Stop)
-        )
+        runForeach(_ => ())
```
hearnadam, line 267: I think this is the correct approach:
```suggestion
        ArrowEffect.handle(tag, v.unit)(
            [C] => (input, cont) => cont(Continue())
        )
```

kyo-prelude/shared/src/test/scala/kyo/StreamTest.scala
```diff
@@ -481,6 +481,20 @@ class StreamTest extends Test:
             )
         }
 
+        "works with effects" in run {
```
hearnadam, line 484: How about something like this?
```
        "non-empty stream" in {
            assert(
                Var.run(0) {
                    Stream.init(0 until 100).map(i => Var.update[Int](_ + i)).runDiscard.andThen(Var.get[Int])
                }.eval == 4950
            )
        }
```

lukestephenson: sounds good. Personally I'd prefer the expectation was compared to `(0 until 100).sum`, but I've gone with your suggestion for now.

hearnadam (approved):

fwbrasil:
thank you @lukestephenson!

Issue #651, opened 2024-09-09
ZIO-inspired error handling


fwbrasil:
Review Abort, Result, and related kyo-combinators methods to follow a naming pattern similar to ZIO's since it's quite mature and it would be helpful to people coming from ZIO. We should also review whether the APIs provide similar semantics to ZIO. Context: https://github.com/getkyo/kyo/pull/644#discussion_r1749345497

sideeffffect:
I believe the ZIO model for errors is brilliant and kyo would greatly benefit from it:
 * typed channel for recoverable/expected errors
 * untyped channel for unrecoverable/unexpected failures

sideeffffect:
Btw, have you @fwbrasil considered constraining `Result` errors only to `Exception`? Like
```
opaque type Result[+E <: Exception, +A] >: (Success[A] | Failure[E]) = Success[A] | Failure[E]
```
`E` stands for those _typed, recoverable/expected errors_, right, but in my experience they should still always be subclasses of `Exception` for good interoperability with the JVM and ecosystem around it.

fwbrasil:
> Btw, have you @fwbrasil considered constraining Result errors only to Exception?

I think @hearnadam also suggested that some time ago. The intent was make `Result` have the features of both `Either` and `Try` and allow `Abort` for any value like in ZIO. I haven't much benefit in the approach yet, though. There's only the `swap` method that might not make sense if we add this restriction. @johnhungerford can you share you thoughts? I added `swap` inspired in the equivalent combinator method.

fwbrasil:
> I believe the ZIO model for errors is brilliant and kyo would greatly benefit from it:
> 
> * typed channel for recoverable/expected errors
> * untyped channel for unrecoverable/unexpected failures

I think that's what have already given that `Result` and `Abort` handle both typed (Result.fail) and untyped failures (Result.panic).

johnhungerford:
I definitely would not want to restrict errors to `Exception` in `kyo`.  For instance, we should be able to use `Abort` as an `Option`-like effect by aborting with `Maybe.Empty` (or whatever it's called), which is not an exception.

fwbrasil:
[...]

PR #674, opened 2024-09-20
error handling naming refactoring

fwbrasil:
Initial change for https://github.com/getkyo/kyo/issues/651

- Refactor `Result` hierarchy:  `Error` => `Failure`, `Fail` => `Error`
- Refactor related method names: `success` => `succeed`, `fail` => `error`, `error` => `fail`, `mapFail` => `mapError`

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Result.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Result.scala
@@ -7,14 +7,21 @@ import scala.reflect.ClassTag
 import scala.util.Try
 import scala.util.control.NonFatal
 
-/** Represents a result that can be either a success, a failure or an unexpected panic.
+/** Represents a computation result that can be either successful, failed, or in an unexpected panic state.
+  *
+  * The Result type has the following hierarchy:
+  *   - Result[+E, +A]
+  *     - Success[+A]: Represents a successful result
+  *     - Failure[+E]: Represents a failure
+  *       - Error[+E]: Represents an expected error
+  *       - Panic: Represents an unexpected error (always contains a Throwable)
   *
   * @tparam E
   *   The type of the error in case of failure
   * @tparam A
   *   The type of the value in case of success
   */
-opaque type Result[+E, +A] >: (Success[A] | Error[E]) = Success[A] | Error[E]
+opaque type Result[+E, +A] >: (Success[A] | Failure[E]) = Success[A] | Failure[E]
 
 /** Companion object for Result type */
 object Result:
@@ -46,13 +53,13 @@ object Result:
           * @tparam A
           *   The type of the successful result
           * @return
-          *   A Result containing either the successful value, a Fail with the caught exception, or a Panic for other exceptions
+          *   A Result containing either the successful value, a Error with the caught exception, or a Panic for other exceptions
           */
         inline def apply[A](inline expr: => A)(using ClassTag[E]): Result[E, A] =
             try
                 Success(expr)
             catch
-                case ex: E => Fail(ex)
+                case ex: E => Error(ex)
                 case ex    => Panic(ex)
     end CatchingOps
 
@@ -69,7 +76,7 @@ object Result:
       * @return
       *   A successful Result
       */
-    inline def success[E, A](inline value: A): Result[E, A] = Success(value)
+    inline def succeed[E, A](inline value: A): Result[E, A] = Success(value)
 
     /** Creates a failed Result.
       *
@@ -82,7 +89,7 @@ object Result:
       * @return
       *   A failed Result
       */
-    inline def fail[E, A](inline error: E): Result[E, A] = Fail(error)
+    inline def error[E, A](inline error: E): Result[E, A] = Error(error)
 
     /** Creates a Result representing a panic situation.
       *
@@ -141,7 +148,7 @@ object Result:
       *   A Result equivalent to the input Either
       */
     def fromEither[E, A](either: Either[E, A]): Result[E, A] =
-        either.fold(fail, success)
+        either.fold(error, succeed)
 
     /** Converts a Try to a Result.
       *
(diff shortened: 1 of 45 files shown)


kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -7,14 +7,21 @@ import scala.reflect.ClassTag
 import scala.util.Try
 import scala.util.control.NonFatal
 
-/** Represents a result that can be either a success, a failure or an unexpected panic.
+/** Represents a computation result that can be either successful, failed, or in an unexpected panic state.
```
steinybot, line 10: How about:
```suggestion
/** Represents a computation result that can be either a success, an error, or an unexpected panic state.
```

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -46,13 +53,13 @@ object Result:
           * @tparam A
           *   The type of the successful result
           * @return
-          *   A Result containing either the successful value, a Fail with the caught exception, or a Panic for other exceptions
+          *   A Result containing either the successful value, a Error with the caught exception, or a Panic for other exceptions
```
steinybot, line 56:
```suggestion
          *   A Result containing either the successful value, an Error with the caught exception, or a Panic for other exceptions
```

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -192,27 +199,27 @@ object Result:
...
         def unapply[E, A](self: Result[E, A]): Maybe.Ops[E | Throwable] =
             self match
-                case error: Error[E] @unchecked => Maybe(error.getFailure)
-                case _                          => Maybe.empty
-    end Error
+                case error: Failure[E] @unchecked => Maybe(error.getFailure)
+                case _                            => Maybe.empty
+    end Failure
 
     /** Represents a failure in a Result. */
```
steinybot, line 217:
```suggestion
    /** Represents an expected error in a Result. */
```

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -249,8 +256,8 @@ object Result:
                 throw exception
     end Panic
 
-    extension [E](self: Error[E])
-        /** Gets the exception from an Error.
+    extension [E](self: Failure[E])
+        /** Gets the exception from an Failure.
```
steinybot, line 260: Opposite problem
```suggestion
        /** Gets the exception from a Failure.
```

kyo-data/shared/src/main/scala/kyo/Result.scala
steinybot: There are a few of these. Maybe a regex replace would be better than me commenting individually.

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -276,16 +283,16 @@ object Result:
...
+                case _             => true
 
-        /** Checks if the Result is a Fail.
+        /** Checks if the Result is a Error.
           *
           * @return
-          *   true if the Result is a Fail, false otherwise
+          *   true if the Result is a Error, false otherwise
           */
         def isFail =
```
steinybot, line 294: This needs to be renamed too right?

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -302,19 +309,19 @@ object Result:
...
+                case self             => Maybe(self.asInstanceOf[A])
 
         /** Gets the error value if present.
           *
           * @return
-          *   A Maybe containing the error value, or empty for non-Fail Results
+          *   A Maybe containing the error value, or empty for non-Error Results
           */
         @targetName("maybeError")
         def failure: Maybe[E] =
```
steinybot, line 321: And rename this?

kyo-data/shared/src/main/scala/kyo/Result.scala
steinybot, line 294: Is there no `isError`?

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -69,7 +76,7 @@ object Result:
       * @return
       *   A successful Result
       */
-    inline def success[E, A](inline value: A): Result[E, A] = Success(value)
+    inline def succeed[E, A](inline value: A): Result[E, A] = Success(value)
```
steinybot, line 79: Is there a particular reason that these return `Result` rather than the more specific types?

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -82,7 +89,7 @@ object Result:
       * @return
       *   A failed Result
       */
-    inline def fail[E, A](inline error: E): Result[E, A] = Fail(error)
+    inline def error[E, A](inline error: E): Result[E, A] = Error(error)
```
steinybot, line 92: Unfortunate that there isn't a good verb for this

fwbrasil:
@steinybot thank you for the review! I was talking to @hearnadam and we thought that actually the naming we have in `main` seems good enough: `Success`/`Error`/`Fail`/`Panic` for the types and `succeed`/`error`/`fail`/`panic` for their respective methods. We'd just need to rename `success` to `succeed`. The most common methods are verbs with the exception of `error` but it should not be common to use it. Wdyt?

fwbrasil:
> Failure is a much better antonym to success as it can be defined literally as the lack of success. Error sounds more like an expected bad state. 

@hearnadam also suggested `Success`/`Failure`/`Fail`/`Panic` in private but I thought `Failure` was too close to `Fail` but it might a good option given your arguments. I also don't like `Error` because in JVM/Java the term represents unrecoverable failures.

> But is it worth the switch? Not sure but it's now or never and you have done the work.

I think we should go for what we think is best regardless of the work. We might not have another opportunity to review before 1.0.

fwbrasil:
Recovering this branch would be challenging. I'm closing it for now, let's continue the discussion in https://github.com/getkyo/kyo/issues/651

PR #676, opened 2024-09-21
tag: fix inference in generic scopes

### LINKED ISSUE #671: Unable to derive tag of container unless child tag is inline
I am trying to create a class `Foo[T: Tag]` which internally uses `Emit.andMap` but it fails to derive the `Tag[Emit[T]]`:

```
class Foo[T: Tag]:
    summon[Tag[Emit[T]]]
end Foo

object Foo:
    def foo[T: Tag] =
        summon[Tag[Emit[T]]]
end Foo
```

```
[error] -- Error: /Users/jason/src/kyo2/kyo-prelude/shared/src/test/scala/kyo/EmitTest.scala:6:24
[error] 6 |    summon[Tag[Emit[T]]]
[error]   |                        ^
[error]   |                 Please provide an implicit kyo.Tag[Foo.this.T] parameter.
[error] -- Error: /Users/jason/src/kyo2/kyo-prelude/shared/src/test/scala/kyo/EmitTest.scala:11:28
[error] 11 |        summon[Tag[Emit[T]]]
[error]    |                            ^
[error]    |                        Please provide an implicit kyo.Tag[T] parameter.
```

I can't easily make it inline. It doesn't make sense to require an `Emit[Tag[T]]` as that is exposing too much implementation detail. As a workaround I probably could wrap the logic that uses `Emit.andMap` although that is unnecessary complexity.

fwbrasil:
Fixes #671.

This was a though one to debug. It seems something changed in the compiler along the way and we didn't have tests for tags involving generic parameters. The cause was `Expr.summon[Tag[tpe]]` in the macro being interpreted by the compiler as `Expr.summon[String]` since it's in a scope where `opaque type Tag` is known to be a `String`. The fix moves the macro to another scope so it doesn't have visibility of the underlying type of `Tag`.


kyo-tag/shared/src/main/scala/kyo/internal/TagMacro.scala
```diff
@@ -0,0 +1,174 @@
...
+                    )
+                end if
+            case _ =>
+                tpe.asType match
+                    case '[tpe] =>
+                        Expr.summon[Tag[tpe]] match
+                            case None =>
+                                failMissing(TypeRepr.of[tpe])
+                            case Some(value) =>
+                                '{ $value.raw }
```
fwbrasil, line 106: This line was the only change in the macro to transform the tag into a string.

kyo-tag/shared/src/test/scala/kyo/TagTest.scala
```diff
@@ -1,4 +1,4 @@
-package kyoTest
+package kyo
```
fwbrasil, line 1: cleanup

kyo-tag/shared/src/main/scala/kyo/internal/TagMacro.scala
```diff
@@ -0,0 +1,174 @@
+package kyo.internal
+
+import kyo.*
+import kyo.Tag.*
+import scala.quoted.*
+
+object TagMacro:
```
hearnadam, line 7: Can this be `private [kyo]`? or is that an issue with inlining?

hearnadam (approved):

PR #677, opened 2024-09-22
cats/zios: reproduce interruption issues

fwbrasil:
The recent CI instability with the `cats-effect` tests is a symptom of a bug. I was expecting the interruption thunk provided to `IO.async` to be evaluated independently of the fiber yielding the execution to the scheduler but that's not the case. Additionally, interruptions from Kyo aren't being propagated to cats-effect.

This issue wasn't detected earlier because the tests copied from `kyo-zio` were incorrect. For ZIO, the interruption from ZIO to Kyo via the provided thunk to `ZIO.asyncInterrupt` is working properly but propagating interruptions from Kyo to ZIO is also not working.

This PR fixes the tests to show the issue.


fwbrasil:
Merging to solve the CI instability

PR #685, opened 2024-09-24
core: async fixes

fwbrasil:
I noticed that https://github.com/getkyo/kyo/pull/666 missed tests for `fiber.mapResult` and, when I added them, it reproduced a bug. I've also introduced added a new `IOPromiseTest` that was missing, which also uncovered a bug with `interrupts`.

### DIFF
--- a/kyo-cats/shared/src/main/scala/kyo/Cats.scala
+++ b/kyo-cats/shared/src/main/scala/kyo/Cats.scala
@@ -34,7 +34,7 @@ object Cats:
                 .map { fiber =>
                     CatsIO.async[CatsIO[A]] { cb =>
                         CatsIO {
-                            fiber.unsafe.onComplete(r => cb(r.toEither))
+                            fiber.unsafe.onResult(r => cb(r.toEither))
                             Some(CatsIO(fiber.unsafe.interrupt(Result.Panic(Fiber.Interrupted(frame)))).void)
                         }
                     }
--- a/kyo-cats/shared/src/test/scala/kyo/CatsTest.scala
+++ b/kyo-cats/shared/src/test/scala/kyo/CatsTest.scala
@@ -91,7 +91,7 @@ class CatsTest extends Test:
                     _ <- f.cancel
                     r <- f.join
                 yield
-                    assert(cdl.await(10, TimeUnit.MILLISECONDS))
+                    assert(cdl.await(100, TimeUnit.MILLISECONDS))
                     assert(r.isCanceled)
                 end for
             }
@@ -104,7 +104,7 @@ class CatsTest extends Test:
                     _ <- f.interrupt(panic)
                     r <- f.getResult
                 yield
-                    assert(cdl.await(10, TimeUnit.MILLISECONDS))
+                    assert(cdl.await(100, TimeUnit.MILLISECONDS))
                     assert(r == panic)
                 end for
             }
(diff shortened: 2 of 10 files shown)


kyo-cats/shared/src/test/scala/kyo/CatsTest.scala
```diff
@@ -91,7 +91,7 @@ class CatsTest extends Test:
                     _ <- f.cancel
                     r <- f.join
                 yield
-                    assert(cdl.await(10, TimeUnit.MILLISECONDS))
+                    assert(cdl.await(100, TimeUnit.MILLISECONDS))
```
fwbrasil, line 94: unrelated change to avoid intermittent test failures

kyo-cats/shared/src/main/scala/kyo/Cats.scala
```diff
@@ -34,7 +34,7 @@ object Cats:
                 .map { fiber =>
                     CatsIO.async[CatsIO[A]] { cb =>
                         CatsIO {
-                            fiber.unsafe.onComplete(r => cb(r.toEither))
+                            fiber.unsafe.onResult(r => cb(r.toEither))
```
fwbrasil, line 37: I've renamed `IOPromise.onComplete` to `onResult` so it doesn't conflict with the `Fiber.onComplete` extension method in `Async.scala`.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -272,7 +272,7 @@ object Async:
             def toFuture(using E <:< Throwable, Frame): Future[A] < IO =
                 IO {
                     val r = scala.concurrent.Promise[A]()
-                    self.onComplete { v =>
```
fwbrasil, line 275: `toFuture`, `map`, `flatMap`, and `mapResult` were all incorrect. The compiler was resolving the `onComplete` extension method for `Fiber` and not the correct `IOPromise.onComplete`, which produced an `Unit < IO` and didn't execute anything. The rename to `onResult` resolves the ambiguity.

fwbrasil:
merging to work on the change to merge the `IO` and `Async` effects

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
```diff
@@ -529,4 +557,176 @@ class AsyncTest extends Test:
...
+                    _ <- fiber.complete(Result.success(42))
+                    completedAfterWait = completed
+                yield
+                    assert(notCompletedYet.isEmpty)
+                    assert(completedAfterWait == Maybe(Result.success(42)))
+                end for
+            }
+        }
+
+        "block" - {
```
hearnadam, line 716: Should we have Fiber specific tests for interrupt?

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -228,23 +239,45 @@ private[kyo] object IOPromise extends IOPromisePlatformSpecific:
 
         final def interrupts(p: IOPromise[?, ?]): Pending[E, A] =
             new Pending[E, A]:
+                def interrupt(panic: Panic): Pending[E, A] =
+                    p.interrupt(panic)
+                    self
```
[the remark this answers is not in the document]
fwbrasil: It enables flushing results of the nested `Pending` instances in a stack-safe manner. Take a look at `Pending.flush`/`Pending.flushInterrupt`, they keep flushing while a `Pending.Empty` is not seen. Each instance in the `Pending` chain can have different behaviors to propagate an interrupt or a result.

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
fwbrasil: Given that `Fiber` is just wrapper of `IOPromise`, I don't think we need to replicate the tests. There are a couple of tests in this file that use `Fiber.interrupt` as well.

Issue #687, opened 2024-09-24
Bug: Unexpected pending effect


steinybot:
I was trying to add the following test to `StreamTest` to verify that it pulls the correct amount:
```
        "exact amount" in {
            def emit(remaining: Int)(ack: Ack): Ack < Emit[Chunk[Int]] =
                if remaining <= 0 then Reducible.eliminate[Abort[Nothing]](Abort.panic(new Exception("Boom!")))
                else
                    ack match
                        case Stop        => Stop
                        case Continue(_) => Emit.andMap(Chunk(remaining))(emit(remaining - 1))
            assert(
                Stream(Emit.andMap(Chunk.empty[Int])(emit(5))).take(5).run.eval == (5 to 1)
            )
        }
```

But it fails with:
```
[info] - exact amount *** FAILED *** (7 milliseconds)
[info]   kyo.bug$KyoBugException: Unexpected pending effect while handling Tag[scala.Any]: Kyo(Tag[kyo.Abort], Input(Panic(java.lang.Exception: Boom!)), StreamTest.scala:79:75, mit.andMap(Chunk.empty[Int])(emit(5))).take(5).run) Please open an issue 🥹  https://github.com/getkyo/kyo/issues
[info]   at kyo.bug$KyoBugException$.apply(prelude.scala:17)
[info]   at kyo.bug$.apply(prelude.scala:29)
[info]   at kyo.bug$.failTag(prelude.scala:23)
[info]   at kyo.StreamTest.evalLoop$10(StreamTest.scala:21)
[info]   at kyo.StreamTest.$init$$$anonfun$3$$anonfun$6(StreamTest.scala:78)
[info]   at org.scalatest.freespec.AsyncFreeSpecLike.transformToOutcomeParam$1(AsyncFreeSpecLike.scala:148)
[info]   at org.scalatest.freespec.AsyncFreeSpecLike.registerTestToRun$$anonfun$1(AsyncFreeSpecLike.scala:149)
[info]   at org.scalatest.AsyncTestSuite.transformToOutcome$$anonfun$1(AsyncTestSuite.scala:240)
[info]   at org.scalatest.freespec.AsyncFreeSpecLike$$anon$1.apply(AsyncFreeSpecLike.scala:425)
[info]   at org.scalatest.AsyncTestSuite.withFixture(AsyncTestSuite.scala:316)
```

steinybot:
Not a bug. It is not legal to elide the effect like this.

PR #688, opened 2024-09-24
Add test to very that stream pulls the exact amount

steinybot:
This reproduces https://github.com/getkyo/kyo/issues/687

### DIFF
--- a/kyo-prelude/shared/src/test/scala/kyo/StreamTest.scala
+++ b/kyo-prelude/shared/src/test/scala/kyo/StreamTest.scala
@@ -1,5 +1,9 @@
 package kyo
 
+import kyo.Emit.Ack
+import kyo.Emit.Ack.*
+import kyo.kernel.Reducible
+
 class StreamTest extends Test:
 
     val n = 10000
@@ -61,6 +65,18 @@ class StreamTest extends Test:
             assert(
                 Stream.init(Seq.fill(100000)(1)).take(5).run.eval ==
                     Seq.fill(5)(1)
+            )
+        }
+
+        "exact amount" in pendingUntilFixed {
+            def emit(remaining: Int)(ack: Ack): Ack < Emit[Chunk[Int]] =
+                if remaining <= 0 then Reducible.eliminate[Abort[Nothing]](Abort.panic(new Exception("Boom!")))
+                else
+                    ack match
+                        case Stop        => Stop
+                        case Continue(_) => Emit.andMap(Chunk(remaining))(emit(remaining - 1))
+            assert(
+                Stream(Emit.andMap(Chunk.empty[Int])(emit(5))).take(5).run.eval == (5 to 1)
             )
         }
     }


kyo-prelude/shared/src/test/scala/kyo/StreamTest.scala
```diff
@@ -63,6 +67,18 @@ class StreamTest extends Test:
                     Seq.fill(5)(1)
             )
         }
+
+        "exact amount" in pendingUntilFixed {
+            def emit(remaining: Int)(ack: Ack): Ack < Emit[Chunk[Int]] =
+                if remaining <= 0 then Reducible.eliminate[Abort[Nothing]](Abort.panic(new Exception("Boom!")))
```
steinybot, line 73: Having to eliminate the `Abort[Nothing]` is kinda annoying.

fwbrasil: `Reducible.apply` should be accessible only within Kyo's package. Ideally, we should make it `private[kyo]`. Was the goal to panic without having a pending `Abort[Nothing]`? That's not possible since the effect needs to be tracked and handled, even if it's a panic. Another way to produce a panic is just using `throw` directly but I'd reserve that only for unexpected behaviors. Have you observed issues with the stream not pulling the correct amount as well?

steinybot: > Was the goal to panic without having a pending Abort[Nothing]? That's not possible since the effect needs to be tracked and handled, even if it's a panic.

It was yeah. Woops, I thought panics were some special case that were somehow handled during `eval`.

For this test I think I can just throw. I'll update it. Is the test actually valid?

On a different branch where I was trying to do the `ZStream` to `Stream` I was testing a `ZStream` that failed after the 5th element. I did a `take(5)` and would get the error instead of the first 5 elements. I suspect I did something wrong or my understanding isn't right but I was trying to add this test just to validate the base case.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -140,7 +140,7 @@ final case class Stream[V, -S](v: Ack < (Emit[Chunk[V]] & S)):
                         else
                             val c   = input.take(state)
                             val nst = state - c.size
-                            Emit.andMap(c)(ack => (nst, cont(ack.maxItems(nst))))
+                            Emit.andMap(c)(ack => (nst, cont(if nst == 0 then Stop else ack.maxItems(nst))))
```
steinybot, line 143: I suspect that this might be unnecessary when the issue with run always passing `Int.MaxValue` is fixed.

kyo-prelude/shared/src/test/scala/kyo/StreamTest.scala
```diff
@@ -63,6 +67,53 @@ class StreamTest extends Test:
...
+                count == 5 && chunk == (1 to 5)
+            )
+        }
+
+        "exact amount" in {
+            def emit(ack: Ack): Ack < (Emit[Chunk[Int]] & Var[Int]) =
+                ack match
+                    case Stop        => Stop: Ack < Any
+                    case Continue(0) => Emit.andMap(Chunk.empty)(emit)
+                    case Continue(maxItems) =>
```
steinybot, line 98: `maxItems` is always `Int.MaxValue` which results in overflow in this case:
```
2147483647
exact amount: start = 1, end = 2147483647, maxItems = 2147483647
2147483647
exact amount: start = -2147483648, end = -2, maxItems = 2147483647
[info] - exact amount *** FAILED *** (4 milliseconds)
[info]   -2 did not equal 5 (StreamTest.scala:113)
```

steinybot: I'm not too sure what the fix is for this.

steinybot: Ahh I think `Continue.maxItems` is wrong. It is taking the max not the min.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
steinybot: Yeah this isn't necessary now. The emit implementation should check for `0`.

kyo-prelude/shared/src/test/scala/kyo/StreamTest.scala
```diff
@@ -63,6 +67,48 @@ class StreamTest extends Test:
                     Seq.fill(5)(1)
             )
         }
+
+        "emit one at a time" in {
+            def emit(ack: Ack): Ack < (Emit[Chunk[Int]] & Var[Int]) =
+                ack match
+                    case Stop        => Stop: Ack < Any
+                    case Continue(0) => Emit.andMap(Chunk.empty)(emit)
```
steinybot, line 75: Kinda weird that we always get `Continue(0)` before `Stop`. Not sure if it is worth doing anything about that.

hearnadam: I believe this is intentional '0' indexing of the Continue. However, I think we could change `Stop = 0` and adjust the related Continue functions.

steinybot: That would make more sense. I can't think of a valid use case for `Continue(0)` other than stopping.

fwbrasil: Agreed, `Stop == 0` seems to make more sense.

PR #690, opened 2024-09-24
prelude: eliminate Abort[Nothing] for panic (#689)

### LINKED ISSUE #689: Improve API for using Abort.panic when there are no normal failures
If you just use `Abort.panic` to fatally abort something you end up with `Abort[Nothing]` which you have to then eliminate using `Reducible.eliminate[Abort[Nothing]]` which is a bit cumbersome. For example:
https://github.com/getkyo/kyo/blob/a62912b34b064c8b860828fc5a8dcb553bbc76af/kyo-prelude/shared/src/test/scala/kyo/StreamTest.scala#L72-L77

steinybot:
Fixes #689.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Abort.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Abort.scala
@@ -30,7 +30,8 @@ object Abort:
       * @return
       *   A computation that immediately fails with the given value
       */
-    inline def fail[E](inline value: E)(using inline frame: Frame): Nothing < Abort[E] = error(Fail(value))
+    inline def fail[E](inline value: E)(using inline frame: Frame): Nothing < Abort[E] =
+        error(Fail(value))
 
     /** Fails the computation with a panic value (unchecked exception).
       *
@@ -39,7 +40,8 @@ object Abort:
       * @return
       *   A computation that immediately fails with the given exception
       */
-    inline def panic[E](inline ex: Throwable)(using inline frame: Frame): Nothing < Abort[E] = error(Panic(ex))
+    inline def panic(inline ex: Throwable)(using inline frame: Frame): Nothing < Any =
+        Reducible.eliminate[Abort[Nothing]](error(Panic(ex)))
 
     inline def error[E](inline e: Error[E])(using inline frame: Frame): Nothing < Abort[E] =
         ArrowEffect.suspendMap[Any](erasedTag[E], e)(_ => ???)
--- a/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
@@ -252,7 +252,7 @@ class AsyncTest extends Test:
             val ex = new Exception
             Async.race(
                 Async.sleep(1.milli).andThen(42),
-                Abort.panic[Exception](ex)
+                Abort.panic(ex)
             ).map { r =>
                 assert(r == 42)
             }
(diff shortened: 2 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -39,7 +40,8 @@ object Abort:
       * @return
       *   A computation that immediately fails with the given exception
       */
-    inline def panic[E](inline ex: Throwable)(using inline frame: Frame): Nothing < Abort[E] = error(Panic(ex))
+    inline def panic(inline ex: Throwable)(using inline frame: Frame): Nothing < Any =
```
steinybot, line 43: I'm not sure if `Nothing < Any` is the best. It looks a bit odd but I think it is correct. Alternatively it could be `Nothing < Abort[Nothing]` (or even keep the `E`) and have some other way of reducing it.

fwbrasil: I see now that your intent is to be able to panic without having to track `Abort[Nothing]`. We'd need a kernel change for that and introduce something like a `ShortCircuitEffect` in addition to `ArrowEffect` and `ContextEffect`. It seems a viable design. If we do that, it's the same as not tracking `Abort[Throwable]`, which is essentially a `Result.Panic`. The `Nothing` is just a product of how `Result` is encoded. Perhaps we should work on this as part of https://github.com/getkyo/kyo/issues/651? I'd prefer to avoid having to track `Abort[Throwable]` as well.

steinybot: Yeah so actually the root of this (I think) originated when I was dealing with `ZIO[Any, Nothing, A]` and I was ending up with an `Abort[Nothing]` but you can't use `Abort.run` on it because you can't get a `ClassTag[Nothing]` or a `Tag[Nothing]`. I don't actually mind the `Abort[Nothing]` so much if it was easier to work with.

> Perhaps we should work on this as part of https://github.com/getkyo/kyo/issues/651? I'd prefer to avoid having to track Abort[Throwable] as well.

Sounds good to me.

As a temporary solution should I add a `run` that works when `E` is nothing?

fwbrasil: Thank you for the clarifications! We could enable tags for `Nothing` but, in many cases, inferring a `Tag[Nothing]` is an indication of misuse. `ClassTag` is another challenge. I hand't realized that it's not possible to handle `Abort[Nothing]` and I believe it should be supported since it's the equivalent of a `Panic` without any tracked exceptions. It seems a dedicated API could cover this scenario.

sideeffffect: > I'm not sure if `Nothing < Any` is the best.

I think it is. It is in line with _ZIO-inspired error handling_ as discussed in
 * https://github.com/getkyo/kyo/issues/651

If something is an _unrecoverable/unexpected failure_, it shouldn't be reflected in the type. So `Nothing < Any` is indeed the correct type in this.

To put it in contrast, _recoverable/expected errors_ should be reflected in the type, and so `Nothing < Abort[E]` is the correct type for _that_ situation, where `E` is the error.

fwbrasil: We've landed on a solution to allow handling `Abort[Nothing]` in https://github.com/getkyo/kyo/pull/695. I was taking a look at what would be necessary to make `Abort.panic` return `Nothing < Any` but I can't see a simple way to encode it in the kernel. I'd require handling possible panics on each `map` call and we need to avoid increasing the inlined bytecode size. Let's wait and see if just fixing the handling of `Abort[Nothing]` is enough. @steinybot thank you for the report and work on this issue!

PR #693, opened 2024-09-25
build: enable nonunit-statement compiler option

fwbrasil:
The compiler needs this other flag to detect the issue discussed in https://github.com/getkyo/kyo/pull/685. I've fixed the warnings and didn't notice bugs besides the expected ones in `Async.scala`

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,23 @@ Replace `<version>` with the latest version: ![Version](https://img.shields.io/m
 
 Kyo utilizes features from the latest Scala 3 versions that are not yet properly supported by IntelliJ IDEA. For the best development experience and to ensure all Kyo features are correctly recognized, we recommend using a [Metals-based](https://scalameta.org/metals/) IDE for your Kyo projects.
 
+### Recommended Compiler Flags
+
+We strongly recommend enabling two Scala compiler flags when working with Kyo:
+
+1. `-Wvalue-discard`: Warns when non-Unit expression results are unused.
+2. `-Wnonunit-statement`: Warns when non-Unit expressions are used in statement position.
+
+Add these to your `build.sbt`:
+
+```scala 
+scalacOptions ++= Seq("-Wvalue-discard", "-Wnonunit-statement")
+```
+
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -74,6 +74,23 @@ Replace `<version>` with the latest version: ![Version](https://img.shields.io/m
 
 Kyo utilizes features from the latest Scala 3 versions that are not yet properly supported by IntelliJ IDEA. For the best development experience and to ensure all Kyo features are correctly recognized, we recommend using a [Metals-based](https://scalameta.org/metals/) IDE for your Kyo projects.
 
+### Recommended Compiler Flags
```
fwbrasil, line 77: Adding to the readme since it's an important aspect for usability.

build.sbt
```diff
@@ -49,6 +50,7 @@ lazy val `kyo-settings` = Seq(
     scalaVersion       := scala3Version,
     crossScalaVersions := List(scala3Version),
     scalacOptions ++= scalacOptionTokens(compilerOptions).value,
+    Test / scalacOptions --= scalacOptionTokens(Set(ScalacOptions.warnNonUnitStatement)).value,
```
fwbrasil, line 53: The tests generate a lot of warnings because there are many `assert` calls that don't use the result so I'm disabling the flag for them.

README.md
```diff
@@ -74,6 +74,23 @@ Replace `<version>` with the latest version: ![Version](https://img.shields.io/m
...
+
+Add these to your `build.sbt`:
+
+```scala 
+scalacOptions ++= Seq("-Wvalue-discard", "-Wnonunit-statement")
+```
+
+These flags help catch unintended effect discards, enforce proper effect sequencing, and prevent subtle bugs in Kyo applications.
+
+Note: In test code, you might want to selectively disable these warnings, as tests often assert side effects without using returned values.
```
hearnadam, line 92: zio-test asserts will only be used if returned - not using exceptions for flow control 😝

hearnadam (approved):

PR #695, opened 2024-09-26
prelude: support handling Abort[Nothing]

fwbrasil:
A solution for https://github.com/getkyo/kyo/issues/689 inspired by @steinybot's approach using a type class in https://github.com/getkyo/kyo/pull/692. I noticed that it's possible to simplify it by encoding only the necessary runtime information in the type class instead of the actual method implementation.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Abort.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Abort.scala
@@ -20,8 +20,9 @@ sealed trait Abort[-E] extends ArrowEffect[Const[Error[E]], Const[Unit]]
 
 object Abort:
 
+    import internal.*
+
     given eliminateAbort: Reducible.Eliminable[Abort[Nothing]] with {}
-    private inline def erasedTag[E]: Tag[Abort[E]] = Tag[Abort[Any]].asInstanceOf[Tag[Abort[E]]]
 
     /** Fails the computation with the given value.
       *
@@ -161,11 +162,11 @@ object Abort:
           */
         def apply[A: Flat, S, ER](v: => A < (Abort[E | ER] & S))(
             using
-            ct: ClassTag[E],
-            tag: Tag[E],
             frame: Frame,
+            abortable: Abortable[E],
             reduce: Reducible[Abort[ER]]
         ): Result[E, A] < (S & reduce.SReduced) =
+            given ct: ClassTag[E] = abortable.classTag
             reduce {
                 ArrowEffect.handle.catching[
                     Const[Error[E]],
@@ -194,11 +195,12 @@ object Abort:
                         case fail => Result.panic(fail)
                 )
             }
+        end apply
     end RunOps
 
     /** Runs an Abort effect. This operation handles the Abort effect, converting it into a Result type.
       */
-    inline def run[E >: Nothing]: RunOps[E] = RunOps(())
+    inline def run[E]: RunOps[E] = RunOps(())
 
     final class CatchingOps[E <: Throwable](dummy: Unit) extends AnyVal:
         /** Catches exceptions of type E and converts them to Abort failures.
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -227,4 +229,16 @@ object Abort:
       */
     inline def catching[E <: Throwable]: CatchingOps[E] = CatchingOps(())
 
+    object internal:
+        final case class Abortable[E] private (classTag: ClassTag[E]) extends AnyVal
+
+        object Abortable:
+            inline given Abortable[Nothing] = Abortable[Nothing](ClassTag.Nothing)
+            inline given [E](using inline ct: ClassTag[E], inline tag: Tag[E]): Abortable[E] =
```
fwbrasil, line 237: It's nice that `tag` is erased by the compiler with this new encoding since it's `inline` and unused.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -20,8 +20,9 @@ sealed trait Abort[-E] extends ArrowEffect[Const[Error[E]], Const[Unit]]
 
 object Abort:
 
+    import internal.*
+
     given eliminateAbort: Reducible.Eliminable[Abort[Nothing]] with {}
-    private inline def erasedTag[E]: Tag[Abort[E]] = Tag[Abort[Any]].asInstanceOf[Tag[Abort[E]]]
```
fwbrasil, line 24: just moved to `internal`

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
steinybot: Sweet!

Are you using javap to check or do you just know? I haven't built up a good intuition for inline yet.

If it is unused why do we have it?

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
```diff
@@ -628,4 +628,39 @@ class AbortsTest extends Test:
         succeed
     }
 
+    "Abortable" - {
+        "handle Abort[Nothing]" in {
+            val computation: Int < Abort[Nothing] = Abort.panic(new RuntimeException("Panic!"))
+            val result                            = Abort.run(computation).eval
+            val _: Result[Nothing, Int]           = result
+            assert(result.isPanic)
```
steinybot, line 636: Is it worth checking that it is the same exception?

steinybot (approved):
Nice! Much better than my attempts.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
fwbrasil: An `inline` parameter gets spliced on each of use. For example:

```scala
inline def test(inline p1: Int): Int = p1 + p1

test(1) // expands to 1 + 1
test(Math.sqrt(42)) // expands to Math.sqrt(42) + Math.sqrt(42)
```

Semantically, an `inline` parameter is equivalent to a by-name parameter in a regular method. If the parameter is never used, it's never spliced and doesn't make it to the generated bytecode.

fwbrasil: > If it is unused why do we have it?

The `Tag` here is used to disallow handling multiple aborts at once like in `Abort.run[Int | String]`. In those cases, the class tag is inferred as `Any`, which ends up handling all aborts. We might be able to support this scenario eventually but it'd require `Tag` to efficiently encode type unions and provide a way to check tags against the runtime classes of failure values.

steinybot: > If the parameter is never used, it's never spliced and doesn't make it to the generated bytecode.

Right that should have been obvious. On a bit of a tangent now but I guess that means the compiler still runs the macro but then it gets eliminated in a later phase.

> The Tag here is used to disallow handling multiple aborts at once like in Abort.run[Int | String] since the macro currently doesn't materialize tags for union and intersection types. In those cases, the class tag is inferred as Any, which ends up handling all aborts. We might be able to support this scenario eventually but it'd require Tag to efficiently encode union types and provide a way to check tags against the runtime classes of failure values.

Ah good point. I wonder if we should alias it or something when it is used for this purpose? Might be clearer to anyone reading it in future. Also if `Tag` ever changes to support unions it might be tough to find them all.

steinybot (approved):

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
fwbrasil: > Right that should have been obvious. On a bit of a tangent now but I guess that means the compiler still runs the macro but then it gets eliminated in a later phase.

Yep, it's essentially free abstraction :) The tag will only exist at compile time and `Abortable` also won't be in the generated bytecode given that it's an `AnyVal`.

> Ah good point. I wonder if we should alias it or something when it is used for this purpose? Might be clearer to anyone reading it in future. Also if Tag ever changes to support unions it might be tough to find them all.

Makes sense, I'll add a comment clarifying that for now.

PR #697, opened 2024-09-27
Upgrade Caliban to 2.9.0

ghostdogpr:
`JsonCodec` from Tapir are no longer needed since it's now handled internally by Caliban.


steinybot (approved):

fwbrasil:
that was quick :) thank you @ghostdogpr!

PR #700, opened 2024-09-27
prelude: fix order of empty Stream type parameters

steinybot:


kyo-prelude/shared/src/test/scala/kyo/StreamTest.scala
```diff
@@ -8,6 +8,12 @@ class StreamTest extends Test:
 
     val n = 10000
 
+    "empty" in {
+        assert(
+            Stream.empty[Int].run.eval == Seq.empty
```
steinybot, line 13: This previously didn't compile since it ended up as `Chunk[Any] < Int` rather than `Chunk[Int] < Any`

hearnadam (approved):

PR #702, opened 2024-09-27
core: first-class support for unsafe APIs

fwbrasil:
We currently don't offer user-facing unsafe APIs. This PR introduces unsafe counterparts for most of the primitives in  `kyo-core`, which are meant to be used in more low-level code that require unsafe execution for performance or integration with other libraries.

To provide some measure of safety, all unsafe APIs, including `IO.run`, require a new `AllowUnsafe` evidence, which can be provided by `IO.appy` or by importing `AllowUnsafe.embrace.danger`.

I haven't added unsafe APIs for `Channel`, `Meter`,  and `Hub`. It's possible but a bit tricky because `Channel` has optimizations that require mixing unsafe execution with regular composition. I'll follow up on them separately.


kyo-combinators/shared/src/test/scala/kyo/ChoiceCombinatorTest.scala
```diff
@@ -64,46 +64,50 @@ class ChoiceCombinatorTest extends Test:
         }
 
         "iteration" - {
-            "should iterate using foreach" in {
+            "should iterate using foreach" in run {
```
fwbrasil, line 67: I've changed tests to avoid `IO.run` if possible

kyo-core/shared/src/main/scala/kyo/Adder.scala
```diff
@@ -13,56 +13,75 @@ class LongAdder private[kyo] (private val ref: j.LongAdder) extends AnyVal:
...
+
     /** Creates a new LongAdder instance.
       *
       * @return
       *   A new LongAdder
       */
-    def init(using Frame): LongAdder < IO = IO(LongAdder(new j.LongAdder))
+    def init(using frame: Frame): LongAdder < IO = IO(LongAdder(Unsafe.init()))
+
+    opaque type Unsafe = j.LongAdder
```
fwbrasil, line 64: The pattern is offering unsafe APIs like `LongAdder.Unsafe` and `Latch.Unsafe` with methods `unsafe`/`safe` to switch between the regular API and the `Unsafe` version.

kyo-core/shared/src/main/scala/kyo/Adder.scala
```diff
@@ -13,56 +13,75 @@ class LongAdder private[kyo] (private val ref: j.LongAdder) extends AnyVal:
...
+
+    object Unsafe:
+        given Flat[Unsafe]                    = Flat.unsafe.bypass
+        def init()(using AllowUnsafe): Unsafe = new j.LongAdder
```
fwbrasil, line 68: I haven't added scaladocs for unsafe APIs because the source files are getting large but I can add them if we judge it's necessary

kyo-core/shared/src/main/scala/kyo/Adder.scala
```diff
@@ -71,28 +90,29 @@ class DoubleAdder private[kyo] (private val ref: j.DoubleAdder) extends AnyVal:
       * @return
       *   Unit
       */
-    inline def add(v: Double)(using Frame): Unit < IO = IO(ref.add(v))
+    inline def add(v: Double)(using inline frame: Frame): Unit < IO = IO(unsafe.add(v))
```
fwbrasil, line 93: minor cleanup: there was no reason not to inline `frame`, which avoids an extra field in the `IO` suspension.

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -1,75 +1,150 @@
...
 
+/** A clock that provides time-related operations.
+  */
 abstract class Clock:
-    def now(using Frame): Instant < IO
-    def stopwatch(using Frame): Clock.Stopwatch < IO
-    def deadline(duration: Duration)(using Frame): Clock.Deadline < IO
     def unsafe: Clock.Unsafe
+
+    /** Gets the current time as an Instant.
```
fwbrasil, line 12: added missing scaladocs

kyo-direct/shared/src/test/scala/kyo/HygieneTest.scala
```diff
@@ -5,10 +5,6 @@ import org.scalatest.freespec.AnyFreeSpec
 
 class HygieneTest extends AnyFreeSpec with Assertions:
 
-    "ok" in {
```
fwbrasil, line 8: this test doesn't seem to make sense so I've removed it

kyo-core/shared/src/main/scala/kyo/Console.scala
```diff
@@ -48,20 +50,21 @@ object Console:
...
     val live: Console =
-        new Console:
-            def readln(using Frame) =
-                IO {
+        Console(
+            new Unsafe:
+                def readln()(using AllowUnsafe) =
                     val line = scala.Console.in.readLine()
                     if line == null then
                         throw new EOFException("Consoles.readln failed.")
```
hearnadam, line 58: pre-existing: Should we use Abort or Maybe?

kyo-core/shared/src/main/scala/kyo/Adder.scala
fwbrasil: Unsafe in this context is an operation that can break referential transparency, which is the case for the initialization of an object containing mutable state. I think the mental model should be that these are advanced low-level APIs meant only for scenarios like libraries where performance is a major concern or integrations with other solutions. 

>  Perhaps just a boilerplate on object Unsafe that says something to the effect that every function in there has a side effect so only use if necessary and within IO.apply would be good enough?

Do you mean that `AllowUnsafe` isn't necessary? I think having the extra protection is important so newcomers aren't tempted to use these APIs.

kyo-core/shared/src/main/scala/kyo/AllowUnsafe.scala
```diff
@@ -0,0 +1,28 @@
...
+
+@implicitNotFound("""
+Unsafe operation detected!
+This code requires explicit permission to perform unsafe operations that
+break referential transparency.
+
+Recommended approach:
+Wrap your unsafe code in IO(...) to properly manage side effects:
+
+IO { /* your unsafe code here */ }
```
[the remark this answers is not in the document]
fwbrasil: I'm having second thoughts on `IO.apply` providing the `AllowUnsafe` evidence. It'd encourage the use of unsafe APIs that are meant only for more advanced users for performance or integrations. I'm considering adding a new `IO.Unsafe.apply` method instead. We could also move `IO.run` to `IO.Unsafe.run`. Wdyt?

@steinybot I think you're leaning toward being more relaxed in relation to referential transparency but that's an essential aspect to ensure Kyo's effects compose correctly. Have you considered using the direct syntax? It drops referential transparency but keeps composability since it gets translated to referentially-transparent code.

kyo-core/shared/src/main/scala/kyo/AllowUnsafe.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import AllowUnsafe.embrace.danger // Avoid!
+
+WARNING: Using AllowUnsafe directly bypasses important safety mechanisms.
+Ensure you fully understand the risks before proceeding this way.""")
+opaque type AllowUnsafe = Null
+
+object AllowUnsafe:
+    object embrace:
+        inline given danger: AllowUnsafe = null
```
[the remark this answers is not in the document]
fwbrasil: Is the idea that users would need to add `given AllowUnsafe = AllowUnsafe.embrace.danger` instead of `import AllowUnsafe.embrace.danger`? We'd still need a way to allow multiple unsafe calls in a scope without introducing `IO` suspensions. For example, a `Channel.Unsafe` impl will need to call multiple `Queue.Unsafe` methods without suspension and `IO` would be introduce by the `Channel(unsafe: UnsafeChannel)` wrapper.

kyo-core/shared/src/main/scala/kyo/AllowUnsafe.scala
```diff
@@ -0,0 +1,28 @@
...
+If you must proceed without IO, you have to explicitly embrace the danger:
+
+import AllowUnsafe.embrace.danger // Avoid!
+
+WARNING: Using AllowUnsafe directly bypasses important safety mechanisms.
+Ensure you fully understand the risks before proceeding this way.""")
+opaque type AllowUnsafe = Null
+
+object AllowUnsafe:
+    object embrace:
```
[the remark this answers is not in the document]
fwbrasil: We need the nesting because adding the `given` to the `AllowUnsafe` companion object directly would make the implicit available everywhere

kyo-core/shared/src/main/scala/kyo/IO.scala
```diff
@@ -78,7 +79,7 @@ object IO:
       * @return
       *   The result of the IO effect after executing its side effects.
       */
-    def run[A: Flat, S](v: A < IO)(using Frame): A < Any =
+    def run[A: Flat, S](v: A < IO)(using Frame, AllowUnsafe): A < Any =
```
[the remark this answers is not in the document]
fwbrasil: Yes, I'll move it to an `Unsafe` object

kyo-core/shared/src/main/scala/kyo/Adder.scala
```diff
@@ -4,7 +4,7 @@ import java.util.concurrent.atomic as j
 
 /** A wrapper for Java's LongAdder.
   */
-class LongAdder private[kyo] (private val ref: j.LongAdder) extends AnyVal:
+final case class LongAdder private (unsafe: LongAdder.Unsafe) extends AnyVal:
```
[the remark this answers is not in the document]
fwbrasil: I tried that initially but it's problematic because the compiler has trouble to know which extension method should be used. The extensions for both the regular and unsafe versions become valid in the scope.

kyo-core/shared/src/main/scala/kyo/Adder.scala
steinybot: No I was meaning that instead of documenting each method there could be a boilerplate copy/pasted doc on the object that says basically what you just said, that it is low level, avoid using, if you do use then put it inside `IO`.

kyo-core/shared/src/main/scala/kyo/Adder.scala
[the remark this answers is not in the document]
steinybot: Yeah I had a tinker and found the same thing. Some of the `Unsafe`s are class like `Barrier` and `Latch` so I think those can be type aliases.

kyo-core/shared/src/main/scala/kyo/Adder.scala
fwbrasil: I had missed the context from my initial comment 🤦🏽 thanks for clarifying! I've added a top-level warning scaladoc to all unsafe APIs.

kyo-core/shared/src/main/scala/kyo/AllowUnsafe.scala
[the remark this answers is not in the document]
fwbrasil: I've introduced `IO.Unsafe.apply`, which is similar to `IO.apply` but provides the `AllowUnsafe` evidence, and moved the IO run methods to it.

fwbrasil:
folks, anything pending here? I'm avoiding changes in `kyo-core` because of this PR

kyo-core/shared/src/main/scala/kyo/AllowUnsafe.scala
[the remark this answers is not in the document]
fwbrasil: That's a valid concern but passing the evidence manually on each call feels too cumbersome if we consider unsafe access a first-class feature. Depending on where the unsafe code is used, it might not be trivial to pass the evidence as parameter as well. For example, it might not be possible to do that if we're implementing some interface required by an integration.

I feel the compile-time error message paired with the explicit indication of danger of the import is enough to avoid users overlooking the use of unsafe APIs. We can review the approach if there's indication of misuse due to convenience.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -270,7 +289,9 @@ object Async:
               * @param f
               *   The callback function
               */
-            def onComplete(f: Result[E, A] => Unit < IO)(using Frame): Unit < IO = IO(self.onComplete(r => IO.run(f(r)).eval))
+            def onComplete(f: Result[E, A] => Unit < IO)(using Frame): Unit < IO =
+                import AllowUnsafe.embrace.danger
+                IO(self.onComplete(r => IO.Unsafe.run(f(r)).eval))
```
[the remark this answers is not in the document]
fwbrasil: In this case, the `onComplete` function is already in an unsafe execution context so the suspension doesn't seem necessary

PR #707, opened 2024-09-29
core: async masking

fwbrasil:
Introduces a feature similar to `uninterruptible` in other effect systems. It's a mechanism similar to `IO.ensure` but encoded in the async execution.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -202,7 +202,7 @@ object Async:
           * @return
           *   A Fiber that completes with the result of the Future
           */
-        def fromFuture[A: Flat](f: Future[A])(using frame: Frame): Fiber[Throwable, A] < IO =
+        def fromFuture[A](f: Future[A])(using frame: Frame): Fiber[Throwable, A] < IO =
             import scala.util.*
             IO {
                 val p = new IOPromise[Throwable, A] with (Try[A] => Unit):
@@ -270,7 +270,19 @@ object Async:
               * @param f
               *   The callback function
               */
-            def onComplete(f: Result[E, A] => Unit < IO)(using Frame): Unit < IO = IO(self.onResult(r => IO.run(f(r)).eval))
+            def onComplete(f: Result[E, A] => Unit < IO)(using Frame): Unit < IO = IO(self.onComplete(r => IO.run(f(r)).eval))
+
+            /** Registers a callback to be called when the Fiber is interrupted.
+              *
+              * This method allows you to specify a callback that will be executed if the Fiber is interrupted. The callback receives the
+              * Panic value that caused the interruption.
+              *
+              * @param f
+              *   The callback function to be executed on interruption
+              * @return
+              *   A unit value wrapped in IO, representing the registration of the callback
+              */
+            def onInterrupt(f: Panic => Unit < IO)(using Frame): Unit < IO = IO(self.onInterrupt(r => IO.run(f(r)).eval))
 
             /** Blocks until the Fiber completes or the timeout is reached.
               *
@@ -289,7 +301,7 @@ object Async:
             def toFuture(using E <:< Throwable, Frame): Future[A] < IO =
                 IO {
                     val r = scala.concurrent.Promise[A]()
-                    self.onResult { v =>
+                    self.onComplete { v =>
                         r.complete(v.toTry)
                     }
                     r.future
@@ -306,7 +318,7 @@ object Async:
                 IO {
                     val p = new IOPromise[E, B](interrupts = self) with (Result[E, A] => Unit):
                         def apply(v: Result[E, A]) = completeUnit(v.map(f))
-                    self.onResult(p)
+                    self.onComplete(p)
                     p
                 }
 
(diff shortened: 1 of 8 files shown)


fwbrasil:
I've just pushed additional changes:

1. Introduce `fiber.onInterrupt` to be used for the new zio/cats-effect integrations
2. Rename `IOPromise.onResult` back to `onComplete` since https://github.com/getkyo/kyo/pull/693 will make bugs like https://github.com/getkyo/kyo/pull/685 become compile-time warnings and I couldn't think of an alternative name for `IOPromise.onInterrupt`.
3. Remove `Flat` requirement from `fromFuture` methods. It's not necessary since the method doesn't handle effects

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -246,21 +264,27 @@ private[kyo] object IOPromise extends IOPromisePlatformSpecific:
...
-                        tail
-                    case p: Pending[E, A] =>
-                        runLoop(p.run(v), v)
+                    case _ if (p eq Pending.Empty) => tail
+                    case p: Pending[E, A]          => runLoop(p.run(v), v)
 
             @tailrec def interruptLoop(p: Pending[E, A], panic: Panic): Pending[E, A] =
                 p match
-                    case _ if (p eq Pending.Empty) =>
-                        self
```
fwbrasil, line 261: There was a bug here. It should return `tail` to continue flushing, not `self`.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -74,6 +74,24 @@ object Async:
...
+      * This method executes the given computation in a context where interrupts are not propagated to previous "steps" of the computation.
+      * The returned computation can still be interrupted, but the interruption won't affect the masked portion. This is useful for ensuring
+      * that cleanup operations or critical sections complete even if an interrupt occurs.
+      *
+      * @param v
+      *   The computation to run with interrupt masking
+      * @return
+      *   The result of the computation, which can still be interrupted
+      */
+    def mask[E, A: Flat, Ctx](v: => A < (Abort[E] & Async & Ctx))(
```
hearnadam, line 88: Is there a reason you didn't go with `uninterruptible`?

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -128,7 +134,19 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
                             given Frame = Frame.internal
                             Log.unsafe.error("uncaught exception", ex)
         onCompleteLoop(this)
-    end onResult
+    end onComplete
+
+    inline def onInterrupt(inline f: Panic => Unit): Unit =
```
hearnadam, line 139: Nice, this will be helpful for zio/cats integrations.

kyo-core/shared/src/test/scala/kyo/scheduler/IOPromiseTest.scala
```diff
@@ -291,4 +291,400 @@ class IOPromiseTest extends Test:
...
+            val masked   = original.mask
+
+            var originalCompleted                         = false
+            var maskedResult: Maybe[Result[Nothing, Int]] = Maybe.Empty
+
+            original.onComplete(_ => originalCompleted = true)
+            masked.onComplete(r => maskedResult = Maybe(r))
+
+            assert(masked.interrupt(Result.Panic(new Exception("Interrupted"))))
+            assert(maskedResult.exists(_.isPanic))
```
hearnadam, line 306: I think this may be surprising to people coming from ZIO. I generally think the mask should probably cover the masked fiber as well...

kyo-core/shared/src/main/scala/kyo/Async.scala
fwbrasil: I normally follow the naming and semantics of the Twitter Future for async primitives and it uses `mask` for this. I'm not a fan of `uninterruptible` because seems a mouthful but I'd be ok with renaming if you strongly about it.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
fwbrasil: yeah, I started working on them and saw that it'd be required otherwise we'd need that workaround like you did with the two promises

kyo-core/shared/src/test/scala/kyo/scheduler/IOPromiseTest.scala
fwbrasil: I was working on changing it but it's not as simple as I thought. Even if we had `onComplete`/`onResult`/`onInterrupt` as discussed privately, we'd still need to have a special `IOPromise` state indicating that it's not interruptible. I'm not sure the complexity and performance cost is worth it. The current implementation follows the behavior of the Twitter Future so it might be ok.

kyo-core/shared/src/main/scala/kyo/Async.scala
hearnadam: Let's proceed with `mask` to avoid users confusing the interrupt-ability of this Fiber.

kyo-core/shared/src/test/scala/kyo/scheduler/IOPromiseTest.scala
hearnadam: Sounds good. We should proceed with this and we can potentially reassess after user feedback.

hearnadam (approved):

PR #708, opened 2024-09-29
prelude: allow union types in Abort.run

### LINKED ISSUE #698: Feature: Add ability to run an Abort with a union of errors
It is a bit cumbersome running an `Abort` that has a union. You have to `run` it repeatedly for each one and then repeatedly `flatten` the `Result`. For example:
```
Abort.run[StatusRuntimeException](Abort.run[StatusException](request)).map(_.flatten)
```

`Abort.run` specifically prevents handling multiple errors at once (see the comment on `Abortable`) so I'm not sure what the solution to this would look like without `Tag` support. Maybe `n` overloads of `run` with increasing number of type parameters (and tediously multiple `RunOps` too)?

fwbrasil:
Fixes #698

My plan was to extend the functionality of `Tag` to allow matching tags against runtime values in use cases like `Abort` but that's something that will take time since we'd need a new implementation to encode union and intersection types in the tag string and figure out a way to safely compare tags to runtime values/classes.

This PR introduces a new `SafeClassTag` as a simpler alternative to solve #698. It has a functionality similar to the standard library `ClassTag` based on runtime classes but without quirks like falling back to `Any` for unions/intersections and it also properly handles special types like `Nothing` and `AnyVal`.

The macro doesn't allow tags for types that aren't known to be a class type like in `Tag`, and, to ensure correctness, it's not possible to obtain a `SafeClassTag` for parametrized types like `List[Int]`. This means that it won't be possible to abort with parametrized types, which I assume isn't very common. We should eventually be able to support parametrized types with a new `Tag` implementation mentioned initially.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/System.scala
+++ b/kyo-core/shared/src/main/scala/kyo/System.scala
@@ -2,6 +2,9 @@ package kyo
 
 import System.Parser
 import java.lang.System as JSystem
+import java.net.MalformedURLException
+import java.net.URISyntaxException
+import java.time.format.DateTimeParseException
 import kyo.kernel.Reducible
 
 /** Represents a system environment with various operations.
@@ -184,32 +187,32 @@ object System:
     /** Companion object for Parser, containing default implementations. */
     object Parser:
         given Parser[Nothing, String]                    = v => Result.success(v)
-        given Parser[NumberFormatException, Int]         = v => Result.catching(v.toInt)
-        given Parser[NumberFormatException, Long]        = v => Result.catching(v.toLong)
-        given Parser[NumberFormatException, Float]       = v => Result.catching(v.toFloat)
-        given Parser[NumberFormatException, Double]      = v => Result.catching(v.toDouble)
-        given Parser[IllegalArgumentException, Boolean]  = v => Result.catching(v.toBoolean)
-        given Parser[NumberFormatException, Byte]        = v => Result.catching(v.toByte)
-        given Parser[NumberFormatException, Short]       = v => Result.catching(v.toShort)
+        given Parser[NumberFormatException, Int]         = v => Result.catching[NumberFormatException](v.toInt)
+        given Parser[NumberFormatException, Long]        = v => Result.catching[NumberFormatException](v.toLong)
+        given Parser[NumberFormatException, Float]       = v => Result.catching[NumberFormatException](v.toFloat)
+        given Parser[NumberFormatException, Double]      = v => Result.catching[NumberFormatException](v.toDouble)
+        given Parser[IllegalArgumentException, Boolean]  = v => Result.catching[IllegalArgumentException](v.toBoolean)
+        given Parser[NumberFormatException, Byte]        = v => Result.catching[NumberFormatException](v.toByte)
+        given Parser[NumberFormatException, Short]       = v => Result.catching[NumberFormatException](v.toShort)
         given Parser[Duration.InvalidDuration, Duration] = v => Duration.parse(v)
 
         given Parser[IllegalArgumentException, java.util.UUID] =
-            v => Result.catching(java.util.UUID.fromString(v))
+            v => Result.catching[IllegalArgumentException](java.util.UUID.fromString(v))
 
-        given Parser[java.time.format.DateTimeParseException, java.time.LocalDate] =
-            v => Result.catching(java.time.LocalDate.parse(v))
+        given Parser[DateTimeParseException, java.time.LocalDate] =
+            v => Result.catching[DateTimeParseException](java.time.LocalDate.parse(v))
 
-        given Parser[java.time.format.DateTimeParseException, java.time.LocalTime] =
-            v => Result.catching(java.time.LocalTime.parse(v))
+        given Parser[DateTimeParseException, java.time.LocalTime] =
+            v => Result.catching[DateTimeParseException](java.time.LocalTime.parse(v))
 
-        given Parser[java.time.format.DateTimeParseException, java.time.LocalDateTime] =
-            v => Result.catching(java.time.LocalDateTime.parse(v))
+        given Parser[DateTimeParseException, java.time.LocalDateTime] =
+            v => Result.catching[DateTimeParseException](java.time.LocalDateTime.parse(v))
 
-        given Parser[java.net.URISyntaxException, java.net.URI] =
-            v => Result.catching(new java.net.URI(v))
+        given Parser[URISyntaxException, java.net.URI] =
+            v => Result.catching[URISyntaxException](new java.net.URI(v))
 
-        given Parser[java.net.MalformedURLException, java.net.URL] =
-            v => Result.catching(new java.net.URL(v))
+        given Parser[MalformedURLException, java.net.URL] =
+            v => Result.catching[MalformedURLException](new java.net.URL(v))
 
         given [E, A](using p: Parser[E, A], frame: Frame): Parser[E, Seq[A]] =
             v => Result.collect(Chunk.from(v.split(",")).map(v => p(v.trim())))
--- a/kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
+++ b/kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
@@ -37,7 +37,7 @@ object SafeClassTag:
                     case Union(elements)        => elements.exists(_.accepts(value))
                     case Intersection(elements) => elements.forall(_.accepts(value))
                     case NothingTag             => false
-                    case UnitTag                => value == ()
+                    case UnitTag                => value.isInstanceOf[Unit]
                     case AnyValTag =>
                         value match
                             case (_: Int | _: Long | _: Double | _: Float | _: Byte | _: Short | _: Char | _: Boolean) =>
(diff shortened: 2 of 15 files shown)


kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -184,15 +184,17 @@ object Abort:
...
                     recover =
-                        case fail: E if classOf[Throwable].isAssignableFrom(ct.runtimeClass) =>
+                        case ct(fail) if ct <:< SafeClassTag[Throwable] =>
                             Result.fail(fail)
-                        case fail => Result.panic(fail)
+                        case fail =>
+                            Result.panic(fail),
+                    done =
+                        case Result.Panic(ct(e)) if ct <:< SafeClassTag[Throwable] =>
+                            Result.Fail(e)
```
fwbrasil, line 196: I noticed a scenario when working on tests where an exception type that was considered a panic in an inner scope continued as a panic even if an outer scope was handling it, which seems an inconsistent state. This is why some tests have changes to match against `Result.fail` instead of `Result.panic`.

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
```diff
@@ -263,6 +263,39 @@ class AbortsTest extends Test:
                 val ex = new Exception
                 Abort.run[Any](throw ex).map(result => assert(result == Result.panic(ex)))
             }
+            "nested panic to fail" - {
```
fwbrasil, line 266: testing the new behavior translating inner panics to fail in an outer scope

kyo-data/shared/src/main/scala/kyo/internal/SafeClassTagMacro.scala
```diff
@@ -0,0 +1,69 @@
...
+                case '[Short]                         => '{ ShortTag }
+                case '[Char]                          => '{ CharTag }
+                case '[Boolean]                       => '{ BooleanTag }
+                case _ if tpe =:= TypeRepr.of[AnyVal] => '{ AnyValTag }
+                case '[t] =>
+                    val classOfSym = Symbol.requiredMethod("scala.Predef.classOf")
+                    val classOfExpr = Select(Ref(Symbol.requiredModule("scala.Predef")), classOfSym)
+                        .appliedToType(TypeRepr.of[t])
+                    val expr = classOfExpr.asExprOf[Any]
+                    '{ $expr.asInstanceOf[SafeClassTag[Any]] }
```
fwbrasil, line 62: this isn't great but I couldn't find a simpler way to generate `classOf` for a `Type`

kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
```diff
@@ -0,0 +1,113 @@
...
+
+        private def isCustomAnyVal(value: Any): Boolean =
+            if isNull(value) then false
+            else
+                val clazz       = value.getClass
+                val anyValClass = classOf[AnyVal]
+                anyValClass.isAssignableFrom(clazz) && (clazz ne anyValClass)
+        end isCustomAnyVal
+
+        def unapply(value: Any): Maybe.Ops[A] =
```
fwbrasil, line 72: I had put `SafeClassTag` in `kyo-tag` initially but moved to `kyo-data` to be able to use `Maybe.Ops` to avoid allocations. I'm planning to follow up with another PR merging `kyo-tag` into `kyo-data`. I don't think there's much benefit in having them separate.

kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
```diff
@@ -0,0 +1,113 @@
...
+            if accepts(value) then Maybe(value.asInstanceOf[A]) else Maybe.empty
+
+        infix def &[B](that: SafeClassTag[B]): SafeClassTag[A & B] = (self, that) match
```
hearnadam, line 75: Tuple pattern matches are better formulated as nested matches to avoid tuple allocations.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
hearnadam: I am not sure I agree with this change. I would have expected that you would need to explicitly lift a Panic -> Fail, like ZIO's resurrect. If you use `Abort.run[Throwable]`, will this lift all Panics to Fail?

kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
```diff
@@ -0,0 +1,113 @@
...
+    extension [A](self: SafeClassTag[A])
+
+        def accepts(value: Any): Boolean =
+            !isNull(value) && {
+                given CanEqual[Any, Any] = CanEqual.derived
+                self match
+                    case Union(elements)        => elements.exists(_.accepts(value))
+                    case Intersection(elements) => elements.forall(_.accepts(value))
+                    case NothingTag             => false
+                    case UnitTag                => value == ()
```
hearnadam, line 40: `value.isInstanceOf[Unit]`?

kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
```diff
@@ -0,0 +1,113 @@
...
+import scala.quoted.*
+
+opaque type SafeClassTag[A] >: SafeClassTag.Element = Class[?] | SafeClassTag.Element
+
+object SafeClassTag:
+    sealed trait Element
+    case class Union(elements: List[SafeClassTag[Any]])        extends Element
+    case class Intersection(elements: List[SafeClassTag[Any]]) extends Element
+
+    sealed trait Primitive extends Element
```
hearnadam, line 15: I think this could also be encoded as a bitmask on a Long, which would be easier to compress (either for intersections or unions, but probably not both?). Not sure if it's worth it given the need for generic type support...

kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
```diff
@@ -0,0 +1,113 @@
...
+                            case _: Short   => self eq ShortTag
+                            case _: Char    => self eq CharTag
+                            case _: Boolean => self eq BooleanTag
+                            case _          => false
+                    case self: Class[?] => self.isInstance(value)
+                end match
+            }
+        end accepts
+
+        private def isCustomAnyVal(value: Any): Boolean =
```
hearnadam, line 64: unused?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
fwbrasil: Yes, it'd transform panics. It felt more natural for `Abort.run[Throwable]` to "catch" any failures in a computation, even if panics. It seems better to discuss that separately, though. I'll revert this change.

kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
fwbrasil: I think in this case it wouldn't be worth it since comparing these static tags via `eq` or by checking the class pointer is cheap. If we used a `Long`, we'd need to worry about boxing as well.

kyo-core/shared/src/main/scala/kyo/System.scala
```diff
@@ -184,32 +187,32 @@ object System:
...
     object Parser:
         given Parser[Nothing, String]                    = v => Result.success(v)
-        given Parser[NumberFormatException, Int]         = v => Result.catching(v.toInt)
-        given Parser[NumberFormatException, Long]        = v => Result.catching(v.toLong)
-        given Parser[NumberFormatException, Float]       = v => Result.catching(v.toFloat)
-        given Parser[NumberFormatException, Double]      = v => Result.catching(v.toDouble)
-        given Parser[IllegalArgumentException, Boolean]  = v => Result.catching(v.toBoolean)
-        given Parser[NumberFormatException, Byte]        = v => Result.catching(v.toByte)
-        given Parser[NumberFormatException, Short]       = v => Result.catching(v.toShort)
+        given Parser[NumberFormatException, Int]         = v => Result.catching[NumberFormatException](v.toInt)
```
fwbrasil, line 190: After https://github.com/getkyo/kyo/pull/708/commits/076a93f98653672cdfafd214bdc5aeb061ffa8a7, I wasn't sure why this class still needed changes for the tests to pass. It turns out that the compiler was inferring `ClassTag[Nothing]`, which incorrectly becomes `java.lang.Object` at runtime and transforms any exceptions to `Fail`. `SafeClassTag` fixes this issue but we now need the explicit type parameter to avoid a panic.

kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
```diff
@@ -0,0 +1,111 @@
...
+                    that match
+                        case NothingTag             => false
+                        case Union(elements)        => elements.exists(self <:< _)
+                        case Intersection(elements) => elements.forall(self <:< _)
+                        case AnyValTag =>
+                            self match
+                                case AnyValTag | IntTag | LongTag | DoubleTag | FloatTag | ByteTag | ShortTag | CharTag | BooleanTag | UnitTag =>
+                                    true
+                                case _ => false
+                        case UnitTag      => self == UnitTag
```
hearnadam, line 101: can be eq?

kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
```diff
@@ -0,0 +1,111 @@
...
+        end accepts
+
+        def unapply(value: Any): Maybe.Ops[A] =
+            if accepts(value) then Maybe(value.asInstanceOf[A]) else Maybe.empty
+
+        infix def &[B](that: SafeClassTag[B]): SafeClassTag[A & B] =
+            self match
+                case Intersection(e1) => that match
+                        case Intersection(e2) => Intersection(e1 ++ e2)
+                        case _                => Intersection(e1 :+ that)
```
hearnadam, line 71: Is there a reason to not prepend for the constant time complexity? Presumably the order shouldn't matter?

kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
```diff
@@ -0,0 +1,111 @@
+package kyo
+
+import kyo.Maybe
+import kyo.internal.SafeClassTagMacro
+import scala.annotation.tailrec
+import scala.quoted.*
+
+opaque type SafeClassTag[A] >: SafeClassTag.Element = Class[?] | SafeClassTag.Element
+
+object SafeClassTag:
```
hearnadam, line 10: 2 general comments
1. Do you think the naming will confuse users? Seems okay to me, just not fully descriptive... top level Scaladoc?
2. Do we want a `show` method? I don't remember what a Class will be serialized as.

hearnadam (approved):

kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
fwbrasil: The name isn't great but I couldn't find a good alternative. I've pushed scaladocs to explain its purpose + the `show` method.

PR #711, opened 2024-10-01
new encoding for zio and cats-effect integrations

### LINKED ISSUE #682: cats/zios: interruption propagation issues
See https://github.com/getkyo/kyo/pull/677 for more information.

fwbrasil:
Fixes #682

The current encoding of the integrations with ZIO and cats-effect uses custom effects and Kyo's Async has limitations to fork computations with arbitrary pending effects. This PR changes `get` to be based directly on `Async` but keeps the `run` method to interpret Kyo computations to the other effect systems.

This approach was explored in collaboration with @hearnadam

### DIFF
--- a/README.md
+++ b/README.md
@@ -3105,8 +3105,8 @@ import kyo.*
 import cats.effect.IO as CatsIO
 import cats.effect.kernel.Outcome.Succeeded
 
-// Note how Cats includes the IO, Async, and Abort[Throwable] effects:
-val a: Int < (Abort[Throwable] & Async) =
+// Note how Cats includes the IO, Async, and Abort[Nothing] effects:
+val a: Int < (Abort[Nothing] & Async) =
     for
         v1 <- Cats.get(CatsIO.pure(21))
         v2 <- IO(21)
@@ -3115,7 +3115,7 @@ val a: Int < (Abort[Throwable] & Async) =
     yield v1 + v2 + v3
 
 // Using fibers from both libraries:
-val b: Int < (Abort[Throwable] & Async) =
+val b: Int < (Abort[Nothing] & Async) =
     for
         f1 <- Cats.get(CatsIO.pure(21).start)
         f2 <- Async.run(IO(21))
@@ -3124,7 +3124,7 @@ val b: Int < (Abort[Throwable] & Async) =
     yield v1 + v2
 
 // Transforming Cats Effect IO within Kyo computations:
-val c: Int < (Abort[Throwable] & Async) =
+val c: Int < (Abort[Nothing] & Async) =
     Cats.get(CatsIO.pure(21)).map(_ * 2)
 
 // Transforming Kyo effects within Cats Effect IO:
--- a/README.md
+++ b/README.md
@@ -3034,11 +3034,11 @@ import kyo.*
 import zio.*
 
 // Use the 'get' method to extract a 'ZIO' effect
-val a: Int < ZIOs =
+val a: Int < (Abort[Nothing] & Async) =
     ZIOs.get(ZIO.succeed(42))
 
 // 'get' also supports error handling with 'Abort'
-val b: Int < (Abort[String] & ZIOs) =
+val b: Int < (Abort[String] & Async) =
     ZIOs.get(ZIO.fail("error"))
 
 // Handle the 'ZIO' effect to obtain a 'ZIO' effect
@@ -3054,7 +3054,7 @@ import zio.*
 
 // Note how ZIO includes the
 // IO and Async effects
-val a: Int < ZIOs =
+val a: Int < (Abort[Nothing] & Async) =
     for
         v1 <- ZIOs.get(ZIO.succeed(21))
         v2 <- IO(21)
@@ -3062,7 +3062,7 @@ val a: Int < ZIOs =
     yield v1 + v2 + v3
 
 // Using fibers from both libraries
-val b: Int < ZIOs =
+val b: Int < (Abort[Nothing] & Async) =
     for
         f1 <- ZIOs.get(ZIO.succeed(21).fork)
         f2 <- Async.run(IO(21))
@@ -3071,7 +3071,7 @@ val b: Int < ZIOs =
     yield v1 + v2
 
 // Transforming ZIO effects within Kyo computations
-val c: Int < ZIOs =
+val c: Int < (Abort[Nothing] & Async) =
     ZIOs.get(ZIO.succeed(21)).map(_ * 2)
 
 // Transforming Kyo effects within ZIO effects
@@ -3090,7 +3090,7 @@ import kyo.*
 import cats.effect.IO as CatsIO
 
 // Use the 'get' method to extract a 'IO' effect from Cats Effect:
-val a: Int < Cats =
+val a: Int < (Abort[Throwable] & Async) =
     Cats.get(CatsIO.pure(42))
 
 // Handle the 'Cats' effect to obtain a 'CatsIO' effect:
@@ -3106,7 +3106,7 @@ import cats.effect.IO as CatsIO
 import cats.effect.kernel.Outcome.Succeeded
 
 // Note how Cats includes the IO, Async, and Abort[Throwable] effects:
-val a: Int < Cats =
+val a: Int < (Abort[Throwable] & Async) =
     for
         v1 <- Cats.get(CatsIO.pure(21))
         v2 <- IO(21)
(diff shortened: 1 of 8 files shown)


kyo-cats/shared/src/main/scala/kyo/Cats.scala
```diff
@@ -16,31 +15,35 @@ object Cats:
...
       *   A Kyo effect that, when run, will execute the cats.effect.IO
       */
-    def get[A](io: CatsIO[A])(using Frame): A < Cats =
-        ArrowEffect.suspendMap(Tag[GetCatsIO], io)(Abort.get(_))
+    def get[A](io: CatsIO[A])(using Frame): A < (Abort[Throwable] & Async) =
+        IO {
+            import cats.effect.unsafe.implicits.global
+            val p                = new IOPromise[Throwable, A]
+            val (result, cancel) = io.unsafeToFutureCancelable()
+            result.onComplete(t => p.complete(Result.fromTry(t)))(ExecutionContext.parasitic)
```
hearnadam, line 23:
```suggestion
            val (future, cancel) = io.unsafeToFutureCancelable()
            future.onComplete(t => p.complete(Result.fromTry(t)))(ExecutionContext.parasitic)
```

kyo-cats/shared/src/main/scala/kyo/Cats.scala
```diff
@@ -16,31 +15,35 @@ object Cats:
...
+            p.onInterrupt(_ => discard(cancel()))
```
hearnadam, line 24: Do we need to await `cancel`? It returns a `Future[Unit]`?

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -1,50 +1,37 @@
...
+      *   A Kyo effect that, when run, will execute the zio.ZIO
       */
-    def get[A](v: ZIO[Any, Nothing, A])(using Frame): A < ZIOs =
-        ArrowEffect.suspend(Tag[GetZIO], v)
+    def get[E, A](v: ZIO[Any, E, A])(using Frame): A < (Abort[E] & Async) =
+        IO {
+            val p = new IOPromise[E, A]
+            val result =
+                Unsafe.unsafe { implicit unsafe =>
+                    Runtime.default.unsafe.runToFuture(v.either)
```
hearnadam, line 26: You can use `.exit` if you want to retain more information, ie `Die` vs `Fail`. I will admit `Exit` + `Cause` contains a significant number of possible cases.

hearnadam, line 26: My preference would be to directly use the Fiber Runtime to interop instead of Future. It's more work, but will save some allocations. You should also be able to avoid the `.either` transformation IIRC.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -1,50 +1,37 @@
...
-      *   A Kyo effect that will produce A
+      *   A Kyo effect that, when run, will execute the zio.ZIO
       */
-    def get[A](v: ZIO[Any, Nothing, A])(using Frame): A < ZIOs =
-        ArrowEffect.suspend(Tag[GetZIO], v)
+    def get[E, A](v: ZIO[Any, E, A])(using Frame): A < (Abort[E] & Async) =
+        IO {
+            val p = new IOPromise[E, A]
+            val result =
+                Unsafe.unsafe { implicit unsafe =>
```
hearnadam, line 25: You can actually use the `Unsafe` object as the evidence. https://github.com/zio/zio/pull/9047

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -55,47 +42,32 @@ object ZIOs:
...
-      * @tparam E
-      *   The error type
-      * @tparam A
-      *   The result type
       * @return
-      *   A ZIO effect that will produce A or fail with E
+      *   A zio.ZIO that, when run, will execute the Kyo effect
       */
-    def run[E, A](v: => A < (Abort[E] & ZIOs))(using frame: Frame): ZIO[Any, E, A] =
+    def run[A: Flat, E](v: => A < (Abort[E] & Async))(using frame: Frame): ZIO[Any, E, A] =
```
hearnadam, line 56: I believe these functions should accept a `using zio.Trace` to avoid polluting stack traces.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -55,47 +42,32 @@ object ZIOs:
...
-                }.pipe(IO.run).eval
-            catch
-                case ex =>
-                    ZIO.isFatalWith { isFatal =>
-                        if !isFatal(ex) then Exit.die(ex)
-                        else throw ex
+            Async.run(v).map { fiber =>
+                ZIO.asyncInterrupt[Any, E, A] { cb =>
+                    fiber.unsafe.onComplete {
+                        case Result.Success(a) => cb(ZIO.succeed(a))
```
hearnadam, line 61:
```suggestion
                        case Result.Success(a) => cb(Exit.succeed(a))
```
Suspension is not necessary here. `ZIO.die/fail` capture additional tracing information that `Exit.die/fail` do not. Not sure if that's necessary.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -55,47 +42,32 @@ object ZIOs:
...
+                        case Result.Fail(e)    => cb(ZIO.fail(e))
```
hearnadam, line 62: Do you need to propagate interruption specifically?


```scala
case Result.Panic(ex: Fiber.Interrupted) => cb(ZIO.fiberIdWith(id => ZIO.failCause(Cause.interrupt(id))))
```

Not sure if the tests check the Cause of ZIO's failure...

README.md
```diff
@@ -3106,7 +3106,7 @@ import cats.effect.IO as CatsIO
 import cats.effect.kernel.Outcome.Succeeded
 
 // Note how Cats includes the IO, Async, and Abort[Throwable] effects:
-val a: Int < Cats =
+val a: Int < (Abort[Throwable] & Async) =
```
fwbrasil, line 3109: I wonder if we should also use `Abort[Nothing]` like in `ZIOs` and treat failures in cat-effect's execution as panics

kyo-cats/shared/src/main/scala/kyo/Cats.scala
fwbrasil: Just evaluating the thunk is enough to dispatch the interrupt. Kyo currently doesn't implement backpressure for interrupts, unlike other effect systems. I'm not convinced it's a good behavior and it's also not the case in the Twitter Future. When a system is under pressure, it's best to offload internal workload as soon as possible and ensure there's proper backpressure on the source of incoming workload via a solution like Kyo's scheduler admission control.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
fwbrasil: I took a look at the APIs but I couldn't find one that provides the thunk to interrupt as well. Do you have specific APIs in mind?

fwbrasil: The `Exit` API confusing to me since methods end up producing `ZIO`. I've added tests to check the translation of ZIO failures to Kyo and it seems correct. ZIO's `die` gets propagated as a failed future.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
fwbrasil: Not sure if that was your suggestion but I've changed it to use `Unsafe.unsafely` that uses a context function to propagate the evidence. Btw, it's nice to see ZIO leveraging `inline` as well! I think I had suggested that some time ago

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
fwbrasil: hmm we could do that but interrupt exceptions aren't limited to `Fiber.Interrupt` in Kyo. Maybe they should?

README.md
fwbrasil: changed to `Abort[Nothing]`

hearnadam (approved):
Approving to unblock merge. We can followup to avoid Future in ZIO interop.

PR #717, opened 2024-10-06
prelude: Flat macro fix

fwbrasil:
Addressing a TODO. The `Flat` macro was incorrectly not considering existing evidences in scope.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Batch.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Batch.scala
@@ -110,8 +110,6 @@ object Batch:
 
         case class Cont(op: Op[?, S], cont: Any => (Cont | A) < (Batch[S] & S & S2))
         def runCont(v: (Cont | A) < (Batch[S] & S & S2)): (Cont | A) < (S & S2) =
-            // TODO workaround, Flat macro isn't inferring correctly with nested classes
-            import Flat.unsafe.bypass
             ArrowEffect.handle(erasedTag[S], v) {
                 [C] => (input, cont) => Cont(input, v => cont(v.asInstanceOf[C]))
             }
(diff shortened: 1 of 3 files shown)


kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -110,8 +110,6 @@ object Batch:
 
         case class Cont(op: Op[?, S], cont: Any => (Cont | A) < (Batch[S] & S & S2))
         def runCont(v: (Cont | A) < (Batch[S] & S & S2)): (Cont | A) < (S & S2) =
-            // TODO workaround, Flat macro isn't inferring correctly with nested classes
-            import Flat.unsafe.bypass
```
fwbrasil, line 114: I thought it was something related to nested classes/methods but it was due to the `Macro` trying to infer for `Cont | A` but then not summoning the existing `Flat[A]` in scope.

kyo-prelude/shared/src/main/scala/kyo/kernel/Flat.scala
```diff
@@ -79,10 +79,10 @@ private object FlatMacro:
                     base =:= TypeRepr.of[Duration]
                 case _ => false
 
-        def hasTag(t: TypeRepr): Boolean =
+        def canDerive(t: TypeRepr): Boolean =
             t.asType match
                 case '[t] =>
-                    Expr.summon[Tag.Full[t]].isDefined
+                    Expr.summon[Tag.Full[t]].isDefined || Expr.summon[Flat[t]].isDefined
```
hearnadam, line 85: Doesn't it make sense to choose the evidence first?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/kernel/Flat.scala
fwbrasil: good point! It should be more lightweight than invoking the tag macro first

PR #718, opened 2024-10-06
bench: Batch

fwbrasil:
Initial step to optimize the `Batch` effect. It's a port of a [benchmark](https://github.com/zio/zio-query/blob/0ea6d62fc482f0f03ee5419af2899f6d1531836d/benchmarks/src/main/scala/zio/query/DataSourceBenchmark.scala#L1) in `zio-query`. [Current results](https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/4192850428178d3dc0f4842a0b1bdf5b/raw/4921d952bb86be51d55537973d09ec0d6f14be4a/jmh-result.json):

image

I''m not sure why cats-effect/Fetch are so behind. Looking at the profiling session it seems due to a large number of allocations building hash sets:

flame_graph


fwbrasil:
I'm working on optimizing `Batch` and I've reached a point where the main bottleneck are allocations outside of the TLAB. The JVM has a fast path that uses a thread-local buffer (TLAB) for allocations without having to touch the heap directly, which has much better performance. The fallback to the slow path happens in a few scenarios, which includes the allocation of large arrays. Example flamegraph of allocations outside the TLAB:

flame_graph

The issue is that the benchmark uses a large batch with 1k items. In my experience using similar batching solutions, the typical batch size is in the order of hundreds of items per batch. I've updated the benchmark to use `100` instead of `1000`, which also made the cats-effect/fetch results more reasonable to reach proper JIT compilation. [Results](https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/583833c3f85bc986bcbe4ef0a2b8d2d5/raw/3f7da2c86f9e344618c72a04a975e4dadaa22e3b/jmh-result.json) after the change but only including a couple of the optimizations I'm working on:

image

hearnadam (approved):

PR #719, opened 2024-10-06
data: Chunk optimizations

fwbrasil:
This is part of the work to optimize the `Batch` effect. I'm changing it to use `Chunk` but there are operations currently using the default implementations from `Seq`, which introduce some performance overhead.


kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -30,7 +30,9 @@ sealed abstract class Chunk[A] extends Seq[A] derives CanEqual:
       * @return
       *   true if the Chunk contains no elements, false otherwise
       */
-    final override def isEmpty: Boolean = size == 0
+    final override def isEmpty: Boolean = length == 0
+
+    override def length: Int
```
fwbrasil, line 35: I noticed interface dispatches (itable stub) to obtain the size of a chunk. This PR introduces this class method so all internal uses can dispatch directly to it and avoid the indirection in `Seq`.

kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -397,13 +400,24 @@ object Chunk:
...
                 this match
-                    case Tail(chunk, offset, size) =>
-                        Tail(chunk, offset + 1, size - 1)
+                    case Tail(chunk, offset, length) =>
+                        Tail(chunk, offset + 1, length - 1)
                     case c =>
-                        Tail(c, 1, size - 1)
+                        Tail(c, 1, length - 1)
+
+        override def iterator: Iterator[A] =
```
fwbrasil, line 411: Specialized iterator for better performance

kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -180,7 +182,7 @@ sealed abstract class Chunk[A] extends Seq[A] derives CanEqual:
       * @return
       *   an Iterator[A] over the elements of the Chunk
       */
-    def iterator: Iterator[A] = toArray.iterator
+    def iterator: Iterator[A] = toIndexed.iterator
```
fwbrasil, line 185: This method was quite expensive, always creating a new array just to provide the iterator

kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -30,7 +30,9 @@ sealed abstract class Chunk[A] extends Seq[A] derives CanEqual:
       * @return
       *   true if the Chunk contains no elements, false otherwise
       */
-    final override def isEmpty: Boolean = size == 0
+    final override def isEmpty: Boolean = length == 0
```
hearnadam, line 33: Do you want this to be private/protected?

hearnadam (approved):

kyo-data/shared/src/main/scala/kyo/Chunk.scala
fwbrasil: It's not possible because it overrides the super class method

PR #720, opened 2024-10-06
prelude: propagate safepoints in Loop

fwbrasil:
Kyo's `Safepoint` mechanism is responsible for providing stack safety, tracing, and interceptors for features like preemption and the `Debug` effect. The `Safepoint` is held in a `ThreadLocal` but, given that every transformation in Kyo has to use it, always reading from the `ThreadLocal` can introduce significant performance overhead. To mitigate it, Kyo's kernel also propagates the safepoint via implicits but the `Loop` implementation is missing it.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/kernel/Loop.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/kernel/Loop.scala
@@ -73,9 +73,12 @@ object Loop:
             def _3 = v3
             def _4 = v4
 
-    inline def apply[A, O, S](inline input: A)(inline run: A => Outcome[A, O] < S)(using inline _frame: Frame): O < S =
+    inline def apply[A, O, S](inline input: A)(inline run: Safepoint ?=> A => Outcome[A, O] < S)(using
+        inline _frame: Frame,
+        safepoint: Safepoint
+    ): O < S =
         @nowarn("msg=anonymous")
-        @tailrec def loop(i1: A)(v: Outcome[A, O] < S = run(i1)): O < S =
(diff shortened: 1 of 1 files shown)


kyo-prelude/shared/src/main/scala/kyo/kernel/Loop.scala
```diff
@@ -141,9 +148,12 @@ object Loop:
         loop(input1, input2, input3, input4)()
     end apply
 
-    inline def indexed[O, S](inline run: Int => Outcome[Unit, O] < S)(using inline _frame: Frame): O < S =
+    inline def indexed[O, S](inline run: Int => Outcome[Unit, O] < S)(using
```
fwbrasil, line 151: Not sure why the compiler has inference issues if I add the `Safepoint` to `run` specifically in this method.

hearnadam: Is it still an issue if the method is not inline?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/kernel/Loop.scala
fwbrasil: yep, same thing:

```
[error] -- [E081] Type Error: /Users/fwbrasil/workspace/kyo/kyo-prelude/shared/src/main/scala/kyo/Kyo.scala:104:39
[error] 104 |                        Loop.indexed { idx =>
[error]     |                                       ^^^
[error]     |                       Missing parameter type
[error]     |
[error]     |                       I could not infer the type of the parameter idx
[error]     |                       Expected type for the whole anonymous function:
[error]     |                         Any
```

PR #722, opened 2024-10-06
prelude: introduce Kyo.foreachIndexed + propagate safepoints

fwbrasil:
The new method is needed to optimize `Batch` and I've changed the collection methods to propagate safepoints like in https://github.com/getkyo/kyo/pull/720

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Kyo.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Kyo.scala
@@ -2,6 +2,7 @@ package kyo
 
 import kernel.Frame
 import kernel.Loop
+import kyo.kernel.Safepoint
 import scala.annotation.tailrec
 
 /** Object containing utility functions for working with Kyo effects. */
@@ -58,7 +59,7 @@ object Kyo:
       * @return
       *   A new effect that produces a Chunk of results
       */
-    def foreach[A, B, S, S2](seq: Seq[A])(f: A => B < S2)(using Frame): Chunk[B] < (S & S2) =
+    def foreach[A, B, S, S2](seq: Seq[A])(f: Safepoint ?=> A => B < S2)(using Frame, Safepoint): Chunk[B] < (S & S2) =
         seq.knownSize match
             case 0 => Chunk.empty
             case 1 => f(seq(0)).map(Chunk(_))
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -78,6 +79,39 @@ object Kyo:
...
+    /** Applies an effect-producing function to each element of a sequence along with its index.
+      *
+      * @param seq
+      *   The input sequence
+      * @param f
+      *   The effect-producing function to apply to each element and its index
+      * @return
+      *   A new effect that produces a Chunk of results
+      */
+    def foreachIndexed[A, B, S, S2](seq: Seq[A])(f: Safepoint ?=> (Int, A) => B < S2)(using Frame, Safepoint): Chunk[B] < (S & S2) =
```
hearnadam, line 93: Does this introduce boxing to the Int? Do we need a specialized trait?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/Kyo.scala
fwbrasil: Yeah, it boxes but the JIT should be able to avoid allocations in many cases. I think it's best to wait for the compiler to implement specialization for this kind of scenario unless we see it as a bottleneck.

```scala
class Testing:
    val seq  = Seq(1, 2)
    def test = Kyo.foreachIndexed(seq)((idx, v) => idx + v)
```

Decompiled:

```java
public class Testing {
    private final Seq<Object> seq = (Seq)((SeqOps)new .colon.colon((Object)BoxesRunTime.boxToInteger((int)1), (List)new .colon.colon((Object)BoxesRunTime.boxToInteger((int)2), (List)Nil$.MODULE$)));

    public Seq<Object> seq() {
        return this.seq;
    }

    public Object test() {
        Frame$package$ frame$package$ = Frame$package$.MODULE$;
        Frame.package.Frame$ Frame$_this = Frame.package.Frame$.MODULE$;
        Frame$package$ Frame$package$_this = frame$package$;
        return Kyo$.MODULE$.foreachIndexed(this.seq(), (Function1 & Serializable)contextual$1 -> (Function2 & Serializable)(idx, v) -> Testing.test$$anonfun$1$$anonfun$1(BoxesRunTime.unboxToInt((Object)idx), BoxesRunTime.unboxToInt((Object)v)), "kyo.Testing\u00a3test\u00a3KyoTest.scala\u00a38\u00a360\u00a3val seq  = Seq(1, 2)\ndef test = Kyo.foreachIndexed(seq)((idx, v) => idx + v)\ud83d\udccd", Safepoint$.MODULE$.get());
    }

    private static final /* synthetic */ Object test$$anonfun$1$$anonfun$1(int idx, int v) {
        Pending$package$ pending$package$;
        Pending$package$ Pending$package$_this = pending$package$ = Pending$package$.MODULE$;
        int v$proxy5 = idx + v;
        return BoxesRunTime.boxToInteger((int)v$proxy5);
    }
}
```

PR #723, opened 2024-10-06
Kyo grpc

### LINKED ISSUE #390: gRPC Support

steinybot:
Fixes #390

Still a bit to do here but just to show the current progress as it has been a long time.


fwbrasil:
@steinybot can we close this for now? I wanted to avoid PRs open for too long. Let me know if there's anything I can help with!

PR #724, opened 2024-10-06
prelude: Batch optimizations

fwbrasil:
[Benchmark results](https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/5eae45217e6646aa77c193d081093920/raw/07acdb06bb3cf53faa466581f32cbd971afb47c0/jmh-result.json) with this change on top of https://github.com/getkyo/kyo/pull/718 and the other recent optimizations.

image

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Batch.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Batch.scala
@@ -111,10 +111,12 @@ object Batch:
         type SourceAny = Source[Any, Any, S]
         type ContAny   = Any => (ToExpand | Expanded | A) < (Batch[S] & S & S2)
 
-        case class ToExpand(op: Seq[Any], cont: ContAny)
-        case class Expanded(value: Any, source: SourceAny, cont: ContAny)
+        type Item = A | ToExpand | Expanded
+        abstract class Pending
+        case class ToExpand(op: Seq[Any], cont: ContAny)                  extends Pending
+        case class Expanded(value: Any, source: SourceAny, cont: ContAny) extends Pending
 
-        def runCont(v: (ToExpand | Expanded | A) < (Batch[S] & S & S2)): (ToExpand | Expanded | A) < (S & S2) =
+        def capture(v: Item < (Batch[S] & S & S2)): Item < (S & S2) =
             ArrowEffect.handle(erasedTag[S], v) {
                 [C] =>
                     (input, cont) =>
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -83,7 +83,7 @@ object Batch:
       *   A batched operation that produces a single value from the sequence
       */
     inline def eval[A](seq: Seq[A])(using inline frame: Frame): A < Batch[Any] =
-        ArrowEffect.suspend[A](erasedTag[Any], Op.Eval(seq))
+        ArrowEffect.suspend[A](erasedTag[Any], seq)
```
fwbrasil, line 86: Suspend with the `Seq` directly to avoid the `Eval` wrapper allocation. See change to make `Op` a type union at the end of the file.

kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -108,62 +108,63 @@ object Batch:
...
             ArrowEffect.handle(erasedTag[S], v) {
-                [C] => (input, cont) => Cont(input, v => cont(v.asInstanceOf[C]))
+                [C] =>
+                    (input, cont) =>
+                        val contAny = cont.asInstanceOf[ContAny]
+                        input match
+                            case Call(v, source) => Expanded(v, source.asInstanceOf[SourceAny], contAny)
+                            case v: Seq[C]       => ToExpand(v, contAny)
             }
         end runCont
```
fwbrasil, line 126: This method now returns `Expanded` directly to avoid the translation from `Cont` to `Expanded` in the previous impl

kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -108,62 +108,63 @@ object Batch:
...
-                        val uniqueValues             = values.distinct
-
-                        source(uniqueValues).map { map =>
-                            Kyo.foreach(expandedItems.zip(indices)) { case (e, idx) =>
-                                runCont(map(e.v).map(e.cont)).map((_, idx))
-                            }
+                    val pending: Map[SourceAny | Unit, Seq[Expanded | A]] =
+                        expanded.groupBy {
+                            case Expanded(_, source, _) => source
+                            case _                      => ()
```
fwbrasil, line 144: I'm creating an artificial source group with unit as the key. This way we can avoid the partition of the collection.

kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -108,62 +108,63 @@ object Batch:
...
+                        case (_: Unit, items) => Chunk.from(items)
+                        case (source: SourceAny, items: Seq[Expanded] @unchecked) =>
+                            source(items.map(_.value).distinct).map { map =>
+                                Kyo.foreach(items) { e =>
+                                    runCont(map(e.value).map(e.cont))
+                                }
+                            }
                     }.map { results =>
-                        loop(completed.map(_._1) ++ results.flatten.map(_._1))
+                        loop(results.flattenChunk)
```
fwbrasil, line 155: The impl had some logic to ensure ordering is preserved but this line was not considering it. I've decided to remove the ordering mechanism because the execution itself already produces results in the expected order.

hearnadam (approved):

PR #732, opened 2024-10-08
data: Instant

### LINKED ISSUE #641: Introduce optimized `Instant`
The `Clock` effect currently uses Java's `Instance` and its APIs don't integrate well with `Duration`. Context: https://github.com/getkyo/kyo/pull/637#discussion_r1747279331

fwbrasil:
Fixes #641 

My initial plan was to use `opaque type Instant = Long` but I thought the loss in precision in comparison to Java's `Instant` can be confusing.


kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -49,9 +48,8 @@ object Clock:
     object Stopwatch:
         /* WARNING: Low-level API meant for integrations, libraries, and performance-sensitive code. See AllowUnsafe for more details. */
         class Unsafe(start: Instant, clock: Clock.Unsafe):
-            def elapsed()(using AllowUnsafe): Duration =
-                Duration.fromJava(java.time.Duration.between(start, clock.now()))
-            def safe: Stopwatch = Stopwatch(this)
+            def elapsed()(using AllowUnsafe): Duration = clock.now() - start
```
fwbrasil, line 51: `Instant` works well with `Duration` now

kyo-data/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -57,7 +58,8 @@ object Duration:
       * @return
       *   A Duration instance
       */
-    inline def fromNanos(value: Long): Duration = value
+    inline def fromNanos(value: Long): Duration =
+        if value <= 0 then Duration.Zero else value
```
fwbrasil, line 62: There was a bug here allowing negative durations

hearnadam (approved):

PR #733, opened 2024-10-08
data: Schedule

fwbrasil:
Inspired by ZIO's `Schedule` but as pure data structure.

### DIFF
--- a/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
+++ b/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
@@ -72,10 +72,11 @@ extension [A, S](effect: A < S)
       * @return
       *   A computation that produces the result of the last execution
       */
-    def repeat(policy: Retry.Policy)(using Flat[A], Frame): A < (S & Async) =
-        Loop.indexed { i =>
-            if i >= policy.limit then effect.map(Loop.done)
-            else effect.delayed(policy.backoff(i)).as(Loop.continue)
+    def repeat(schedule: Schedule)(using Flat[A], Frame): A < (S & Async) =
+        Loop(schedule) { schedule =>
+            val (delay, nextSchedule) = schedule.next
+            if !delay.isFinite then effect.map(Loop.done)
+            else effect.delayed(delay).as(Loop.continue(nextSchedule))
         }
 
     /** Performs this computation repeatedly with a limit.
@@ -186,8 +187,8 @@ extension [A, S](effect: A < S)
       * @return
       *   A computation that produces the result of this computation with Async and Abort[Throwable] effects
       */
-    def retry(policy: Retry.Policy)(using Flat[A], Frame): A < (S & Async & Abort[Throwable]) =
-        Retry[Throwable](policy)(effect)
+    def retry(schedule: Schedule)(using Flat[A], Frame): A < (S & Async & Abort[Throwable]) =
+        Retry[Throwable](schedule)(effect)
 
     /** Performs this computation repeatedly with a limit in case of failures.
       *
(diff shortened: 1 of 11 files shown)


kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -197,19 +198,7 @@ extension [A, S](effect: A < S)
...
-    /** Performs this computation repeatedly with a backoff policy and a limit in case of failures.
-      *
-      * @param backoff
-      *   The backoff policy to use
-      * @param limit
-      *   The limit to use
-      * @return
-      *   A computation that produces the result of this computation with Async and Abort[Throwable] effects
-      */
-    def retry(backoff: Int => Duration, n: Int)(using Flat[A], Frame): A < (S & Async & Abort[Throwable]) =
```
fwbrasil, line 211: I've removed the "raw" method to reduce API surface

kyo-core/shared/src/main/scala/kyo/Retry.scala
```diff
@@ -6,74 +6,12 @@ import scala.util.*
...
-          */
-        def limit(v: Int): Policy =
-            copy(limit = v)
-    end Policy
-
-    object Policy:
-        /** The default retry policy with no backoff and 3 attempts. */
-        val default = Policy(_ => Duration.Zero, 3)
+    /** The default retry schedule. */
+    val defaultSchedule = Schedule.exponentialBackoff(initial = 100.millis, factor = 2, maxDelay = 5.seconds)
```
fwbrasil, line 10: More sensible defaults

kyo-data/shared/src/main/scala/kyo/Schedule.scala
```diff
@@ -0,0 +1,348 @@
...
+import Schedule.internal.*
+import kyo.Duration
+import kyo.Instant
+
+/** An immutable, composable scheduling policy.
+  *
+  * Schedule provides various combinators for creating complex scheduling policies. It can be used to define retry policies, periodic tasks,
+  * or any other time-based scheduling logic.
+  */
+sealed abstract class Schedule derives CanEqual:
```
hearnadam, line 12: I really like that this is immutable and very generic - not specific to Kyo.

kyo-data/shared/src/main/scala/kyo/Schedule.scala
```diff
@@ -0,0 +1,348 @@
...
+
+    /** Returns the next delay and the updated schedule.
+      *
+      * @return
+      *   a tuple containing the next delay duration and the updated schedule
+      */
+    def next: Maybe[(Duration, Schedule)]
```
hearnadam, line 19: I am not in love with this signature. In general, I am not a fan of Tuples...

kyo-data/shared/src/main/scala/kyo/Schedule.scala
hearnadam, line 12: For consistency, how would you feel about `self =>`?

kyo-data/shared/src/main/scala/kyo/Instant.scala
```diff
@@ -138,6 +138,24 @@ object Instant:
...
             instant.truncatedTo(unit.chronoUnit)
 
+        /** Returns the minimum of this Instant and another.
+          *
+          * @param other
+          *   The other Instant to compare with.
+          * @return
+          *   The earlier of the two Instants.
+          */
+        def min(other: Instant): Instant = if instant.isBefore(other) then instant else other
```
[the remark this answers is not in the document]
fwbrasil: I've added `infix` to the correspondent methods in `Instant` as well

kyo-data/shared/src/main/scala/kyo/Schedule.scala
```diff
@@ -0,0 +1,348 @@
...
+
+        case object Done extends Schedule:
+            def next = Maybe.empty
+            def show = "Schedule.done"
+
+        final case class Fixed(interval: Duration) extends Schedule:
+            val next = Maybe((interval, this))
+            def show = s"Schedule.fixed(${interval.show})"
+
+        final case class Exponential(initial: Duration, factor: Double) extends Schedule:
```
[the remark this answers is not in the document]
fwbrasil: I think it's ok in terms of binary compatibility since these are internal APIs. The goal of the different classes is for `Schedule` to automatically reduce to one of these canonical representations that can be compared for equality given that they're case classes. These tests show the functionality: https://github.com/getkyo/kyo/pull/733/files#diff-66ce2c17e23c73ff33f5e658a7ba3ef1972eb58d007873d50d795dae21a8e52dR578. I'm planning to use this behavior to optimize the execution of a new `Timer` implementation based on `Schedule`.

kyo-data/shared/src/main/scala/kyo/Schedule.scala
fwbrasil: Yeah, I had a `withNext` instead in one of the impls I pushed because of that. I decided to go back to the tuple because this is an API I'd expect users to rarely use directly and the tuple simplifies the implementation.

kyo-data/shared/src/main/scala/kyo/Schedule.scala
fwbrasil: I think in this case it wouldn't matter? There aren't nested classes where `this` would be ambiguous

hearnadam (approved):
LGTM overall. We can revisit the tuple result later if necessary.

PR #734, opened 2024-10-09
data: Rename Maybe.Defined/Empty to Present/Absent and export them to the kyo package

fwbrasil:
I sometimes avoid pattern matching on `Maybe` because it feels too verbose. This PR exports `Maybe.Empty` and `Maybe.Defined` as members in the `kyo` package. It also reduces the syntax overhead for tracking `Abort[Empty]`. It seems reasonable since the type names shouldn't conflict with common types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,11 +102,12 @@ In Kyo, computations are expressed via the infix type `<`, known as "Pending". I
 ```scala 
 import kyo.*
 
-// 'Int' pending 'Abort[Maybe.Empty]'
-Int < Abort[Maybe.Empty]
+// 'Int' pending 'Abort[Absent]'
+// 'Absent' is Kyo's equivalent of 'None' via the 'Maybe' type
+Int < Abort[Absent]
 
-// 'String' pending 'Abort[Maybe.Empty]' and 'IO'
-String < (Abort[Maybe.Empty] & IO)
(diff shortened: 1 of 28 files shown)


hearnadam:
As discussed offline, I am concerned with collisions with `Empty`. We discussed a few potential options,
- `Present`/`Absent` - nice parity
- `Just`/`Void` - Haskell like

Or perhaps we could just export the `Defined`... leaving underscore pattern matches where necessary...

johnhungerford:
+1 `Present`/`Absent`, although frankly I don't think ever seen anyone actually use scala streams or therefore the std `Empty` object

fwbrasil:
I like Present/Absent as well since it'd hardly conflict with other types and feels more natural to communicate meaning to newcomers. Just/Void has the Haskell precedence that helps with understanding but I find it a bit opaque.

fwbrasil:
I've updated the PR with the rename to Present/Absent to see how it feels. It seems ok, the only thing I wasn't sure about is if we should also rename methods like `Maybe.empty` to `Maybe.absent`. I think not since people will instinctively search for methods like `empty`, `isDefined`, etc and they still seem to make sense even with the new naming.

hearnadam (approved):
Perhaps worth renaming PR title for release notes - LGTM!

PR #736, opened 2024-10-09
core: make masked IOPromise not interruptible

fwbrasil:
As pointed out by @hearnadam in https://github.com/getkyo/kyo/pull/707, the fiber returned by `mask` doesn't propagate interrupts to its previous steps but the instance itself can still be interrupted. It's a behavior that is different from other effect systems and makes the masking feature less useful. This PR fixes that by using a custom `IOPromise` class with a method override.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -23,7 +23,8 @@ object Fiber extends FiberPlatformSpecific:
 
     inline given [E, A]: Flat[Fiber[E, A]] = Flat.unsafe.bypass
 
-    private val _unit = success(())
+    private val _unit  = success(()).mask
+    private val _never = IOPromise[Nothing, Unit]().mask
 
     private[kyo] inline def fromTask[E, A](inline ioTask: IOTask[?, E, A]): Fiber[E, A] = ioTask
 
@@ -39,7 +40,7 @@ object Fiber extends FiberPlatformSpecific:
       * @return
       *   A Fiber that never completes
       */
-    def never: Fiber[Nothing, Unit] = IOPromise[Nothing, Unit]()
+    def never[E]: Fiber[E, Unit] = _never.asInstanceOf[Fiber[E, Unit]]
 
     /** Creates a successful Fiber.
       *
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -23,7 +23,8 @@ object Fiber extends FiberPlatformSpecific:
 
     inline given [E, A]: Flat[Fiber[E, A]] = Flat.unsafe.bypass
 
-    private val _unit = success(())
+    private val _unit  = success(()).mask
+    private val _never = IOPromise[Nothing, Unit]().mask
```
fwbrasil, line 27: These constant weren't safe before since they could still be interrupted. With the new `mask` behavior, they're effectively final unless the user uses type casts.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -63,12 +63,13 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
     end interrupts
 
     final def mask: IOPromise[E, A] =
-        val p = IOPromise[E, A]()
+        val p = new IOPromise[E, A]:
+            override def interrupt(error: Panic): Boolean = false
```
fwbrasil, line 67: I had avoided implementing this behavior before because I thought we'd need to add a new type to `IOPromise.State` but, given that `interrupt` is the only method that exposes interruption, we can simply override it here.

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
```diff
@@ -777,8 +777,8 @@ class AsyncTest extends Test:
                         for
                             _ <- start.release
                             _ <- run.await
-                            _ <- stop.release
                             _ <- result.set(42)
+                            _ <- stop.release
```
fwbrasil, line 781: This test had a race condition and was failing from time to time. It wasn't guaranteed that `result` is set before the other thread is released.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
hearnadam: Ah yes, this is exactly what I was trying to communicate when I was wanting to convert these to synthetic classes. This is better though.

hearnadam (approved):

PR #738, opened 2024-10-09
prelude/core: rename Ensures to Finalizers

fwbrasil:
The term "ensure" came from the `IO.ensure` method and seems too opaque regarding its purpose.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/scheduler/Finalizers.scala
+++ b/kyo-core/shared/src/main/scala/kyo/scheduler/Finalizers.scala
@@ -6,20 +6,20 @@ import kyo.discard
 import org.jctools.queues.MpmcArrayQueue
 import scala.annotation.tailrec
 
-private[kyo] opaque type Ensures = Ensures.Empty.type | (() => Unit) | ArrayDeque[() => Unit]
+private[kyo] opaque type Finalizers = Finalizers.Empty.type | (() => Unit) | ArrayDeque[() => Unit]
 
-private[kyo] object Ensures:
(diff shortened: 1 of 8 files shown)


kyo-core/shared/src/main/scala/kyo/scheduler/Finalizers.scala
```diff
@@ -31,7 +31,7 @@ private[kyo] object Ensures:
                     arr.add(f)
                     arr
 
-        def remove(f: () => Unit): Ensures =
+        def remove(f: () => Unit): Finalizers =
```
hearnadam, line 34: Not sure how often this API will be used directly, but worth a comment that the function reference must be the same (not just the value) IIUC.

kyo-core/shared/src/main/scala/kyo/scheduler/Finalizers.scala
```diff
@@ -6,20 +6,20 @@ import kyo.discard
...
+private[kyo] opaque type Finalizers = Finalizers.Empty.type | (() => Unit) | ArrayDeque[() => Unit]
 
-private[kyo] object Ensures:
+private[kyo] object Finalizers:
     case object Empty derives CanEqual
 
-    val empty: Ensures = Empty
+    val empty: Finalizers = Empty
 
     private val bufferCache = new MpmcArrayQueue[ArrayDeque[() => Unit]](1000)
```
hearnadam, line 16: 1024?

hearnadam (approved):

Issue #739, opened 2024-10-09
Abort with literal strings


fwbrasil:
Users currently need to create new types in case they need to distinguish different failure scenarios. Allowing literal strings seems a good way to make the effect more convenient:

```scala
def test(v: Int): Int < Abort["negative" | "positive"] =
    if v < 0 then Abort.failLiteral("negative")
    else if v > 0 then Abort.failLiteral("positive")
    else 42

val a: Result["negative" | "positive", Int] < Any = 
    Abort.run(test(1))

val b: Result["negative", Int] < Abort["positive"] = 
    Abort.run["negative"](test(1))
```

I think we'd need a separate method like `failLiteral` to avoid inference issues and fail compilation if the string isn't statically known. We could allow other literal types like integers as well but I'm not sure it'd make much sense.

The change in `Abort` seems easy but we'd need to make `SafeClassTag` support literal types, which is a more involved change.

PR #740, opened 2024-10-09
data: support literal types in SafeClassTag

fwbrasil:
first step for https://github.com/getkyo/kyo/issues/739

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
+++ b/kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
@@ -26,7 +26,7 @@ object SafeClassTag:
 
     case class Union(elements: List[SafeClassTag[Any]])        extends Element
     case class Intersection(elements: List[SafeClassTag[Any]]) extends Element
-    case class LiteralTag[A](value: A)                         extends Element
+    case class LiteralTag(value: Any)                          extends Element
 
     sealed trait Primitive extends Element
     case object IntTag     extends Primitive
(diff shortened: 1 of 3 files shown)


kyo-data/shared/src/main/scala/kyo/internal/SafeClassTagMacro.scala
```diff
@@ -55,24 +55,38 @@ private[kyo] object SafeClassTagMacro:
...
+                case ConstantType(const) =>
+                    val value =
+                        const.value match
+                            case x: Int     => Expr(x)
+                            case x: Long    => Expr(x)
+                            case x: Float   => Expr(x)
+                            case x: Double  => Expr(x)
+                            case x: Boolean => Expr(x)
+                            case x: Char    => Expr(x)
+                            case x: String  => Expr(x)
```
hearnadam, line 68: Do we need individual matches here? or can it be:

`_: Int | Long |... => Expr(const.value)`

kyo-data/shared/src/main/scala/kyo/SafeClassTag.scala
```diff
@@ -26,6 +26,7 @@ object SafeClassTag:
 
     case class Union(elements: List[SafeClassTag[Any]])        extends Element
     case class Intersection(elements: List[SafeClassTag[Any]]) extends Element
+    case class LiteralTag[A](value: A)                         extends Element
```
hearnadam, line 29: Will this always be type erased? Should it just be `Any`?

hearnadam (approved):

kyo-data/shared/src/main/scala/kyo/internal/SafeClassTagMacro.scala
fwbrasil: It's not possible because `Expr` requires a `ToExpr` evidence of the specific type. I suspect there might be a better way to do this but I don't have much experience with Scala 3 macros.

PR #743, opened 2024-10-11
core: Fiber improvements and fixes

fwbrasil:


kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -17,14 +17,14 @@ import scala.util.NotGiven
 import scala.util.control.NonFatal
 import scala.util.control.NoStackTrace
 
-opaque type Fiber[E, A] = IOPromise[E, A]
+opaque type Fiber[+E, +A] = IOPromise[E, A]
```
fwbrasil, line 20: I hadn't added variance in earlier versions because of compiler limitations but never revisited it.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -17,14 +17,14 @@ import scala.util.NotGiven
...
 
 object Fiber extends FiberPlatformSpecific:
 
     inline given [E, A]: Flat[Fiber[E, A]] = Flat.unsafe.bypass
 
-    private val _unit  = success(()).mask
-    private val _never = IOPromise[Nothing, Unit]().mask
+    private val _unit  = IOPromise(Result.unit).mask()
+    private val _never = IOPromise[Nothing, Unit]().mask()
```
fwbrasil, line 27: `IOPromise.mask` was missing the `()` since it performs side effects.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -49,7 +49,7 @@ object Fiber extends FiberPlatformSpecific:
       * @return
       *   A Fiber that completes successfully with the given value
       */
-    def success[E, A](v: A): Fiber[E, A] = result(Result.success(v))
+    def success[E, A](v: A)(using Frame): Fiber[E, A] < IO = result(Result.success(v))
```
fwbrasil, line 52: These methods were missing the `IO` suspension since they initialize objects with mutable state.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -83,23 +83,10 @@ object Fiber extends FiberPlatformSpecific:
...
-        IO {
-            val p = new IOPromise[Throwable, A] with (Try[A] => Unit):
-                def apply(result: Try[A]) =
-                    result match
-                        case Success(v) =>
-                            completeDiscard(Result.success(v))
-                        case Failure(ex) =>
-                            completeDiscard(Result.fail(ex))
+    def fromFuture[A](future: Future[A])(using frame: Frame): Fiber[Throwable, A] < IO =
+        IO.Unsafe(Unsafe.fromFuture(future))
```
fwbrasil, line 87: Moved the impl of a few methods to `Unsafe`.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -248,7 +216,7 @@ object Fiber extends FiberPlatformSpecific:
           * @return
           *   A new Fiber that runs with interrupt masking
           */
-        def mask(using Frame): Fiber[E, A] < IO = IO(self.mask)
+        def mask(using Frame): Fiber[E, A] < IO = IO.Unsafe(Unsafe.mask(self)())
```
fwbrasil, line 219: Explicitly call the `mask` extension method via its companion object to avoid the compiler inferring the incorrect version.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -198,13 +179,9 @@ object Fiber extends FiberPlatformSpecific:
...
           *   A new Fiber with the mapped result
           */
-        def map[B](f: A => B)(using Frame): Fiber[E, B] < IO =
-            IO {
-                val p = new IOPromise[E, B](interrupts = self) with (Result[E, A] => Unit):
-                    def apply(v: Result[E, A]) = completeDiscard(v.map(f))
-                self.onComplete(p)
-                p
-            }
+        def map[B: Flat](f: A => B < IO)(using Frame): Fiber[E, B] < IO =
```
fwbrasil, line 182: The function can now return a computation pending `IO` like in `onComplete`.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -277,29 +277,29 @@ private[kyo] object IOPromise extends IOPromisePlatformSpecific:
...
                 p match
                     case _ if (p eq Pending.Empty) => tail
                     case p: Pending[E, A]          => interruptLoop(p.interrupt(panic), panic)
 
-            new Pending[E, A]:
-                def waiters: Int            = self.waiters + 1
-                def interrupt(panic: Panic) = interruptLoop(self, panic)
-                def run(v: Result[E, A])    = runLoop(self, v)
+            new Pending[E2, A2]:
+                def waiters: Int                               = self.waiters + tail.waiters
```
fwbrasil, line 296: The method was incorrectly not considering the number of waiters in the tail. The waiter count is only used by `toString`, though.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
fwbrasil: Actually, I've just realized that the returned values are effectively final. `IOPromise` can't change its state after it's complete, even in case of interruption. I'll remove the suspension.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -49,7 +49,7 @@ object Fiber extends FiberPlatformSpecific:
       * @return
       *   A Fiber that completes successfully with the given value
       */
-    def success[E, A](v: A): Fiber[E, A] = result(Result.success(v))
+    def success[E, A](v: A)(using Frame): Fiber[E, A] = result(Result.success(v))
```
hearnadam, line 52: Do you know why the macro was not failing expansion before?

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -83,23 +83,10 @@ object Fiber extends FiberPlatformSpecific:
...
-        import scala.util.*
-        IO {
-            val p = new IOPromise[Throwable, A] with (Try[A] => Unit):
-                def apply(result: Try[A]) =
-                    result match
-                        case Success(v) =>
-                            completeDiscard(Result.success(v))
-                        case Failure(ex) =>
-                            completeDiscard(Result.fail(ex))
+    def fromFuture[A](future: Future[A])(using frame: Frame): Fiber[Throwable, A] < IO =
```
hearnadam, line 86: I might have mentioned this before, but I think this should be by name.

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/Fiber.scala
fwbrasil: ah, it's actually unnecessary. I had added it with the `IO` suspension but later realized that it's not necessary because the returned values are effectively final

PR #744, opened 2024-10-11
core: fix unsafe APIs scaladocs

fwbrasil:
The unsafe warnings are not showing up in scaladocs because they're formatted as a regular comment. I've also added the warning to a few APIs that were missing it.


fwbrasil:
merging since it's a trivial cleanup

PR #745, opened 2024-10-11
improved logo

fwbrasil:
I've removed the outer ring, processed it with an AI enhancer and then an AI vectorizer.


hearnadam:
It might be good to replace the Kyo text with a font to ensure the edges are crisp.

fwbrasil:
I've manually edited the file to replace the text and added a shadow around the logo

PR #753, opened 2024-10-13
prelude: introduce Abort.recover and Kyo.pure

fwbrasil:
I took a look at ZIO's error handling as part of https://github.com/getkyo/kyo/issues/651 and it's essentially based on user-defined runtime pattern matching without a mechanism like Kyo's to safely handle multiple failure types at once via tags. I'm not sure there's much we can incorporate from ZIO but something that is a major advantage is how easy it is to recover and transform errors in ZIO compared to Kyo. Users currently need to use `Abort.run`, which returns a `Result`, and then perform additional transformations.

This PR introduces `Abort.recover` to directly recover specific failures. It provides two versions:

1. `Abort.recover[FailureType](onFail)(computation)`: Only handles `Fail` suspensions and leaves a pending `Abort[Nothing]` in case of panics.
2. `Abort.recover[FailureType](onFail, onPanic)(computation)`: Handles both `Fail` and `Panic` suspensions, not leaving an `Abort[Nothing]` since panics are also handled.

In addition, I had a type inference issue with if/else as recently elaborated by @steinybot, which prompted the addition of `Kyo.pure`.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Abort.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Abort.scala
@@ -43,6 +43,13 @@ object Abort:
       */
     inline def panic[E](inline ex: Throwable)(using inline frame: Frame): Nothing < Abort[E] = error(Panic(ex))
 
+    /** Fails the computation with the given error value (failure or panic).
+      *
+      * @param e
+      *   The error value to fail with
+      * @return
+      *   A computation that immediately fails with the given error value
+      */
     inline def error[E](inline e: Error[E])(using inline frame: Frame): Nothing < Abort[E] =
         ArrowEffect.suspendMap[Any](erasedTag[E], e)(_ => ???)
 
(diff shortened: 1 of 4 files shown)


kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -200,6 +200,111 @@ object Abort:
...
+          * @return
+          *   A computation that either succeeds with the original value or the recovered value
+          */
+        def apply[A: Flat, B: Flat, S, ER](onFail: E => B < S)(v: => A < (Abort[E | ER] & S))(
+            using
+            frame: Frame,
+            ct: SafeClassTag[E],
+            reduce: Reducible[Abort[ER]]
+        ): (A | B) < (S & Abort[ER]) =
+            ArrowEffect.handle.catching[
```
fwbrasil, line 223: I couldn't avoid the code repetition because of inference issues that require `asInstanceOf`.

hearnadam (approved):

PR #754, opened 2024-10-14
data: improve duration.show

fwbrasil:


kyo-data/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -174,13 +174,24 @@ object Duration:
...
+        /** Converts the Duration to a human-readable string at the most coarse possible resolution without losing information.
           *
           * @return
           *   A string representation of the Duration
           */
-        inline def show: String = s"Duration($self ns)"
-
+        def show: String =
+            if self == Zero then "Duration.Zero"
+            else if self == Infinity then "Duration.Infinity"
```
sideeffffect, line 184: Is there also `-Infinity`?

fwbrasil: no, the current impl doesn't allow negative durations

hearnadam (approved):

PR #757, opened 2024-10-15
use: ZIO Fibers directly to reduce interop overhead

hearnadam:


kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -67,4 +73,28 @@ object ZIOs:
         }
     end run
 
+    private object internal:
+        extension [E, A](exit: zio.Exit[E, A])
```
fwbrasil, line 77: Could you introduce tests for these extensions? I wonder if we should make them public, I imagine they could be useful in user code as well

kyo-zio/shared/src/main/scala/zio/FiberInterop.scala
```diff
@@ -0,0 +1,5 @@
+package zio
```
fwbrasil, line 1: What do you think about creating an issue in ZIO to provide this API? I'm ok with the workaround but we could migrate to a proper API if the issue gets addressed

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -17,15 +19,19 @@ object ZIOs:
       * @return
       *   A Kyo effect that, when run, will execute the zio.ZIO
       */
-    def get[E, A](v: => ZIO[Any, E, A])(using Frame, zio.Trace): A < (Abort[E] & Async) =
+    def get[E, A](v: => ZIO[Any, E, A])(using f: Frame, t: zio.Trace): A < (Abort[E] & Async) =
```
fwbrasil, line 22: it seems naming the implicits isn't necessary

kyo-zio/shared/src/main/scala/zio/FiberInterop.scala
hearnadam: I will open a PR later this week.

hearnadam: https://github.com/zio/zio/pull/9246

PR #758, opened 2024-10-15
fix: Tag compactness + cross project tests

hearnadam:


build.sbt
```diff
@@ -378,6 +378,17 @@ lazy val `kyo-examples` =
             libraryDependencies += "com.softwaremill.sttp.tapir" %% "tapir-json-zio" % "1.11.7"
         )
 
+lazy val `kyo-tests` =
```
hearnadam, line 381: Open to better naming.

fwbrasil: Could we put the new tests in `kyo-core` instead?

Issue #716, opened 2024-10-05
Channel APIs should include Abort[Closed]


steinybot:
I think the `Channel` APIs should include `Abort[Closed]`. It seems pretty normal to repeatedly take from a channel until the writer closes it and so `Closed` should be an expected fail.

PR #765, opened 2024-10-18
core: Queue/Channel/Meter improvements and fixes

fwbrasil:
Fixes https://github.com/getkyo/kyo/issues/716

Apologies for the large PR. It started as a follow up of https://github.com/getkyo/kyo/pull/702 to introduce missing unsafe APIs but it wouldn't be trivial to split the change since the implementations depend on each other. I also identified two bugs, more details in the comments.

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/Bench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/Bench.scala
@@ -48,7 +48,7 @@ object Bench:
 
     abstract class Base[A](expectedResult: A) extends Bench[A](expectedResult):
         def zioBench(): zio.UIO[A]
-        def kyoBenchFiber(): kyo.<[A, kyo.Async] = kyoBench()
+        def kyoBenchFiber(): kyo.<[A, kyo.Async & kyo.Abort[Throwable]] = kyoBench()
         def kyoBench(): kyo.<[A, kyo.IO]
         def catsBench(): cats.effect.IO[A]
     end Base
--- a/kyo-bench/src/main/scala/kyo/bench/CollectParBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/CollectParBench.scala
@@ -10,7 +10,9 @@ class CollectParBench extends Bench.ForkOnly(Seq.fill(1000)(1)):
     override def kyoBenchFiber() =
         import kyo.*
 
-        Async.parallel(kyoTasks)
+        // TODO inference issue
+        val x = Async.parallel(kyoTasks)
+        x
     end kyoBenchFiber
 
     def catsBench() =
--- a/kyo-bench/src/main/scala/kyo/bench/HttpClientBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/HttpClientBench.scala
@@ -35,7 +35,7 @@ class HttpClientBench extends Bench.ForkOnly("pong"):
     override def kyoBenchFiber() =
         import kyo.*
 
-        Abort.run(Requests(_.get(kyoUrl))).map(_.getOrThrow)
+        Requests(_.get(kyoUrl))
     end kyoBenchFiber
 
     val zioUrl =
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -246,7 +246,7 @@ object Channel:
                         flush()
                     else if queueSize > 0 && !takesEmpty then
                         // Attempt to transfer a value from the queue to
-                        // a waiting take opeation.
+                        // a waiting take operation.
                         Maybe(takes.poll()).foreach { promise =>
                             queue.poll() match
                                 case Result.Success(Present(value)) =>
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -179,8 +179,6 @@ object Channel:
             access: Access = Access.MultiProducerMultiConsumer
         )(using initFrame: Frame, allow: AllowUnsafe): Unsafe[A] =
             new Unsafe[A]:
-                import AllowUnsafe.embrace.danger
-
                 val queue = Queue.Unsafe.init[A](_capacity, access)
                 val takes = new MpmcUnboundedXaddArrayQueue[Promise.Unsafe[Closed, A]](8)
                 val puts  = new MpmcUnboundedXaddArrayQueue[(A, Promise.Unsafe[Closed, Unit])](8)
@@ -233,8 +231,6 @@ object Channel:
                 def closed()(using AllowUnsafe) = queue.closed()
 
                 @tailrec private def flush(): Unit =
-                    import AllowUnsafe.embrace.danger
-
                     // This method ensures that all values are processed
                     // and handles interrupted fibers by discarding them.
                     val queueClosed = queue.closed()
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -1,5 +1,6 @@
 package kyo
 
+import org.jctools.queues.MessagePassingQueue.Consumer
 import org.jctools.queues.MpmcUnboundedXaddArrayQueue
 import scala.annotation.tailrec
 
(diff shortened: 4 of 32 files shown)


kyo-bench/src/main/scala/kyo/bench/Bench.scala
```diff
@@ -48,7 +48,7 @@ object Bench:
 
     abstract class Base[A](expectedResult: A) extends Bench[A](expectedResult):
         def zioBench(): zio.UIO[A]
-        def kyoBenchFiber(): kyo.<[A, kyo.Async] = kyoBench()
+        def kyoBenchFiber(): kyo.<[A, kyo.Async & kyo.Abort[Throwable]] = kyoBench()
```
fwbrasil, line 51: Allow `Abort` in benchmarks since channels and meters can now about with `Closed`.

kyo-bench/src/main/scala/kyo/bench/HttpClientBench.scala
```diff
@@ -35,7 +35,7 @@ class HttpClientBench extends Bench.ForkOnly("pong"):
     override def kyoBenchFiber() =
         import kyo.*
 
-        Abort.run(Requests(_.get(kyoUrl))).map(_.getOrThrow)
+        Requests(_.get(kyoUrl))
```
fwbrasil, line 38: No need to handle `Abort` directly anymore since it's handled in `Bench`

kyo-bench/src/main/scala/kyo/bench/CollectParBench.scala
```diff
@@ -10,7 +10,9 @@ class CollectParBench extends Bench.ForkOnly(Seq.fill(1000)(1)):
     override def kyoBenchFiber() =
         import kyo.*
 
-        Async.parallel(kyoTasks)
+        // TODO inference issue
+        val x = Async.parallel(kyoTasks)
+        x
```
fwbrasil, line 15: The compilation fails reporting that `Abort[Throwable]` isn't a context effect. It seems a bad interaction with effect widening. It's pre-existing, I'll follow up on this separately.

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -8,334 +9,295 @@ import scala.annotation.tailrec
...
-      *   The initial frame for the channel
       * @tparam A
       *   The type of elements in the channel
       * @return
       *   A new Channel instance
+      *
+      * @note
+      *   The actual capacity will be rounded up to the next power of two.
+      * @warning
+      *   The actual capacity may be larger than the specified capacity due to rounding.
```
fwbrasil, line 149: One of the two bugs I identified is due to JCTools rounding up the size of queues to the next power of two. I've decided to keep rounding for `Channel` and `Queue` and document the behavior. For `Meter`, the rounding is a major issue since, if the user requests 100 req/s, it'd become 128 req/s. I had to reimplement `Meter` with a different approach to fix it. More details below.

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -8,334 +9,295 @@ import scala.annotation.tailrec
...
+                    queue.close().map { backlog =>
+                        flush()
+                        backlog
+                    }
+
+                def empty()(using AllowUnsafe)  = queue.empty()
+                def full()(using AllowUnsafe)   = queue.full()
+                def closed()(using AllowUnsafe) = queue.closed()
+
+                @tailrec private def flush(): Unit =
```
fwbrasil, line 235: the logic here is the same but using `Result` and `Maybe` since we don't need to worry about them allocating.=

kyo-core/shared/src/main/scala/kyo/Meter.scala
```diff
@@ -1,25 +1,13 @@
...
-      *   The number of available permits as an Int effect.
-      */
-    def available(using Frame): Int < IO
-
-    /** Checks if there are any available permits.
-      *
-      * @return
-      *   A Boolean effect indicating whether permits are available.
-      */
-    def isAvailable(using Frame): Boolean < IO =
```
fwbrasil, line 20: I've removed `isAvailable` since it doesn't seem very useful, renamed `available` to `availablePermits`, and moved the method to the end of the class.

kyo-core/shared/src/main/scala/kyo/Meter.scala
```diff
@@ -225,16 +221,123 @@ object Meter:
...
+
+                def closed(using Frame): Boolean < IO =
+                    Kyo.foreach(meters)(_.closed).map(_.exists(identity))
             end new
         }
 
-    private def offer[A](n: Int, chan: Channel[A], v: A)(using Frame): Unit < IO =
-        Loop.indexed { idx =>
-            if idx == n then Loop.done
+    abstract private class Base(permits: Int)(using initFrame: Frame, allow: AllowUnsafe) extends Meter:
```
fwbrasil, line 230: New `Meter` implementation to avoid the issue with JCTools rounding up the capacity to the next power of two. Instead of managing the permits with a channel, the new approach uses an atomic and a queue for waiters. It's inspired in Twitter Util's `AsyncSemaphore` I [worked on](https://github.com/twitter/util/blob/17c8187d948846756fabcdbb2ff6838b3665968b/util-core/src/main/scala/com/twitter/concurrent/AsyncSemaphore.scala#L1).

kyo-core/shared/src/main/scala/kyo/Queue.scala
```diff
@@ -13,189 +12,332 @@ import scala.annotation.tailrec
...
+
+            protected inline def offerOp[A](inline f: => Boolean, inline raceRepair: => Boolean): Result[Closed, Boolean] =
+                _closed.get().getOrElse {
+                    val result = f
+                    if result && _closed.get().isDefined then
+                        Result(raceRepair)
+                    else
+                        Result(result)
+                    end if
+                }
```
fwbrasil, line 310: This is part of the fix for the other bug I identified. There's a race between the queue getting closed and a new item being added concurrently. The thread that is adding the new item might see the queue still not closed but it gets closed before the item is added to the underlying queue. To resolve the race, an extra check is performed after the operation and a repair is performed.

kyo-core/shared/src/main/scala/kyo/Queue.scala
```diff
@@ -210,118 +352,38 @@ object Queue:
...
+                    offerOp(
+                        q.offer(v),
+                        try !q.remove(v)
+                        catch
+                            case _: UnsupportedOperationException =>
+                                // TODO the race repair should use '!q.remove(v)' but JCTools doesn't support the operation.
+                                // In rare cases, items may be left in the queue permanently after closing due to this limitation.
+                                // The item will only be removed when the queue object itself is garbage collected.
+                                !q.contains(v)
+                    )
```
fwbrasil, line 372: This is the repair logic for the race condition. There's an issue with JCTools not supporting `remove` as described in the comment. It seems reasonable since the race condition should be quite rare.

kyo-core/shared/src/test/scala/kyo/QueueTest.scala
```diff
@@ -201,39 +203,161 @@ class QueueTest extends Test:
...
+            )
+                .pipe(Choice.run).unit
+                .repeat(repeats)
+                .as(succeed)
+        }
+    }
 
-    end TestUnsafeQueue
+    if Platform.isJVM then
+        "non-power-of-two capacity" - {
```
fwbrasil, line 336: jvm pending tests showing the behavior with queues rounding up the capacity to the next power of two

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -8,334 +9,295 @@ import scala.annotation.tailrec
...
+
+                val queue = Queue.Unsafe.init[A](_capacity, access)
+                val takes = new MpmcUnboundedXaddArrayQueue[Promise.Unsafe[Closed, A]](8)
+                val puts  = new MpmcUnboundedXaddArrayQueue[(A, Promise.Unsafe[Closed, Unit])](8)
+
+                def capacity = _capacity
+
+                def size()(using AllowUnsafe) = queue.size()
+
+                def offer(value: A)(using AllowUnsafe) =
```
fwbrasil, line 192: I've optimized these methods to only call `flush` if the operation was able to make a change, otherwise it isn't necessary.

kyo-core/shared/src/main/scala/kyo/Meter.scala
```diff
@@ -115,25 +108,23 @@ object Meter:
...
+
+                def dispatch[A, S](v: => A < S) =
+                    // Don't release a permit since it's managed by the timer task
+                    v
+
+                @tailrec def replenish(i: Int = 0): Unit =
+                    if i < rate then
+                        release()
+                        replenish(i + 1)
+
```
hearnadam, line 126: I could be wrong, but I suspect this may lead to over-replenishing if the permits are not fully used in a given period.

fwbrasil: Good catch! Fixed 🙏

kyo-core/shared/src/main/scala/kyo/Meter.scala
```diff
@@ -225,16 +221,126 @@ object Meter:
...
+                    else
+                        // CAS failed, retry
+                        loop()
+                    end if
+                end loop
+                loop()
+            }
+        end run
+
+        final def tryRun[A, S](v: => A < S)(using Frame): Maybe[A] < (S & Async & Abort[Closed]) =
```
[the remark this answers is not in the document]
fwbrasil: I considered that, `offer` in `Channel` and `Queue` is a similar scenario. Aborting seems a safer behavior since it could highlight potential issues but I don't have much experience with tracking failures statically. @steinybot could you share your thoughts?

fwbrasil:
merging to avoid conflicts with other changes but please feel free to provide feedback!

PR #767, opened 2024-10-20
kyo-monix

### LINKED ISSUE #763: Monix Interop
Enable users to switch from Monix to Kyo

hearnadam:
Put this together on the plane. I think I will need to handle interruption still.

What do you think about naming @fwbrasil? Is `Monixs` too awkward?

/fixes #763


fwbrasil:
could it maybe be just `Monix`?

kyo-monix/shared/src/main/scala/kyo/Monix.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import kyo.*
+import monix.eval.Task
+
+object Monix:
+
+    def get[A: Flat](task: Task[A])(using monix.execution.Scheduler, Frame): A < (Abort[Throwable] & Async) =
+        IO.Unsafe {
+            val p = Promise.Unsafe.init[Throwable, A]()
+            val cancelable = task.runAsync { (e: Either[Throwable, A]) =>
```
fwbrasil, line 11: I was debugging the failing interruption tests and noticed that `task.runAsync` evaluates the task synchronously in the same thread (kyo worker), which makes it not possible to interrupt the execution since the `onInterrupt` function isn't even registered. 

```suggestion
            val cancelable =  task.executeAsync.runAsync { (e: Either[Throwable, A]) =>
```

fwbrasil (approved):

PR #771, opened 2024-10-21
core: fix Fiber.race interrupts

fwbrasil:
We recently changed fibers to propagate interrupts separate from completions (see `IOPromise`'s `onComplete` vs `onInterrupt`) but forgot to update `Fiber.race` to ensure loser fibers are interrupted. Since the `interrupts` link now only triggers if there's an interruption, which isn't the case if a fiber wins the race, all raced fibers are running to completion.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -278,21 +278,17 @@ object Fiber extends FiberPlatformSpecific:
             end State
             val state = new State
             import state.*
-            foreach(seq)((idx, io) => io.evalNow.foreach(v => state(Result.success(v))))
-            if state.done() then
-                state
-            else
-                boundary { (trace, context) =>
-                    IO {
-                        foreach(seq) { (_, v) =>
-                            val fiber = IOTask(v, safepoint.copyTrace(trace), context)
-                            state.interrupts(fiber)
-                            fiber.onComplete(state)
-                        }
-                        state
+            boundary { (trace, context) =>
+                IO {
+                    val interruptPanic = Result.Panic(Fiber.Interrupted(frame))
+                    foreach(seq) { (_, v) =>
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -278,21 +278,17 @@ object Fiber extends FiberPlatformSpecific:
...
-                            state.interrupts(fiber)
-                            fiber.onComplete(state)
-                        }
-                        state
+            boundary { (trace, context) =>
+                IO {
+                    val interruptPanic = Result.Panic(Fiber.Interrupted(frame))
+                    foreach(seq) { (_, v) =>
+                        val fiber = IOTask(v, safepoint.copyTrace(trace), context)
+                        state.onComplete(_ => discard(fiber.interrupt(interruptPanic)))
```
fwbrasil, line 286: This is the main fix. Instead of using `state.interrupts`, which would trigger only if the state promise itself gets interrupted, the fix uses `onComplete` to interrupt loser fibers as soon as the `state` is complete. If the fiber is already completed, `interrupt` will return `false`, which is ok to ignore.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -278,21 +278,17 @@ object Fiber extends FiberPlatformSpecific:
             end State
             val state = new State
             import state.*
-            foreach(seq)((idx, io) => io.evalNow.foreach(v => state(Result.success(v))))
```
fwbrasil, line 281: I've removed this optimization that attempts to short-circuit the execution in case one of the computations is already done. The semantics of interruption wouldn't be consistent if we kept it. For example, in the test scenario "promise + plain value", it wouldn't interrupt the pending promise.

hearnadam (approved):

PR #772, opened 2024-10-22
prelude/core: fix boundary inference with Abort

fwbrasil:
As mentioned in https://github.com/getkyo/kyo/pull/765/files#r1807072333, the inference of `Boundary` in `Async` is failing when the compiler infers a widened `Abort` from the return type of an operation. This issue is also affecting [easyracer](https://github.com/jamesward/easyracer).

### DIFF
--- a/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
+++ b/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
@@ -640,7 +640,7 @@ extension [A, E, Ctx](effect: A < (Abort[E] & Async & Ctx))
     def fork(
         using
         flat: Flat[A],
-        boundary: Boundary[Ctx, IO],
+        boundary: Boundary[Ctx, IO & Abort[E]],
         reduce: Reducible[Abort[E]],
         frame: Frame
     ): Fiber[E, A] < (IO & Ctx) =
@@ -656,7 +656,7 @@ extension [A, E, Ctx](effect: A < (Abort[E] & Async & Ctx))
     def forkScoped(
         using
         flat: Flat[A],
-        boundary: Boundary[Ctx, IO],
+        boundary: Boundary[Ctx, IO & Abort[E]],
         reduce: Reducible[Abort[E]],
         frame: Frame
     ): Fiber[E, A] < (IO & Ctx & Resource) =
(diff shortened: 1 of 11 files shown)


kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -640,7 +640,7 @@ extension [A, E, Ctx](effect: A < (Abort[E] & Async & Ctx))
     def fork(
         using
         flat: Flat[A],
-        boundary: Boundary[Ctx, IO],
+        boundary: Boundary[Ctx, IO & Abort[E]],
```
fwbrasil, line 643: The second type parameter of `Boundary` represents effects that are allowed in addition to `ContextEffect`s. `Abort` was missing, which is the main cause of the bug.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -83,9 +82,10 @@ object Async:
       */
     def mask[E, A: Flat, Ctx](v: => A < (Abort[E] & Async & Ctx))(
         using
-        boundary: Boundary[Ctx, IO],
+        reduce: Reducible[Abort[E]],
+        boundary: Boundary[Ctx, Async & Abort[E]],
         frame: Frame
-    ): A < (Abort[E] & Async & Ctx) =
+    ): A < (reduce.SReduced & Async & Ctx) =
```
fwbrasil, line 88: related fix: `Reducible` was missing here

kyo-prelude/shared/src/main/scala/kyo/kernel/Boundary.scala
```diff
@@ -37,7 +37,7 @@ private[kyo] object Boundary:
 
         val s = flatten(TypeRepr.of[S])
 
-        val r = flatten(TypeRepr.of[Ctx]).filter(tpe => !s.exists(_ <:< tpe))
+        val r = flatten(TypeRepr.of[Ctx]).filter(tpe => !s.exists(tpe <:< _))
```
fwbrasil, line 40: There was another bug here, this check was inverted.

hearnadam (approved):

PR #774, opened 2024-10-22
data: Text

fwbrasil:
This PR introduces `Text` in `kyo-data` to provide more efficient handling of `String` operations. The main optimization is representing substrings via a `Cut` wrapper instead of copying strings.

This is part of the work to introduce a new `Parse` effect in `kyo-prelude` using effect handling for a more convenient backtracking mechanism.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Text.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Text.scala
@@ -1,0 +1,288 @@
+package kyo
+
+import Text.internal.*
+import scala.annotation.tailrec
+
+opaque type Text >: String = String | Op
+
+object Text:
+
+    def apply(s: String): Text = s
+
+    def empty: Text = ""
+
+    extension (self: Text)
+
+        def isEmpty: Boolean =
+            self match
+                case s: String => s.isEmpty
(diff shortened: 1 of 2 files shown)


kyo-data/shared/src/main/scala/kyo/Text.scala
```diff
@@ -0,0 +1,288 @@
+package kyo
+
+import Text.internal.*
+import scala.annotation.tailrec
+
+opaque type Text >: String = String | Op
```
fwbrasil, line 6: Note how `Text` is marked as a super class of `String`, which enables passing strings where `Text` is expected.

kyo-data/shared/src/main/scala/kyo/Text.scala
```diff
@@ -0,0 +1,288 @@
...
+
+object Text:
+
+    def apply(s: String): Text = s
```
fwbrasil, line 10: Just a helper in cases where inference doesn't transform a `String` into `Text`, similarly to `Kyo.pure`.

fwbrasil:
merging to post the PR with the new `Parse` effect

PR #775, opened 2024-10-23
[prelude][core] isolated locals and reentrant meters

fwbrasil:
Kyo's `Meter`s are currently non-reentrant. That's not an ideal default because it can easily lead to deadlocks. If a fiber tries to obtain the same mutex multiple times for example, it won't ever resume. This PR is the initial step to enable reentrancy by providing a new mechanism that isolates locals so they aren't inherited by forked fibers.

The solution uses a new `ContextEffect.Isolated` marker trait to filter entries from `Context` when a computation is being forked via a `Boundary`. The `Local` implementation uses two effect types to differentiate locals that are isolated or not.

With this new mechanism, I'm planning to change `Meter` to propagate an isolated local indicating that the fiber is currently holding a permit for a specific meter. Since locals are scoped, they will be cleared up as soon as the computation is finished, right before the permit is released. Forked fibers also won't see the local since it'll be isolated. The goal is to allow users to customize if a meter should be reentrant or not.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Meter.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Meter.scala
@@ -4,11 +4,16 @@ import org.jctools.queues.MpmcUnboundedXaddArrayQueue
 import scala.annotation.tailrec
 
 /** A Meter is an abstract class that represents a mechanism for controlling concurrency and rate limiting.
+  *
+  * Meters can be configured to be reentrant or non-reentrant. Reentrant meters allow nested calls from the same fiber, while non-reentrant
+  * meters will block or fail on nested calls.
   */
 abstract class Meter:
     self =>
 
     /** Runs an effect after acquiring a permit.
+      *
+      * If the meter is reentrant, nested calls from the same fiber will be allowed. If non-reentrant, nested calls will block or fail.
       *
       * @param v
       *   The effect to run.
@@ -66,6 +71,8 @@ end Meter
 
 object Meter:
 
+    import internal.*
+
     /** A no-op Meter that always allows operations and can't be closed. */
     case object Noop extends Meter:
         def availablePermits(using Frame)          = Int.MaxValue
@@ -76,24 +83,36 @@ object Meter:
         def closed(using Frame): Boolean < IO      = false
     end Noop
 
-    /** Creates a Meter that acts as a mutex (binary semaphore).
+    /** Creates a **reentrant** Meter that acts as a mutex (binary semaphore).
       *
       * @return
       *   A Meter effect that represents a mutex.
       */
     def initMutex(using Frame): Meter < IO =
-        initSemaphore(1)
+        initMutex(true)
+
+    /** Creates a Meter that acts as a mutex (binary semaphore).
+      *
+      * @param reentrant
+      *   If true, allows nested calls from the same fiber. Default is true.
+      * @return
+      *   A Meter effect that represents a mutex.
+      */
+    def initMutex(reentrant: Boolean = true)(using Frame): Meter < IO =
+        initSemaphore(1, reentrant)
 
     /** Creates a Meter that acts as a semaphore with the specified concurrency.
       *
       * @param concurrency
       *   The number of concurrent operations allowed.
+      * @param reentrant
+      *   If true, allows nested calls from the same fiber. Default is true.
       * @return
       *   A Meter effect that represents a semaphore.
       */
-    def initSemaphore(concurrency: Int)(using Frame): Meter < IO =
+    def initSemaphore(concurrency: Int, reentrant: Boolean = true)(using Frame): Meter < IO =
         IO.Unsafe {
-            new Base(concurrency):
+            new Base(concurrency, reentrant):
                 def dispatch[A, S](v: => A < S) =
                     // Release the permit right after the computation
                     IO.ensure(discard(release()))(v)
@@ -106,12 +125,14 @@ object Meter:
       *   The number of operations allowed per period.
       * @param period
       *   The duration of each period.
+      * @param reentrant
+      *   If true, allows nested calls from the same fiber. Default is true.
       * @return
       *   A Meter effect that represents a rate limiter.
       */
-    def initRateLimiter(rate: Int, period: Duration)(using initFrame: Frame): Meter < IO =
+    def initRateLimiter(rate: Int, period: Duration, reentrant: Boolean = true)(using initFrame: Frame): Meter < IO =
         IO.Unsafe {
-            new Base(rate):
+            new Base(rate, reentrant):
                 val timerTask =
                     // Schedule periodic task to replenish permits
                     IO.Unsafe.run(Clock.repeatAtInterval(period)(replenish())).eval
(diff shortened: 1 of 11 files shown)


kyo-prelude/shared/src/main/scala/kyo/Local.scala
```diff
@@ -77,8 +67,41 @@ object Local:
...
+        new Base[A, IsolatedState]:
+            def tag             = Tag[IsolatedState]
+            lazy val default: A = defaultValue
+
+    object internal:
+
+        sealed private[Local] trait State         extends ContextEffect[Map[Local[?], AnyRef]]
+        sealed private[Local] trait IsolatedState extends ContextEffect[Map[Local[?], AnyRef]] with ContextEffect.Isolated
+
+        abstract class Base[A, E <: ContextEffect[Map[Local[?], AnyRef]]] extends Local[A]:
```
fwbrasil, line 85: needs to be public because the `init*` methods are `inline`

fwbrasil:
I've updated this PR to include the `Meter` changes to support reentrancy.

kyo-prelude/shared/src/main/scala/kyo/kernel/Context.scala
```diff
@@ -16,11 +16,14 @@ object Context:
         inline def contains[A, E <: ContextEffect[A]](tag: Tag[E]): Boolean =
             context.contains(tag.erased)
 
+        inline def inherit: Context =
+            context.filterNot(_._1 <:< Tag[ContextEffect.Isolated])
```
hearnadam, line 20: To generalize this approach a bit more, do you think `ContextEffect` should have an `onFork` (or equivalent) function? It seems like this could be useful for other context effects.

'def onFork' can essentially become `identity | default` in this case.

fwbrasil:
@hearnadam anything pending to merge this?

kyo-core/shared/src/main/scala/kyo/Meter.scala
```diff
@@ -76,24 +83,36 @@ object Meter:
         def closed(using Frame): Boolean < IO      = false
     end Noop
 
-    /** Creates a Meter that acts as a mutex (binary semaphore).
+    /** Creates a **reentrant** Meter that acts as a mutex (binary semaphore).
       *
       * @return
       *   A Meter effect that represents a mutex.
       */
     def initMutex(using Frame): Meter < IO =
```
hearnadam, line 91: Why not use default param? worse binary compat?

kyo-core/shared/src/main/scala/kyo/Meter.scala
```diff
@@ -239,8 +262,23 @@ object Meter:
...
+                acquiredMeters.use { meters =>
+                    if meters.contains(this) then reenter
+                    else IO.Unsafe(acquire)
+                }
+            else
+                IO.Unsafe(acquire)
+
+        private inline def withAcquiredMeter[A, S](inline v: => A < S) =
+            if reentrant then
+                acquiredMeters.update(_ + this)(v)
```
hearnadam, line 276: Couldn't this be inlined to line 269?

kyo-prelude/shared/src/main/scala/kyo/Local.scala
```diff
@@ -57,28 +55,68 @@ abstract class Local[A]:
...
+    @nowarn("msg=anonymous")
+    inline def initIsolated[A](inline defaultValue: A): Local[A] =
+        new Base[A, IsolatedState]:
+            def tag             = Tag[IsolatedState]
+            lazy val default: A = defaultValue
+
+    object internal:
+
+        sealed private[kyo] trait State         extends ContextEffect[Map[Local[?], AnyRef]]
+        sealed private[kyo] trait IsolatedState extends ContextEffect[Map[Local[?], AnyRef]] with ContextEffect.Isolated
```
hearnadam, line 98: Why does this need to be a `trait`?

kyo-prelude/shared/src/main/scala/kyo/Local.scala
```diff
@@ -57,28 +55,68 @@ abstract class Local[A]:
...
+      *   The default value for the Local
+      * @return
+      *   A new isolated Local instance
+      */
+    @nowarn("msg=anonymous")
+    inline def initIsolated[A](inline defaultValue: A): Local[A] =
+        new Base[A, IsolatedState]:
+            def tag             = Tag[IsolatedState]
+            lazy val default: A = defaultValue
+
```
hearnadam, line 94: How about using a boolean for `isolated`?

kyo-prelude/shared/src/main/scala/kyo/Local.scala
```diff
@@ -57,28 +55,68 @@ abstract class Local[A]:
...
+      * inherit from their parent fiber.
+      *
+      * @param defaultValue
+      *   The default value for the Local
+      * @return
+      *   A new isolated Local instance
+      */
+    @nowarn("msg=anonymous")
+    inline def initIsolated[A](inline defaultValue: A): Local[A] =
+        new Base[A, IsolatedState]:
```
hearnadam, line 91: Does the type actually need to be reflected in Base? Or is it just important for the tag?

kyo-prelude/shared/src/main/scala/kyo/kernel/Context.scala
hearnadam: Ah. This seems like it could be a limitation in the future... I'll think about this problem more.

fwbrasil: It seems a reasonable given that the alternative requires code to cache instances in effects like `Abort`, which also requires `asInstanceOf`. I think we should avoid changing core aspects like this one before 1.0.

kyo-core/shared/src/main/scala/kyo/Meter.scala
fwbrasil: With a default param, it's not possible to call `Meter.initMutex` without an empty `()` at the end.

kyo-core/shared/src/main/scala/kyo/Meter.scala
fwbrasil: This method is used at the end when a computation is being effectively dispatched to update the local with the acquired meter. `withReentry` is used at another point at the start to check if the fiber is reentering. Between these two points, the computation might stay suspended in case there are no available permits

kyo-prelude/shared/src/main/scala/kyo/Local.scala
fwbrasil: All effect types are encoded as `sealed trait` without any sub-classes since they aren't instantiated, being only used for matching tags. It could be `sealed abstract class` instead.

kyo-prelude/shared/src/main/scala/kyo/Local.scala
fwbrasil: do you mean `def init(isolated: Boolean)`? I think it'd be less clear

kyo-prelude/shared/src/main/scala/kyo/Local.scala
fwbrasil: Yes, it's just so it's possible to use the two different tags

PR #778, opened 2024-10-26
feat: setting scala LTS version for relevant submodules

### LINKED ISSUE #715: Try and keep some modules on LTS Scala
I'm enjoying using Kyo, but as a library maintainer and enterprise user, the dependency on non-LTS Scala means my users would also have to go on a non-LTS Scala version.

Perhaps some analysis of what can remain on LTS can be done, for example:

- `kyo-data` - doesn't seem to have anything that requires 3.5.x

mobley-trent:
Closes #715
/claim #715


build.sbt
```diff
@@ -6,6 +6,7 @@ import org.typelevel.scalacoptions.ScalaVersion
 import sbtdynver.DynVerPlugin.autoImport.*
 
 val scala3Version   = "3.5.2"
+val scala333Version = "3.3.3"
```
sideeffffect, line 9:
```suggestion
val scala333Version = "3.3.4"
```

PR #780, opened 2024-10-27
core: simplify KyoApp

fwbrasil:
More details in comments

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
+++ b/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
@@ -9,22 +9,14 @@ import scala.collection.mutable.ListBuffer
   *
   * Note: This class and its methods are unsafe and should only be used as the entrypoint of an application.
   */
-abstract class KyoApp extends KyoApp.Base[KyoApp.Effects]:
-    def log: Log       = Log.live
-    def random: Random = Random.live
-    def clock: Clock   = Clock.live
+abstract class KyoApp extends KyoApp.Base[Async & Resource & Abort[Throwable]]:
+    def timeout: Duration = Duration.Infinity
 
-    override protected def handle[A: Flat](v: A < KyoApp.Effects)(using Frame): Unit =
+    override protected def handle[A: Flat](v: A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
         import AllowUnsafe.embrace.danger
-        v.map { v =>
-            if (()).equals(v) then ()
-            else Console.println(v)
-        }.pipe(
-            Clock.let(clock),
-            Random.let(random),
-            Log.let(log),
-            KyoApp.Unsafe.run
-        )
+        val result = KyoApp.Unsafe.runAndBlock(timeout)(v)
+        if !result.exists(().equals(_)) then
+            println(pprint.apply(result).plainText)
     end handle
 end KyoApp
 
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -9,22 +9,13 @@ import scala.collection.mutable.ListBuffer
   *
   * Note: This class and its methods are unsafe and should only be used as the entrypoint of an application.
   */
-abstract class KyoApp extends KyoApp.Base[KyoApp.Effects]:
-    def log: Log.Unsafe = Log.unsafe
-    def random: Random  = Random.live
-    def clock: Clock    = Clock.live
```
fwbrasil, line 15: I'm working on new `Clock` and `Timer` implementations that will offer [APIs to control time](https://github.com/getkyo/kyo/blob/cb97b5d74131740572098d5ec3b611144acf4bde/kyo-core/shared/src/main/scala/kyo/Clock.scala#L132). Overriding the default services via these high-level combinators is a more user-friendly approach than having to implement them manually. `Random` and `Log` can already be overriden via their `let` method but I'll also look into providing convenient behavior customization APIs for them.

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -55,9 +46,6 @@ object KyoApp:
             initCode += (() => handle(v))
     end Base
 
-    /** The combined effect type used by KyoApp. */
-    type Effects = Async & Resource & Abort[Throwable]
```
fwbrasil, line 59: I think this indirection can be confusing to users. Splicing the type union on each use will allow users to directly see what effects are supported by `run` for example.

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -74,82 +62,12 @@ object KyoApp:
...
-          *   Evidence that A is Flat.
-          * @param frame
-          *   The implicit Frame.
-          * @return
-          *   A Fiber representing the running effect.
+          *   A Result containing either the computed value or a failure.
           */
-        def runFiber[A: Flat](timeout: Duration)(v: A < Effects)(using Frame, AllowUnsafe): Fiber[Throwable, A] =
-            v.pipe(Resource.run, Async.timeout(timeout), Async.run, IO.Unsafe.run).eval
+        def runAndBlock[A: Flat](timeout: Duration)(
```
fwbrasil, line 67: Simplified all methods to a single one that indicates it's blocking (something important that was missing) and returns a `Result`. The other methods didn't seem very useful and their behavior can be implemented by the user via regular effect handling before calling `runAndBlock`.

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -74,82 +62,12 @@ object KyoApp:
           * @param frame
           *   The implicit Frame.
           * @return
-          *   A Result containing either the computed value or a Throwable.
-          */
-        def attempt[A: Flat](timeout: Duration)(v: A < Effects)(using Frame, AllowUnsafe): Result[Throwable, A] =
-            IO.Unsafe.run(runFiber(timeout)(v).block(timeout)).eval
```
fwbrasil, line 80: This method was unnecessarily using two timeouts for the same computation

fwbrasil:
By the way, my initial goal with this change was exploring if we should provide `Layer` integration in `KyoApp`. My conclusion was that using the regular provided APIs to wire layers is good enough and there's no need to replicate them in `KyoApp`.

hearnadam (approved):

PR #783, opened 2024-10-27
[core] Log improvements

fwbrasil:
Following up on https://github.com/getkyo/kyo/pull/780#discussion_r1818130462

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Log.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Log.scala
@@ -19,15 +19,16 @@ end Log
 /** Logging utility object for Kyo applications. */
 object Log extends LogPlatformSpecific:
 
-    case class Level private (priority: Int) extends AnyVal:
+    final class Level private (private val priority: Int) extends AnyVal:
         def enabled(minLevel: Level) = priority >= minLevel.priority
+
     object Level:
         val trace: Level = Level(10)
         val debug: Level = Level(20)
         val info: Level  = Level(30)
         val warn: Level  = Level(40)
         val error: Level = Level(50)
-    end Level
+    end Level   
 
     private val local = Local.init(live)
 
--- a/kyo-core/shared/src/main/scala/kyo/Log.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Log.scala
@@ -28,7 +28,7 @@ object Log extends LogPlatformSpecific:
         val info: Level  = Level(30)
         val warn: Level  = Level(40)
         val error: Level = Level(50)
-    end Level   
+    end Level
 
     private val local = Local.init(live)
 
--- a/kyo-core/shared/src/main/scala/kyo/Log.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Log.scala
@@ -19,9 +19,15 @@ end Log
 /** Logging utility object for Kyo applications. */
 object Log extends LogPlatformSpecific:
 
-    enum Level:
-        case Trace, Debug, Info, Warn, Error
-        def enabled(minLevel: Level): Boolean = this.ordinal >= minLevel.ordinal
+    case class Level private (priority: Int) extends AnyVal:
+        def enabled(minLevel: Level) = priority >= minLevel.priority
+    object Level:
+        val trace: Level = Level(10)
+        val debug: Level = Level(20)
+        val info: Level  = Level(30)
+        val warn: Level  = Level(40)
+        val error: Level = Level(50)
+    end Level
 
     private val local = Local.init(live)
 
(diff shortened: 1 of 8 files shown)


kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -2,10 +2,28 @@ package kyo
 
 import kyo.internal.LogPlatformSpecific
 
+final case class Log(unsafe: Log.Unsafe) extends AnyVal:
```
fwbrasil, line 5: Follow the pattern of other effects with a `Log` and `Log.Unsafe` pair

kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -2,10 +2,28 @@ package kyo
...
+    inline def warn(inline msg: => String, t: => Throwable)(using inline frame: Frame): Unit < IO         = IO.Unsafe(unsafe.warn(msg, t))
+    inline def error(inline msg: => String)(using inline frame: Frame): Unit < IO                         = IO.Unsafe(unsafe.error(msg))
+    inline def error(inline msg: => String, t: => Throwable)(using inline frame: Frame): Unit < IO        = IO.Unsafe(unsafe.error(msg, t))
+end Log
+
 /** Logging utility object for Kyo applications. */
 object Log extends LogPlatformSpecific:
 
-    private val local = Local.init[Unsafe](unsafe)
+    enum Level:
```
fwbrasil, line 22: More high-level handling of log levels

kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -16,16 +34,52 @@ object Log extends LogPlatformSpecific:
...
+      *   The name to use for the console logger
+      * @param level
+      *   The log level
+      * @param v
+      *   The effect to execute with the console logger
+      * @return
+      *   The result of executing the effect with the console logger
+      */
+    def withConsoleLogger[A, S](name: String = "kyo.logs", level: Level = Level.Debug)(v: A < S)(using Frame): A < S =
+        let(Log(Unsafe.ConsoleLogger(name, level)))(v)
```
fwbrasil, line 78: Convenience methods to redirect logs to the console

kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -37,89 +91,82 @@ object Log extends LogPlatformSpecific:
         def warn(msg: => String, t: => Throwable)(using frame: Frame, allow: AllowUnsafe): Unit
         def error(msg: => String)(using frame: Frame, allow: AllowUnsafe): Unit
         def error(msg: => String, t: => Throwable)(using frame: Frame, allow: AllowUnsafe): Unit
+
+        def safe: Log = Log(this)
```
hearnadam, line 95: Not sure I see the value in this to be honest. It's unlike other designs where the APIs are hidden behind an opaque type.

kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -2,10 +2,28 @@ package kyo
...
+    inline def error(inline msg: => String, t: => Throwable)(using inline frame: Frame): Unit < IO        = IO.Unsafe(unsafe.error(msg, t))
+end Log
+
 /** Logging utility object for Kyo applications. */
 object Log extends LogPlatformSpecific:
 
-    private val local = Local.init[Unsafe](unsafe)
+    enum Level:
+        case Trace, Debug, Info, Warn, Error
+        def enabled(minLevel: Level): Boolean = this.ordinal >= minLevel.ordinal
```
hearnadam, line 24: Will this have any issues if we need to add/remove levels?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/Log.scala
fwbrasil: The main reason for adding it was so it's possible to have `Log.get` and `Local.use` methods. It'd be strange to add them with a `Log.Unsafe` type. That's something useful because computations can get the `Log` instance once and avoid further `Local` lookups. Another use case is passing the instance to some unsafe execution context like library integrations. While not essential, adding it keeps the consistency with similar APIs and provides these benefits

hearnadam: Can the local store the Safe variant?

fwbrasil: Yes, I've changed it to use safe variant

kyo-core/shared/src/main/scala/kyo/Log.scala
fwbrasil: I've made it be based on an `Int` with some padding so we can make changes in the future

PR #785, opened 2024-10-28
[core] Console improvements

fwbrasil:


kyo-core/shared/src/main/scala/kyo/Console.scala
```diff
@@ -1,46 +1,52 @@
 package kyo
 
 import java.io.EOFException
+import java.io.IOException
 
 /** Represents a console for input and output operations.
   */
-abstract class Console:
-    def unsafe: Console.Unsafe
+final case class Console(unsafe: Console.Unsafe) extends AnyVal:
```
fwbrasil, line 8: update the pattern like in the other effects

kyo-core/shared/src/main/scala/kyo/Console.scala
```diff
@@ -1,46 +1,52 @@
...
 
     /** Reads a line from the console.
       *
       * @return
       *   A String representing the line read from the console.
       */
-    def readln(using Frame): String < IO
+    def readln(using Frame): String < (IO & Abort[IOException]) = IO.Unsafe(Abort.get(unsafe.readln()))
```
fwbrasil, line 15: Added tracking of `IOException`.

kyo-core/shared/src/main/scala/kyo/Console.scala
```diff
@@ -49,22 +55,19 @@ object Console:
...
-                def printlnErr(s: String)(using AllowUnsafe) = scala.Console.err.println(s)
-        )
+    val live: Console = Console(
+        new Unsafe:
+            def readln()(using AllowUnsafe) =
+                Result.catching(Maybe(scala.Console.in.readLine())).flatMap {
+                    case Absent         => Result.fail(new EOFException("Consoles.readln failed."))
+                    case Present(value) => Result.success(value)
+                }
+            def print(s: String)(using AllowUnsafe)      = Result.catching(scala.Console.out.print(s))
```
hearnadam, line 65: Will this catch only the type in the signature?

kyo-core/shared/src/main/scala/kyo/Console.scala
```diff
@@ -49,22 +55,19 @@ object Console:
...
-                def printErr(s: String)(using AllowUnsafe)   = scala.Console.err.print(s)
-                def println(s: String)(using AllowUnsafe)    = scala.Console.out.println(s)
-                def printlnErr(s: String)(using AllowUnsafe) = scala.Console.err.println(s)
-        )
+    val live: Console = Console(
+        new Unsafe:
+            def readln()(using AllowUnsafe) =
+                Result.catching(Maybe(scala.Console.in.readLine())).flatMap {
+                    case Absent         => Result.fail(new EOFException("Consoles.readln failed."))
+                    case Present(value) => Result.success(value)
```
hearnadam, line 63: `Maybe` could have a method `toSuccess[E](error: => E)`?

kyo-core/shared/src/main/scala/kyo/Console.scala
```diff
@@ -80,12 +83,96 @@ object Console:
...
+      *   A tuple containing the captured output (Out) and the computation result.
+      */
+    def withOut[A, S](v: A < S)(using Frame): (Out, A) < (IO & S) =
+        use { console =>
+            IO {
+                val stdOut = new StringBuffer
+                val stdErr = new StringBuffer
+                val proxy =
+                    new Proxy(console.unsafe):
+                        override def print(s: String)(using AllowUnsafe)      = Result.success(stdOut.append(s)).unit
```
hearnadam, line 161: Will this actually output?

kyo-core/shared/src/main/scala/kyo/Console.scala
```diff
@@ -80,12 +83,96 @@ object Console:
...
+      * @param v
+      *   The computation to execute.
+      * @tparam A
+      *   The type of the computation result.
+      * @tparam S
+      *   The type of effects in the computation.
+      * @return
+      *   A tuple containing the captured output (Out) and the computation result.
+      */
+    def withOut[A, S](v: A < S)(using Frame): (Out, A) < (IO & S) =
```
hearnadam, line 154: `withCapturedOut`?

sideeffffect (commented):
Since this PR is about Console and errors, I'd like to ask, whether Kyo is able to surface (instead of swallowing) the error(s) described in

 * https://blog.sunfishcode.online/bugs-in-hello-world/
 * https://github.com/sunfishcode/hello-world-vs-io-errors

?

kyo-core/shared/src/main/scala/kyo/Console.scala
fwbrasil: Yes, the exception type is inferred but I've added explicit type params to avoid ambiguity

kyo-core/shared/src/main/scala/kyo/Console.scala
fwbrasil: good idea! I've added `maybe.toResult` methods.

kyo-core/shared/src/main/scala/kyo/Console.scala
fwbrasil: No, the idea is to only collect the output

kyo-core/shared/src/main/scala/kyo/Console.scala
fwbrasil: I think I prefer the shorter version

fwbrasil:
> Since this PR is about Console and errors, I'd like to ask, whether Kyo is able to surface (instead of swallowing) the error(s) described in
> 
> * https://blog.sunfishcode.online/bugs-in-hello-world/
> * https://github.com/sunfishcode/hello-world-vs-io-errors
> 
> ?

No, we don't handle exit codes yet. Do you mind creating an issue for it?

sideeffffect:
> we don't handle exit codes yet. Do you mind creating an issue for it?

Happy to help:
 * https://github.com/getkyo/kyo/issues/790

But the issue the blog post describes is even more about silently swallowing failures when writing to stream/pipe. So I thought it's relevant to this PR, which is changing things around `Console`. Does (Kyo) Console swallow errors like that? Because errors passing silently is a _bad thing ™_, right?

hearnadam (approved):

PR #786, opened 2024-10-28
[tapir] propagate shutdown timeout to netty

fwbrasil:
Tests that start a tapir server are taking ~2s to finalize because Netty waits for some time before shutting down. This PR makes the shut down process respect the timeout defined by the user.


kyo-tapir/shared/src/main/scala/kyo/server/NettyKyoServer.scala
```diff
@@ -153,15 +153,16 @@ case class NettyKyoServer(
         gracefulShutdownTimeout: Option[FiniteDuration]
     ): KyoSttpMonad.M[Unit] =
         isShuttingDown.set(true)
+        val timeout = gracefulShutdownTimeout.map(_.toNanos).getOrElse(Long.MaxValue)
```
hearnadam, line 156:
```suggestion
        val timeout = gracefulShutdownTimeout.fold(Long.MaxValue)(_.toNanos)
```

hearnadam (approved):

PR #789, opened 2024-10-29
[core] time shift and control

fwbrasil:
This PR uses the new [Schedule](https://github.com/getkyo/kyo/pull/733) in `kyo-data` to reimplement timers in `kyo-core`. The new design provides two mechanisms to control time: `Clock.withTimeShift` speeds up or slows down the time for a computation based on a factor and `Clock.withTimeControl` provides an API for users to manually set the time. Both methods not only affect `Clock.now` but also scheduled operations in `Timer` and `Async.sleep` calls.


kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -87,7 +100,18 @@ object Clock:
     val live: Clock =
         Clock(
             new Unsafe:
-                def now()(using AllowUnsafe): Instant = Instant.fromJava(java.time.Instant.now())
+                val executor                 = Executors.newScheduledThreadPool(2, Threads("kyo-core-clock-executor"))
+                def now()(using AllowUnsafe) = Instant.fromJava(java.time.Instant.now())
+                def sleep(duration: Duration) =
```
fwbrasil, line 105: I had to move the management of timed tasks to `Clock` so it's possible to customize the behavior for time shifting and control

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -104,21 +128,158 @@ object Clock:
...
+                    @volatile var current = Instant.Epoch
+
+                    case class Task(deadline: Instant) extends IOPromise[Nothing, Unit] with Delayed:
+                        def getDelay(unit: TimeUnit): Long =
+                            (deadline - current).to(unit)
+                        def compareTo(other: Delayed): Int =
+                            deadline.toJava.compareTo(other.asInstanceOf[Task].deadline.toJava)
+                    end Task
+
+                    val queue = new DelayQueue[Task]
```
fwbrasil, line 232: As simple timer implementation using `DelayQueue`. The queue is `synchronized` so `sleep` can be called concurrently by multiple fibers

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -104,21 +128,158 @@ object Clock:
...
+                            tick()
+                        }
+
+                    def advance(duration: Duration) =
+                        IO {
+                            current = current + duration
+                            tick()
+                        }
+
+                    def tick(): Unit =
```
fwbrasil, line 254: When the user changes the time, this method is called to execute the delayed tasks if it's time. The `DelayQueue` returns `null` in case there are no items ready for execution

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -365,9 +365,10 @@ object Fiber extends FiberPlatformSpecific:
             def onComplete(f: Result[E, A] => Unit)(using AllowUnsafe): Unit                             = self.onComplete(f)
             def onInterrupt(f: Panic => Unit)(using Frame): Unit                                         = self.onInterrupt(f)
             def block(deadline: Clock.Deadline.Unsafe)(using AllowUnsafe, Frame): Result[E | Timeout, A] = self.block(deadline)
-            def interrupt(error: Panic)(using AllowUnsafe): Boolean                                      = self.interrupt(error)
-            def interruptDiscard(error: Panic)(using AllowUnsafe): Unit                                  = discard(self.interrupt(error))
-            def mask()(using AllowUnsafe): Unsafe[E, A]                                                  = self.mask()
+            def interrupt()(using frame: Frame, allow: AllowUnsafe): Boolean = self.interrupt(Panic(Interrupted(frame)))
```
fwbrasil, line 368: A new overloaded `interrupt` without params. The following methods were changed only because of formatting

kyo-core/shared/src/main/scala/kyo/Timer.scala
```diff
@@ -1,288 +1,87 @@
...
+        state: A
+    )(
+        f: A => A < (Async & Abort[E])
+    )(using Frame): Fiber[E, A] < IO =
+        repeatWithDelay(Schedule.delay(startAfter).andThen(Schedule.fixed(delay)), state)(f)
+
+    def repeatWithDelay[E, S](delaySchedule: Schedule)(f: => Unit < (Async & Abort[E]))(using Frame): Fiber[E, Unit] < IO =
+        repeatWithDelay(delaySchedule, ())(_ => f)
+
+    def repeatWithDelay[E, A: Flat, S](
```
fwbrasil, line 28: Instead of relying on the underlying java timer to manage repeated tasks, this new `Timer` implementation spins up new fibers that use the provided `Schedule` to repeat and `Clock` to sleep when necessary. This approach also allows scheduled tasks to have internal state between iterations and return a final value at the end (see the `A` type).

kyo-data/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -58,7 +59,7 @@ object Duration:
       * @return
       *   A Duration instance
       */
-    inline def fromNanos(value: Long): Duration =
```
fwbrasil, line 61: Something that was quite annoying while working on this change were assertion errors reporting inlined trees from `Duration` operations. It becomes a large and confusing error message. Since these methods don't take functions, the benefit of inlining them is low so I've decided to remove the modifier

kyo-data/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -118,44 +119,52 @@ object Duration:
...
-        inline def toMicros: Long  = self.to(Micros)
-        inline def toMillis: Long  = self.to(Millis)
-        inline def toSeconds: Long = self.to(Seconds)
-        inline def toMinutes: Long = self.to(Minutes)
-        inline def toHours: Long   = self.to(Hours)
-        inline def toDays: Long    = self.to(Days)
-        inline def toWeeks: Long   = self.to(Weeks)
-        inline def toMonths: Long  = self.to(Months)
-        inline def toYears: Long   = self.to(Years)
+        def to(timeUnit: TimeUnit): Long =
```
fwbrasil, line 150: new methods to integrate with `TimeUnit` and `ChronoUnit`

kyo-core/shared/src/main/scala/kyo/Clock.scala
fwbrasil: I had to change it to a manually-synchronized `PriorityQueue` because `DelayQueue` isn't available in js

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -104,21 +118,163 @@ object Clock:
...
 
+    /** Gets the current Clock instance from the local context.
+      *
+      * This is the primary way to access the current Clock instance when you only need to use it directly.
+      *
+      * @return
+      *   The current Clock instance
+      */
+    def get(using Frame): Clock < Any =
+        use(identity)
```
hearnadam, line 129: Any reason to avoid local.get?
```suggestion
        local.get
```

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -104,21 +118,163 @@ object Clock:
...
+      *
+      * @param factor
+      *   The time scaling factor. Values > 1 speed up time, values < 1 slow down time
+      * @param v
+      *   The effect to run with scaled time
+      * @return
+      *   The result of running the effect with scaled time
+      */
+    def withTimeShift[A, S](factor: Double)(v: => A < S)(using Frame): A < (IO & S) =
+        if factor == 1 then v
```
hearnadam, line 154: - Do you foresee any issues with imprecise `1.000001`? in the case of speeding up + slowing down?
- If a user speeds up with timeShift: 2.0, they would then need to run: .5 to get back normal speed. From my understanding the new clock will include multiple factors. Is there a better way to Reset to an 'unshifted' clock?

kyo-core/shared/src/main/scala/kyo/Timer.scala
hearnadam: I think this is much more useful!

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -105,19 +105,9 @@ object Async:
...
-                if d.isFinite then
-                    Timer.schedule(d)(p.completeDiscard(Result.success(()))).map { t =>
-                        IO.ensure(t.cancel.unit)(p.safe.get)
-                    }
-                else
-                    p.safe.get
-                end if
-            }
+    def sleep(duration: Duration)(using Frame): Unit < Async =
+        if duration == Duration.Zero then ()
```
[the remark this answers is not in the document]
fwbrasil: It's a performance optimization to avoid suspensions

kyo-core/shared/src/main/scala/kyo/Clock.scala
fwbrasil: The time shifting is scoped to the `v` computation. As soon as it finishes, the custom shifted clock is cleared and time returns to normal. I don't think precision would be an issue unless users nest time shifts, which I imagine won't be common.

kyo-core/shared/src/main/scala/kyo/Timer.scala
```diff
@@ -1,288 +1,213 @@
...
-/** A timer for scheduling tasks to run after a delay or periodically. */
-abstract class Timer:
-    def unsafe: Timer.Unsafe
+/** A utility for scheduling recurring tasks with different timing strategies.
+  *
+  * Timer provides two main scheduling approaches:
+  *   - `repeatWithDelay`: Executes tasks with a fixed delay between task completions
+  *   - `repeatAtInterval`: Executes tasks at fixed time intervals, regardless of task duration
+  */
+object Timer:
```
hearnadam, line 9: Do we need a distinct object for `Timer`? I generally prefer extension methods (or at least avoiding too many distinct objects) for discoverability.

I think most other effect systems don't have a concept of 'Timer'

kyo-data/shared/src/main/scala/kyo/Duration.scala
hearnadam: Sounds good, I think this is the right move. I've noticed this issue with `loop` as well. If you incorrectly return a non-loop (Done/Continue) value the expanded tree is massive.

kyo-data/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -116,46 +117,72 @@ object Duration:
         val factor: Double = chronoUnit.getDuration.toNanos.toDouble
     end Units
 
+    object Units:
+        def fromJava(chronoUnit: ChronoUnit): Units =
+            Units.values.find(_.chronoUnit.equals(chronoUnit)) match
```
hearnadam, line 122: Should this be a Map?

kyo-core/shared/src/main/scala/kyo/Fiber.scala
hearnadam: This will be useful! I think kyo-zio can use it.

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -128,31 +284,39 @@ object Clock:
...
+        final def safe: Clock = Clock(this)
+    end Unsafe
+
+    object Unsafe:
+        def apply(executor: ScheduledExecutorService)(using AllowUnsafe): Unsafe =
+            new Unsafe:
+                def now()(using AllowUnsafe) = Instant.fromJava(java.time.Instant.now())
+                def sleep(duration: Duration) =
+                    Promise.Unsafe.fromIOPromise {
+                        new IOPromise[Nothing, Unit] with Callable[Unit]:
```
hearnadam, line 312: Any reason to use `Callable[Unit]` vs `Runnable`?

kyo-core/shared/src/main/scala/kyo/Timer.scala
fwbrasil: Yeah, it doesn't feel necessary. Perhaps the methods could be in `Clock` since `sleep` is already there?

fwbrasil: I've made the change, it seems ok

kyo-core/shared/src/main/scala/kyo/Clock.scala
fwbrasil: I had changed it to `Runnable` but then I remembered why I used `Callable`. It's what's supported by the ScalaJS stubs:

```
[info] compiling 27 Scala sources to /home/runner/work/kyo/kyo/kyo-prelude/js/target/scala-3.5.2/test-classes ...
[error] -- [E007] Type Mismatch Error: /home/runner/work/kyo/kyo/kyo-core/shared/src/main/scala/kyo/Clock.scala:516:57 
[error] 516 |                            val task = executor.schedule(this, duration.toNanos, TimeUnit.NANOSECONDS)
[error]     |                                                         ^^^^
[error]     |Found:    ($anon.this : kyo.scheduler.IOPromise[Nothing, Unit] with Runnable {...})
[error]     |Required: java.util.concurrent.Callable[A]
[error]     |
[error]     |where:    A is a type variable
[error]     |
[error]     | longer explanation available when compiling with `-explain`
[error] one error found
```

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -128,31 +284,242 @@ object Clock:
...
+      * executions.
+      *
+      * @param delay
+      *   The duration to wait after each task completion before starting the next execution
+      * @param f
+      *   The task to execute
+      * @return
+      *   A Fiber that can be used to control or interrupt the recurring task
+      */
+    def repeatWithDelay[E, S](delay: Duration)(f: => Unit < (Async & Abort[E]))(using Frame): Fiber[E, Unit] < IO =
```
hearnadam, line 301: I feel this bloats clock, but as discussed offline there really isn't a good line between Async and Clock for time based execution. Seems fine.

hearnadam (approved):

PR #792, opened 2024-10-29
[prelude] Parse effect

fwbrasil:
Provides parsing capabilities with backtracking and lookahead support. The effect assumes there's an input string in scope (`Var`) being parsed and operations consume from it, evaluating multiple possible branches with `Choice` and failing with `Abort` if necessary. It's also possible to mix other effects with `Parse`.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Text.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Text.scala
@@ -7,6 +7,8 @@ opaque type Text >: String = String | Op
 
 object Text:
 
+    given Tag[Text] = Tag.fromRaw("kyo.Text")
+
     def apply(s: String): Text = s
 
     def empty: Text = ""
@@ -229,6 +231,17 @@ object Text:
         def show: String = self.toString()
 
         def compact: Text = self.toString()
+
+        def is(other: Text): Boolean =
+            if self.length != other.length then false
+            else
+                @tailrec
+                def loop(i: Int): Boolean =
+                    if i >= self.length then true
+                    else if self.charAt(i) != other.charAt(i) then false
+                    else loop(i + 1)
+                loop(0)
+        end is
 
     end extension
 
--- a/kyo-prelude/shared/src/main/scala/kyo/Parse.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Parse.scala
@@ -1,0 +1,473 @@
+package kyo
+
+import kyo.Ansi.*
+import kyo.kernel.*
+
+/** The Parse effect combines three fundamental capabilities needed for parsing:
+  *   - State management (Var[Parse.State]) to track input position
+  *   - Choice for handling alternatives and backtracking
+  *   - Error handling (Abort[ParseFailed]) for parse failures
+  *
+  * This combination enables building complex parsers that can handle ambiguous grammars, implement look-ahead, and provide detailed error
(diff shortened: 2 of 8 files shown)


kyo-data/shared/src/main/scala/kyo/Text.scala
```diff
@@ -7,6 +7,8 @@ opaque type Text >: String = String | Op
 
 object Text:
 
+    given Tag[Text] = Tag.fromRaw("kyo.Text")
```
fwbrasil, line 10: A workaround since it's currently not possible to have a `Tag` for a type union. This raw tag will only support exact match and not `<:<`, which is ok since that's what effect handling uses.

kyo-data/shared/src/main/scala/kyo/Text.scala
```diff
@@ -230,6 +232,17 @@ object Text:
 
         def compact: Text = self.toString()
 
+        def is(other: Text): Boolean =
```
fwbrasil, line 235: New method to check for equality. I've just noticed that `Text` is missing scaladocs, I'll follow up separatelly.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -0,0 +1,473 @@
...
+  *   - State management (Var[Parse.State]) to track input position
+  *   - Choice for handling alternatives and backtracking
+  *   - Error handling (Abort[ParseFailed]) for parse failures
+  *
+  * This combination enables building complex parsers that can handle ambiguous grammars, implement look-ahead, and provide detailed error
+  * messages.
+  */
+opaque type Parse <: (Var[Parse.State] & Choice & Abort[ParseFailed]) = Var[Parse.State] & Choice & Abort[ParseFailed]
+
+object Parse:
```
fwbrasil, line 16: The effect isn't very optimized yet. We could, for example, merge the features needed from `Choice`, `Var`, and `Abort` into a new `ArrowEffect` for better performance. I'm planning to follow up on benchmarking and optimization

kyo-prelude/shared/src/main/scala/kyo/kernel/Flat.scala
```diff
@@ -75,9 +76,10 @@ private object FlatMacro:
                 case AppliedType(base, _) =>
                     base =:= TypeRepr.of[Maybe] ||
                     base =:= TypeRepr.of[Result] ||
-                    base =:= TypeRepr.of[TypeMap] ||
-                    base =:= TypeRepr.of[Duration]
```
fwbrasil, line 79: `Duration` was incorrect here since it doesn't have type parameters and isn't an `AppliedType`

kyo-prelude/shared/src/main/scala/kyo/kernel/Flat.scala
```diff
@@ -75,9 +76,10 @@ private object FlatMacro:
...
-                case _ => false
+                    base =:= TypeRepr.of[TypeMap]
+                case _ =>
+                    t =:= TypeRepr.of[Duration] ||
+                    t =:= TypeRepr.of[Text]
```
fwbrasil, line 82: added `Text`

hearnadam (approved):
Approving as I don't see anything major. We can followup for future optimization.

PR #797, opened 2024-10-30
[core][sttp] fix timeout interruption

fwbrasil:
Easyracer is having trouble with Kyo not interrupting requests on timeout. It's a bug similar to https://github.com/getkyo/kyo/pull/771. `Async.timeout` is only propagating a completion, which isn't enough to interrupt execution now that we split regular completions from interruptions.

This PR changes `IOPromise.interrupt` and related methods to allow an `Error[E]` so it's possible to interrupt the execution with a `Result.Fail(timeoutError)` in `Async.timeout` instead of a `Panic`.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -130,7 +130,7 @@ object Async:
                     IO.Unsafe {
                         val sleepFiber = clock.unsafe.sleep(after)
                         val task       = IOTask[Ctx, E | Timeout, A](v, trace, context)
-                        sleepFiber.onComplete(_ => discard(task.complete(Result.fail(Timeout(frame)))))
+                        sleepFiber.onComplete(_ => discard(task.interrupt(Result.Fail(Timeout(frame)))))
                         task.onComplete(_ => discard(sleepFiber.interrupt()))
                         Async.get(task)
                     }
--- a/kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
+++ b/kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
@@ -3,7 +3,7 @@ package kyo.scheduler
 import IOPromise.*
 import java.util.concurrent.locks.LockSupport
 import kyo.*
-import kyo.Result.Panic
+import kyo.Result.Error
 import kyo.kernel.Safepoint
 import scala.annotation.nowarn
 import scala.annotation.tailrec
@@ -64,12 +64,12 @@ private[kyo] class IOPromise[+E, +A](init: State[E, A]) extends Safepoint.Interc
 
     final def mask(): IOPromise[E, A] =
         val p = new IOPromise[E, A]:
-            override def interrupt(error: Panic): Boolean = false
+            override def interrupt[E2 >: E](error: Error[E2]): Boolean = false
         onComplete(p.completeDiscard)
         p
     end mask
 
-    def interrupt(error: Panic): Boolean =
+    def interrupt[E2 >: E](error: Error[E2]): Boolean =
         @tailrec def interruptLoop(promise: IOPromise[E, A]): Boolean =
             promise.state match
                 case p: Pending[E, A] @unchecked =>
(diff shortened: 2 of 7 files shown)


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -130,7 +130,7 @@ object Async:
                     IO.Unsafe {
                         val sleepFiber = clock.unsafe.sleep(after)
                         val task       = IOTask[Ctx, E | Timeout, A](v, trace, context)
-                        sleepFiber.onComplete(_ => discard(task.complete(Result.fail(Timeout(frame)))))
+                        sleepFiber.onComplete(_ => discard(task.interrupt(Result.Fail(Timeout(frame)))))
```
fwbrasil, line 133: main fix is here: interrupt with the timeout failure instead of just completing the task

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
```diff
@@ -429,4 +429,43 @@ class AsyncTest extends Test:
...
+        }
+
+        "interrupts computation" in run {
+            for
+                flag  <- AtomicBoolean.init(false)
+                fiber <- Promise.init[Nothing, Int]
+                _     <- fiber.onInterrupt(_ => flag.set(true))
+                // TODO Boundary inference issue
+                v = Async.timeout(1.millis)(fiber.get)
+                result <- Async.run(v)
```
fwbrasil, line 464: I get a compilation error with `Boundary` if I inline `v` here. I'll follow up on it.

kyo-sttp/shared/src/main/scala/kyo/internal/KyoSttpMonad.scala
```diff
@@ -55,9 +55,8 @@ class KyoSttpMonad extends MonadAsyncError[M]:
                     case Left(t)  => discard(p.complete(Result.panic(t)))
                     case Right(t) => discard(p.complete(Result.success(t)))
                 }
-            p.onComplete { r =>
-                if r.isPanic then
-                    canceller.cancel()
+            p.onInterrupt { _ =>
+                canceller.cancel()
```
fwbrasil, line 59: Use the new `onInterrupt` to cancel only in case of interruptions, not regular completions.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -263,18 +263,18 @@ private[kyo] object IOPromise extends IOPromisePlatformSpecific:
...
                     self
 
         @nowarn("msg=anonymous")
-        inline def onInterrupt(inline f: Panic => Unit): Pending[E, A] =
+        inline def onInterrupt(inline f: Error[E] => Unit): Pending[E, A] =
             new Pending[E, A]:
-                def interrupt(panic: Panic): Pending[E, A] =
-                    f(panic)
+                def interrupt[E2 >: E](error: Error[E2]) =
+                    f(error.asInstanceOf[Error[E]])
```
hearnadam, line 277: Is there a reason the casting is necessary? Perhaps we can just coerce it with the `: Error[E]`?

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
hearnadam: Is it an issue with a type union?

hearnadam (approved):

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
fwbrasil: yeah, it might be that

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
fwbrasil: Just coercing isn't enough. I've just tried a few things: removing variance from `IOPromise` but keeping it in `Fiber`, updating the type parameters, etc but nothing worked. Variance is a nightmare

fwbrasil: I've created https://github.com/getkyo/kyo/issues/801

PR #800, opened 2024-10-31
[prelude][core] Frame and Log optimizations

fwbrasil:
I noticed a major performance regression in `LoggingBench`. The main cause is a bug introduced by https://github.com/getkyo/kyo/pull/783 that makes the benchmark consider `Level.error` enabled when it isn't. I fixed the bug but also worked on a few optimizations to improve performance in case the log is enabled.

Benchmark [results](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/b7686518b8168f6fcba4ab2787d0ff41/raw/29e149f7bb688672cfdad844e308cde4df999e0e/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/b7686518b8168f6fcba4ab2787d0ff41/raw/29e149f7bb688672cfdad844e308cde4df999e0e/jmh-result-candidate.json):

image

### DIFF
--- a/kyo-core/jvm/src/main/scala/kyo/internal/LogPlatformSpecific.scala
+++ b/kyo-core/jvm/src/main/scala/kyo/internal/LogPlatformSpecific.scala
@@ -17,42 +17,42 @@ object LogPlatformSpecific:
             def apply(name: String) = new SLF4J(org.slf4j.LoggerFactory.getLogger(name))
 
         class SLF4J(logger: org.slf4j.Logger) extends Log.Unsafe:
-            def level =
-                if logger.isTraceEnabled() then Level.trace
-                else if logger.isDebugEnabled() then Level.debug
-                else if logger.isInfoEnabled() then Level.info
+            val level =
+                if logger.isErrorEnabled() then Level.error
                 else if logger.isWarnEnabled() then Level.warn
-                else Level.error
+                else if logger.isInfoEnabled() then Level.info
+                else if logger.isDebugEnabled() then Level.debug
+                else Level.trace
 
             inline def trace(msg: => String)(using frame: Frame, allow: AllowUnsafe): Unit =
-                if Level.trace.enabled(level) then logger.trace(s"[${frame.parse.position}] $msg")
+                if Level.trace.enabled(level) then logger.trace(s"[${frame.position.show}] $msg")
 
             inline def trace(msg: => String, t: => Throwable)(using frame: Frame, allow: AllowUnsafe): Unit =
-                if Level.trace.enabled(level) then logger.trace(s"[${frame.parse.position}] $msg", t)
+                if Level.trace.enabled(level) then logger.trace(s"[${frame.position.show}] $msg", t)
 
             inline def debug(msg: => String)(using frame: Frame, allow: AllowUnsafe): Unit =
-                if Level.debug.enabled(level) then logger.debug(s"[${frame.parse.position}] $msg")
+                if Level.debug.enabled(level) then logger.debug(s"[${frame.position.show}] $msg")
 
             inline def debug(msg: => String, t: => Throwable)(using frame: Frame, allow: AllowUnsafe): Unit =
-                if Level.debug.enabled(level) then logger.debug(s"[${frame.parse.position}] $msg", t)
+                if Level.debug.enabled(level) then logger.debug(s"[${frame.position.show}] $msg", t)
 
             inline def info(msg: => String)(using frame: Frame, allow: AllowUnsafe): Unit =
-                if Level.info.enabled(level) then logger.info(s"[${frame.parse.position}] $msg")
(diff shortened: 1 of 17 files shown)


kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -2,7 +2,7 @@ package kyo
 
 import kyo.internal.LogPlatformSpecific
 
-final case class Log(unsafe: Log.Unsafe) extends AnyVal:
+final case class Log(unsafe: Log.Unsafe):
```
fwbrasil, line 5: I noticed several `Log` allocations in the benchmark. There are scenarios where both the compiler and the JIT fail to prevent allocation, triggering the boxing every time `Log` is used. By removing `AnyVal`, the `Log` is always allocated once when created and then reused. I'll take a look at making this change for the other effects as well.

kyo-data/shared/src/main/scala/kyo/Ansi.scala
```diff
@@ -67,18 +67,19 @@ object Ansi:
           */
         def apply(header: String, code: String, trailer: String, startLine: Int = 1): String =
             try
-                val headerLines  = if header.nonEmpty then header.split("\n") else Array.empty[String]
```
fwbrasil, line 70: A few cosmetic improvements

kyo-prelude/shared/src/main/scala/kyo/Check.scala
```diff
@@ -14,15 +14,7 @@ import kyo.kernel.*
...
-    override def getMessage() =
-        Seq(
-            "",
-            "──────────────────────────────".dim,
-            "Check failed! ".red.bold + message,
-            "──────────────────────────────".dim,
-            frame.parse.show,
-            "──────────────────────────────".dim
-        ).mkString("\n")
+    override def getMessage() = frame.render("Check failed! ".red.bold + message)
```
fwbrasil, line 17: New convenience `frame.render` for this kind of use case.

kyo-prelude/shared/src/main/scala/kyo/Local.scala
```diff
@@ -23,8 +23,8 @@ abstract class Local[A]:
       * @return
       *   An effect that produces the current value
       */
-    final def get(using Frame): A < Any =
```
fwbrasil, line 26: These weren't inlined I believe because of compilation issues since we typically inline suspension methods. I've made `State` public and was able to inline the methods.

kyo-prelude/shared/src/main/scala/kyo/kernel/Frame.scala
```diff
@@ -9,45 +9,88 @@ opaque type Frame <: AnyRef = String
...
-                arr(5)
+    extension (self: Frame)
+        private def findNextSeparator(from: Int): Int =
+            self.indexOf('|', from)
+
+        private def parseSection(start: Int, end: Int): String =
+            if end < 0 then self.substring(start)
+            else self.substring(start, end)
+
+        def position: Position = self
```
fwbrasil, line 51: Note that there's no allocation when `frame.position` is called. It's just a wrapper that provides methods to read the specific fields.

kyo-prelude/shared/src/main/scala/kyo/kernel/Frame.scala
```diff
@@ -9,45 +9,88 @@ opaque type Frame <: AnyRef = String
...
+        end methodName
+
+        def snippet: String =
+            val firstSep  = findNextSeparator(1)
+            val secondSep = findNextSeparator(firstSep + 1)
+            val start     = findNextSeparator(secondSep + 1) + 1
+            self.substring(start)
+        end snippet
+
+        def snippetShort: String = snippet.split("📍")(0).reverse.take(snippetShortMaxChars).takeWhile(_ != '\n').trim.reverse
```
fwbrasil, line 70: this expensive string transformation was being done on each benchmark iteration 😱

kyo-core/jvm/src/main/scala/kyo/internal/LogPlatformSpecific.scala
```diff
@@ -17,42 +17,42 @@ object LogPlatformSpecific:
             def apply(name: String) = new SLF4J(org.slf4j.LoggerFactory.getLogger(name))
 
         class SLF4J(logger: org.slf4j.Logger) extends Log.Unsafe:
-            def level =
```
fwbrasil, line 20: The main bug was here. The logic was inverted.

fwbrasil: I think it'd be an issue only if the SLF4J config dynamically changes at runtime. I've never used such a feature, I'm not sure if it's possible or how useful it'd be. I made it a `val` because, in theory, we'd need an `IO` suspension to read the level.

hearnadam (approved):
Nice work!

PR #802, opened 2024-10-31
[data] Text improvements

fwbrasil:
This PR adds a `Predicate` abstract class to avoid boxing on each character read since Scala 3 still doesn't specialize functions + scaladocs + a few new methods

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Text.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Text.scala
@@ -7,39 +7,111 @@ opaque type Text >: String = String | Op
 
 object Text:
 
+    /** Creates a Tag instance for Text type */
     given Tag[Text] = Tag.fromRaw("kyo.Text")
 
+    /** Creates a new Text from a String
+      *
+      * @param s
+      *   the string to convert to Text
+      * @return
+      *   a new Text instance
+      */
     def apply(s: String): Text = s
 
+    /** Returns an empty Text instance */
     def empty: Text = ""
+
+    /** Abstract class for character predicates used in Text operations */
+    abstract class Predicate:
+        /** Tests if a character matches the predicate
+          *
+          * @param char
+          *   the character to test
+          * @return
+          *   true if the character matches the predicate
+          */
+        def apply(char: Char): Boolean
+    end Predicate
 
     extension (self: Text)
 
+        /** Tests if this Text is empty
+          *
+          * @return
+          *   true if the Text contains no characters
+          */
         def isEmpty: Boolean =
             self match
                 case s: String => s.isEmpty
                 case op: Op    => op.isEmpty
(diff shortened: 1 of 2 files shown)


kyo-data/shared/src/main/scala/kyo/Text.scala
```diff
@@ -296,6 +607,24 @@ object Text:
...
+
+        case class Reverse(payload: Text) extends Op:
+            def isEmpty: Boolean         = payload.isEmpty
+            def length: Int              = payload.length
+            def charAt(index: Int): Char = payload.charAt(length - 1 - index)
+            def substring(from: Int, until: Int): Text =
+                if from >= until then Text.empty
+                else if from == 0 && until == length then this
+                else Reverse(payload.substring(length - until, length - from))
+            override def toString: String =
```
hearnadam, line 619: Should this be lazy val?

hearnadam (approved):

kyo-data/shared/src/main/scala/kyo/Text.scala
```diff
@@ -7,39 +7,111 @@ opaque type Text >: String = String | Op
...
+      * @return
+      *   a new Text instance
+      */
     def apply(s: String): Text = s
 
+    /** Returns an empty Text instance */
     def empty: Text = ""
 
+    /** Abstract class for character predicates used in Text operations */
+    abstract class Predicate:
```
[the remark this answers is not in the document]
fwbrasil: If the method is inlined, which is common in isolated benchmarking but might not be the case in real codebases, it doesn't matter if it's an interface or a class. If the method isn't inlined, a virtual dispatch (vtable stub in profiling results) is less expensive than an interface dispatch (itable stub). The JVM has to perform a linear search on the interfaces of the object before locating the method. That's one of the reasons Scala collections have poor performance. A single call to a collection type might require a linear search on tens of interfaces. If you benchmark in isolation, it's difficult to observe this, but it's quite common in real systems.

There's a tradeoff with using abstract class, though. With an interface, the compiler generates an anonymous lambda that gets expanded into a class by the LambdaMetafactory at runtime. With a class, the compiler generates the class at compile time, which increases compilation times and jar sizes.

kyo-data/shared/src/main/scala/kyo/Text.scala
fwbrasil: I'm not sure, I think most instances of `Cut` will be transient won't ever produce a `String`. For example, `Parse` operates on `Text` without having to produce intermediate strings. The extra memory footprint for the two fields that `lazy val` desugars to doesn't seem worth it.

hearnadam: Makes sense to me. I agree.

PR #803, opened 2024-10-31
[core] avoid AnyVal

fwbrasil:
Following up on https://github.com/getkyo/kyo/pull/800#discussion_r1824897284. Since Scala 3 doesn't have specialization `AnyVal` values can keep boxing on every transformation.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Adder.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Adder.scala
@@ -4,7 +4,7 @@ import java.util.concurrent.atomic as j
 
 /** A wrapper for Java's LongAdder.
   */
-final case class LongAdder private (unsafe: LongAdder.Unsafe) extends AnyVal:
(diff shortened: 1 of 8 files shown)


kyo-core/shared/src/main/scala/kyo/Adder.scala
```diff
@@ -4,7 +4,7 @@ import java.util.concurrent.atomic as j
 
 /** A wrapper for Java's LongAdder.
   */
-final case class LongAdder private (unsafe: LongAdder.Unsafe) extends AnyVal:
```
fwbrasil, line 7: I've tried `opaque type LongAdder = LongAdder.Unsafe` but had trouble with compilation issues with `inline`. It's not ideal that we'll have these wrapper allocations but we can follow up later in case we see them as relevant in profiling sessions.

hearnadam (approved):

Issue #804, opened 2024-10-31
Inference issue with nested async boundaries


fwbrasil:
The inference of `Boundary` when two operations are directly nested like `Async.run(Async.timeout(1.second)(v))` is failing to compile. Background: https://github.com/getkyo/kyo/pull/797#discussion_r1823422587

hearnadam:
I believe the issue is caused by `using` automatically propagating Boundaries between functions where the type may have changed, eg `timeout` (adds a failure type).

The fix is to derive the boundary within the function where the type already contains the expected value. The only issue is that the function must now be `inline` to expose `Ctx` type to the underlying macro.

```scala
    inline def timeout[E, A: Flat, Ctx](after: Duration)(v: => A < (Abort[E] & Async & Ctx))(
        using frame: Frame
    ): A < (Abort[E | Timeout] & Async & Ctx) =
        if !after.isFinite then v
        else
            summon[Boundary[Ctx, Async & Abort[E | Timeout]]] { (trace, context) =>
                Clock.use { clock =>
                    IO.Unsafe {
                        val sleepFiber = clock.unsafe.sleep(after)
                        val task       = IOTask[Ctx, E | Timeout, A](v, trace, context)
                        sleepFiber.onComplete(_ => discard(task.complete(Result.fail(Timeout(frame)))))
                        task.onComplete(_ => discard(sleepFiber.interrupt()))
                        Async.get(task)
                    }
                }
            }
    end timeout
```

hearnadam:
/bounty $75 for a fix that doesn't require inlining the full body of each function

fwbrasil:
already fixed

PR #805, opened 2024-10-31
[core] reproduce issue with nested boundaries + provide better error messages

fwbrasil:
I couldn't find a fix for https://github.com/getkyo/kyo/issues/804. We might need to restructure how we encode boundaries to support it. This PR adds pending tests and better error reporting in case of issues to infer the boundary.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/kernel/Boundary.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/kernel/Boundary.scala
@@ -1,9 +1,24 @@
 package kyo.kernel
 
 import internal.*
+import kyo.Ansi.*
+import scala.annotation.implicitNotFound
 import scala.annotation.nowarn
 import scala.quoted.*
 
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo/kernel/Boundary.scala
```diff
@@ -1,9 +1,24 @@
...
+
+Note: If you're seeing this error with nested Async operations (like Async.run(Async.run(v))),
+this is due to a current limitation. As a workaround, you can break them into separate statements:
+
+Instead of:
+  Async.run(Async.run(v))
+
+Use:
+  val x = Async.run(v)
+  Async.run(x)
```
fwbrasil, line 20: The macro doesn't even get invoked and the compiler complains of not being able to locate the implicit for the case  where boundaries are nested.

hearnadam (approved):

PR #807, opened 2024-11-01
fix: ScalaJS's KyoApp

### LINKED ISSUE #784: ScalaJS with KyoApp
I tried starting a scalajs project with `KyoApp`, but code as simple as below doesn't work:
```scala
import kyo.*

object ScalaJsApp extends KyoApp {
  run(for {
    _ <- println("1")
    _ <- println("2")
  } yield {})
}
```
running that in browser shows this error after printing `1`:
```ERROR kyo.logs -- [IOPromise.scala:212:21] uncaught exception java.lang.UnsupportedOperationException: fiber.block is not supported in ScalaJS``` and it never prints `2`
originally I was at `0.12.0`, the same also happens on `0.13.0`

updating to `0.13.1` will see another error at linking time, complaining about ```Referring to non-existent method constructor java.util.concurrent.ThreadFactory.<init>()void```

Is `KyoApp` intended for use in scalajs? Am I missing something? please help

### PR DESCRIPTION
**Pull request by HollandDM**
resolve #784

### DIFF
--- a/kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
+++ b/kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
@@ -1,0 +1,26 @@
+package kyo
+
+import kyo.Maybe.Absent
+import kyo.Maybe.Present
+
+abstract class KyoAppPlatformSpecific extends KyoApp.Base[Async & Resource & Abort[Throwable]]:
+
+    private var maybePreviousAsync: Maybe[Unit < (Async & Abort[Throwable])] = Absent
+
+    final override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+        import AllowUnsafe.embrace.danger
+        val currentAsync: Unit < (Async & Abort[Throwable]) =
+            Abort.run(handle(v)).map(result => IO(printResult(result)).andThen(Abort.get(result)).unit)
+        maybePreviousAsync = maybePreviousAsync match
+            case Absent                 => Present(currentAsync)
+            case Present(previousAsync) => Present(previousAsync.map(_ => currentAsync))
+        initCode = maybePreviousAsync.map { previousAsync => () =>
+            val racedAsyncIO = Clock.repeatWithDelay(1.hour)(()).map { fiber =>
+                val race = Async.race(Seq(fiber.get, previousAsync))
+                Async.timeout(timeout)(race)
+            }
+            val _ = IO.Unsafe.run(Async.run(racedAsyncIO)).eval
+        }.toList
+    end run
+
+end KyoAppPlatformSpecific
--- a/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
+++ b/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
@@ -1,7 +1,5 @@
 package kyo
 
-import scala.collection.mutable.ListBuffer
-
 /** An abstract base class for Kyo applications.
   *
   * This class provides a foundation for building applications using the Kyo framework, with built-in support for logging, random number
@@ -9,14 +7,11 @@ import scala.collection.mutable.ListBuffer
   *
   * Note: This class and its methods are unsafe and should only be used as the entrypoint of an application.
   */
-abstract class KyoApp extends KyoApp.Base[Async & Resource & Abort[Throwable]]:
-    def timeout: Duration = Duration.Infinity
+abstract class KyoApp extends KyoAppPlatformSpecific:
+    override def timeout: Duration = Duration.Infinity
 
-    override protected def handle[A: Flat](v: A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
-        import AllowUnsafe.embrace.danger
-        val result = KyoApp.Unsafe.runAndBlock(timeout)(v)
-        if !result.exists(().equals(_)) then
-            println(pprint.apply(result).plainText)
+    override protected def handle[A: Flat](v: A < (Async & Resource & Abort[Throwable]))(using Frame): A < (Async & Abort[Throwable]) =
+        Resource.run(v)
     end handle
 end KyoApp
 
@@ -30,21 +25,26 @@ object KyoApp:
       *   The effect type used by the application.
       */
     abstract class Base[S]:
+        protected def timeout: Duration
 
-        protected def handle[A: Flat](v: A < S)(using Frame): Unit
+        protected def handle[A: Flat](v: A < S)(using Frame): A < (Async & Abort[Throwable])
 
         final protected def args: Array[String] = _args
 
         private var _args: Array[String] = null
 
-        private val initCode = new ListBuffer[() => Unit]
+        protected var initCode: List[() => Unit] = List.empty
 
         final def main(args: Array[String]) =
             this._args = args
             for proc <- initCode do proc()
+        end main
 
-        protected def run[A: Flat](v: => A < S)(using Frame): Unit =
-            initCode += (() => handle(v))
+        protected def run[A: Flat](v: => A < S)(using Frame): Unit
+
+        protected def printResult(result: Result[Any, Any]): Unit =
+            if !result.exists(().equals(_)) then println(pprint.apply(result).plainText)
+        end printResult
     end Base
 
     /** WARNING: Low-level API meant for integrations, libraries, and performance-sensitive code. See AllowUnsafe for more details. */
--- a/kyo-core/shared/src/test/scala/kyo/KyoAppTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/KyoAppTest.scala
@@ -1,6 +1,13 @@
 package kyo
 
 import Tagged.*
+import kyo.Clock.Deadline
+import kyo.Clock.Stopwatch
+import kyo.Clock.Unsafe
+import kyo.internal.LayerMacros.Validated.succeed
+import org.scalatest.compatible.Assertion
+import scala.collection.mutable.ListBuffer
+import scala.util.Try
 
 class KyoAppTest extends Test:
 
@@ -27,7 +34,18 @@ class KyoAppTest extends Test:
             runs <- ref.get
         yield assert(runs == 3)
     }
-    "effects" taggedAs jvmOnly in {
+    "ordered runs" in {
+        val x       = new ListBuffer[Int]
+        val promise = scala.concurrent.Promise[Assertion]()
+        val app = new KyoApp:
+            run { Async.delay(10.millis)(IO(x += 1)) }
+            run { Async.delay(10.millis)(IO(x += 2)) }
+            run { Async.delay(10.millis)(IO(x += 3)) }
+            run { IO(promise.complete(Try(assert(x.toList == List(1, 2, 3))))) }
+        app.main(Array.empty)
+        promise.future
+    }
+    "effects in JVM" taggedAs jvmOnly in {
         def run: Int < (Async & Resource & Abort[Throwable]) =
             for
                 _ <- Clock.repeatAtInterval(1.second, 1.second)(())
@@ -35,11 +53,27 @@ class KyoAppTest extends Test:
                 _ <- Console.printLine(s"$i")
                 _ <- Clock.now
                 _ <- Resource.ensure(())
-                _ <- Async.run(())
+                _ <- Async.sleep(1.second)
             yield 1
 
         import AllowUnsafe.embrace.danger
         assert(KyoApp.Unsafe.runAndBlock(Duration.Infinity)(run) == Result.success(1))
+    }
+    "effects in JS" taggedAs jsOnly in {
+        val promise = scala.concurrent.Promise[Assertion]()
+        val app = new KyoApp:
+            run {
+                for
+                    _ <- Clock.repeatAtInterval(1.second, 1.second)(())
+                    i <- Random.nextInt
+                    _ <- Console.printLine(s"$i")
+                    _ <- Clock.now
+                    _ <- Resource.ensure(())
+                    _ <- Async.sleep(1.second)
+                yield promise.complete(Try(succeed))
+            }
+        app.main(Array.empty)
+        promise.future
     }
     "failing effects" taggedAs jvmOnly in {
         def run: Unit < (Async & Resource & Abort[Throwable]) =
--- a/kyo-core/jvm-native/src/main/scala/kyo/KyoAppPlatformSpecific.scala
+++ b/kyo-core/jvm-native/src/main/scala/kyo/KyoAppPlatformSpecific.scala
@@ -1,0 +1,15 @@
+package kyo
+
+import scala.collection.mutable.ListBuffer
+
+abstract class KyoAppPlatformSpecific extends KyoApp.Base[Async & Resource & Abort[Throwable]]:
+
+    final override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+        import AllowUnsafe.embrace.danger
+        initCode = initCode.appended(() =>
+            val result = IO.Unsafe.run(Abort.run(Async.runAndBlock(timeout)(handle(v)))).eval
+            printResult(result)
+        )
+    end run
+
+end KyoAppPlatformSpecific
--- a/kyo-scheduler/js/src/main/scala/java/util/concurrent/ThreadFactory.scala
+++ b/kyo-scheduler/js/src/main/scala/java/util/concurrent/ThreadFactory.scala
@@ -1,7 +1,0 @@
-package java.util.concurrent
-
-class ThreadFactory
-
-object ThreadFactoryStub {
-    val get = new ThreadFactory
-}
--- a/kyo-scheduler/js/src/main/scala/kyo/scheduler/util/Threads.scala
+++ b/kyo-scheduler/js/src/main/scala/kyo/scheduler/util/Threads.scala
@@ -1,13 +1,15 @@
 package kyo.scheduler.util
 
+import java.lang.Thread
 import java.util.concurrent.ThreadFactory
-import java.util.concurrent.ThreadFactoryStub
 
 object Threads {
 
-    def apply(name: String): ThreadFactory =
-        ThreadFactoryStub.get
+    def apply(name: String): ThreadFactory = apply(name, _ => Thread.currentThread())
 
     def apply(name: String, create: Runnable => Thread): ThreadFactory =
-        ThreadFactoryStub.get
+        new ThreadFactory {
+            override def newThread(runnable: Runnable): Thread = create(runnable)
+        }
+
 }

### CODE REVIEW

**Review by hearnadam — COMMENTED** [review]
Thanks for taking this on!

File: kyo-core/js/src/main/scala/kyo/KyoAppBasePlatformSpecific.scala [hunk 1]
```diff
@@ -0,0 +1,42 @@
+package kyo
+
+abstract class KyoAppBasePlatformSpecific extends KyoApp.Base[Throwable, Async & Resource]:
```
fwbrasil, line 3: Any specific reason why the error is in a separate type parameter? Could it be a single type param `Abort[Throwable] & Async & Resource`?

kyo-core/js/src/main/scala/kyo/KyoAppBasePlatformSpecific.scala
```diff
@@ -0,0 +1,42 @@
...
+        Resource.run(v)
+
+    final override protected def args: Array[String] = _args
+
+    private var _args: Array[String] = null
+
+    private var concatenatedAsync: Any < (Async & Abort[Throwable]) = scala.compiletime.uninitialized
+    private var initCode: () => Unit                                = scala.compiletime.uninitialized
+
+    private def keepAlive(using Frame): Unit < (Async & Abort[Throwable]) = Async.sleep(timeout).map(_ => keepAlive)
```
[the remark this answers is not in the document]
fwbrasil: It'd be good to make the keep alive configurable as a method that can be overriden like `timeout`. There are new APIs in `Clock` for scheduled execution as well: 

```scala
val keepAliveTask = Clock.repeatWithDelay(keepAliveInterval)(())
```

It returns a `Fiber`, I think we'd also need to interrupt it once the execution has finished.

kyo-core/js/src/main/scala/kyo/KyoAppBasePlatformSpecific.scala
```diff
@@ -0,0 +1,42 @@
...
+
+    protected def handle[A: Flat](v: A < (Async & Resource & Abort[Throwable]))(using Frame): A < (Async & Abort[Throwable]) =
+        Resource.run(v)
+
+    final override protected def args: Array[String] = _args
+
+    private var _args: Array[String] = null
+
+    private var concatenatedAsync: Any < (Async & Abort[Throwable]) = scala.compiletime.uninitialized
+    private var initCode: () => Unit                                = scala.compiletime.uninitialized
```
fwbrasil, line 14: Can you use `Maybe` to avoid `null`?

kyo-core/js/src/main/scala/kyo/KyoAppBasePlatformSpecific.scala
```diff
@@ -0,0 +1,42 @@
...
+        if null eq concatenatedAsync.asInstanceOf[Object] then
+            concatenatedAsync = handle(v)
+        else
+            concatenatedAsync = concatenatedAsync.map(_ => handle(v))
+        end if
+
+        initCode = (
+            () =>
+                if null ne concatenatedAsync.asInstanceOf[Object] then
+                    val finalAsync = Async.race(concatenatedAsync.map(_ => ()), keepAlive)
```
fwbrasil, line 34: Nice, I see now that `keepAlive` will be interrupted by the race

kyo-core/js/src/main/scala/kyo/KyoAppBasePlatformSpecific.scala
```diff
@@ -0,0 +1,42 @@
...
+    private var _args: Array[String] = null
+
+    private var concatenatedAsync: Any < (Async & Abort[Throwable]) = scala.compiletime.uninitialized
+    private var initCode: () => Unit                                = scala.compiletime.uninitialized
+
+    private def keepAlive(using Frame): Unit < (Async & Abort[Throwable]) = Async.sleep(timeout).map(_ => keepAlive)
+
+    final def main(args: Array[String]) =
+        this._args = args
+        if initCode ne null then initCode()
```
fwbrasil, line 20: I think we could throw an exception in case the `App` has no `run` calls. I imagine some users might try to define `def run` like in other effect systems and it'd just not execute anything silently.

kyo-core/js/src/main/scala/kyo/KyoAppBasePlatformSpecific.scala
```diff
@@ -0,0 +1,42 @@
...
+
+    override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+        import AllowUnsafe.embrace.danger
+
+        if null eq concatenatedAsync.asInstanceOf[Object] then
+            concatenatedAsync = handle(v)
+        else
+            concatenatedAsync = concatenatedAsync.map(_ => handle(v))
```
fwbrasil, line 28: I like this workaround to chain `run` calls!

kyo-core/jvm/src/main/scala/kyo/KyoAppBasePlatformSpecific.scala
```diff
@@ -0,0 +1,28 @@
...
+
+    final def main(args: Array[String]) =
+        this._args = args
+        for proc <- initCode do proc()
+
+    override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+        import AllowUnsafe.embrace.danger
+        initCode += (() =>
+            val result = IO.Unsafe.run(Abort.run(Async.runAndBlock(timeout)(handle(v)))).eval
+            if !result.exists(().equals(_)) then println(pprint.apply(result).plainText)
```
fwbrasil, line 25: `!result.contains(())`

kyo-core/jvm/src/main/scala/kyo/KyoAppBasePlatformSpecific.scala
```diff
@@ -0,0 +1,28 @@
+package kyo
+
+import scala.collection.mutable.ListBuffer
+
+abstract class KyoAppBasePlatformSpecific extends KyoApp.Base[Throwable, Async & Resource]:
+    override def timeout: Duration = Duration.Infinity
+
+    protected def handle[A: Flat](v: A < (Async & Resource & Abort[Throwable]))(using Frame): A < (Async & Abort[Throwable]) =
```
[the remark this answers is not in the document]
HollandDM: Yes, I’ll update this

kyo-core/js/src/main/scala/kyo/KyoAppBasePlatformSpecific.scala
HollandDM: Yes, I think its because of my initial attempt in which handle initially returned Fiber[E, A] < IO, so I encode both E and A on the signature. It should just encodes A and lock E to Throwable

kyo-core/js/src/main/scala/kyo/KyoAppBasePlatformSpecific.scala
```diff
@@ -0,0 +1,42 @@
...
+        end if
+
+        initCode = (
+            () =>
+                if null ne concatenatedAsync.asInstanceOf[Object] then
+                    val finalAsync = Async.race(concatenatedAsync.map(_ => ()), keepAlive)
+                    val _          = IO.Unsafe.run(Async.run(finalAsync)).eval
+                else
+                    (
+                )
```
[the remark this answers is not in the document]
HollandDM: In JVM we have blocking, so initCode can be sequentially run. In JS we can’t block, so if we run them like we did in JVM, they’ll be spawned as JS promise by the runtime, hence the order of execution is not guaranteed. That why in JS code I concatenate the initCode into a single one.
I didn’t do this concatenation in main method because it requires Frames implicitly, and I was not sure if I want to use Frames.internal or not

kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
```diff
@@ -0,0 +1,41 @@
...
+    final override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+        import AllowUnsafe.embrace.danger
+        val currentFiberIO =
+            for
+                currentFiber <- handle(v).flatMap(_.mapResult { result =>
+                    IO(Result(if !result.exists(().equals(_)) then println(pprint.apply(result).plainText)))
+                })
+                sleepFiber <- Async.run(Async.sleep(timeout))
+                raceFiber  <- Fiber.race(Seq(currentFiber.get, sleepFiber.get))
+            yield raceFiber
```
HollandDM, line 19: I try to get the same behavior on the JVM side, by racing this with a `Async.sleep`

kyo-core/jvm/src/main/scala/kyo/KyoAppPlatformSpecific.scala
```diff
@@ -0,0 +1,25 @@
...
+    private val procListBuffer = new ListBuffer[() => Unit]
+
+    final override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+        import AllowUnsafe.embrace.danger
+        procListBuffer += (() =>
+            val resultEffect = handle(v).map { fiber =>
+                fiber.block(timeout).map(Abort.get(_))
+            }
+            val result = IO.Unsafe.run(Abort.run(resultEffect)).eval
+            if !result.exists(().equals(_)) then println(pprint.apply(result).plainText)
```
HollandDM, line 16: Compiler will complain about strict equality of `A | Unit` if we use `!result.contains(_)` here, i guess that why it's been like this from the start.

kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
```diff
@@ -0,0 +1,41 @@
...
+            maybeProc = Present(() =>
+                previousFiberIO match
+                    case Present(fiberIO) =>
+                        val combinedFiberIO =
+                            for
+                                keepAliveFiber <- Clock.repeatWithDelay(1.hours)(())
+                                fiber          <- fiberIO
+                                combineFiber   <- Fiber.race(Seq(keepAliveFiber.get, fiber.get))
+                            yield combineFiber
+                        val _ = IO.Unsafe.run(Async.run(combinedFiberIO.get)).eval
```
hearnadam, line 35: I don't think you need to use `Async.run` or `combinedFiberIO.get` but I could be wrong.

kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
```diff
@@ -0,0 +1,41 @@
...
+
+abstract class KyoAppPlatformSpecific extends KyoApp.Base[Async & Resource & Abort[Throwable]]:
+
+    private var previousFiberIO: Maybe[Fiber[Throwable, Unit] < IO] = Absent
+
+    final override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+        import AllowUnsafe.embrace.danger
+        val currentFiberIO =
+            for
+                currentFiber <- handle(v).flatMap(_.mapResult { result =>
```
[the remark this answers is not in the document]
hearnadam: I think that's probably a good thing. If a user cancels a run, they will want to cancel all.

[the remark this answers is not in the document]
hearnadam: You can probably also use `.unsafe.onComplete(printResult)` to avoid needing the `IO(Result(...`

kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
```diff
@@ -0,0 +1,41 @@
...
+                    IO(Result(if !result.exists(().equals(_)) then println(pprint.apply(result).plainText)))
+                })
+                sleepFiber <- Async.run(Async.sleep(timeout))
+                raceFiber  <- Fiber.race(Seq(currentFiber.get, sleepFiber.get))
```
hearnadam, line 18: `Async.timeout`?

hearnadam (commented):
Do you mind adding a test with multiple `run`s if we don't have one already?

I think it can be as simple as building a list of runs

```
val x = ListBuffer...
object TestApp extends KyoApp:
  run { IO(x += 1) }
  run { IO(x += 2) }
TestApp.main(Array.empty)
assert(x == Seq(1, 2, 3....)
```

kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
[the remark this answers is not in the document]
HollandDM: `.unsafe.onComplete(printResult)` gives me `Unit < IO`, but i need `Fiber[Throwable, Unit] < IO` here

kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
HollandDM: most of these back and forth `fiber.get` and `Async.run` are just to convert between `Fiber[E, A]` and `A < (Async & Abort[E])` to use the method. I feel like some APIs can be update here:
- `Fiber.race` should receive `Fiber` seq, currently it's receiving `A < (Async & Abort[E])`
- There is no `timeout` or `sleep` for `Fiber`, only available now on `A < (Async & Abort[E])`

HollandDM:
Updated the code a bit, and added a small test for KyoApp in js

kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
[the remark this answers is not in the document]
HollandDM: I changed the code to use Async, which is friendlier IMO

HollandDM:
Updated, fix and add one more test for KyoApp on JS

HollandDM:
Updated, squash commits

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -30,21 +25,28 @@ object KyoApp:
...
-        private val initCode = new ListBuffer[() => Unit]
+        protected var maybeProc: Maybe[() => Unit] = Absent
 
         final def main(args: Array[String]) =
             this._args = args
-            for proc <- initCode do proc()
+            maybeProc match
+                case Present(proc) => proc()
+                case Absent        => ()
+        end main
```
hearnadam, line 43:
```suggestion
            for proc <- initCode do proc()
```

kyo-core/shared/src/test/scala/kyo/KyoAppTest.scala
```diff
@@ -30,6 +34,17 @@ class KyoAppTest extends Test:
             runs <- ref.get
         yield assert(runs == 3)
     }
+    "multiple runs on JS" taggedAs jsOnly in {
```
hearnadam, line 37: This test is necessary for JVM and JS. We can delete the other one.

```suggestion
    "ordered runs" in {
```

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -30,21 +25,28 @@ object KyoApp:
...
 
-        protected def handle[A: Flat](v: A < S)(using Frame): Unit
+        protected def handle[A: Flat](v: A < S)(using Frame): A < (Async & Abort[Throwable])
 
         final protected def args: Array[String] = _args
 
         private var _args: Array[String] = null
 
-        private val initCode = new ListBuffer[() => Unit]
+        protected var maybeProc: Maybe[() => Unit] = Absent
```
hearnadam, line 36: Can this be `Iterable[() => Unit]`? Or `IterableOnce`? I think the changes in this PR break multiple runs on JVM.

kyo-core/jvm/src/main/scala/kyo/KyoAppPlatformSpecific.scala
```diff
@@ -0,0 +1,22 @@
...
+
+    private val procListBuffer = new ListBuffer[() => Unit]
+
+    final override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+        import AllowUnsafe.embrace.danger
+        procListBuffer += (() =>
+            val result = IO.Unsafe.run(Abort.run(Async.runAndBlock(timeout)(handle(v)))).eval
+            printResult(result)
+        )
+        if maybeProc.isEmpty then
```
hearnadam, line 15: Won't this only be `true` once? I think we can make the `maybeProc` a generic Iterable:

`private var initCode: Iterable[() => Unit] = Nil`

Then we can just always update it. Both versions should be able to conform to this signature.

Scala JS:
`initCode = maybeProc.toList`
jvm
`initCode = procListBuffer.toList`

HollandDM: I think I understand what you're trying to say now, I'll try to update this.

kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
```diff
@@ -0,0 +1,30 @@
...
+
+    private var maybePreviousAsync: Maybe[Unit < (Async & Abort[Throwable])] = Absent
+
+    final override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+        import AllowUnsafe.embrace.danger
+        val currentAsync: Unit < (Async & Abort[Throwable]) =
+            Async.timeout(timeout)(handle(v)).map(value => printResult(Result.success(value)))
+        maybePreviousAsync = maybePreviousAsync match
+            case Absent                 => Present(currentAsync)
+            case Present(previousAsync) => Present(previousAsync.map(_ => currentAsync))
```
HollandDM, line 16: updated

kyo-core/jvm/src/main/scala/kyo/KyoAppPlatformSpecific.scala
```diff
@@ -0,0 +1,15 @@
+package kyo
+
+import scala.collection.mutable.ListBuffer
+
+abstract class KyoAppPlatformSpecific extends KyoApp.Base[Async & Resource & Abort[Throwable]]:
+
+    final override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+        import AllowUnsafe.embrace.danger
+        initCode += (() =>
```
HollandDM, line 9: JVM side, code is pretty much the same

kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
```diff
@@ -0,0 +1,30 @@
...
+import kyo.Maybe.Present
+
+abstract class KyoAppPlatformSpecific extends KyoApp.Base[Async & Resource & Abort[Throwable]]:
+
+    private var maybePreviousAsync: Maybe[Unit < (Async & Abort[Throwable])] = Absent
+
+    final override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+        import AllowUnsafe.embrace.danger
+        val currentAsync: Unit < (Async & Abort[Throwable]) =
+            Async.timeout(timeout)(handle(v)).map(value => printResult(Result.success(value)))
```
hearnadam, line 13: If you put the `timeout` within the initiation of the Unit function, (when you update `initCode`) you will avoid some sleep fibers and potentially racy behavior.

I think it makes sense to timeout the whole application anyways.

kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
```diff
@@ -0,0 +1,30 @@
...
+        maybePreviousAsync = maybePreviousAsync match
+            case Absent                 => Present(currentAsync)
+            case Present(previousAsync) => Present(previousAsync.map(_ => currentAsync))
+        if initCode.isEmpty then
```
hearnadam, line 17: I was initially wondering **why** this would work, when only the first fiber would be initialized, but then I realized the first fiber would begin the execution of the next...

In my testing, I think this code also causes `repeatWithDelay` to run forever. The `run` fiber is already complete when the `timeout` expires.

Regardless, I re-wrote this a bit which I think is a simplification:

```scala
package kyo

import kyo.Maybe.Absent
import kyo.Maybe.Present

abstract class KyoAppPlatformSpecific extends KyoApp.Base[Async & Resource & Abort[Throwable]]:

    private var previous: Maybe[Unit < (Async & Abort[Throwable])] = Absent

    final override protected def run[A: Flat](v: => A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
        val current: Unit < (Async & Abort[Throwable]) =
            Abort.run(handle(v)).map(result => IO(printResult(result)).andThen(Abort.get(result)).unit)

        previous = previous match
            case Absent            => Present(current)
            case Present(p) => Present(p.map(_ => current))

        initCode = previous.map { async => () =>
            val raced = Clock.repeatWithDelay(1.hour)(()).map { fiber =>
                val race = Async.race(fiber.get, async)
                Async.timeout(timeout)(race)
            }
            val _ = IO.Unsafe.run(Async.run(raced)).eval
        }.toList

    end run

end KyoAppPlatformSpecific
```

In `KyoApp` I made `initCode` a variable with a List for ease of conversion with Maybe.
```scala
protected var initCode: List[() => Unit] = List.empty
```

You can check if the Clock fiber is still running by changing `()` to a `Console.println` and increase the frequency of heartbeat.

kyo-core/shared/src/test/scala/kyo/KyoAppTest.scala
```diff
@@ -30,20 +34,47 @@ class KyoAppTest extends Test:
...
         yield assert(runs == 3)
     }
-    "effects" taggedAs jvmOnly in {
+    "orderred runs" in {
+        val x       = new ListBuffer[Int]
+        val promise = scala.concurrent.Promise[Assertion]()
+        val app = new KyoApp:
+            run { IO(x += 1) }
+            run { IO(x += 2) }
+            run { IO(x += 3) }
```
hearnadam, line 43: Let's mix in `Async` just for posterity:

```scala
            run { Abort.delay(10.millis)(IO(x += 1)) }
            run { Async.delay(10.millis)(IO(x += 2)) }
            run { Async.delay(10.millis)(IO(x += 3)) }
```

kyo-core/js/src/main/scala/kyo/KyoAppPlatformSpecific.scala
HollandDM: I mimiced the code in JVM (timeout for each `run`s). Will updated this

fwbrasil:
folks, anything pending to merge this? We can follow up with improvements if needed. It'd be nice to get the issue fixed.

HollandDM:
I'll try to finalize this in the upcoming days, I think there aer only some little code conventions/styles left to adjust

HollandDM:
@fwbrasil @hearnadam I've updated the code, it should be good now

hearnadam (approved):

kyo-core/jvm/src/main/scala/kyo/KyoAppPlatformSpecific.scala
```diff
@@ -0,0 +1,15 @@
+package kyo
+
+import scala.collection.mutable.ListBuffer
+
+abstract class KyoAppPlatformSpecific extends KyoApp.Base[Async & Resource & Abort[Throwable]]:
```
fwbrasil, line 5: The build is failing in Native. I think it's just a matter of moving this file to `kyo-core/jvm-native/src/`

HollandDM: updated and tested, should work now

hearnadam (approved):

PR #809, opened 2024-11-03
[core] fix flaky fiber tests

fwbrasil:
I noticed the tests failing in CI

### DIFF
--- a/kyo-core/shared/src/test/scala/kyo/FiberTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/FiberTest.scala
@@ -137,25 +137,21 @@ class FiberTest extends Test:
         "interrupts losers" - {
             "promise + plain value" in runJVM {
                 for
-                    flag        <- AtomicBoolean.init(false)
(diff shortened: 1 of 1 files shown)


kyo-core/shared/src/test/scala/kyo/FiberTest.scala
```diff
@@ -137,25 +137,21 @@ class FiberTest extends Test:
         "interrupts losers" - {
             "promise + plain value" in runJVM {
                 for
-                    flag        <- AtomicBoolean.init(false)
-                    promise     <- Promise.init[Nothing, Int]
-                    _           <- promise.onInterrupt(_ => flag.set(true))
-                    result      <- Async.race(promise.get, 42)
-                    interrupted <- flag.get
```
fwbrasil, line 144: There was no guarantee that the atomic set by the `onInterrupt` function is already executed before the flag is read. The fix uses a latch to park execution and wait for the `onInterrupt` function to execute.

hearnadam (approved):

PR #811, opened 2024-11-03
[core] provide init with default values for atomics

fwbrasil:


kyo-core/shared/src/main/scala/kyo/Atomic.scala
```diff
@@ -106,7 +113,9 @@ object AtomicInt:
     object Unsafe:
         given Flat[Unsafe] = Flat.unsafe.bypass
 
-        def init(v: Int)(using allow: AllowUnsafe): Unsafe = new j.AtomicInteger(v)
+        def init(using AllowUnsafe): Unsafe = init(0)
```
hearnadam, line 116: I think it's probably less useful for Unsafe apis...

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/Atomic.scala
fwbrasil: yeah but I use them often in Kyo's codebase, it'll be helpful to have this method

PR #814, opened 2024-11-04
[prelude] Parse streaming

fwbrasil:


kyo-prelude/shared/src/main/scala/kyo/Emit.scala
```diff
@@ -109,21 +109,22 @@ object Emit:
...
-          *   The folding function
+          *   The folding function that takes the current accumulator and emitted value, and returns a tuple of the new accumulator and an Ack
+          *   to control further emissions
           * @param v
           *   The computation with Emit effect
           * @return
           *   A tuple of the final accumulator value and the result of the computation
           */
-        def apply[A, S, B: Flat, S2](acc: A)(f: (A, V) => A < S)(v: B < (Emit[V] & S2))(
+        def apply[A, S, B: Flat, S2](acc: A)(f: (A, V) => (A, Ack) < S)(v: B < (Emit[V] & S2))(
```
fwbrasil, line 119: I've updated this method to be more generic and allow the function to determine if the stream should continue.

kyo-prelude/shared/src/test/scala/kyo/EmitTest.scala
```diff
@@ -199,6 +199,83 @@ class EmitTest extends Test:
         }
     }
 
+    "runFold" - {
```
fwbrasil, line 202: These methods were missing test coverage.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -248,6 +249,70 @@ object Parse:
...
+      * @tparam A
+      *   Type of parsed result
+      * @tparam S
+      *   Effects required by input stream
+      * @tparam S2
+      *   Effects required by parser
+      * @return
+      *   Stream of successfully parsed results, which can abort with ParseFailed
+      */
+    def run[A: Flat, S, S2](input: Stream[Text, S])(v: A < (Parse & S2))(
```
fwbrasil, line 268: This initial implementation isn't very efficient. Instead of always restarting the parser execution from start, it could use a suspension to "pause" the parsing where more input is needed so intermediate parsings aren't discarded. We'll need a custom `ArrowEffect` to optimize it.

kyo-prelude/shared/src/test/scala/kyo/ParseTest.scala
```diff
@@ -559,4 +559,207 @@ class ParseTest extends Test:
...
+                    _         <- Parse.attempt(Parse.char(separator))
+                yield r
+            val input = Stream.init(Seq[Text]("1", "2 3", "4 5"))
+            Env.run(' ')(Parse.run(input)(parser).run).map { result =>
+                assert(result == Chunk(12, 34, 5))
+            }
+        }
+    }
+
+    "ParseFailed" - {
```
fwbrasil, line 739: I had some trouble with SBT reporting that `ParseFailed` isn't serializable. I'm not sure why since this test shows it is.

fwbrasil: I've removed the test since it was making the JS build fail.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
fwbrasil: yeah, makes sense. I can play with it when optimizing the effect.

PR #815, opened 2024-11-05
[core] include Abort[Nothing] in Async

fwbrasil:
`Async` already handles `Abort[Nothing]` but the effect wasn't included in the type intersection.

### DIFF
--- a/kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
+++ b/kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
@@ -114,9 +114,8 @@ extension (kyoObject: Kyo.type)
         using
         flat: Flat[A1],
         boundary: Boundary[Ctx, Async & Abort[E]],
-        reduce: Reducible[Abort[E]],
         frame: Frame
-    ): Seq[A1] < (reduce.SReduced & Async & Ctx) =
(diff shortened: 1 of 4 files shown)


sideeffffect (commented):
`Abort[Nothing]` corresponds to a fatal, unexpected failure/defect/bug, right?

In that case, shouldn't it already be subsumed by `IO` itself?

hearnadam (approved):

hearnadam:
@fwbrasil after checking the build, it looks like this is breaking.

fwbrasil:
> `Abort[Nothing]` corresponds to a fatal, unexpected failure/defect/bug, right?

Yes, it represents a `Result.Panic`. 

> In that case, shouldn't it already be subsumed by `IO` itself?

That's a good point. It seems it'd be more convenient but then `IO.Unsafe.run` will need to return a `Result`. Do you mind creating an issue for it so we can follow up separately? It'll be trickier to implement since `IO` doesn't handle `Abort` currently.

fwbrasil:
Build should be fixed now. I hadn't realized that the `Reducible` evidences aren't necessary anymore since `Abort[Nothing]` is included in `Async`.

sideeffffect:
created the new issues for this
* https://github.com/getkyo/kyo/issues/836

PR #816, opened 2024-11-05
[prelude] Introduce Emit.Ack.next and missing Flat evidences

fwbrasil:


kyo-prelude/shared/src/main/scala/kyo/Emit.scala
```diff
@@ -22,8 +22,10 @@ object Emit:
       */
     opaque type Ack = Int
     object Ack:
-        given CanEqual[Ack, Ack] = CanEqual.derived
-        inline given Flat[Ack]   = Flat.unsafe.bypass
+        given CanEqual[Ack, Ack]         = CanEqual.derived
+        inline given Flat[Ack]           = Flat.unsafe.bypass
+        inline given Flat[Ack.Stop.type] = Flat.unsafe.bypass
+        inline given Flat[Ack.Continue]  = Flat.unsafe.bypass
```
fwbrasil, line 28: Since `Flat` is invariant, we have to provide explicit evidences for sub-classes. I've also tried making `Flat` covariant but I had trouble with the compiler inferring the wrong necessary evidence in a few scenarios, especially with `Async` operations.

kyo-prelude/shared/src/main/scala/kyo/Emit.scala
```diff
@@ -46,7 +48,25 @@ object Emit:
...
+              *   The function to execute to get the next acknowledgement
+              * @return
+              *   [[Stop]] if current acknowledgement is [[Stop]], otherwise the result of `f`
+              */
+            inline def next[S](inline f: => Ack < S): Ack < S =
+                if isStop then Stop
+                else f
+
+            // Workaround for compiler issue with inlined `next`
+            private def isStop: Boolean = self == Stop
```
hearnadam, line 68:
```suggestion
            private def stop: Boolean = self == Stop
```

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/Emit.scala
```diff
@@ -46,7 +48,25 @@ object Emit:
...
+              * If the current acknowledgement is [[Stop]], returns [[Stop]] immediately. Otherwise, executes the provided function to get
+              * the next acknowledgement.
+              *
+              * @param f
+              *   The function to execute to get the next acknowledgement
+              * @return
+              *   [[Stop]] if current acknowledgement is [[Stop]], otherwise the result of `f`
+              */
+            inline def next[S](inline f: => Ack < S): Ack < S =
+                if isStop then Stop
```
hearnadam, line 64:
```suggestion
                if stop then Stop
```

PR #820, opened 2024-11-05
[core] Clock.nowMonotonic + fix flaky tests in JS

fwbrasil:
Time-based tests are flaky in JS. It seems the default timer implementation has precision issues. This PR introduces `Clock.nowMonotonic` to provide more a more accurate `Stopwatch` and updates test thresholds to account for the timer imprecision.

### DIFF
--- a/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
@@ -519,7 +519,7 @@ class AsyncTest extends Test:
             yield assert(result == 42)
         }
 
-        "interrupts computation" in run {
+        "interrupts computation" in runJVM {
             for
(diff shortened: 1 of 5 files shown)


kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
```diff
@@ -519,7 +519,7 @@ class AsyncTest extends Test:
             yield assert(result == 42)
         }
 
-        "interrupts computation" in run {
+        "interrupts computation" in runJVM {
```
fwbrasil, line 522: Unrelated to the `Clock` changes. This test sometimes fails in JS, I think the issue is the ordering of the execution of the callbacks that is not deterministic.

hearnadam (approved):

PR #823, opened 2024-11-06
[prelude] Parse improvements

fwbrasil:


kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -66,23 +66,27 @@ object Parse:
     private def fail(states: Seq[State], message: String)(using frame: Frame): Nothing < Abort[ParseFailed] =
         Abort.fail(ParseFailed(frame, states, message))
 
-    /** Tries all parsers in sequence and returns results from all successful parses. Unlike firstOf, this will evaluate all parsers even
```
fwbrasil, line 69: This scaladoc was incorrect. I think I changed the behavior of the method and forgot to update it.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -94,6 +98,76 @@ object Parse:
...
         }
 
+    /** Parses a sequence of parsers in order and collects their results
+      *
+      * @param parsers
+      *   Sequence of parsers to execute sequentially
+      * @return
+      *   Chunk containing all parsed results in order. The parse branch is dropped if any parser fails
+      */
+    def inOrder[A: Flat, S](parsers: Seq[A < (Parse & S)])(using Frame): Chunk[A] < (Parse & S) =
```
fwbrasil, line 108: This combinator is normally named `sequence` in other parsing libraries. `Parse.inOrder` seems more descriptive.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -167,18 +241,23 @@ object Parse:
...
+      * This operation is sometimes known as a "cut" in other parsing libraries, as it cuts off the possibility of backtracking.
       *
       * @param v
-      *   Parser to attempt
+      *   Parser to run
       * @return
       *   Parser result, fails if parser fails with no possibility of backtracking
       */
-    def cut[A: Flat, S](v: A < (Parse & S))(using Frame): A < (Parse & S) =
+    def require[A: Flat, S](v: A < (Parse & S))(using Frame): A < (Parse & S) =
```
fwbrasil, line 254: I had used `cut` because that's what other libraries use but `require` seems more clear.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -94,6 +98,80 @@ object Parse:
...
+      *   Third parser to execute
+      * @param p4
+      *   Fourth parser to execute
+      * @return
+      *   Tuple containing all four parsed results in order. The parse branch is dropped if any parser fails
+      */
+    def inOrder[A, B, C, D, S](p1: A < (Parse & S), p2: B < (Parse & S), p3: C < (Parse & S), p4: D < (Parse & S))(using
+        Frame
+    ): (A, B, C, D) < (Parse & S) =
+        inOrder(Seq(p1, p2, p3, p4))(using Flat.unsafe.bypass).map { s =>
```
hearnadam, line 171:
```suggestion
        inOrder(Chunk(p1, p2, p3, p4))(using Flat.unsafe.bypass).map { s =>
```

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -94,6 +98,80 @@ object Parse:
...
+      *
+      * @param p1
+      *   First parser to execute
+      * @param p2
+      *   Second parser to execute
+      * @return
+      *   Tuple containing both parsed results in order. The parse branch is dropped if either parser fails
+      */
+    def inOrder[A, B, S](p1: A < (Parse & S), p2: B < (Parse & S))(using Frame): (A, B) < (Parse & S) =
+        inOrder(Seq(p1, p2))(using Flat.unsafe.bypass).map { s =>
```
hearnadam, line 133:
```suggestion
        inOrder(Chunk(p1, p2))(using Flat.unsafe.bypass).map { s =>
```

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -229,6 +312,67 @@ object Parse:
...
+      *   Parser for left delimiter
+      * @param content
+      *   Parser for content between delimiters
+      * @param right
+      *   Parser for right delimiter
+      * @return
+      *   The parsed content
+      */
+    def between[A: Flat, S](
+        left: Unit < (Parse & S),
```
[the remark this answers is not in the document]
fwbrasil: I normally prefer to require `Unit` to make it more difficult for users to accidentally discard values but I'm starting to think that it doesn't have much benefit. For instance, I find the distinction between `as` and `andThen` not very relevant and have been wondering if we should have only `andThen` that always discards, regardless if the computation produces a `Unit` or not. I feels like the semantics of the APIs is a strong enough contract for people to form a mental model of when values are discarded.

hearnadam (approved):

PR #827, opened 2024-11-08
[prelude][core] Remove `as` from the pending type + make `andThen` always discard

fwbrasil:
As elaborated in https://github.com/getkyo/kyo/pull/823#discussion_r1833693316, I don't think we need both methods. The `as` name also doesn't feel very intuitive since it doesn't indicate chaining of computations.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/kernel/Pending.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/kernel/Pending.scala
@@ -38,20 +38,11 @@ object `<`:
 
         inline def andThen[B, S2](inline f: Safepoint ?=> B < S2)(
             using
-            inline ev: A => Unit,
             inline frame: Frame,
             inline flatA: WeakFlat[A],
             inline flatB: WeakFlat[B]
         ): B < (S & S2) =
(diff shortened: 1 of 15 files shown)


kyo-prelude/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -38,21 +38,12 @@ object `<`:
...
         inline def andThen[B, S2](inline f: Safepoint ?=> B < S2)(
             using
-            inline ev: A => Unit,
             inline frame: Frame,
             inline flatA: Flat.Weak[A],
             inline flatB: Flat.Weak[B]
         ): B < (S & S2) =
             map(_ => f)
 
-        inline def as[B, S2](other: B < S2)(
```
hearnadam, line 48: Can this be moved to `kyo-combinators`?

Would it make sense to keep this if it only supported pure values?

fwbrasil: I think unnecessary given that `andThen` now offers the same functionality as `as`?

PR #828, opened 2024-11-09
[data][prelude][core] kernel cleanups

fwbrasil:
- `Frame` moved to `kyo-data`. I'm planning to make it a user-facing API given how convenient it is to produce more detailed information of a computation.
- `Flat` also moved to `kyo-data`. The macro had a workaround hardcoding the `kyo-data` types. Moving `Flat` to `kyo-data` avoids the workaround and provides proper `Flat` support for types that use an internal flat encoding like `Maybe`.
- Reduced the visibility of kernel APIs where possible
- Refactor APIs to follow the style of the rest of the codebase. For instance, `ArrowEffect.handleState` instead of `ArrowEffect.handle.state`.
- Removed `repeat` from the pending type since we have `Loop.repeat` and it doesn't seem a common API to use.

See the second commit for a clean diff considering renamings.


kyo-data/shared/src/main/scala/kyo/Flat.scala
```diff
@@ -15,56 +9,23 @@ opaque type Flat[A] = Null
 object Flat:
     object unsafe:
         inline given bypass[A]: Flat[A] = null
-    end unsafe
-
-    inline given infer[A]: Flat[A] = FlatMacro.infer
-
-    @implicitNotFound("Detected nested Kyo computation '${A}', please call '.flatten' first.")
-    case class Weak[A](dummy: Null) extends AnyVal
```
fwbrasil: Moved to a new `WeakFlat` in `kyo-prelude`.

kyo-data/shared/src/main/scala/kyo/Flat.scala
```diff
@@ -15,56 +9,23 @@ opaque type Flat[A] = Null
...
 
-    inline def infer[A]: Flat[A] = ${ macroImpl[A] }
+    inline def derive[A]: Flat[A] = ${ macroImpl[A] }
 
     def macroImpl[A: Type](using Quotes): Expr[Flat[A]] =
         import quotes.reflect.*
 
         val t = TypeRepr.of[A].dealias
 
-        object Kyo:
```
fwbrasil: This is an old mechanism that used to provide more detailed error messages when pending effects don't match. It isn't functional since the last major redesign so it's fine to remove.

kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -16,6 +16,7 @@ export Maybe.Present
 /** Companion object for Maybe type */
 object Maybe:
     inline given [A, B](using inline ce: CanEqual[A, B]): CanEqual[Maybe[A], Maybe[B]] = CanEqual.derived
+    inline given [A: Flat]: Flat[Maybe[A]]                                             = Flat.unsafe.bypass
```
fwbrasil, line 19: The workaround hardcoding the type in the macro was incorrect. A `Maybe[A]` can only be considered flat if `A` is flat type given that the internal representation is flat type union.

kyo-prelude/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -161,134 +163,130 @@ object ArrowEffect:
             end handle3Loop
             handle3Loop(v, Context.empty)
         end apply
+    end handle
 
-        private[kyo] inline def partial[
```
fwbrasil, line 165: No implementation changes here, only API refactorings.

kyo-prelude/shared/src/main/scala/kyo/kernel/Context.scala
```diff
@@ -12,12 +13,12 @@ object Context:
     val empty: Context = Map.empty
 
     extension (self: Context)
-        inline def isEmpty = self eq empty
```
fwbrasil, line 15: no need to inline these methods

build.sbt
```diff
@@ -180,6 +180,7 @@ lazy val `kyo-data` =
         .in(file("kyo-data"))
         .settings(
             `kyo-settings`,
+            libraryDependencies += "com.lihaoyi"   %%% "pprint"        % "0.9.0",
```
fwbrasil, line 183: Used by `Frame`

hearnadam: Should we consider replicating the bare minimum functionality to avoid this dependency? I think we only need JCTools as a hard dependency.

kyo-prelude/shared/src/main/scala/kyo/kernel/Trace.scala
```diff
@@ -1,11 +1,12 @@
 package kyo.kernel
 
 import internal.*
+import kyo.Frame
 import scala.annotation.tailrec
 import scala.util.control.NonFatal
 import scala.util.control.NoStackTrace
 
-class Trace(
+final class Trace(
```
hearnadam, line 9: I wonder if we could enforce that all classes are either final/abstract/sealed...

hearnadam (approved):

PR #834, opened 2024-11-10
[prelude] kernel scaladocs + cleanups

fwbrasil:
Most of the changes are scaladocs but I also reduced the visibility of additional APIs and removed a few unnecessary parameters.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
@@ -8,97 +8,187 @@ import scala.annotation.nowarn
 import scala.annotation.tailrec
 import scala.util.control.NonFatal
 
-abstract class ArrowEffect[-I[_], +O[_]] extends Effect
+/** Represents abstract functions whose implementations are provided later by a handler.
+  *
+  * ArrowEffect captures the shape of a function without specifying its implementation. It describes a transformation from Input[A] to
+  * Output[A] for any type A, but defers how that transformation actually happens until a handler interprets it. This makes it a powerful
+  * way to write code that is abstract over how its operations are performed.
+  *
+  * ArrowEffect supports multi-shot continuations, meaning that handlers can invoke the continuation function multiple times or not at all.
+  * This enables powerful control flow effects like backtracking, non-determinism, or early returns. For example, a choice effect could
+  * invoke its continuation multiple times with different values to explore multiple execution paths.
+  *
+  * The type parameters Input[_] and Output[_] define the "shape" of the function being abstracted:
+  *
+  * @tparam Input
+  *   The input type constructor - what arguments the function takes
+  * @tparam Output
+  *   The output type constructor - what results the function produces
+  *
+  * Every use of an ArrowEffect creates a suspended function call. This suspended call contains all the information needed to perform the
+  * operation, but doesn't specify how to perform it.
+  *
+  * A handler then provides the actual function implementation that determines what happens when that suspended call is executed. Each
+  * handler takes two parameters: an input value of type I[C] that contains the input of the operation, and a continuation function
+  * representing the remainder of the computation from the point where the effect was suspended to the point where it's being handled.
+  *
+  * When defining concrete effects, ArrowEffect is commonly used with two special type constructors: Const and Id. The Const[X] type
+  * constructor ignores its type parameter and always returns X, while Id[X] simply returns X unchanged. For instance, an effect that needs
+  * to fail with errors of type E would use Const[E] as its input type - it only needs the error value itself, not any type parameters.
+  * Similarly, an effect for making choices among values would use Id as its output type - it passes through the chosen value unchanged.
+  */
+abstract class ArrowEffect[-Input[_], +Output[_]] extends Effect
 
 object ArrowEffect:
 
     final class SuspendOps[A](dummy: Unit) extends AnyVal:
 
+        /** Creates a suspended computation that requests a function implementation from an arrow effect. This establishes a requirement for
+          * a function that must be satisfied by a handler higher up in the program. The requirement becomes part of the effect type,
+          * ensuring that handlers must provide the requested function before the program can execute.
+          *
+          * @param effectTag
+          *   Identifies which arrow effect to request the function from
+          * @param funcionInput
+          *   The input value to be transformed by the function
+          * @return
+          *   A computation that will receive the requested function when executed
+          */
         @nowarn("msg=anonymous")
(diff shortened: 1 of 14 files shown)


kyo-prelude/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -163,21 +277,123 @@ object ArrowEffect:
             end handle3Loop
             handle3Loop(v, Context.empty)
         end apply
+
```
fwbrasil, line 280: new overloading with 4 params

kyo-prelude/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -8,97 +8,187 @@ import scala.annotation.nowarn
...
+          *   The computation result with the function implementation provided
+          */
         inline def apply[I[_], O[_], E <: ArrowEffect[I, O], A, B, S, S2, S3](
-            inline tag: Tag[E],
+            inline effectTag: Tag[E],
             v: A < (E & S)
         )(
-            inline handle: [C] => (I[C], Safepoint ?=> O[C] => A < (E & S & S2)) => A < (E & S & S2),
-            inline done: A => B < S3 = (v: A) => v,
-            inline accept: [C] => I[C] => Boolean = [C] => (v: I[C]) => true
```
fwbrasil, line 60: The `accept` function is dangerous and shouldn't be available to users and `done` was unused. I've removed them.

kyo-prelude/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -192,19 +408,24 @@ object ArrowEffect:
         handleLoop(state, v, Context.empty)
     end handleState
 
-    inline def handleCatching[I[_], O[_], E <: ArrowEffect[I, O], A, B, S, S2, S3](
-        inline tag: Tag[E],
+    private[kyo] inline def handleCatching[I[_], O[_], E <: ArrowEffect[I, O], A, B, S, S2, S3](
```
fwbrasil, line 411: This functionality can be achieved indirectly via `Effect.catching` + `ArrowEffect.handle` and the API isn't very safe given that it has `accept` so I've made the method `private[kyo]`.

kyo-prelude/shared/src/main/scala/kyo/kernel/Effect.scala
```diff
@@ -42,4 +61,10 @@ object Effect:
                 f(ex)
         end try
     end catching
+
+    private[kyo] def defer[A, S](f: Safepoint ?=> A < S)(using _frame: Frame): A < S =
```
fwbrasil, line 65: I decided to make this method `private[kyo]` because I can't see a use case that would make sense for the user call it directly. Stack safety is provided automatically and side effects should be suspended via `IO`.

kyo-prelude/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -163,21 +277,123 @@ object ArrowEffect:
...
-        inline tag: Tag[E],
+        inline effectTag: Tag[E],
         inline state: State,
         v: A < (E & S)
     )(
         inline handle: [C] => (I[C], State, Safepoint ?=> O[C] => A < (E & S & S2)) => (State, A < (E & S & S2)) < S3,
-        inline done: (State, A) => B < (S & S2 & S3) = (_: State, v: A) => v,
-        inline accept: [C] => I[C] => Boolean = [C] => (v: I[C]) => true
-    )(using inline _frame: Frame, inline flat: Flat[A], safepoint: Safepoint): B < (S & S2 & S3) =
+        inline done: (State, A) => B < (S & S2 & S3) = (_: State, v: A) => v
```
hearnadam, line 386: Do you need `done` here?

kyo-prelude/shared/src/main/scala/kyo/kernel/Loop.scala
```diff
@@ -175,6 +401,18 @@ object Loop:
...
+      * current index and state value, enabling more complex stateful computations with index tracking.
+      *
+      * @param input
+      *   The initial state value
+      * @param run
+      *   The function to execute for each iteration, receiving the current index and state, and producing an outcome
+      * @return
+      *   The final result after loop completion
+      */
     inline def indexed[A, O, S](input: A)(inline run: Safepoint ?=> (Int, A) => Outcome[A, O] < S)(using
```
hearnadam, line 416: Do we need to specialize this function?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -84,11 +66,53 @@ object `<`:
             mapLoop(v)
         end map
 
-        inline def evalNow(using inline flat: Flat[A]): Maybe[A] =
```
fwbrasil, line 87: Moved to the end + made `private[kyo]`. I don't envision users needing this API and it exposes too much of Kyo's current internal implementation, which might change over time.

kyo-prelude/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -7,60 +7,42 @@ import scala.annotation.nowarn
...
+  *
+  * The pending type has a single fundamental operation - the monadic bind, which is exposed as both `map` and `flatMap` (for
+  * for-comprehension support). Plain values are automatically lifted into the effect context, which means `map` can serve as both `map` and
+  * `flatMap`. This allows writing effectful code typically without having to distinguish map from flatMap or manually lifting values.
+  */
 opaque type <[+A, -S] = A | Kyo[A, S]
 
 object `<`:
 
-    implicit private[kernel] inline def fromKyo[A, S](v: Kyo[A, S]): A < S      = v
```
fwbrasil, line 14: moved to the end of the file

kyo-prelude/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
fwbrasil: Yes, it's used in `Var.run` for example.

kyo-prelude/shared/src/main/scala/kyo/kernel/Loop.scala
fwbrasil: It's not strictly necessary because a regular loop where the user updates the index manually should not box as well but it's more of a usability feature. Indexed loops are common.

PR #839, opened 2024-11-11
[prelude] more generic Aspect

fwbrasil:


kyo-prelude/shared/src/main/scala/kyo/Aspect.scala
```diff
@@ -4,180 +4,223 @@ import Aspect.*
...
+  * @tparam Input
+  *   The input type constructor - what values the aspect can process
+  * @tparam Output
+  *   The output type constructor - what results the aspect can produce
   * @tparam S
-  *   The effect type of the aspect
+  *   The effect type - what effects the aspect can perform
   */
-final class Aspect[A, B, S] private[kyo] (default: Cut[A, B, S])(using Frame) extends Cut[A, B, S]:
+final class Aspect[Input[_], Output[_], S] private[kyo] (
```
fwbrasil, line 30: I want to introduce an aspect to `Parse.read` to provide a `Parse.spaced` combinator that automatically consumes  whitespaces between parsers. It's not possible with the current implementation because `Parse.read` has a generic `A` type parameter.

This change uses an approach similar to `ArrowEffect` with two `Input[_]` and `Output[_]` type parameters to support aspects in generic contexts like `Read.parse`.

kyo-prelude/shared/src/main/scala/kyo/Aspect.scala
```diff
@@ -4,180 +4,223 @@ import Aspect.*
...
+        /** Represents a binding function that implements a cut.
+          *
+          * @tparam I
+          *   The input type constructor
+          * @tparam O
+          *   The output type constructor
+          * @tparam S
+          *   The effect type
+          */
+        type Binding[I[_], O[_], S] = [C] => (I[C], I[C] => O[C] < S) => O[C] < S
```
fwbrasil, line 162: I've added this new `Cut.Binding` generic function to make it easier to create cuts with anonymous functions.

kyo-prelude/shared/src/main/scala/kyo/kernel/package.scala
```diff
@@ -2,26 +2,6 @@ package kyo
 
 package object kernel:
 
-    /** Identity type constructor.
```
fwbrasil, line 5: I've moved these type helpers to the `kyo` package since `Aspect` declarations can now use them as well.

kyo-prelude/shared/src/test/scala/kyo/AspectTest.scala
```diff
@@ -5,42 +5,30 @@ import kyo.Aspect.Cut
...
         "default" in run {
             test(1).map(v => assert(v == 2))
         }
 
         "with cut" in run {
-            val cut = new Cut[Int, Int, Any]:
-                def apply(v: Int)(f: Int => Int < Any) =
-                    f(v + 1)
-            aspect.let[Assertion, Any](cut) {
+            aspect.let([C] => (input, cont) => cont(input + 1)) {
```
fwbrasil, line 16: More lightweight way to define cuts with `Cut.Binding`. I hope Scala will eventually allow omitting [C].

kyo-prelude/shared/src/test/scala/kyo/AspectTest.scala
```diff
@@ -267,4 +201,115 @@ class AspectTest extends Test:
         }
     }
 
+    "non-Const type parameters" - {
+        case class Wrapped[+A, B](value: A, meta: B) derives CanEqual
+        case class Container[+A](value: A, meta: String) derives CanEqual
+
+        // TODO SIGSEGV in Scala Native
```
fwbrasil, line 208: It seems a bug in Scala Native since JVM and JS pass. I couldn't isolate it but all these tests fail in Native.

fwbrasil: I've changed the effect to make `Cut` the generic function directly and now all tests are failing in Native. It seems the issue is the use of the generic functions. I'll open a ticket to follow up.

kyo-prelude/shared/src/main/scala/kyo/Aspect.scala
fwbrasil: I've decided to remove `Binding` and define `Cut` as the generic function directly to simplify the effect.

fwbrasil:
The test still fails to compile in native even when the tests are marked with `taggedAs notNative`. I've moved the test file so it executes only for the JVM as a workaround.

fwbrasil:
merging to post the `Parse` PR based on this change. Please feel free to provide feedback!

PR #843, opened 2024-11-13
[prelude] Parse: readAspect + spaced + readWhile + by-name parsers

fwbrasil:


kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -28,9 +37,9 @@ object Parse:
       * @return
       *   Parsed value if successful, drops the current parse branch if unsuccessful
       */
-    def read[A, S](f: Text => Maybe[(Text, A)] < S)(using Frame): A < (Parse & S) =
```
fwbrasil, line 31: I've removed the support for extra `S` effects in the read function since `Aspect` needs to know what effects will be used beforehand. It was also unused.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -28,9 +37,9 @@ object Parse:
...
+    def read[A](f: Text => Maybe[(Text, A)])(using Frame): A < Parse =
         Var.use[State] { state =>
-            f(state.remaining).map {
+            readAspect(state.remaining)(f).map {
```
fwbrasil, line 42: The aspect is applied only to the read function. It enables users to customize the reading behavior for an entire computation like in the new `Parse.spaced`.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -77,43 +120,162 @@ object Parse:
     def anyOf[A, S](seq: (A < (Parse & S))*)(using Frame): A < (Parse & S) =
         Choice.eval(seq)(identity)
 
+    private def firstOf[A: Flat, S](seq: Seq[() => A < (Parse & S)])(using Frame): A < (Parse & S) =
```
fwbrasil, line 123: I had issues in a parser I've been working on with recursion generating stack overflows. The issue isn't in the monadic execution but in wiring parsers with circular dependencies via methods that might short circuit execution like `firstOf`. I've updated the methods to take by-name computations instead and made the version taking `Seq` private since the usage isn't straightforward with thunks.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -122,90 +284,202 @@ object Parse:
...
-      * @param seq
-      *   Sequence of parsers to try - all parsers will be evaluated
-      * @param f
-      *   Selection function that receives pairs of successful parses (with their states) and decides which one to keep. Returns Maybe.empty
-      *   to skip both.
       * @return
-      *   The parse result selected by the selection function
+      *   Tuple containing all five parsed results in order. The parse branch is dropped if any parser fails
       */
-    def select[A: Flat, S, S2](seq: (A < (Parse & S))*)(f: ((State, A), (State, A)) => Maybe[(State, A)] < S2)(
```
fwbrasil, line 186: I've decided to remove `Parse.select`, `Parse.shortest`, and `Parse.longest`. I though I was going to need it but I won't. It's also possible to achieve the same behavior via `Parse.attempt`.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -499,41 +773,13 @@ object Parse:
...
-                if state2.position > state1.position then Maybe(next)
-                else Maybe(curr)
-        }
-
     /** Consumes whitespace characters
       *
       * @return
       *   Unit after consuming whitespace
       */
-    def whitespaces(using Frame): Unit < Parse =
```
fwbrasil, line 535: I had made a few methods return `Unit` so users wouldn't need to discard values manually. We're now discarding values in the APIs themselves like in `Parse.between` and `andThen` works even for non-`Unit` computations so it's more ergonomic to return values.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -588,15 +834,20 @@ object Parse:
...
+            else
+                Maybe.empty
+        }
 
     /** Consumes any single character
       *
       * @return
       *   The character consumed
       */
-    def anyChar(using Frame): Char < Parse =
```
fwbrasil, line 599: A few renamings so it's easier to find parsers via auto complete.

fwbrasil:
The build for Scala Native is failing with the same issue as https://github.com/getkyo/kyo/pull/839 because of the use of `Aspect` in `Parse`. I've moved `ParseTest` to JVM only as a workaround.

hearnadam (approved):

PR #849, opened 2024-11-16
[data] make `Tag[Defer]` compact

hearnadam:


kyo-prelude/shared/src/main/scala/kyo/kernel/package.scala
```diff
@@ -5,7 +5,7 @@ package object kernel:
     enum Mode derives CanEqual:
         case Development, Staging, Production
 
-    private object internal:
+    private[kyo] object internal:
```
fwbrasil, line 8: I'd prefer to avoid making this visible to the rest of the codebase. An important aspect of the last major redesign was to avoid having code that uses the internal representation of the pending type outside of the kernel. It makes it easier to evolve the kernel in isolation. Can you make only `Defer` `private[kyo]`? It could be in the package object instead.

hearnadam: I could also create a test in the kernel namespace.

PR #850, opened 2024-11-18
[core][combinators] workaround issue with nested boundaries

fwbrasil:
This PR applies the workaround identified by @hearnadam with an `inline` indirection. I think we should still keep #804 open to explore a more proper solution. Users might need to define methods with boundaries and the workaround can be confusing.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -39,6 +39,11 @@ object Async:
       *   A Fiber representing the running computation
       */
     inline def run[E, A: Flat, Ctx](inline v: => A < (Abort[E] & Async & Ctx))(
+        using frame: Frame
+    ): Fiber[E, A] < (IO & Ctx) =
+        _run(v)
+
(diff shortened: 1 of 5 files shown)


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -39,6 +39,11 @@ object Async:
       *   A Fiber representing the running computation
       */
     inline def run[E, A: Flat, Ctx](inline v: => A < (Abort[E] & Async & Ctx))(
+        using frame: Frame
+    ): Fiber[E, A] < (IO & Ctx) =
+        _run(v)
+
+    private[kyo] inline def _run[E, A: Flat, Ctx](inline v: => A < (Abort[E] & Async & Ctx))(
```
hearnadam, line 46: As discussed offline, I don't think these can be `private` at all due to the `inline`... We may need to find a way to test only public `inline` functions in another namespace...

fwbrasil: I've just checked by moving `AsyncTest` to a `kyoTest` package and it seems ok.

hearnadam (approved):

Issue #679, opened 2024-09-23
Improve API of Streams


calvinlfer:
The current API of streams is quite low level requiring the user of the API to drop down to the underlying representation of Emit in order to deal with complexity (if you take a look at Virgil Kyo to see me struggling 😂). The underlying representation feels akin to ZIO’s channel or FS2’s Pull but it would be much nicer to offer more higher level combinators that prevent/delay the user of the API to reach for lower levels immediately and allows one to stay at a very high level of abstraction until one cannot and requires fine grained control and performance. The idea at least naively would be to eventually replicate the API surface (or provide similar or better expressivity) of the other 2 streaming libraries.

maujim:
Can you please clarify what the acceptance criteria are? I would like to /attempt #679

calvinlfer:
Given my experience of having used FS2 Streams & ZIO Streams which are pull-based solutions. I would love to see the following, @hearnadam & @fwbrasil please feel free to add or subtract since you folks have way more expertise:
* The ability to evaluate multiple effects in parallel (mapZIOPar and mapAsync/parEvalMap) preserving order
* The ability to evaluate multiple effects in parallel (mapZIOParUnordered and mapAsync/parEvalMapUnordered) without preserving order
* The ability to perform rechunking on the stream elements
* The ability to group multiple elements together using time and space (groupedWithin(x.seconds, y.sizeOfChunk))
* The ability to merge multiple Streams together (not sure if this makes sense)
* An API like Pull/ZChannel to model Streams - we may already have this
* The ability to have different ways to evaluate streams (runDrain/runCollect/etc.)
* More high level operators to create Streams - like paginate/ZStreams.async
* Integration with the rest of the Kyo ecosystem (Queue/Channel/etc.)

DamianReeves:
I'd suggest adding a `merge` operation as well.

johnhungerford:
Thinking about what work is left to do on stream, I thought I'd give a quick update of @calvinlfer 's feature list:

#### Finished
- The ability to perform rechunking on the stream elements
- Integration with the rest of the Kyo ecosystem (Queue/Channel/etc.)
  - Streaming from `Channel` and `Hub`
  - `Queue` integration was ruled out

#### Todo
- The ability to evaluate multiple effects in parallel (mapZIOPar and mapAsync/parEvalMap) preserving order
- The ability to evaluate multiple effects in parallel (mapZIOParUnordered and mapAsync/parEvalMapUnordered) without preserving order
[...]

PR #851, opened 2024-11-19
[prelude] make Stream lazier

hearnadam:
- stop using class member, enforcing accidental eagerness
- add new method: Stream.range to simplify creation of streams
- init/range: accept `chunkSize` to limit size of Chunks- Chunk.from (Array): remove AnyRef type bound

The first PR to address #679. This change is important to ensure the laziness of streams initialized from non-effectful sources.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
@@ -337,7 +337,8 @@ object Stream:
     private val _empty           = Stream(Stop)
     def empty[V]: Stream[V, Any] = _empty.asInstanceOf[Stream[V, Any]]
 
-    private inline def DefaultChunkSize = 4096
+    /** The default chunk size for streams. */
+    inline def DefaultChunkSize: Int = 4096
 
     /** Creates a stream from a sequence of values.
       *
@@ -408,7 +409,7 @@ object Stream:
                                         else
                                             ((current - end - 1) / step.abs).abs + 1
                                     val size  = (n min _chunkSize) min remaining
-                                    val chunk = Chunk.from(Array.range(current, current + size * step, step))
+                                    val chunk = Chunk.from(Range(current, current + size * step, step))
                                     Emit.andMap(chunk)(ack => Loop.continue(current + step * size, ack))
                                 end if
                     }
--- a/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
@@ -4,6 +4,7 @@ import kyo.Emit.Ack
 import kyo.Emit.Ack.*
 import kyo.Tag
 import kyo.kernel.ArrowEffect
+import scala.annotation.nowarn
 import scala.annotation.targetName
 
 /** Represents a stream of values of type `V` with effects of type `S`.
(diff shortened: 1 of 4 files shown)


kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -328,25 +329,52 @@ final case class Stream[V, -S](v: Ack < (Emit[Chunk[V]] & S)):
 end Stream
 
 object Stream:
+    @nowarn("msg=anonymous")
+    inline def apply[V, S](v: => Ack < (Emit[Chunk[V]] & S)): Stream[V, S] =
```
fwbrasil, line 333: we can avoid the function allocation for `v`:

```suggestion
    inline def apply[V, S](inline v: => Ack < (Emit[Chunk[V]] & S)): Stream[V, S] =
```

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -328,25 +329,52 @@ final case class Stream[V, -S](v: Ack < (Emit[Chunk[V]] & S)):
...
+                        if c.isEmpty then Loop.done(Ack.Continue())
+                        else
+                            val i = n min chunkSize
+                            Emit.andMap(c.take(i))(ack => Loop.continue(c.dropLeft(i), ack))
+            }
+        }
+
+    private def fromIterable[V, S](iterable: Iterable[V], chunkSize: Int)(using tag: Tag[Emit[Chunk[V]]], frame: Frame) =
+        Emit.andMap(Chunk.empty[V]) { ack =>
+            val iterator = iterable.iterator
```
fwbrasil, line 364: I'm not sure about this since using the iterator is a side effect

hearnadam: I removed it for now. Let's see if it's necessary later.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -328,25 +329,90 @@ final case class Stream[V, -S](v: Ack < (Emit[Chunk[V]] & S)):
...
+    @nowarn("msg=anonymous")
+    inline def apply[V, S](inline v: => Ack < (Emit[Chunk[V]] & S)): Stream[V, S] =
+        new Stream[V, S]:
+            def emit: Ack < (Emit[Chunk[V]] & S) = v
 
     private val _empty           = Stream(Stop)
     def empty[V]: Stream[V, Any] = _empty.asInstanceOf[Stream[V, Any]]
 
-    def init[V, S](seq: Seq[V] < S)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Stream[V, S] =
+    private inline def DefaultChunkSize = 4096
```
fwbrasil, line 340: What do you think about making this public? It'd be nice to document it and allow users to inspect it.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -328,25 +329,90 @@ final case class Stream[V, -S](v: Ack < (Emit[Chunk[V]] & S)):
...
+
+                                if !continue then Loop.done(Stop)
+                                else
+                                    val remaining =
+                                        if step > 0 then
+                                            ((end - current - 1) / step).abs + 1
+                                        else
+                                            ((current - end - 1) / step.abs).abs + 1
+                                    val size  = (n min _chunkSize) min remaining
+                                    val chunk = Chunk.from(Array.range(current, current + size * step, step))
```
fwbrasil, line 411: Could we use `collection.immutable.Range` here? It represents a range without allocating the individual items.

fwbrasil (approved):

PR #852, opened 2024-11-19
[prelude] remove type parameter from Batch

fwbrasil:
I've realized that we can resume the suspension from a source call with another computation that has the pending effects from the source. There's no need to track source effects in the effect type.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Batch.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Batch.scala
@@ -8,17 +8,12 @@ import kyo.kernel.*
   *
   * Batch is used to group multiple operations together and execute them in a single batch, which can lead to performance improvements,
   * especially when dealing with external systems or databases.
-  *
-  * @tparam S
-  *   Effects from batch sources
   */
-sealed trait Batch[+S] extends ArrowEffect[Op[*, S], Id]
+sealed trait Batch extends ArrowEffect[Op, Id]
 
 object Batch:
 
     import internal.*
-
-    private inline def erasedTag[S]: Tag[Batch[S]] = Tag[Batch[Any]].asInstanceOf[Tag[Batch[S]]]
 
     /** Creates a batched computation from a source function.
       *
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -36,8 +31,8 @@ object Batch:
       * This method allows for efficient batching of operations by processing multiple inputs at once, while still providing individual
       * results for each input.
       */
-    inline def source[A, B, S](f: Seq[A] => (A => B < S) < S)(using inline frame: Frame): A => B < Batch[S] =
-        (v: A) => ArrowEffect.suspend[B](erasedTag[S], Call(v, f))
+    inline def source[A, B, S](f: Seq[A] => (A => B < S) < S)(using inline frame: Frame): A => B < (Batch & S) =
+        (v: A) => ArrowEffect.suspendAndMap(Tag[Batch], Call(v, f))(identity)
```
fwbrasil, line 35: The trick is here: the continuation will receive a computation of type `B < (Batch & S)` containing `S` effects from the source. We just flatten need to flatten (`identity`).

kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -160,14 +155,14 @@ object Batch:
                             // Reassemble the results by iterating on the original collection
                             Kyo.foreach(items) { e =>
                                 // Note how each value can have its own effects
-                                capture(results(e.value).map(e.cont))
+                                capture(e.cont(results(e.value)))
```
fwbrasil, line 158: Note how the continuation now receives a computation instead of a value.

kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -191,8 +186,9 @@ object Batch:
     object internal:
         type Source[A, B, S] = Seq[A] => (A => B < S) < S
 
-        type Op[V, -S] = Seq[V] | Call[?, V, S]
-        case class Call[A, B, -S](v: A, source: Source[A, B, S])
+        sealed trait Op[A]
+        case class Eval[A](seq: Seq[A])                         extends Op[A]
+        case class Call[A, B, S](v: A, source: Source[A, B, S]) extends Op[B < (Batch & S)]
```
fwbrasil, line 191: I couldn't keep the optimization to avoid allocations for `Batch.eval` since it's necessary to indicate from the `Op` that the result is a computation in the case of `Call`.

kyo-prelude/shared/src/test/scala/kyo/BatchTest.scala
```diff
@@ -767,4 +767,87 @@ class BatchTest extends Test:
         }
     }
 
+    "source effects handled before Batch.run" - {
```
fwbrasil, line 770: Checks that computations after the source call have the correct pending effects that can be handled before `Batch.run`.

hearnadam (approved):
Awesome improvement!

PR #853, opened 2024-11-20
[prelude][core] effect isolates

fwbrasil:
Please see the `Isolate` scaladoc for more background. Ideally, we should find a way to merge `Boundary` and `Isolate` but that's challenging with the inference issues we've been having and it's difficult to match the current performance of `Boundary` since it's optimized by using more low-level kernel APIs. This new mechanism is focused only on effects that maintain state but don't transform values.

Main benefits of this change:

 - It allows users to fork computations with `Var`,  `Emit`, `Memo`, and `Check`. 
 - Users can leverage `isolate.run` directly to enforce the effect handling for a computation
 - Isolates provide transaction-like behavior via interaction with `Abort`. This behavior will be necessary for an STM-based effect since it'll need to rollback state to retry computations if necessary.

`Isolate` can't support effects like `Choice` since it transforms `A` into `Seq[A]`.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Isolate.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Isolate.scala
@@ -1,0 +1,141 @@
+package kyo
+
+import internal.*
+import kyo.Flat
+import kyo.Frame
+import kyo.Kyo
+import kyo.Var.isolate
+import kyo.kernel.ContextEffect
+import kyo.kernel.Safepoint.Interceptor
+import scala.quoted.*
+
+/** Provides fine-grained control over state isolation and propagation within computations.
+  *
+  * The `Isolate` mechanism allows you to control how effect state is handled, contained, and propagated during computation. This is
+  * essential for implementing effects that need careful control over their state visibility and ordering of operations.
+  *
+  * While Isolate provides a low-level API through `use`, `resume`, and `restore` for advanced use cases, most users should rely on the
+  * high-level `run` method which composes these operations safely and handles all the complexity of state management internally.
+  *
+  * Effects typically provide their isolation strategies through their companion objects (e.g. `Var.isolate.update`, `Check.isolate.merge`).
+  * These isolates can be composed using `andThen`, which enforces the ordering of effect handling - ensuring effects are handled in the
+  * specified sequence.
+  *
+  * Key use cases include:
+  *   - Transactional effects where state changes may need to be contained until commit
+  *   - Effects that require explicit control over state propagation order
+  *   - Parallel computations where state needs to be merged in specific ways
+  *   - Scenarios where effect state should be temporarily scoped or discarded
+  *   - Error handling where partial effects must be rolled back atomically
+  *   - Complex workflows requiring nested transaction boundaries
+  *
+  * Implementations typically choose one of three strategies:
+  *   - Merge: Combine states when multiple computations need to reconcile their effects
+  *   - Update: Unconditionally updates the state with the last available value
+  *   - Discard: Prevent state from propagating beyond a defined scope
+  *
+  * A key feature of Isolate is its interaction with error handling through the Abort effect. When an Abort occurs within an isolated scope,
+  * the isolation acts as a transaction boundary - isolated effects within the isolated scope are rolled back or discarded, maintaining
+  * consistency of the outer scope. This transactional behavior applies across all effect types:
+  *   - State changes are rolled back to their pre-isolation values
+  *   - Emitted values are discarded rather than being merged
+  *   - Cached computations remain isolated and don't propagate
+  *   - Nested isolations maintain these guarantees at each level
+  *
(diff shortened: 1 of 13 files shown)


kyo-prelude/shared/src/main/scala/kyo/kernel/Boundary.scala
```diff
@@ -22,19 +22,6 @@ import scala.quoted.*
   * @tparam S
   *   The source effects that must be handled before crossing
   */
-@implicitNotFound("""
-Could not create an Async boundary for effects: ${Ctx}
```
fwbrasil, line 26: This scenario doesn't happen anymore after https://github.com/getkyo/kyo/pull/850

kyo-prelude/shared/src/main/scala/kyo/kernel/Boundary.scala
```diff
@@ -75,23 +62,26 @@ object Boundary:
...
+                    |1. Handle the effects before the async operation:
+                    |   Instead of:
+                    |     Async.race(computation1, computation2) // where computations have pending MyEffect
                     |
-                    |Handle the effect first:
-                    |  Async.run(MyEffect.run(computation))
+                    |   Handle the effect first:
+                    |     Async.race(MyEffect.run(computation1), MyEffect.run(computation2))
                     |
-                    |Note: There's currently a limitation with nested Async operations (like Async.run(Async.run(v))).
```
fwbrasil, line 86: I've also removed the text about the limitation with nested boundaries here

kyo-prelude/shared/src/main/scala/kyo/Isolate.scala
```diff
@@ -0,0 +1,141 @@
...
+  *   - Cached computations remain isolated and don't propagate
+  *   - Nested isolations maintain these guarantees at each level
+  *
+  * This combination of isolation strategies and error handling makes Isolate particularly powerful for implementing robust,
+  * transaction-like behaviors across different effect types while maintaining clean separation of concerns.
+  *
+  * @tparam S
+  *   The type of effect being isolated
+  */
+abstract class Isolate[S]:
```
hearnadam, line 51: As discussed offline, I think it would be better if we could redesign this class to be an 'EffectScope' or 'Scope' etc. We should be able to then design a general `Scope.Strategy` to determine the strategy for managing Context effects within forks. I believe this would then enable broader application, combination, and more fluent operators.

I can imagine for most usecases, effects could define a typeclass to encode the required state handling for each 'strategy'. When necessary, we could ask the user to define something (Merge).

We could specialize the behavior on `Nothing`/`Any` to avoid the cost when no management within forks is required. Ideally this would also be merged with Boundary.


To avoid blocking this functionality, I will approve for now.

kyo-prelude/shared/src/main/scala/kyo/Memo.scala
```diff
@@ -57,4 +57,39 @@ object Memo:
...
+
+        /** Creates an isolate that overwrites the memoization cache.
+          *
+          * When the isolation ends, replaces the outer cache with the cache from the isolated computation. Earlier cached results are
+          * discarded in favor of any new results computed in isolation.
+          *
+          * @return
+          *   An isolate that updates the cache with isolated results
+          */
+        def update: Isolate[Memo] =
```
hearnadam, line 81: Any reason these aren't constants?

kyo-prelude/shared/src/main/scala/kyo/Var.scala
```diff
@@ -151,6 +151,62 @@ object Var:
...
+          *
+          * @tparam V
+          *   The type of value in the Var
+          * @return
+          *   An isolate that updates the Var with its isolated value
+          */
+        def update[V: Tag]: Isolate[Var[V]] =
+            new Base[V]:
+                def restore[A: Flat, S2](state: V, v: A < S2)(using Frame) =
+                    Var.set(state).andThen(v)
```
hearnadam, line 174: If the tag was provided in the function itself, could we use a cached instance?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/Isolate.scala
fwbrasil: Makes sense. I think having a more explicit mechanism without implicit resolution initially will be beneficial for us to identify patterns of how to safely make the mechanism more convenient. The trouble with inference is that effect handling ordering is important but we'll probably find usage patterns we can optimize for.

kyo-prelude/shared/src/main/scala/kyo/Var.scala
fwbrasil: It could be we'd need to extend `Isolate` to pipe the type parameter of the state, which would increase the complexity of the abstraction. I think it's ok to have the allocation since users can cache instances and reuse.

PR #854, opened 2024-11-21
[prelude] introduce `Stream#rechunk`

hearnadam:
PR 2 for #679 

Rechunk was a bit tricky to implement as the remaining 'buffer' would get dropped unless the chunks perfectly aligned with the number of emits. Because of this, it's necessary to Emit a final empty chunk to ensure the buffer is flushed. This can sometimes drop the number of elements in a given chunk if a user Emits an empty chunk, but that shouldn't be a problem.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
@@ -272,7 +272,8 @@ sealed abstract class Stream[V, -S]:
       *   A new stream with elements regrouped into chunks of the specified size
       */
     def rechunk(chunkSize: Int)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Stream[V, S] =
-        Stream.sized[V, S](chunkSize)(cs =>
+        Stream[V, S]:
+            val _chunkSize = chunkSize max 1
             ArrowEffect.handleState(tag, Chunk.empty[V], emit.andThen(Emit(Chunk.empty[V])))(
                 [C] =>
                     (input, buffer, cont) =>
@@ -280,23 +281,22 @@ sealed abstract class Stream[V, -S]:
                             Emit.andMap(buffer)(ack => (Chunk.empty, cont(ack)))
                         else
                             val combined = buffer.concat(input)
-                            if combined.size < cs then
+                            if combined.size < _chunkSize then
                                 (combined, cont(Continue()))
                             else
                                 Loop(combined: Chunk[V], Continue(): Ack) { (current, ack) =>
                                     ack match
                                         case Stop => Loop.done((current, cont(Stop)))
                                         case Continue(_) =>
-                                            if current.size < cs then
+                                            if current.size < _chunkSize then
                                                 Loop.done((current, cont(Continue())))
                                             else
-                                                Emit.andMap(current.take(cs)) { nextAck =>
-                                                    Loop.continue(current.dropLeft(cs), nextAck)
+                                                Emit.andMap(current.take(_chunkSize)) { nextAck =>
+                                                    Loop.continue(current.dropLeft(_chunkSize), nextAck)
                                                 }
                                 }
                             end if
             )
-        )
     end rechunk
 
     /** Runs the stream and discards all emitted values.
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -259,6 +259,46 @@ sealed abstract class Stream[V, -S]:
...
+      *   - When there aren't enough remaining elements to form a complete chunk
+      *   - When the input stream emits an empty chunk
+      *
+      * @param chunkSize
+      *   The target size for each chunk. Must be positive - negative values will be treated as 1.
+      * @return
+      *   A new stream with elements regrouped into chunks of the specified size
+      */
+    def rechunk(chunkSize: Int)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Stream[V, S] =
+        Stream.sized[V, S](chunkSize)(cs =>
```
fwbrasil, line 275: I find it a bit confusing that this method only transforms the `chunkSize`. This kind of API pattern typically transforms the computation as well. Could it be a method that returns `Int` instead?

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -259,6 +259,46 @@ sealed abstract class Stream[V, -S]:
...
+            ArrowEffect.handleState(tag, Chunk.empty[V], emit.andThen(Emit(Chunk.empty[V])))(
+                [C] =>
+                    (input, buffer, cont) =>
+                        if input.isEmpty && buffer.nonEmpty then
+                            Emit.andMap(buffer)(ack => (Chunk.empty, cont(ack)))
+                        else
+                            val combined = Chunk.from(buffer.appendedAll(input))
```
fwbrasil, line 282: It seems this could be
```suggestion
                            val combined = buffer.concat(input)
```

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
fwbrasil: Also, how about using `chunkSize` instead of `cs` for clarity?

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -259,6 +259,46 @@ sealed abstract class Stream[V, -S]:
...
+      *   - When the input stream emits an empty chunk
+      *
+      * @param chunkSize
+      *   The target size for each chunk. Must be positive - negative values will be treated as 1.
+      * @return
+      *   A new stream with elements regrouped into chunks of the specified size
+      */
+    def rechunk(chunkSize: Int)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Stream[V, S] =
+        Stream.sized[V, S](chunkSize)(cs =>
+            ArrowEffect.handleState(tag, Chunk.empty[V], emit.andThen(Emit(Chunk.empty[V])))(
```
fwbrasil, line 276: I see now why you need to emit the empty `Chunk` at the end. I wonder if it'd be useful to change `Emit` to:

```scala
sealed trait Op[V]
sealed trait Emit[V] extends ArrowEffect[Const[Op[V]], Const[Emit.Ack]]
```

and then provide different `Op` subclasses for the different needs. Just an idea in case it's helpful.

hearnadam: I don't see how that would be helpful here, but I might be missing your vision 😄

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
hearnadam: Removing the function for now.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
fwbrasil: Not sure if it's useful but I thought it was worth mentioning :) It could allow more advanced patterns by providing more information on each emission

fwbrasil (approved):

PR #859, opened 2024-11-22
[direct] review direct syntax

fwbrasil:
I've been using the direct module and I find the syntax overhead still too high. For example, code with for-comprehension:

```scala
for
  a <- doSomething1
  b <- doSomething2
  _ <- doSomething3(a, b)
  c <- doSomething4
yield a + b + c
```

becomes 

```scala
val a = await(doSomething1)
val b = await(doSomething2)
await(doSomething3(a, b))
val c = await(doSometihng4)
a + b + c
```

It should be easier to follow for newcomers but it has more syntax overhead than the `for` version. This PR renames `await` to an unary `~` to make the syntax more lightweight. Example:

```scala
val a = ~doSomething1
val b = ~doSomething2
~doSomething3(a, b)
val c = ~doSometihng4
a + b + c
```


sideeffffect (commented):
Anything that replace `await` is welcome improvement. `await` requires wrapping inside parentheses, which is unnecessarily painful. 

Have you also considered `.run`? That's what ZIO Direct uses AFAIK. It's advantage over prefix `~` is that it's alphanumeric (i.e. non-symbolic) and postfix, which makes it more idiomatic in Scala. It also can be listed in auto-complete suggestion list.

fwbrasil:
@sideeffffect Yeah, it'd be best to avoid a symbol. I'd prefer to avoid `run` since the term is used in Kyo to indicate effect handling. How about `.now` to complement `defer`? Another option is `.get` but I think it might be too ambiguous.

regiskuckaertz:
I don't suppose `effect.!` would be adequate? iirc, it's a naming convention in lisp dialects, but now the Roc language has lifted it to mean exactly what you're after.
```
Stdout.line! "Hello!"
```

sideeffffect:
> How about `.now` to complement `defer`?

Oh, I like that! That would make a lot of sense.

`!` also won't be too bad. Not alphanumeric name. But may actually stand out in code.

johnhungerford:
I actually think the unary `~` is a good idea. While I agree with @sideeffffect that it's not idiomatic, I think `defer` blocks are unique situations such that it's fitting to have an out-of-the-ordinary syntax. It's a useful reminder that we're in macro land and things aren't working exactly how we're used to.

Other advantages are:
1. Absolute minimal syntax overhead, even when chaining effects
2. The indication that we're binding from an effect is up front, as it were, and hard to miss

hearnadam:
Let's try it! Can we keep both `~` and `await`?

DamianReeves:
Anything that simplifies is helpful but want to add my vote for `!`

sideeffffect:
Since we're voicing support here is the list of my priorities, from the most important to the least important

1. most important is that it doesn't need wrapping in `()`
2. it's postfix (not prefix), which is more Scala idiomatic:
  * we.write.Scala.like.this
  * this(like(not))
3. not as important, but still nice: it's alphabetical, not symbolic, again, more idiomatic in Scala

fwbrasil:
Thanks for the feedback, folks! I've been considering how to proceed and how to address https://github.com/getkyo/kyo/issues/908 as well. 

New suggestion:  `.now` to replace `await` and a new `.later` method to enable the use of Kyo computations in `defer` blocks.

```scala
defer {
  val a: Int = doSomething1.now
  val b: Int = doSomething2.now
  val c: Int < IO = doSomething3(a, b).later 
  a + b + c.now
}
```

I wanted to avoid allowing mixing monadic and direct control flows but it seems an important limitation. This new approach seems to make the control flow more clear and we could document that `later` as an advanced API.

johnhungerford:
Nice

kyo-direct/shared/src/main/scala/kyo/Direct.scala
```diff
@@ -4,40 +4,100 @@ import cps.CpsMonad
...
+      *
+      * Must be used within a `defer` block. This operation tells the direct syntax to execute the effect at this point in the computation,
+      * allowing you to use its result in subsequent operations.
+      *
+      * @return
+      *   The unwrapped value of type `A` from the effect
+      * @throws RuntimeException
+      *   if used outside a `defer` block
+      */
+    inline def now: A = ${ nowImpl('self) }
```
fwbrasil, line 43: I've made these a macro so it's possible to format the error message with `Ansi`. `compiletime.error` doesn't support that.

fwbrasil:
PR updated with the new approach

hearnadam (approved):
LGTM. I do question if we should keep `defer` if we plan to use `Defer` as the 'IO' type later...

fwbrasil:
good point. I hadn't considered that but let's follow up separately

PR #860, opened 2024-11-23
[core] async gather

fwbrasil:
New `gather` methods in `Fiber` and `Async` to collect successful computations out of a sequence. It also allows specifying a max number of expected results. See scaladocs for more information.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -153,6 +153,8 @@ object Async:
     /** Races multiple computations and returns the result of the first to complete. When one computation completes, all other computations
       * are interrupted.
       *
+      * WARNING: Executes all computations in parallel without bounds. Use with caution on large sequences to avoid resource exhaustion.
+      *
       * @param seq
       *   The sequence of computations to race
       * @return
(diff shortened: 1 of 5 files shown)


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -153,6 +153,8 @@ object Async:
     /** Races multiple computations and returns the result of the first to complete. When one computation completes, all other computations
       * are interrupted.
       *
+      * WARNING: Executes all computations in parallel without bounds. Use with caution on large sequences to avoid resource exhaustion.
```
fwbrasil, line 156: I've added this warning to `race` and `gather` methods since they don't have a parallelism limit.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -271,9 +273,9 @@ object Fiber extends FiberPlatformSpecific:
         frame: Frame,
         safepoint: Safepoint
     ): Fiber[E, A] < (IO & Ctx) =
-        IO {
+        IO.Unsafe {
             class State extends IOPromise[E, A] with Function1[Result[E, A], Unit]:
-                val pending = new AtomicInteger(seq.size)
+                val pending = AtomicInt.Unsafe.init(seq.size)
```
fwbrasil, line 278: unrelated cleanup

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -283,7 +285,7 @@ object Fiber extends FiberPlatformSpecific:
             import state.*
             boundary { (trace, context) =>
                 IO {
-                    val interruptPanic = Result.Panic(Fiber.Interrupted(frame))
+                    inline def interruptPanic = Result.Panic(Fiber.Interrupted(frame))
```
fwbrasil, line 288: unrelated cleanup: avoid allocations in the happy path

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -294,6 +296,166 @@ object Fiber extends FiberPlatformSpecific:
...
+                                    completeDiscard(Result.success(Chunk.from(results).take(size)))
+                                end if
+                            end if
+                        end loop
+                        loop()
+                    end apply
+                end State
+                val state = new State
+                import state.*
+                boundary { (trace, context) =>
```
fwbrasil, line 406: other methods have a very similar code. I'll follow up to reduce duplication

kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -42,7 +42,8 @@ sealed abstract class Chunk[A] extends Seq[A] derives CanEqual:
       *   a new Chunk containing the first n elements
       */
     override def take(n: Int): Chunk[A] =
-        dropLeftAndRight(0, length - Math.min(Math.max(0, n), length))
+        if n == length then this
+        else dropLeftAndRight(0, length - Math.min(Math.max(0, n), length))
```
fwbrasil, line 46: optimization to avoid unnecessary allocation in `Fiber._gather`

kyo-core/shared/src/test/scala/kyo/FiberTest.scala
```diff
@@ -730,4 +730,385 @@ class FiberTest extends Test:
...
+                    val indices = Array[Int](3, 1, 4, 2)
+                    val results = Array[AnyRef]("c", "a", "d", "b")
+                    Fiber.quickSort(indices, results, 4)
+                    assert(indices.sameElements(Array(1, 2, 3, 4)))
+                    assert(results.sameElements(Array("a", "b", "c", "d")))
+                }
+            }
+
+            "can handle larger arrays" - {
+                val size = 1000
```
fwbrasil, line 1075: `quickSort` isn't stack-safe but the implementation can handle larger arrays with a relatively low stack depth. Also, I think we can consider gathering a large number of computations as misuse.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -294,6 +296,189 @@ object Fiber extends FiberPlatformSpecific:
...
+
+                    // Helper array to store original indices to maintain ordering
+                    // Initialized to Int.MaxValue to handle partial results
+                    val indices = new Array[Int](max)
+                    Arrays.fill(indices, Int.MaxValue)
+
+                    // Packed representation to avoid allocations and ensure atomicity
+                    // - lower 32 bits  => successful results count (ok)
+                    // - higher 32 bits => failed results count (nok)
+                    val packed = AtomicLong.Unsafe.init(0)
```
hearnadam, line 363: It should be possible to use an opaque type here, right? Not necessary for this PR.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -294,6 +296,189 @@ object Fiber extends FiberPlatformSpecific:
...
+
+                                    // Handle race condition
+                                    waitForResults(results, size)
+
+                                    // Restore original ordering but limit size since later
+                                    // results might still arrive and we want to avoid races
+                                    quickSort(indices, results, size)
+
+                                    // Limit final result to max successful results
+                                    completeDiscard(Result.success(Chunk.from(results).take(size)))
```
hearnadam, line 401: We should be able to avoid copying the buffer here as well.

hearnadam (approved):
LGTM overall

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -356,15 +541,15 @@ object Fiber extends FiberPlatformSpecific:
...
+                IO.Unsafe {
                     class State extends IOPromise[E, Seq[A]] with ((Int, Result[E, A]) => Unit):
                         val results = (new Array[Any](seq.size)).asInstanceOf[Array[A]]
-                        val pending = new AtomicInteger(seq.size)
+                        val pending = AtomicInt.Unsafe.init(seq.size)
                         def apply(idx: Int, result: Result[E, A]): Unit =
                             result.fold(this.interruptDiscard) { value =>
                                 results(idx) = value
                                 if pending.decrementAndGet() == 0 then
-                                    this.completeDiscard(Result.success(ArraySeq.unsafeWrapArray(results)))
```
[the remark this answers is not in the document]
fwbrasil: good catch! I've added `Chunk.fromNoCopy` but made it `private[kyo]` since it's unsafe.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -294,6 +296,189 @@ object Fiber extends FiberPlatformSpecific:
...
+        if total == 0 || max <= 0 then Fiber.success(Chunk.empty)
+        else
+            IO.Unsafe {
+                class State extends IOPromise[E, Chunk[A]] with Function2[Int, Result[E, A], Unit]:
+                    val results = new Array[AnyRef](max)
+
+                    // Helper array to store original indices to maintain ordering
+                    // Initialized to Int.MaxValue to handle partial results
+                    val indices = new Array[Int](max)
+                    Arrays.fill(indices, Int.MaxValue)
```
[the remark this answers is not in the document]
fwbrasil: It's definitely more heavyweight since it allocates thunks and functions + has to dispatch them. `Arrays.fill` is a simple for loop amenable to auto vectorization. I imagine JIT compilers might even replace the method with intrinsics.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
fwbrasil: I thought it wasn't worth the extra code since the implementation is relatively long already. I can revisit when refactoring these methods.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -188,6 +190,95 @@ object Async:
...
+      *   Context requirements
+      * @param seq
+      *   Sequence of effects to execute
+      * @return
+      *   Successful results as a Chunk
+      */
+    inline def gather[E, A: Flat, Ctx](seq: Seq[A < (Abort[E] & Async & Ctx)])(
+        using frame: Frame
+    ): Chunk[A] < (Abort[E] & Async & Ctx) =
+        _gather(seq.size)(seq)
```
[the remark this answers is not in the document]
fwbrasil: I think building chunk would have the same issue since it needs to know the size

PR #865, opened 2024-11-25
[scheduler] Admission rotation window + improve hash distribution

fwbrasil:
The current mechanism will always penalize the same ids. This PR introduces a time-based window mechanism to rotate which IDs are rejected over time. I've also added a multiplication with a large prime to improve the hash distribution in case the incoming IDs aren't well distributed.

### DIFF
--- a/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/regulator/Admission.scala
+++ b/kyo-scheduler/jvm/src/main/scala/kyo/scheduler/regulator/Admission.scala
@@ -25,15 +25,28 @@ import scala.util.hashing.MurmurHash3
   *
   * ==Rejection Mechanism==
   *
-  * Tasks are rejected using a deterministic hashing mechanism that provides stable and consistent admission decisions. Each task key
-  * (string or integer) is hashed to a value between 0-99, and tasks with hash values above the current admission percentage are rejected.
-  * This approach ensures the same keys will be consistently rejected until the admission percentage changes.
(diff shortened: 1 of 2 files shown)


kyo-scheduler/jvm/src/main/scala/kyo/scheduler/regulator/Admission.scala
```diff
@@ -76,9 +91,12 @@ final class Admission(
     schedule: Task => Unit,
     nowMillis: LongSupplier,
     timer: InternalTimer,
-    config: Config = Admission.defaultConfig
+    config: Config = Admission.defaultConfig,
+    rotationWindow: Duration = Flag("admission.rotationWindowMinutes", 60).minutes
```
hearnadam, line 95: What about 5 minutes? Any reason the scheduler would need such a long window?

hearnadam (approved):

kyo-scheduler/jvm/src/main/scala/kyo/scheduler/regulator/Admission.scala
fwbrasil: It's not much about the scheduler itself. On each new rotation window, a new set of ids are accepted. If a system is under pressure, the shift in incoming load to a new set of users can increase the pressure on shared resources. For example, data might need to be loaded from databases to caches and cache evictions can compound the pressure. The longer window favors stability. The goal is to avoid the same set of users always being penalized across outages, not necessarily provide user-level fairness during a specific outage. It's always possible to tune the config, though.

PR #873, opened 2024-11-28
[core][direct] scala native support

fwbrasil:


build.sbt
```diff
@@ -225,12 +226,10 @@ lazy val `kyo-core` =
             libraryDependencies += "org.jctools"    % "jctools-core"    % "4.0.5",
             libraryDependencies += "org.slf4j"      % "slf4j-api"       % "2.0.16",
             libraryDependencies += "dev.dirs"       % "directories"     % "26",
-            libraryDependencies += "dev.zio"      %%% "zio-laws-laws"   % "1.0.0-RC35" % Test,
```
fwbrasil, line 228: These dependencies were unused.

kyo-core/native/src/main/scala/kyo/addersStubs.scala
```diff
@@ -0,0 +1,29 @@
+package java.util.concurrent.atomic
+
+class DoubleAdder():
```
fwbrasil, line 3: I've created these quick stubs but we should follow up later optimizing them since `synchronized` isn't ideal.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -380,6 +381,7 @@ object Fiber extends FiberPlatformSpecific:
                                             // Store successful result and its original index for ordering
                                             indices(okInt - 1) = idx
                                             results(okInt - 1) = v.asInstanceOf[AnyRef]
+                                            discard(VarHandle.storeStoreFence())
```
fwbrasil, line 384: I had trouble with flaky tests in Native. The fence was already technically missing here but the JVM probably introduces fences right after so the bug wasn't visible. That doesn't seem the case for Native.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -12,7 +13,7 @@ import scala.util.control.NoStackTrace
 
 private[kyo] class IOPromise[+E, +A](init: State[E, A]) extends Safepoint.Interceptor:
 
-    @volatile private var state: State[E, A] = init
+    private val state = new AtomicReference[State[Any, Any]](init)
```
fwbrasil, line 16: I couldn't find a supported way to perform atomic operations on a field in Native without allocating an `AtomicReference`. This will introduce a performance regression with a new allocation for each fiber. I'll create a ticket to follow up.

kyo-core/shared/src/main/scala/kyo/stats/internal/TraceReceiver.scala
```diff
@@ -37,6 +28,15 @@ object TraceReceiver:
             )(using Frame) =
                 Span.noop
 
+    val get: TraceReceiver =
```
fwbrasil, line 31: I was facing compiler crashes in Native. The issue seems the combination of `lazy val` with `ServiceLoader`. The Scala Native developers will investigate. As a workaround, I've removed `lazy`. It's not ideal because failure to load the service instances will result in the class itself failing to load but it seems reasonable since it won't be a silent failure.

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
```diff
@@ -892,7 +892,7 @@ class AsyncTest extends Test:
                     Async.gather(1, emitIsolate)(
                         for
                             _ <- Emit("a1")
-                            _ <- Async.sleep(2.millis)
+                            _ <- Async.sleep(10.millis)
```
fwbrasil, line 895: I've made a few test threshold adjustments for tests that were flaky in Native.

hearnadam (approved):

kyo-core/native/src/main/scala/kyo/queuesStubs.scala
```diff
@@ -0,0 +1,37 @@
+package org.jctools.queues
+
+import java.util.ArrayDeque
+import java.util.concurrent.ArrayBlockingQueue
+import java.util.concurrent.LinkedBlockingQueue
+import scala.annotation.tailrec
+
+class StubQueue[A](capacity: Int) extends LinkedBlockingQueue[A](capacity):
```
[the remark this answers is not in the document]
fwbrasil: agreed!

PR #876, opened 2024-11-29
[sttp][combinators] scala native support

fwbrasil:


kyo-sttp/native/src/main/scala/kyo/PlatformBackend.scala
```diff
@@ -0,0 +1,16 @@
+package kyo
+
+import kyo.Requests.Backend
+import sttp.client3.*
+
+object PlatformBackend:
```
fwbrasil, line 6: `kyo-sttp` doesn't have tests for JS and I'm also not introducing tests for Native. The challenge is how to provide a server for the tests. I'll open a ticket for it.

kyo-sttp/native/src/main/scala/kyo/PlatformBackend.scala
```diff
@@ -0,0 +1,16 @@
...
+    val default =
+        new Backend:
+            val b = CurlBackend()
+            def send[A: Flat](r: Request[A, Any]) =
+                given Frame = Frame.internal
+                def call    = r.send(b)
+                Abort.run[Throwable](IO(call))
```
hearnadam, line 13: Does this need `Abort.catching`?

hearnadam (approved):

kyo-sttp/native/src/main/scala/kyo/PlatformBackend.scala
fwbrasil: I don't think so. `Abort.run` already catches

PR #878, opened 2024-11-29
[core] STM effect

fwbrasil:
See scaladocs for more info


kyo-core/js/src/main/scala/kyo/safePublish.scala
```diff
@@ -0,0 +1,12 @@
...
+import scala.annotation.StaticAnnotation
+
+/** Enforces Java Memory Model's safe publication guarantees for final fields (vals).
+ *
+ * When applied to a field or class, ensures strict JMM semantics rather than Scala Native's
+ * default relaxed memory model. Has no effect on JVM or JavaScript platforms.
+ *
+ * @see [[scala.scalanative.annotation.safePublish]]
+ */
+class safePublish extends StaticAnnotation
```
fwbrasil, line 12: I made the annotation available on all platforms so it can be used from a `shared` source folder.

kyo-core/shared/src/test/scala/kyo/Test.scala
```diff
@@ -16,7 +16,11 @@ abstract class Test extends AsyncFreeSpec with BaseKyoTest[Abort[Any] & Async &
 
     def run(v: Future[Assertion] < (Abort[Any] & Async & Resource)): Future[Assertion] =
         import AllowUnsafe.embrace.danger
-        val a = Async.run(Abort.run(Resource.run(v)).map(_.fold(e => throw new IllegalStateException(s"Test aborted with $e"))(identity)))
+        val a = Async.run(Abort.run(Resource.run(v)).map(_.fold {
+            _.getFailure match
+                case ex: Throwable => throw ex
+                case e             => throw new IllegalStateException(s"Test aborted with $e")
```
fwbrasil, line 22: This helped debug failures

build.sbt
```diff
@@ -533,7 +547,7 @@ lazy val readme =
 lazy val `native-settings` = Seq(
     fork                                        := false,
     bspEnabled                                  := false,
-    libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.6.0" % "provided"
+    libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.6.0"
```
fwbrasil, line 550: Unrelated change: I was testing scripts in Scala native with scala-cli and they fail with the missing Java time classes

build.sbt
```diff
@@ -250,6 +253,17 @@ lazy val `kyo-direct` =
         .nativeSettings(`native-settings`)
         .jsSettings(`js-settings`)
 
+lazy val `kyo-stm` =
```
fwbrasil, line 256: I've moved the implementation to a separate module since we'll likely have multiple STM-based implementations like `TPromise`, `TChunk`, `TQueue`, etc. It'll also help reduce compilation times of `kyo-core`.

kyo-stm/shared/src/main/scala/kyo/TMap.scala
```diff
@@ -0,0 +1,280 @@
...
+  *
+  * This architecture is particularly effective at reducing retries in concurrent scenarios by limiting the scope of conflicts between
+  * transactions. Updates to different keys can proceed in parallel, while reads can execute concurrently with writes to different keys. The
+  * implementation maintains strong consistency guarantees while allowing maximum concurrency for non-conflicting operations.
+  *
+  * Operations that modify existing values have low contention characteristics, while structural modifications experience higher contention.
+  * This makes TMap particularly well-suited for scenarios with high concurrent access to different keys where most operations are reads or
+  * updates to existing values.
+  */
+opaque type TMap[K, V] = TRef[Map[K, TRef[V]]]
```
fwbrasil, line 18: This approach is different from ZIO. It provides more granular locking with the goal to reduce transaction conflicts and retries. It seems ZIO's `TMap` requires locking the map itself in all modifications: https://github.com/zio/zio/blob/series/2.x/core/shared/src/main/scala/zio/stm/TMap.scala

ghostdogpr: Are you referring to the lock in the zio implementation? It's only used on resizing the array which doesn't happen on all modifications.

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -0,0 +1,170 @@
...
+                            Var.setAndThen(log + (refAny -> entry))(f(state.value))
+                        end if
+                    }
+                }
+            end if
+        }
+
+    def set(v: A)(using Frame): Unit < STM =
+        Var.use[RefLog] { log =>
+            val refAny = this.asInstanceOf[TRef[Any]]
```
hearnadam, line 95: Could we have an inline implicit conversion for this?

kyo-stm/shared/src/test/scala/kyo/Test.scala
```diff
@@ -0,0 +1,32 @@
+package kyo
```
hearnadam, line 1: Can this be a symbolic link to core?

kyo-stm/shared/src/main/scala/kyo/TMap.scala
fwbrasil: Based on my reading of ZIO's code, any modifications to `TMap` require a write to the underlying `TArray` reference, which implies a lock at commit time of the map and essentially serializes modifications during commit. In this new encoding of `TMap`, each map entry has its own `TRef`. 

For example, updating an existing key requires only a read lock of the outer map `TRef` (to validate the structure) and a write lock of just that entry's specific `TRef`. This means that concurrent modifications to different existing keys can commit in parallel while reads of other keys can also commit in parallel, with only structural changes forcing full commit-time serialization. The tradeoff is memory overhead since each value needs its own `TRef` but this initial `TRef` implementation is already quite lightweight. 

I'm planning to work on benchmarks to validate these tradeoffs, particularly around concurrent access patterns.

kyo-prelude/shared/src/main/scala/kyo/Var.scala
```diff
@@ -60,6 +60,21 @@ object Var:
...
+    /** Sets a new value and then executes another computation.
+      *
+      * @param value
+      *   The new value to set in the Var
+      * @param f
+      *   The computation to execute after setting the value
+      * @return
+      *   The result of the computation after setting the new value
+      */
+    inline def setAndThen[V, A, S](inline value: V)(inline f: => A < S)(using
```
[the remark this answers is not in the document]
fwbrasil: that's a good point. I added it actually as a minor performance optimization since it reduces from two allocations in the equivalent `Var.set(..).andThen` to a single one. I'd be ok with making it `private[kyo]` but we might eventually need to provide this kind of API to library developers for best performance.

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -0,0 +1,200 @@
...
+  * concurrent readers, while write-heavy workloads may experience more contention due to the need for exclusive write access. The
+  * fine-grained locking strategy means that transactions only conflict if they actually touch the same references, allowing for high
+  * concurrency when different transactions operate on different refs.
+  */
+opaque type STM <: (Var[STM.RefLog] & Abort[FailedTransaction] & Async) =
+    Var[STM.RefLog] & Abort[FailedTransaction] & Async
+
+object STM:
+
+    // The default retry schedule for failed transactions
```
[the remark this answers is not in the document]
fwbrasil: Fixed, it should be a scaladoc

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -0,0 +1,200 @@
...
+    /** Initializes a new transactional reference (TRef) with an initial value. The reference is created within the current transaction.
+      *
+      * @param value
+      *   The initial value to store in the reference
+      * @return
+      *   A new transactional reference containing the initial value
+      */
+    def initRef[A](value: A)(using Frame): TRef[A] < STM =
+        useRequiredTid { tid =>
+            Var.use[RefLog] { log =>
```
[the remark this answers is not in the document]
fwbrasil: yep! I think it should be possible to support effects in the update function using an approach like https://github.com/getkyo/kyo/pull/852. I can follow up on that.

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -0,0 +1,200 @@
...
+                // Can't return `result` directly since it has a pending STM effect
+                // but it's safe to cast because, if there's a parent transaction,
+                // then there's a frame upper in the stack that will handle the
+                // STM effect in the parent transaction's `run`.
+                result.asInstanceOf[A < (Async & Abort[E | FailedTransaction])]
+        }
+
+    end run
+
+    type RefLog = Map[TRef[Any], Entry[Any]]
```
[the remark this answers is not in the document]
fwbrasil: It wasn't trivial because `RefLog` is accessed from multiple source files but I moved it to a proper implementation with an opaque type + extension methods marked as `private[kyo]`.

kyo-stm/shared/src/main/scala/kyo/TRef.scala
fwbrasil: The new `RefLog` extension methods are more type safe and don't require the cast anymore.

kyo-stm/shared/src/test/scala/kyo/Test.scala
fwbrasil: yeah, we need to revisit this approach

fwbrasil:
The tests for Native are flaky. It sometimes reports inconsistent results that shouldn't be possible unless there's an memory model issue. I might have misunderstood what `@safePublish` provides. I'm removing `@safePublish` and disabling Scala Native for the module for now. I can create a follow up ticket for it.

kyo-stm/shared/src/main/scala/kyo/RefLog.scala
```diff
@@ -0,0 +1,25 @@
...
+    val empty: RefLog = Map.empty
+
+    extension (self: RefLog)
+
+        def put[A](ref: TRef[A], entry: Entry[A]): RefLog =
+            self.updated(ref.asInstanceOf[TRef[Any]], entry.asInstanceOf[Entry[Any]])
+
+        def get[A](ref: TRef[A]): Maybe[Entry[A]] =
+            val refAny = ref.asInstanceOf[TRef[Any]]
+            Maybe.when(self.contains(refAny))(self(refAny).asInstanceOf[Entry[A]])
```
hearnadam, line 20: You can also do `Maybe(self.getOrElse(refAny, null))` to avoid the `contains` lookup

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -0,0 +1,180 @@
...
+                                // Attempt to acquire locks and commit the transaction
+                                val (locked, unlocked) =
+                                    // Sort references by identity to prevent deadlocks
+                                    log.toSeq.sortBy((ref, _) => ref.hashCode)
+                                        .span((ref, entry) => ref.lock(entry))
+
+                                if unlocked.nonEmpty then
+                                    // Failed to acquire some locks - rollback and retry
+                                    locked.foreach((ref, entry) => ref.unlock(entry))
+                                    Abort.fail(FailedTransaction(frame))
```
hearnadam, line 123: `retry`?

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -0,0 +1,190 @@
...
+      *
+      * Use this only for static initialization or when you specifically need non-transactional creation. For most cases, prefer `init`.
+      *
+      * @param value
+      *   The initial value for the reference
+      * @return
+      *   A new transactional reference containing the value, within the IO effect
+      */
+    def initNow[A](value: A)(using Frame): TRef[A] < IO =
+        IO.Unsafe(TRef.Unsafe.initNow(value))
```
hearnadam, line 180: I'm not yet sure how important this is to expose outside of the `Unsafe` API.

hearnadam (approved):
I haven't used STM in other effect systems heavily, so I am unclear on the semantics being perfect...

However, the code is quite simple and seems reasonable. Perhaps @ghostdogpr has more feedback.

kyo-stm/shared/src/main/scala/kyo/RefLog.scala
fwbrasil: I believe `getOrElse` will call `contains` anyway. I'd prefer to avoid `null`.

kyo-stm/shared/src/main/scala/kyo/STM.scala
fwbrasil: It's not possible to use it directly because `retry` adds a pending `STM` instead of just `Abort[FailedTransaction]`. It could return `Abort` but returning `STM` simplifies the user-facing effect tracking.

kyo-stm/shared/src/main/scala/kyo/TRef.scala
fwbrasil: It's useful in regular safe code as well. The tests have examples of a ref being initialized first outside of a transaction and then the following transactions using it.

ghostdogpr (approved):
LGTM! Curious if this approach is actually faster than the ZIO one, because the code is quite a lot simpler. We could potentially adopt it in ZIO if benchmarks look better 😄

fwbrasil:
@ghostdogpr thanks for the review! It'd be nice to see the projects exchanging more :)

PR #884, opened 2024-12-02
[stm] TChunk and TSchedule

fwbrasil:
Transactional versions for a couple of `kyo-data` APIs

### DIFF
--- a/kyo-stm/shared/src/main/scala/kyo/TChunk.scala
+++ b/kyo-stm/shared/src/main/scala/kyo/TChunk.scala
@@ -1,0 +1,212 @@
+package kyo
+
+/** A transactional wrapper around Chunk that provides atomic operations on sequences. All operations are performed within STM transactions
+  * to ensure consistency.
+  *
+  * @tparam A
+  *   the type of elements in the chunk
+  */
+opaque type TChunk[A] = TRef[Chunk[A]]
+
+object TChunk:
+
+    given [A]: Flat[TChunk[A]] = Flat.derive[TRef[Chunk[A]]]
+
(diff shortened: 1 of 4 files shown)


kyo-stm/shared/src/main/scala/kyo/TChunk.scala
```diff
@@ -0,0 +1,212 @@
...
+      * Use this only for static initialization or when you specifically need non-transactional creation. For most cases, prefer `init`.
+      *
+      * @param chunk
+      *   The initial chunk to wrap
+      * @return
+      *   A new transactional chunk containing the values, within the IO effect
+      */
+    def initNow[A](chunk: Chunk[A])(using Frame): TChunk[A] < IO = TRef.initNow(chunk)
+
+    extension [A](self: TChunk[A])
```
hearnadam, line 89: Do we need a `use` method?

fwbrasil: The `use` in `TMap` provides key access. Would it be indexed-based for a `Chunk`? It doesn't seem very safe.

hearnadam: My suggestion, was simply `def use[B](f: Chunk[A] => B)` to enable accessing non-exposed methods.

fwbrasil: Thanks for the clarification! Updated

fwbrasil:
I'm starting an optimization round of `kyo-stm` and having this change merged will be helpful to avoid conflicts. Please feel free to provide feedback and I'll address it!

PR #885, opened 2024-12-02
[stm][bench] Ref log optimization + benchmarks

fwbrasil:
Introduces an initial set of benchmarks for the STM module and optimizes ref log usage. I was surprised by how well optimized ZIO's STM is :)

[Benchmark results](https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/308dc726e9a4f5913170d7d49f7a2dcf/raw/19d284bd0fff6556f83a408db993d4f814b56a89/jmh-result.json):

image

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/TRefMultiBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/TRefMultiBench.scala
@@ -14,8 +14,7 @@ class TRefMultiBench(parallelism: Int) extends Bench.ForkOnly(parallelism):
         STM.runtime[IO].flatMap { stm =>
             for
                 refs <- stm.commit(Seq.fill(parallelism)(stm.TVar.of(0)).sequence)
-                x: Seq[IO[Unit]] = refs.map(ref => stm.commit(ref.modify(_ + 1)))
-                _      <- x.parSequence_
+                _      <- refs.map(ref => stm.commit(ref.modify(_ + 1))).parSequence_
                 result <- stm.commit(refs.traverse(_.get).map(_.sum))
             yield result
         }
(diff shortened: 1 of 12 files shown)


kyo-stm/shared/src/main/scala/kyo/TRefLog.scala
```diff
@@ -2,25 +2,23 @@ package kyo
...
   * @note
   *   This is a private implementation detail of the STM system
   */
-opaque type RefLog = Map[TRef[Any], RefLog.Entry[Any]]
+opaque type TRefLog = Map[TRef[Any], TRefLog.Entry[Any]]
 
-private[kyo] object RefLog:
+private[kyo] object TRefLog:
 
-    given Tag[RefLog] = Tag[Map[TRef[Any], Entry[Any]]]
```
fwbrasil, line 17: There was an issue with the handling of `Var[RefLog]` producing a dynamic tag with the `RefLog` type bound at runtime. I've moved all methods that use the tag to the companion object so the tag can be fully inferred statically. I've also played with replacing `Map` with `TreeMap` but the results weren't encouraging.

kyo-stm/shared/src/main/scala/kyo/TRefLog.scala
```diff
@@ -2,25 +2,23 @@ package kyo
...
   * This type is used internally by the STM implementation and should not be accessed directly by application code.
   *
   * @note
   *   This is a private implementation detail of the STM system
   */
-opaque type RefLog = Map[TRef[Any], RefLog.Entry[Any]]
+opaque type TRefLog = Map[TRef[Any], TRefLog.Entry[Any]]
 
-private[kyo] object RefLog:
+private[kyo] object TRefLog:
```
fwbrasil, line 15: Renamed for consistency.

hearnadam (approved):

kyo-bench/src/main/scala/kyo/bench/TRefMultiBench.scala
```diff
@@ -0,0 +1,46 @@
...
+        end for
+    end kyoBenchFiber
+
+    def zioBench() =
+        import zio.*
+        import zio.stm.*
+
+        for
+            refs   <- ZIO.collectAll(Seq.fill(parallelism)(TRef.make(0).commit))
+            _      <- ZIO.collectAllPar(refs.map(_.update(_ + 1).commit))
```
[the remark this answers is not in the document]
fwbrasil: that's better! thanks

PR #889, opened 2024-12-03
[data] add `ChunkBuilder` to create StrictOptimizedSeqFactory[Chunk]

hearnadam:


kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -4,6 +4,8 @@ import Chunk.Indexed
 import java.util.Arrays
 import scala.annotation.tailrec
 import scala.annotation.targetName
+import scala.collection.SeqFactory
```
sideeffffect, line 7: Where is this used?

fwbrasil (approved):

kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -354,9 +355,7 @@ sealed abstract class Chunk[A] extends Seq[A] derives CanEqual:
                 c.toArray
 
 end Chunk
```
sideeffffect, line 357: Have you also considered `Chunk` extending `StrictOptimizedSeqOps`?
It seems like the sibling of `StrictOptimizedSeqFactory`, but for the class itself, instead of the companion object.

For the record, ZIO's Chunk extends
 * `IndexedSeq`
 * `IndexedSeqOps`
 * `StrictOptimizedSeqOps`
 * `IterableFactoryDefaults`

https://github.com/zio/zio/blob/series/2.x/core/shared/src/main/scala-2.13%2B/zio/ChunkLike.scala#L40-L43

hearnadam: We definitely should be extending `IndexedSeq`, not `Seq`. However, I will need to take a look at the implications of these other traits. From a quick look `StrictOptimizedSeqOps` simply uses the Builder to optimize the cost of each operator.

This is worth a followup to avoid further changes in this PR.

fwbrasil: Let's make sure we need those extensions. One of the main costs of scala collections is polymorphic interface dispatches. They require a linear search on the interfaces implemented by the object. More interfaces can incur into more overhead.

PR #890, opened 2024-12-03
[core] exit KyoApp on failure

### LINKED ISSUE #790: Return non-zero exit code on Kyo program failure
* https://github.com/getkyo/kyo/pull/785#issuecomment-2443249265

@fwbrasil 

> No, we don't handle exit codes yet. Do you mind creating an issue for it?

hearnadam:
Closes #790

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
+++ b/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
@@ -41,7 +41,7 @@ object KyoApp:
             this._args = args
             for proc <- initCode do proc()
 
-        protected def exit(code: Int): Unit = java.lang.System.exit(code)
+        protected def exit(code: Int): Unit = kernel.Platform.exit(code)
 
         protected def onResult(result: Result[Any, Any]): Unit =
             if !result.exists(().equals(_)) then
--- a/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
+++ b/kyo-core/shared/src/main/scala/kyo/KyoApp.scala
@@ -12,11 +12,9 @@ import scala.collection.mutable.ListBuffer
 abstract class KyoApp extends KyoApp.Base[Async & Resource & Abort[Throwable]]:
     def timeout: Duration = Duration.Infinity
 
-    override protected def handle[A: Flat](v: A < (Async & Resource & Abort[Throwable]))(using Frame): Unit =
+    override protected def handle[A: Flat](v: A < (Async & Resource & Abort[Throwable]))(using Frame): Result[Any, A] =
         import AllowUnsafe.embrace.danger
-        val result = KyoApp.Unsafe.runAndBlock(timeout)(v)
-        if !result.exists(().equals(_)) then
-            println(pprint.apply(result).plainText)
+        KyoApp.Unsafe.runAndBlock(timeout)(v)
     end handle
 end KyoApp
 
@@ -31,7 +29,7 @@ object KyoApp:
       */
     abstract class Base[S]:
 
-        protected def handle[A: Flat](v: A < S)(using Frame): Unit
+        protected def handle[A: Flat](v: A < S)(using Frame): Result[Any, A]
 
         final protected def args: Array[String] = _args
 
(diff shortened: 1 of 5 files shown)


hearnadam:
@sideeffffect

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -43,8 +41,18 @@ object KyoApp:
             this._args = args
             for proc <- initCode do proc()
 
+        protected def exit(code: Int): Unit = kernel.Platform.exit(code)
+
+        protected def onResult(result: Result[Any, Any]): Unit =
+            if !result.exists(().equals(_)) then
+                println(pprint.apply(result).plainText)
```
sideeffffect, line 48: Logging (JUL) instead of raw println?

hearnadam: I think we can followup. I prefer that it's not a logger as it works without configuration + doesn't print `[INFO]`... or whatever.

sideeffffect: I may be wrong about this, but I've always thought that logger would work even without any explicit configuration.

 Generally speaking,
 * the traditional place to put diagnostic information like this on Unix has been the _Standard error output_.
 * More modern approach is to use some sort of logger, and I would expect that (unless configured otherwise) it will spill it out to STDERR as well. But the advantage is that it is more easily instrumentable.

STDOUT just didn't feel principled to me, so that's why I pointed it out in my comment above.

> we can followup

No worries. I'm writing this message to explain more my reasoning for this, not to argue. My original message was rather terse.

sideeffffect (approved):

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -43,8 +41,18 @@ object KyoApp:
...
+            result match
+                case Result.Error(_) => exit(1)
+                case _               => ()
```
sideeffffect, line 51:
```suggestion
                case _               => exit(0)
```
Maybe this could be like this to make it super regular and predictable?

hearnadam: A `KyoApp` can contain multiple runs. We should only exit on failure to prevent prematurely exiting.

sideeffffect: > can contain multiple runs

Oh, that caught me by surprise, I didn't know that could be the case.

On another note, it's it preferable to throw an exception in a situation like this? Such exception shouldn't be handled in the program and should result in its termination. I assume that all our target runtime (JVM, Scala Native and Scala.js) will return a correct, i.e. non-zero, code on terminating like this.

hearnadam: The exception will already be printed via the pretty print. We could consider printing only success and re-throwing exceptions.

The issue is `Result[Any, A]` may contain other `Fail` values than `Throwable`.

fwbrasil: > The issue is Result[Any, A] may contain other Fail values than Throwable.

It's possible to pattern match to handle exceptions differently:
```scala
result match
    case Result.Error(ex: Throwable) => throw ex
    case Result.Error(_: ) => exit(1)
    case _               => ()
```

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
hearnadam: My original idea was actually to make this effectful, then use the Kyo `Log` local to fetch the logger, then invoke the APIs using `Unsafe`...but they use `IO` - not general effects...Then I also realized the default slf4j logger may not output without a config file.

JUL will work without config, but it's not suited for high performance applications. Disabling a logging level does not reduce the cost significantly (if at all). It's also very expensive to translate JUL -> slf4j

sideeffffect: > JUL will work without config, but it's not suited for high performance applications.

> It's also very expensive to translate JUL -> slf4j

My assumption around this is that these error logs will happen very rarely. If this gets triggered only once (or a few times) per application lifetime, you don't need to worry about "high performance" or "expensive to translate". At that point, it's much more important to be interoperable.

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
sideeffffect: This could work well :+1: 

I just want to point out that having to branch based on whether the error is `<: Throwable` is the unfortunate downside of allowing Result to have error which aren't `<: Throwable`. If all errors in were `<: Throwable`, we could `throw` all of them.

hearnadam:
Canceling this as my branch is quite stale. Will followup later.

Issue #491, opened 2024-06-20
Shutdown gracefully on SIGINT


steinybot:
Whilst working on #390 I had this:
```
  run {
    for {
      // TODO: Get the shutdown working properly.
      _ <- Consoles.println(s"Server is running on port $port. Press Ctrl-C to stop.")
      server <- Resources.acquireRelease(IOs(buildServer(port, services).start())) { (server: Server) =>
        IOs.run(Consoles.run(Consoles.print("Shutting down...")))
        // TODO: Add a timeout.
        server.shutdown().awaitTermination()
        IOs.run(Consoles.run(Consoles.println("Done.")))
      }
      _ <- Fibers.sleep(Duration.Infinity)
    } yield {
      "Goodbye!"
    }
  }
```

Which doesn't currently stop on Ctrl-C (SIGINT).

We had a brief chat about it in https://discord.com/channels/1087005439859904574/1245754417840394313/1253415536138653788 and @fwbrasil suggestion was to add something to `Resources` to handle this.

steinybot:
Workaround:
```
  def waitForInterrupt: Unit < Fibers =
    for {
      promise <- Fibers.initPromise[Unit]
      _ <- IOs(Signal.handle(new Signal("INT"), _ => IOs.run(promise.complete(())))).unit
      _ <- IOs(Signal.handle(new Signal("TERM"), _ => promise.complete(IOs.fail("Terminated.")))).unit
      _ <- promise.get
    } yield ()
```

PR #893, opened 2024-12-04
[core] add `Signal.Handler` class to install shutdown hook

hearnadam:
This is step 1 of solving #491. Next, we need a hook to shutdown the scheduler/interrupt tasks.

### DIFF
--- a/kyo-core/jvm/src/main/scala/kyo/internal/SignalPlatformSpecific.scala
+++ b/kyo-core/jvm/src/main/scala/kyo/internal/SignalPlatformSpecific.scala
@@ -1,0 +1,89 @@
+package kyo.internal
+
+import Signal.Handler
+import java.lang.invoke.MethodHandle
+import java.lang.invoke.MethodHandles
+import java.lang.invoke.MethodType
+import java.lang.reflect.InvocationHandler
+import java.lang.reflect.Method
+import java.lang.reflect.Proxy
+import java.util.logging.Logger
+import kyo.Result
+
+/** This class provides a jvm-specific implementation of signal handling. It uses reflection to install signal handlers as they may not be
+  * available on all implementations of the JVM.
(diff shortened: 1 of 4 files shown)


hearnadam:
The JVM reflection code is based on ZIO's implementation by @kyri-petrou https://github.com/zio/zio/blob/series/2.x/core/jvm/src/main/scala/zio/internal/Signal.scala

I took the liberty of using `Result` to properly route errors to the log message.

kyri-petrou:
As a side note, as this approach relies on java reflection, it might be worth adding these files to make the signal handler compatible with graalvm's native-image compilation

hearnadam:
Yes, I noticed that. My plan is to tackle all our configuration for reflection at once, since I think we use reflection elsewhere.

kyo-core/jvm/src/main/scala/kyo/internal/SignalPlatformSpecific.scala
```diff
@@ -0,0 +1,89 @@
...
+    private val logger = Logger.getLogger("kyo.internal.Signal")
+    val handle: Handler = {
+        for
+            signalClass        <- findClass("sun.misc.Signal")
+            signalHandlerClass <- findClass("sun.misc.SignalHandler")
+            lookup = MethodHandles.lookup()
+            constructorHandle  <- initSignalConstructorMethodHandle(lookup, signalClass)
+            staticMethodHandle <- initHandleStaticMethodHandle(lookup, signalClass, signalHandlerClass)
+        yield SunMisc(signalHandlerClass, constructorHandle, staticMethodHandle)
+    }.fold { error =>
```
fwbrasil, line 26: It seems `Abort.recover` wrapping the `for` would be cleaner

fwbrasil (approved):

PR #895, opened 2024-12-04
[core] add System.workingDirectory

### LINKED ISSUE #655: Neither Path nor Process provides a means to get cwd/pwd
While `Process` allows you to set the `cwd`, I don't see a way to get the `cwd`, either through `Process` or `path`.

hearnadam:
Resolves  #655


fwbrasil:
The build is failing because `Path` isn't available in JS. It's under the JVM folder. I think this might already be covered: https://github.com/getkyo/kyo/blob/5d5f0c7d16a29624c8a207d021cbf5f6426a7a63/kyo-core/jvm/src/main/scala/kyo/Path.scala#L372

Issue #881, opened 2024-11-30
Stream: interop java.util.concurrent.Flow Reactive Stream


calvinlfer:
Implement support to interop with java.util.concurrent.Flow's Publisher/Subscriber/Subscription to and from Kyo effects

This is a follow-up to #679 

For example, see how [FS2 Streams](https://github.com/typelevel/fs2/tree/main/core/shared/src/main/scala/fs2/interop/flow) does it

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/896

PR #896, opened 2024-12-04
Reactive-stream/Flow Interop

### LINKED ISSUE #764: Requests: support sttp streaming
Streaming HTTP responses is quite useful. It should be a standard feature of `kyo-sttp`

HollandDM:
Originally, this was created just to resolve #764, but since #881 was requested, I took sometime to create a dedicated PR for it.
This took both ZIO's and Fs2's approaches as the inspiration, and passes the [reactive-stream-tck](https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck-flow) testcases. I also add some test cases derived from Fs2's codebase. Hopefully it covers enough.
With this, solving #764 will be more trivial.


kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscription.scala
```diff
@@ -0,0 +1,255 @@
...
+        }
+        sideEffect()
+    end subscribe
+
+    private[reactivestreams] def consume(
+        using
+        tag: Tag[Emit[Chunk[V]]],
+        frame: Frame,
+        safepoint: Safepoint
+    ): Fiber[Nothing, StreamFinishState] < (IO & Ctx) =
```
HollandDM, line 117: Basically, we need to fork this into a new fiber, which will run the stream until completion. Maybe Poll can help us eliminate the need to fork, which results in eliminating boundary

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscription.scala
```diff
@@ -0,0 +1,148 @@
...
+    end request
+
+    override def cancel(): Unit =
+        given Frame = Frame.internal
+        discard(requestChannel.close())
+    end cancel
+
+    private[reactivestreams] inline def subscribe: Unit < IO = IO(subscriber.onSubscribe(this))
+
+    private[reactivestreams] def poll: StreamFinishState < (Async & Poll[Chunk[V]]) =
```
HollandDM, line 38: `Poll` is great, is make the logic much cleaner then the previous attemp

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/package.scala
```diff
@@ -0,0 +1,33 @@
...
+import kyo.*
+import kyo.interop.reactivestreams.*
+import kyo.kernel.Boundary
+import org.reactivestreams.*
+import org.reactivestreams.FlowAdapters
+
+package object reactivestreams:
+    def fromPublisher[T](
+        publisher: Publisher[T],
+        bufferSize: Int
```
calvinlfer, line 12: I had a question regarding this, do we wait for the entire buffer to be filled up before emitting them as a Chunk on the Stream? Is it possible to have some kind of choice with regards to size & time?

The reason I ask this is because on an infinite stream (let's say we're integrating with a reactive streams kafka library like Helidon or Vertx Kafka), if this buffer size > number of messages on the topic unconsumed then the messages will be held in the buffer until number of messages on the topic unconsumed >= buffer size.

I observed this limitation with the Helidon Kafka client when I was integrating with the FS2 Streams Flow interop where if you used a low buffer size, you can overcome this limitation at the cost of performance since chunk-size = 1 has much worse performance

HollandDM: Current implementation tries to emit the value to stream as soon as it receives one. Buffer size was used mostly as a params of subscription’s ’request’.

HollandDM: I did notice this behavior of fs2 stream and I’m also not a fan of it. So I took the Zio approach. But maybe we can have some kind of “strategy” or different implementations so that users can  choose between them. WDYT

calvinlfer: That would be excellent! Amazing work 🙇

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/package.scala
```diff
@@ -0,0 +1,33 @@
+package kyo.interop
+
+import java.util.concurrent.Flow.*
+import kyo.*
+import kyo.interop.reactivestreams
+import kyo.kernel.Boundary
+import org.reactivestreams.FlowAdapters
+
+package object flow:
```
fwbrasil, line 9: What do you think about exposing these operations as `Stream` extensions in the `kyo` package instead? I'd be nice to keep a companion object as well since people might want to import it in isolation:

```scala
package kyo

object StreamReactiveStreamsExtensions:
     extension [V, S](self: Stream[V, S]):
         // instance methods
     extension (self: Stream.type):
         // companion methods
         
 // export to the kyo package
 export StreamReactiveStreamsExtensions.*
```

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/package.scala
```diff
@@ -0,0 +1,33 @@
...
+    ): Stream[T, Async] < IO = reactivestreams.fromPublisher(FlowAdapters.toPublisher(publisher), bufferSize)
+
+    def subscribeToStream[T, Ctx](
+        stream: Stream[T, Ctx],
+        subscriber: Subscriber[? >: T]
+    )(
+        using
+        Boundary[Ctx, IO],
+        Frame,
+        Tag[T]
```
fwbrasil, line 26: It'd be ideal to take `Tag[Emit[Chunk[T]]]` evidences directly otherwise the tag macro has to dynamically compose tags, which requires allocation.

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/package.scala
```diff
@@ -0,0 +1,33 @@
...
+    ): Stream[T, Async] < IO = reactivestreams.fromPublisher(FlowAdapters.toPublisher(publisher), bufferSize)
+
+    def subscribeToStream[T, Ctx](
+        stream: Stream[T, Ctx],
+        subscriber: Subscriber[? >: T]
+    )(
+        using
+        Boundary[Ctx, IO],
```
fwbrasil, line 24: We have some inference issues when using `Boundary` directly. We've been using a workaround with a user-facing `inline` method calling another private one that then requires the `Boundary` evidence. See `Async` for examples.

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamPublisher.scala
```diff
@@ -0,0 +1,124 @@
...
+            subscriber.onComplete()
+        end discardSubscriber
+
+        def consumeChannel(
+            channel: Channel[Subscriber[? >: V]],
+            supervisorPromise: Fiber.Promise[Nothing, Unit]
+        ): Unit < (Async & Ctx) =
+            Loop(()) { _ =>
+                channel.closed.map {
+                    if _ then
```
fwbrasil, line 52: can you use `case true => ..` `case false => ...` instead?

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscriber.scala
```diff
@@ -0,0 +1,209 @@
...
+
+    private val state = AtomicRef.Unsafe.init(
+        UpstreamState.Uninitialized -> Maybe.empty[Fiber.Promise.Unsafe[Nothing, Unit]]
+    )
+
+    private def throwIfNull[A](b: A): Unit = if isNull(b) then throw new NullPointerException()
+
+    override def onSubscribe(subscription: Subscription): Unit =
+        throwIfNull(subscription)
+        var sideEffect: () => Unit = () => ()
```
fwbrasil, line 29: would it be possible to avoid the mutable value by using `state.get` and `state.set` directly instead of `state.update`?

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscription.scala
fwbrasil: cool!!

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscription.scala
fwbrasil: Yeah, I can't see why forking is necessary. Wouldn't `Poll.run(stream.emit)(poll).map(_._2)` directly work?

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscription.scala
```diff
@@ -0,0 +1,148 @@
...
+
+    private[reactivestreams] def consume(
+        using
+        emitTag: Tag[Emit[Chunk[V]]],
+        pollTag: Tag[Poll[Chunk[V]]],
+        frame: Frame,
+        safepoint: Safepoint
+    ): Fiber[Nothing, StreamFinishState] < (IO & Ctx) =
+        boundary { (trace, context) =>
+            val fiber = Fiber.fromTask(IOTask(Poll.run(stream.emit)(poll).map(_._2), safepoint.copyTrace(trace), context))
```
fwbrasil, line 96: `IOTask` is an internal detail of `Fiber`/`Async`. Ideally, we should rely on the existing APIs instead of using it directly

build.sbt
```diff
@@ -307,6 +308,22 @@ lazy val `kyo-cache` =
...
+        .withoutSuffixFor(JVMPlatform)
+        .crossType(CrossType.Full)
+        .in(file("kyo-reactive-streams"))
+        .dependsOn(`kyo-core`)
+        .settings(
+            `kyo-settings`,
+            libraryDependencies ++= Seq(
+                "org.reactivestreams" % "reactive-streams" % "1.0.4",
+                "org.reactivestreams" % "reactive-streams-tck" % "1.0.4" % Test,
+                "org.scalatestplus" %% "testng-7-5" % "3.2.17.0" % Test
```
fwbrasil, line 322: Is this required to reuse the tck tests? Would it be possible to use scalatest like the other modules?

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscription.scala
HollandDM: Other lib like ZIO and Fs2 also has to fork. AFAIK, it's mainly because draining stream is a "runForever" task until interrupted or stream is done. Without forking it will block other thing from being done in the current thread.
I'll try to figure out if I can eliminate it

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscriber.scala
HollandDM: yes its can be done, but It'll need to be converted into loop method for correctness

build.sbt
HollandDM: We'll need to reimplement those test in scalatest though.

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/package.scala
HollandDM: sounds good, I'' update this

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscription.scala
HollandDM: this is just a copy of `Async._run` with boundary that does not care about `Abort[E]` (`E =:= Nothing` in this case). Maybe after `https://github.com/getkyo/kyo/pull/906` is merged I can use `_run` instead

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscriber.scala
fwbrasil: Makes sense. If a `Loop` is too cumbersome, another option is introducing an `update` that allows retuning a "tail" computation

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/package.scala
HollandDM: now that I though about it, moving this into kyo package would polute it with reactive-streams dependency. I don't think that's a good ideal, right?

HollandDM:
@fwbrasil updated, please recheck

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscription.scala
```diff
@@ -0,0 +1,173 @@
...
+        )(
+            using
+            AllowUnsafe,
+            Frame,
+            Tag[Emit[Chunk[V]]],
+            Tag[Poll[Chunk[V]]]
+        ): StreamSubscription[V, Ctx] =
+            _subscribe(stream, subscriber)(subscribeCallback)
+
+        private[kyo] inline def _subscribe[V, Ctx](
```
HollandDM, line 154: Same trick with `Async._run`, now we don't have to pass context params explicitly

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscriber.scala
```diff
@@ -0,0 +1,283 @@
...
+    def stream(using Frame, Tag[Emit[Chunk[V]]]): Stream[V, Async] = Stream(emit)
+
+end StreamSubscriber
+
+object StreamSubscriber:
+
+    abstract private[reactivestreams] class SubscriberDone
+    private[reactivestreams] case object SubscriberDone extends SubscriberDone
+
+    enum EmitStrategy derives CanEqual:
```
HollandDM, line 272: introduce `EmitStrategy` so that users can choose the behavior they want

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscription.scala
```diff
@@ -0,0 +1,173 @@
...
+    end poll
+
+    private[interop] def consume(
+        using
+        Tag[Emit[Chunk[V]]],
+        Tag[Poll[Chunk[V]]],
+        Frame,
+        Boundary[Ctx, IO & Abort[Nothing]]
+    ): Fiber[Nothing, StreamFinishState] < (IO & Ctx) =
+        Async
```
HollandDM, line 93: replace `boundary` usage with `Async._run`

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/package.scala
fwbrasil: It'd still be limited to the specific module. `kyo-reactive-streams` needs to be in the classpath for the extension methods to be available

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscriber.scala
fwbrasil: Nice! We could eventually introduce more advanced strategies including timeouts for example

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamSubscriber.scala
```diff
@@ -0,0 +1,283 @@
...
+                    end if
+                case other =>
+                    if !state.compareAndSet(curState, other) then handleComplete()
+            end match
+        end handleComplete
+        handleComplete()
+    end onComplete
+
+    private[interop] def await(using Frame): Boolean < Async =
+        @tailrec def handleAwait: Boolean < Async =
```
fwbrasil, line 110:
```suggestion
        @tailrec def handleAwait(): Boolean < Async =
```

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamPublisher.scala
```diff
@@ -0,0 +1,114 @@
...
+end StreamPublisher
+
+object StreamPublisher:
+
+    def apply[V, Ctx](
+        stream: Stream[V, Ctx],
+        capacity: Int = Int.MaxValue
+    )(
+        using
+        Boundary[Ctx, IO & Abort[Nothing]],
```
fwbrasil, line 34: you probably don't need `Abort[Nothing]` anymore since it's included in `IO`

kyo-reactive-streams/shared/src/main/scala/kyo/interop/reactive-streams/StreamPublisher.scala
```diff
@@ -0,0 +1,114 @@
...
+                                subscriber   <- channel.take
+                                subscription <- IO.Unsafe(new StreamSubscription[V, Ctx](stream, subscriber))
+                                fiber        <- subscription.subscribe.andThen(subscription.consume)
+                                _            <- supervisor.onComplete(_ => discard(fiber.interrupt(interruptPanic)))
+                            yield ()
+                        }.map {
+                            case Result.Success(_) => Loop.continue(())
+                            case _                 => Loop.done
+                        }
+                }
```
fwbrasil, line 68: simplification suggestion:
```scala
                  Loop(()) { _ =>
                         Abort.recover[Closed](_ => Loop.done) {
                            for
                                subscriber   <- channel.take
                                subscription <- IO.Unsafe(new StreamSubscription[V, Ctx](stream, subscriber))
                                fiber        <- subscription.subscribe.andThen(subscription.consume)
                                _            <- supervisor.onComplete(_ => discard(fiber.interrupt(interruptPanic)))
                            yield Loop.continue(())
                        }
                 }
```

kyo-reactive-streams/shared/src/test/scala/kyo/interop/reactive-streams/StreamPublisherTest.scala
```diff
@@ -0,0 +1,49 @@
+package kyo.interop.reactivestreams
+
+import kyo.*
+import kyo.interop.reactivestreams.StreamPublisher
+import org.reactivestreams.tck.PublisherVerification
+import org.reactivestreams.tck.TestEnvironment
+import org.scalatestplus.testng.*
+
+final class StreamPublisherTest extends PublisherVerification[Int](new TestEnvironment(1000L)), TestNGSuiteLike:
```
fwbrasil, line 9: I took at the tck tests and it seems they check mostly for correctness without concurrency. If that's the case, I think we'll need to introduce tests since the code relies heavily on shared state

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/package.scala
HollandDM: I understand what you mean now, will update!

HollandDM:
I moved the implement from extending `reactive-streams` to extending `flow`. I think it's more suitable for `Kyo`

kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/SreamSubscriberTest.scala
```diff
@@ -0,0 +1,86 @@
...
+        requestCount.set(0)
+        discard(CompletableFuture.allOf(
+            List.fill(batchSize)(CompletableFuture.runAsync(loop()))*
+        ).get())
+    end request
+
+    override def cancel(): Unit = isCanceled.set(true)
+end JavaSubscription
+
+final class SreamSubscriberTest extends Test:
```
HollandDM, line 31: I decided to test `StreamSubscriber` part only, as it's also the only part that use shared atomic state

kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
```diff
@@ -0,0 +1,198 @@
...
+            publisher  <- stream.toPublisher
+            subscriber <- streamSubscriber
+            _ = publisher.subscribe(subscriber)
+            _ <- subscriber.stream.take(10).runDiscard
+            _ <- stopPromise.get
+        yield assert(true)
+        end for
+    }
+
+    "single publisher & multiple subscribers" - {
```
HollandDM, line 77: Added some test between publisher and subscribers

kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/SreamSubscriberTest.scala
```diff
@@ -0,0 +1,86 @@
+package kyo.interop.flow
+
+import StreamSubscriber.EmitStrategy
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.Flow.*
+import kyo.*
+import kyo.Maybe.*
+
+final class JavaSubscription(subscriber: Subscriber[? >: Int], batchSize: Int, counter: java.util.concurrent.atomic.AtomicInteger)
```
HollandDM, line 9: We need to test a generic `Publisher`, so I create a generic Java copies

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/StreamSubscription.scala
```diff
@@ -0,0 +1,154 @@
...
+    end request
+
+    override def cancel(): Unit =
+        given Frame = Frame.internal
+        discard(requestChannel.close())
+    end cancel
+
+    private[interop] inline def subscribe(using Frame): Unit < IO = IO(subscriber.onSubscribe(this))
+
+    private[interop] def poll(using Tag[Poll[Chunk[V]]], Frame): StreamFinishState < (Async & Poll[Chunk[V]]) =
```
HollandDM, line 36: I tried to convert this into `channel.stream()`, but this provides more controls over the computation

fwbrasil (approved):
This is a sizable contribution, congrats! LGTM but I think you'll need to resolve conflicts with https://github.com/getkyo/kyo/pull/929

HollandDM:
@fwbrasil thanks, this PR should be good now

PR #897, opened 2024-12-04
[core] more clear Console method names

fwbrasil:


hearnadam (approved):
At some point (closer to 1.0), we should start including scalafix rules.

PR #899, opened 2024-12-04
[monix] scaladocs

fwbrasil:


kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -6,11 +6,6 @@ import scala.annotation.tailrec
 
 /** A transactional reference that can be modified within STM transactions. Provides atomic read and write operations with strong
   * consistency guarantees.
-  *
```
fwbrasil, line 9: minor unrelated scaladoc fix

hearnadam (approved):

PR #901, opened 2024-12-04
[stm] fix race condition during commit + enable Native

### LINKED ISSUE #898: [stm] Scala Native support
The `kyo-stm` tests in Native were flaky during development so the cross compilation was disabled. We need to re-enable the cross compilation and investigate the test failures. The failures weren't very common. The project has a `repeat` command that repeats an operation that can be helpful to reproduce: `sbt> repeat 50 kyo-stm/test`. More background: https://github.com/getkyo/kyo/pull/878#issuecomment-2509605893

Given the failures I was observing, it seems there's some issue with the memory model in Native that doesn't exactly match the JVM's behavior. I tried using `@safePublish` but it didn't fix the issue.

fwbrasil:
Fixes #898

I saw flaky JVM tests in CI and decided to look into it. The test failures in Native were actually due to a race condition during commit.

### DIFF
--- a/kyo-stm/shared/src/test/scala/kyo/STMTest.scala
+++ b/kyo-stm/shared/src/test/scala/kyo/STMTest.scala
@@ -476,11 +476,12 @@ class STMTest extends Test:
 
     "Concurrency" - {
 
-        val repeats = 100
+        val repeats = 10
+        val sizes   = Seq(1, 10, 100, 1000)
 
         "concurrent updates" in run {
             (for
-                size  <- Choice.get(Seq(1, 10, 100))
+                size  <- Choice.get(sizes)
                 ref   <- TRef.initNow(0)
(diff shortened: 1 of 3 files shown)


kyo-stm/shared/src/test/scala/kyo/STMTest.scala
```diff
@@ -476,11 +476,12 @@ class STMTest extends Test:
 
     "Concurrency" - {
 
-        val repeats = 100
+        val repeats = 10
+        val sizes   = Seq(1, 10, 100, 1000)
```
fwbrasil, line 480: I was able to more reliably reproduce the issue in the JVM by increasing the max concurrency of the test to 1000. I've reduced `repeats` so the tests don't take too long

hearnadam (approved):

PR #902, opened 2024-12-05
orDie combinator

johnhungerford:
- Removes `Abort[E]` effect for any `E`
- If `E` is `Throwable`, throw any failures
- If `E` is not `Throwable`, throw a `PanicException` that contains `E`


johnhungerford:
Addresses #651

johnhungerford:
Snuck in a few more changes:

- Supporting union aborts (I didn't realize this was fixed with SafeClassTag, and was still using Tag)
- Improved (I think) handling of some aborts
   - Previously, to handle some aborts, you had to do `effect.handleSomeAbort[ErrorType]()`. The need to call `()` at the end kept throwing me off and I think is going to be a big wart for users. It was even worse for `catchSomeAbort`
   - Now I have a single ops class for all of these `...SomeAbort` methods
      - `effect.someAbort[ErrorType].handled`
      - `effect.someAbort[ErrorType].caught(e => ...)`
      - `effect.someAbort[ErrorType].caughtPartial { case ... => ... }`
      - `effect.someAbort[ErrorType].toEmpty`
      - `effect.someAbort[ErrorType].toChoice`

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -386,6 +376,24 @@ extension [A, S, E](effect: A < (Abort[E] & S))
...
+        ct: SafeClassTag[E],
+        fl: Flat[A],
+        frame: Frame
+    ): A < S =
+        Abort.run[E](effect).map(
+            _.fold(_.getFailure match
+                case thr: Throwable => throw thr
+                case other          => throw PanicException(other)
+            )(identity)
+        )
```
hearnadam, line 396: I don't think we should throw, we should `Abort.panic`. Long term it think core effect handling will need to be aware of Panics...

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -417,90 +425,29 @@ extension [A, S, E](effect: A < (Abort[Absent] & S))
...
-class HandleSomeAbort[A, S, E, E1 <: E](effect: A < (Abort[E] & S)) extends AnyVal:
-
-    /** Handles the Abort[E] effect and returns its result as a `Result[E, A]`.
-      *
-      * @return
-      *   A computation that produces the a `Result[E, A]`
-      */
-    def apply[ER]()(
+class SomeAbortOps[A, S, E, E1 <: E](effect: A < (Abort[E] & S)) extends AnyVal:
+    def handled[ER](
```
hearnadam, line 436: How about calling this `result`? Similar to ZIO's `exit`/`either`?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -536,29 +473,43 @@ class CatchSomeAbortPartialOps[A, S, E, E1 <: E](effect: A < (Abort[E] & S)) ext
...
-class SwapSomeAbortOps[A, S, E, E1 <: E](effect: A < (Abort[E] & S)) extends AnyVal:
+    def toChoice[ER](
+        using
+        ev: E => E1 | ER,
+        ct: SafeClassTag[E1],
+        reduce: Reducible[Abort[ER]],
+        flat: Flat[A],
+        frame: Frame
+    ): A < (S & reduce.SReduced & Choice) =
+        Abort.run[E1](effect.asInstanceOf[A < (Abort[E1 | ER] & S)]).map(e => Choice.get(e.fold(_ => Nil)(List(_))))
```
hearnadam, line 485: Can we avoid the cast with the evidence?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -536,29 +473,43 @@ class CatchSomeAbortPartialOps[A, S, E, E1 <: E](effect: A < (Abort[E] & S)) ext
...
+        ev: E => E1 | ER,
+        ct: SafeClassTag[E1],
+        reduce: Reducible[Abort[ER]],
+        flat: Flat[A],
+        frame: Frame
+    ): A < (S & reduce.SReduced & Abort[Absent]) =
+        Abort.run[E1](effect.asInstanceOf[A < (Abort[E1 | ER] & S)]).map {
+            case Result.Fail(_)        => Abort.get(Result.Fail(Absent))
+            case p @ Result.Panic(e)   => Abort.get(p.asInstanceOf[Result[Nothing, Nothing]])
+            case s @ Result.Success(a) => Abort.get(s.asInstanceOf[Result[Nothing, A]])
```
hearnadam, line 498: `e/a` are unused?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -386,6 +376,24 @@ extension [A, S, E](effect: A < (Abort[E] & S))
     ): A < (S & reduce.SReduced) =
         Abort.run[Throwable](effect.asInstanceOf[A < (Abort[Throwable | ER] & S)])
             .map(_.fold(e => throw e.getFailure)(identity))
+
+    /** Catches any Aborts and panics instead
+      *
+      * @return
+      *   A computation that panics instead of catching Abort effect failures
+      */
+    def orDie(
```
hearnadam, line 385:
```suggestion
    def orPanic(
```

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -408,7 +416,7 @@ extension [A, S, E](effect: A < (Abort[Absent] & S))
       *   A computation that produces the result of this computation with the Abort[Absent] effect translated to Choice
       */
     def emptyAbortToChoice(using Flat[A], Frame): A < (S & Choice) =
```
hearnadam, line 418: Can the evidence be in the extension? that should clean up signatures throughout... or does it reduce clarity of why a method is unavailable?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
johnhungerford: I don't see how. `ev` can convert `E` to `E1 | ER`, another type, of which the `E` is simply a parameter. @fwbrasil any thoughts here?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
johnhungerford: removed e/a

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
johnhungerford: 1. IDEs cannot resolve the methods as quickly
2. The error messages are generally harder to understand when it's on the extension methods

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
hearnadam: Ah you're right. I think the cast is fine.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
johnhungerford: I went through and updated a bunch of the name

handleAbort -> result
catchAbort -> catching
catchAbortPartial -> catchingPartial

someAbort[E].handled -> forAbort[E].result
someAbort[E].caught -> forAbort[E].catching
someAbort[E].caughtPartial -> forAbort[E].catchingPartial

handleEmptyAbort -> maybe

README.md
```diff
@@ -419,7 +417,7 @@ The `defer` method in Kyo mirrors Scala's `for`-comprehensions in providing a co
 
 The `kyo-direct` module is constructed as a wrapper around [dotty-cps-async](https://github.com/rssh/dotty-cps-async).
 
-> Note: `defer` is currently the only user-facing macro in Kyo. All other features use regular language constructs.
+> Note: `defer` is currently one of only two user-facing macros in Kyo. All other features use regular language constructs.
```
fwbrasil, line 420: We have a few macros now. How about removing the note it instead?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
fwbrasil: yep, I think it's not possible to use the `E => E1 | ER ` evidence to transform the pending set

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -227,8 +229,11 @@ extension [A, S](effect: A < S)
       * @return
       *   A computation that produces the result of this computation with Abort[Throwable] effect
       */
-    def explicitThrowable(using Flat[A], Frame): A < (S & Abort[Throwable]) =
-        Abort.catching[Throwable](effect)
+    def resurrect(using Flat[A], Frame): A < (S & Abort[Throwable]) =
+        Abort.run[Throwable](effect: A < (S & Abort[Throwable])).map:
```
fwbrasil, line 233: Is the explicit type annotation for the computation isn't necessary?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -227,8 +229,11 @@ extension [A, S](effect: A < S)
...
+            case Result.Success(v)                      => v
+            case err: Result.Fail[Throwable] @unchecked => Abort.get(err)
+            case err: Result.Panic                      => Abort.fail(err.exception)
```
fwbrasil, line 236: It seems this could be: 
```suggestion
            case Result.Success(v) => v
            case Result.Error(ex)     => Abort.fail(ex)
```

fwbrasil (approved):

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
johnhungerford: Apparently not -- removed

PR #909, opened 2024-12-07
[scheduler] cats-effect integration

fwbrasil:
Benchmark [results](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/24a082e8f25d6ca29e02e0720948af3c/raw/c2eb9471cf6a730ee8a636cec9d887733dc6ef0e/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/24a082e8f25d6ca29e02e0720948af3c/raw/c2eb9471cf6a730ee8a636cec9d887733dc6ef0e/jmh-result-candidate.json) are mixed but use in realistic workloads might yield significant improvements, especially in containerized environments given Kyo's adaptive concurrency control.

image

### DIFF
--- a/kyo-scheduler/shared/src/main/scala/kyo/scheduler/Task.scala
+++ b/kyo-scheduler/shared/src/main/scala/kyo/scheduler/Task.scala
@@ -41,6 +41,14 @@ object Task {
     val Preempted: Result = true
     val Done: Result      = false
 
+    def apply(runnable: Runnable): Task =
+        new Task {
+            def run(startMillis: Long, clock: InternalClock) = {
+                runnable.run()
+                Task.Done
+            }
+        }
+
     def apply(r: => Unit): Task =
         apply(r, 0)
 
--- a/build.sbt
+++ b/build.sbt
@@ -88,6 +88,7 @@ lazy val kyoJVM = project
     .aggregate(
         `kyo-scheduler`.jvm,
         `kyo-scheduler-zio`.jvm,
+        `kyo-scheduler-cats`.jvm,
         `kyo-data`.jvm,
         `kyo-prelude`.jvm,
         `kyo-core`.jvm,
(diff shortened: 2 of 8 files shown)


kyo-scheduler/shared/src/main/scala/kyo/scheduler/Task.scala
```diff
@@ -41,6 +41,14 @@ object Task {
     val Preempted: Result = true
     val Done: Result      = false
 
+    def apply(runnable: Runnable): Task =
```
fwbrasil, line 44: I inspected `SchedulingBench`, the top benchmark with regression, and noticed that we can avoid an allocation with this dedicated constructor. I also noticed that `IOFiber` has a few hot code paths with special treatment for the default work stealing thread pool, which should account for some of the regression.

hearnadam (approved):
LGTM! Can we create a simple mechanisms for users of kyo-zio and kyo-cats use the scheduler? Should those modules depend on the related scheduler modules?

On second thought, do we even need separate modules? Can we just compile the scheduler code for scala2/3 and the effect interop for scala 3?

fwbrasil:
> LGTM! Can we create a simple mechanisms for users of kyo-zio and kyo-cats use the scheduler? Should those modules depend on the related scheduler modules?
> 
> On second thought, do we even need separate modules? Can we just compile the scheduler code for scala2/3 and the effect interop for scala 3?

The goal of the separate artifact is to allow adopting Kyo's scheduler in isolation without the effect system. If we merged the modules, it wouldn't be possible to provide Scala 2 artifacts as well. I think we can address this via documentation mentioning that users can use `KyoSchedulerIOApp` to enable Kyo's scheduler in addition to the regular integration APIs. I'm not sure there's much more we can do other than provide the `App` super class injecting the scheduler automatically.

PR #912, opened 2024-12-09
feat: Enable Stream creation from Queue

### LINKED ISSUE #791: Stream: enable creation from Queue or Channel

mobley-trent:
Closes #791 
/claim #791

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -192,6 +192,7 @@ lazy val `kyo-data` =
         .withoutSuffixFor(JVMPlatform)
         .crossType(CrossType.Full)
         .in(file("kyo-data"))
+        .dependsOn(`kyo-scheduler`)
         .settings(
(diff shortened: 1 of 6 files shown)


mobley-trent:
Hello @fwbrasil, I can't seem to understand why `kyo-prelude` is compiling but `kyo-preludeJS` can't seem to recognize `kyo.scheduler.Queue`. Could there be something I'm missing ?

hearnadam:
`kyo-data` should not depend on `kyo-scheduler`. It's meant to be a standalone module. Additionally, this code is using the wrong Queue. Queue is defined in `kyo-core`, due to requiring IO/Async. The scheduler's queue is optimized for task scheduling, not use with Kyo.

Given the incorrect Queue usage and an implementation that will work for a finite number of elements, I am closing this PR.

Issue #678, opened 2024-09-23
Feature: Poll/Take multiple elements from a Channel


steinybot:
When using a `Channel[A]` to produce a `Stream[A, ?]` via `Ack < Emit[Chunk[A]]` it would be nice to be able to get multiple elements from the channel. I guess there should probably be two versions, one which takes at most `maxItems` (where `maxItems` is from `Continue`) like `poll`, and another which blocks until there are exactly `maxItems` like `take`.

PR #913, opened 2024-12-09
Channel: support taking multiple elements

johnhungerford:
Addresses #678 

Added two methods to `Channel`:
- `drainUpTo`: works basically like drain, but limits the number of elements taken
- `takeExactly`: takes a specified number elements from, semantically blocking to wait for elements to be available

Added one method to `Queue`:
- `drainUpTo`: works as above, and is used in implementation of `Channel#drainUpTo` and `Channel#takeExactly`


kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -180,7 +210,7 @@ object Channel:
         )(using initFrame: Frame, allow: AllowUnsafe): Unsafe[A] =
             new Unsafe[A]:
                 val queue = Queue.Unsafe.init[A](_capacity, access)
-                val takes = new MpmcUnboundedXaddArrayQueue[Promise.Unsafe[Closed, A]](8)
+                val takes = new MpmcUnboundedXaddArrayQueue[Promise.Unsafe[Closed, A] | (Int, Promise.Unsafe[Closed, Seq[A]])](8)
```
johnhungerford, line 213: `takes` element can now be either a request for a single element or for multiple elements as specified by first element of tuple

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -78,6 +80,24 @@ object Channel:
                 }
             }
 
+        /** Takes [[n]] elements from the channel, asynchronously blocking if necessary.
+          *
+          * @return
+          *   The taken element
+          */
+        def takeExactly(n: Int)(using Frame): Seq[A] < (Abort[Closed] & Async) =
+            Loop[Seq[A], Seq[A], Abort[Closed] & Async](Seq.empty): lastChunk =>
```
johnhungerford, line 89: Is there a performance overhead for Loop I should be concerned about?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -78,6 +80,24 @@ object Channel:
...
+                val nextN = n - lastChunk.size
+                for
+                    drainFiberUnsafe <- IO.Unsafe(self.drainUpToFiber(nextN))
```
fwbrasil, line 92: Would it be possible to reuse the safe version of `drainUpTo` here?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -78,6 +80,24 @@ object Channel:
                 }
             }
 
+        /** Takes [[n]] elements from the channel, asynchronously blocking if necessary.
```
fwbrasil, line 83: If you could add a warning that the computation will wait while the channel doesn't provide all values?

kyo-core/shared/src/main/scala/kyo/Channel.scala
fwbrasil: I'm not sure a `drain` operation should block. Isn't just returning what's currently in the channel like in `drain` enough? The outer loop in `takeExactly` can then keep taking more items if the initial drain doesn't have enough elements.

kyo-core/shared/src/main/scala/kyo/Queue.scala
```diff
@@ -372,6 +388,18 @@ object Queue:
...
+                        if i < max then
+                            val value = q.poll()
+                            if !isNull(value) then
+                                b.addOne(value)
+                                loop(i + 1)
+                    end loop
+                    loop(0)
+                    b.result()
+                end _drainUpTo
                 def _drain() =
```
fwbrasil, line 403: It'd be nice to avoid the code duplication. It seems `drainUpTo` can be implemented in terms of `drain` with a max `Int.MaxValue`?

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: Yes. In fact, would it be alright with you to make all of these methods (`Queue#drain`, `Channel#drain`, `Channel#drainUpTo`, etc) explicitly yield `Chunk`? It seems appropriate for relatively low level APIs.

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: The problem is that if we loop over `drain`/`drainUpTo`, each iteration will return with an empty `Seq` when the channel's empty.  I figure we want to avoid unnecessary polling here.

kyo-core/shared/src/main/scala/kyo/Channel.scala
fwbrasil: Yep! I think we should specialize to Chunk where possible

kyo-core/shared/src/main/scala/kyo/Channel.scala
fwbrasil: I had in mind an initial non-blocking drain and then a loop that uses take to block if needed to wait for more results

johnhungerford: So after initial drain, just take one at a time? Isn't it more efficient to take in chunks? Or is having tuples in the `takes` queue more overhead than is worth it?

fwbrasil: My initial concern was more semantic. The current API separates `take` as an operation that may park with `Async` from `drain` that returns synchronously given the `IO`-only restriction. This distinction is useful to reason about how to compose these operations efficiently. It might be more a question of method naming than semantics, though.

Regarding performance, I wouldn't be surprised if taking items individually has similar performance to a batched operation. There's the tuple allocation you mentioned, the additional branching in the channel flushing logic, and the memory barriers to modify the queue are needed for each item individually anyway. Perhaps `takeExactly` could be a loop that starts with an initial `drainUpTo` and then intercalates `take` and `drainUpTo` on each iteration?

fwbrasil: We should probably follow up with batched channel insertion. It could be a good performance improvement if paired with this change.

johnhungerford: Ah good idea. I'll do that.

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: For some reason it won't recognize the safe methods when I try to use them here, presumably because it knows that `Channel[A]` is unsafe. Should I rename the `Unsafe` methods so they don't conflict?

fwbrasil: You can explicitly access the safe version via the companion object: `Channel.drainUpTo(channel)(max)`

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -90,6 +111,37 @@ class ChannelTest extends Test:
...
+                _ <- c.put(1)
+                _ <- c.put(2)
+                r <- c.drainUpTo(4)
+            yield assert(r == Seq(1, 2))
+        }
+        "non-empty drain less than is in the Channel" in run {
+            for
+                c <- Channel.init[Int](4)
+                _ <- Kyo.foreach(1 to 4)(c.put(_))
+                r <- c.drainUpTo(2)
```
fwbrasil, line 141: can you check the channel size as well?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -241,24 +276,26 @@ object Channel:
                     if queueClosed && (!takesEmpty || !putsEmpty) then
                         // Queue is closed, drain all takes and puts
                         val fail = queue.size() // Obtain the failed Result
-                        takes.drain(_.completeDiscard(fail))
+                        takes.drain:
```
fwbrasil, line 279: do we still need changes in `flush`?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -78,6 +80,25 @@ object Channel:
                 }
             }
 
+        /** Takes [[n]] elements from the channel, semantically blocking until enough elements are present. Note that if enough elements are
+          * not added to the channel it can black indefinitely.
```
fwbrasil, line 84:
```suggestion
          * not added to the channel it can block indefinitely.
```

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: good point -- no, I'll revert that

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: Ah good to know! Done.

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -78,6 +80,26 @@ object Channel:
...
             }
 
+        /** Takes [[n]] elements from the channel, semantically blocking until enough elements are present. Note that if enough elements are
+          * not added to the channel it can block indefinitely.
+          *
+          * @return
+          *   Chunk of [[n]] elements
+          */
+        def takeExactly(n: Int)(using Frame): Chunk[A] < (Abort[Closed] & Async) =
+            Loop(Chunk.empty[A]): lastChunk =>
```
fwbrasil, line 90: Can you add an initial if condition to return `Chunk.empty` in case `n <= 0`?

fwbrasil (approved):
LGTM besides the minor issues. Thank you for addressing the feedback!

fwbrasil:
I've opened https://github.com/getkyo/kyo/issues/916 for batched insertion

Issue #907, opened 2024-12-06
Show type class or equivalent?


johnhungerford:
Given the extensive use of opaque types, I think it might make sense to provide some ad-hoc polymorphic utility for generating string representations of objects rather than using `toString`.

Some issues I've run into:
- Logging a `Result` and seeing a `()` or logging a `Maybe` and seeing a `23` instead of a `Present(23)` is confusing on its face when we're used to seeing the wrappers
- It also makes it hard to distinguish logs when in one place we're logging a result and in another we're logging an extracted value. 
- We can't see the level of nesting of the value for `Result` and `Maybe`
- Logging an effect and seeing a pure value can make it look like an effect has been fully handled when it hasn't

The obvious solution is a `Show` type class, with derived instances for your usual types and explicit given instances for kyo's types:
```scala
trait Show[A]:
  inline def show(value: A): String

object Show:
   // This could be replaced with ADT derivation if we want to provide some custom
   // view for case classes
   given default[A]: Show[A] with { inline def show(value: A): String = value.toString }
   
...

object Maybe:
  given [A](using ash: Show[A]): Show[Maybe[A]] with
    inline def show(value: Maybe[A]): String =
      value.fold("Absent")(a => s"Present(${ash.show(a)})")
```

DamianReeves:
I wonder if a Rust style distinction of `Display` and `Debug` would make sense here. Either way it would be great if it was a `zio-prelude`/Rust style debug that had some formatting help. 

Agree with the overall ask either way though.

johnhungerford:
@DamianReeves we just merged a simpler version that can render kyo types to `Text`. We had some chat in the PR about adding some more logic -- maybe supporting multiple modes of text rendering -- but that will be down the line.

johnhungerford:
Closed with #914

PR #914, opened 2024-12-09
Show type class

johnhungerford:
- `Show[A]` type class to generate string representations of types
  - Generic derivation for ADTs
  - Default instance for other instances using `toString`
- `Show` instances for `Maybe`, `Result`, and `<` displaying wrappers
- String interpolator `k` to use `Show` to construct strings
  - E.g., `println(k"hello ${Maybe.Present("world")}")` produces `hello Present(world)` instead of `hello world`


kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -19,6 +19,14 @@ object Maybe:
     inline given [A: Flat]: Flat[Maybe[A]]                                             = Flat.unsafe.bypass
     given [A]: Conversion[Maybe[A], IterableOnce[A]]                                   = _.iterator
 
+    given [A, MA <: Maybe[A]](using sha: Show[A]): Show[MA] with
+        given CanEqual[Absent, MA] = CanEqual.derived
+        def show(value: MA): String = value match
+            case pr @ Present(_) => s"Present(${sha.show(pr.get)})"
```
hearnadam, line 25:
```suggestion
            case Present(a) => s"Present(${sha.show(a)})"
```

kyo-data/shared/src/main/scala/kyo/Show.scala
```diff
@@ -0,0 +1,93 @@
+package kyo
+
+import kyo.Schedule.done
+import scala.language.implicitConversions
+
+/** Provides string representation of a type as an alternative to .toString. Needed for customizing how to display opaque types
+  */
+trait Show[A]:
```
hearnadam, line 8:
```suggestion
abstract class Show[A]:
```

johnhungerford: No objections to the change, but out of curiosity what's the reason for the change?

johnhungerford:
Addresses #907

kyo-data/shared/src/main/scala/kyo/Show.scala
hearnadam: Traits compile to interfaces, which introduce interface dispatch. It won't matter much here, but it's just to promote consistency throughout the codebase.

README.md
```diff
@@ -463,6 +463,30 @@ val b: Result[Throwable, Int] =
...
+
+This can be jarring to new Kyo users, since we would expect a Kyo computation to be something more than just a pure value. However, Kyo's ability to treat pure values as effects is part of what makes it so performant. Nevetheless, the string representations can mislead us as to the compiletime type of a value, which can make it harder to interpret our logs. To make this a bit easier, Kyo provides a string interpolator which will format Kyo types appropriately. To use this interpolater, prefix your interpolated strings with `k` instead of `s`.
+
+```scala
+import kyo.*
+
+val a: Int < Any = 23
+println(k"Kyo effect: $a")
+// Ouputs `Kyo effect: Kyo(23)` to console
+```
```
fwbrasil, line 486: Can you add an example using show directly as well?

kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -19,6 +19,14 @@ object Maybe:
     inline given [A: Flat]: Flat[Maybe[A]]                                             = Flat.unsafe.bypass
     given [A]: Conversion[Maybe[A], IterableOnce[A]]                                   = _.iterator
 
+    given [A, MA <: Maybe[A]](using sha: Show[A]): Show[MA] with
```
fwbrasil, line 22: Could you use more descriptive identifiers? `sha` is a bit confusing

kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -71,8 +79,9 @@ object Maybe:
 
     /** Represents a defined value in a Maybe. */
     opaque type Present[+A] = A | PresentAbsent
-
     object Present:
+        // given presentShow[A](using sha: Show[A]): Show[Present[A]] with
+        //     def show(value: Present[A]): String = s"Present(${sha.show(value.get)})"
```
fwbrasil, line 84: remove?

kyo-data/shared/src/main/scala/kyo/Show.scala
```diff
@@ -0,0 +1,93 @@
...
+object Shown:
+    given [A](using sh: Show[A]): Conversion[A, Shown] with
+        def apply(a: A): Shown =
+            new Shown:
+                private[kyo] def shownValue: String = sh.show(a)
+    end given
+end Shown
+
+extension (sc: StringContext)
+    def k(args: Shown*): String = sc.s(args.map(_.shownValue)*)
```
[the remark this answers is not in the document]
johnhungerford: I don't really understand the purpose of `Text`, is it meant to be a Kyo-wide replacement for `String`? Will `Console.printLine` and `Log.info` accepts `Text` instead of `String`?

If so, one thing we could is replace `Show` with `AsText`, and instead of using the `Conversion[A: Show, Shown]` that I'm using for interpolation, we could use `Conversion[A: AsText, Text]`, and have the interpolator generate a single combined `Text`, as you suggest.

Any thoughts?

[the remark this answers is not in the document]
fwbrasil: > I don't really understand the purpose of Text, is it meant to be a Kyo-wide replacement for String? Will Console.printLine and Log.info accepts Text instead of String?

Yes. It currently only addresses an important performance issue with string operations that require array copying but we'll likely add new string-related features to it and implement more advanced optimizations. We should update `Console` to take `Text`. Note that `Text` is defined as a super type of `String` so it's possible to pass plain strings to methods that take `Text`. @hearnadam could you share your thoughts on this?

> If so, one thing we could is replace Show with AsText, and instead of using the Conversion[A: Show, Shown] that I'm using for interpolation, we could use Conversion[A: AsText, Text], and have the interpolator generate a single combined Text, as you suggest.

It might be better to restrict the conversion to a specific type like `Shown`. Wouldn't the conversion trigger for any type that has a `show` evidence, essentially adding the methods in `Text` to all showable types?

kyo-data/shared/src/main/scala/kyo/Show.scala
```diff
@@ -0,0 +1,93 @@
...
+        def show(value: A): String = value.toString
+
+object Show extends LowPriorityShows:
+    inline def apply[A](using sh: Show[A]): Show[A] = sh
+
+    def show[A](value: A)(using sh: Show[A]): String = sh.show(value)
+
+    import scala.compiletime.*
+
+    private inline def sumShow[A, M <: scala.deriving.Mirror.ProductOf[A]](label: String, mir: M): Show[A] =
```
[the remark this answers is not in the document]
johnhungerford: Yeah. And it's nice to be able to use inlined compile-time operations to do this stuff rather than using implicit resolution

kyo-data/shared/src/main/scala/kyo/Show.scala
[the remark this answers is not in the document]
johnhungerford: Yes good point -- that's why I made the method on `Shown` private.

I think, then, I will:
- Make `Show` as `AsText`, and have it convert to `Text` instead of `String` (this way we won't conflict with `Show` for anyone using cats)
- Update `k` to generate `Text` instead of `String`
- Update `Log` and `Console` to use `Text`

(Let me know if this doesn't make sense)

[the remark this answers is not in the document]
fwbrasil: Sounds good. I think the interpolator could `text` instead of `k`? If you prefer, we can make the Console change separately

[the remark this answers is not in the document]
johnhungerford: Is `txt` okay? `text` can conflict with other methods with that name.

kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -124,6 +130,8 @@ object Maybe:
     /** Represents an empty Maybe instance. */
     sealed abstract class Absent
     case object Absent extends Absent
+    // given Show[Absent] with
+    //     def show(value: Absent): String = "Absent"
```
fwbrasil, line 134: leftover

fwbrasil (approved):

kyo-data/shared/src/main/scala/kyo/Show.scala
```diff
@@ -0,0 +1,93 @@
...
+                    builder.append(showIter.next().show(valIter.next()))
+                    ()
+                end while
+                builder.append(")")
+                builder.toString()
+            end show
+        end new
+    end sumShow
+
+    inline given [A](using mir: scala.deriving.Mirror.Of[A]): Show[A] = inline mir match
```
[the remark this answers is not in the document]
johnhungerford: scala.compiletime makes my palms sweaty so I don't feel comfortable with it in scope ;)

I'll import it though

kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -19,6 +19,14 @@ object Maybe:
     inline given [A: Flat]: Flat[Maybe[A]]                                             = Flat.unsafe.bypass
     given [A]: Conversion[Maybe[A], IterableOnce[A]]                                   = _.iterator
 
+    given [A, MaybeA <: Maybe[A]](using ata: AsText[A]): AsText[MaybeA] with
+        given CanEqual[Absent, MaybeA] = CanEqual.derived
+        def asText(value: MaybeA): String = (value: Maybe[A]) match
+            case Present(a) => s"Present(${ata.asText(a)})"
+            case Absent     => "Absent"
+            case _          => throw IllegalStateException()
```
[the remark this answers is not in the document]
johnhungerford: Exhaustivity warning without it

kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -19,6 +19,14 @@ object Maybe:
     inline given [A: Flat]: Flat[Maybe[A]]                                             = Flat.unsafe.bypass
     given [A]: Conversion[Maybe[A], IterableOnce[A]]                                   = _.iterator
 
+    given [A, MA <: Maybe[A]](using sha: Show[A]): Show[MA] with
+        given CanEqual[Absent, MA] = CanEqual.derived
+        def show(value: MA): String = value match
+            case Present(a) => s"Present(${sha.show(a)})"
+            case Absent          => "Absent"
+            case _               => throw IllegalStateException()
```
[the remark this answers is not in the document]
johnhungerford: Note sure, but there's an exhaustivity warning without it

kyo-data/shared/src/main/scala/kyo/AsText.scala
```diff
@@ -0,0 +1,99 @@
...
+object IsText:
+    given [A](using at: AsText[A]): Conversion[A, IsText] with
+        def apply(a: A): IsText =
+            new IsText:
+                private[kyo] def textValue: Text = at.asText(a)
+    end given
+end IsText
+
+extension (sc: StringContext)
+    def txt(args: IsText*): Text =
```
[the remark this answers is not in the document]
fwbrasil: I think `t` would be better than `txt` but is `text` really problematic? It seems `k` and `t` could conflict with other identifiers more often.

kyo-data/shared/src/main/scala/kyo/Maybe.scala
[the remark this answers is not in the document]
fwbrasil: that's odd, I wouldn't expect it to warn

kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -411,9 +417,7 @@ object Maybe:
         inline def toResult[E](inline ifEmpty: => Result[E, A]): Result[E, A] =
             if isEmpty then ifEmpty else Result.success(get)
 
-        def show: String =
-            if isEmpty then "Absent"
-            else s"Present(${get})"
+        def text(using at: AsText[Maybe[A]]): Text = at.asText(self)
```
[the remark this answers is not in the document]
hearnadam: Ah that's fair.

kyo-data/shared/src/main/scala/kyo/AsText.scala
```diff
@@ -0,0 +1,99 @@
+package kyo
+
+import kyo.Schedule.done
+import scala.language.implicitConversions
+
+/** Provides Text representation of a type. Needed for customizing how to display opaque types as alternative to toString
+  */
+abstract class AsText[A]:
+    def asText(value: A): Text
```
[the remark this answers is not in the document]
hearnadam: How about `Render`? `AsText` is a bit odd as it doesn't store the value 'AsText', it just enables converting to Text

kyo-data/shared/src/main/scala/kyo/AsText.scala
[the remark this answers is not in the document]
johnhungerford: To @hearnadam 's point, having a single-character prefix is easier to write and I think is more consistent with user expectations. (e.g., f-string and s-strings). That said, if you prefer text, I'll just revert the `text` methods to `show`, which I'm now thinking is a better idea, and it shouldn't be a problem.

kyo-data/shared/src/main/scala/kyo/Render.scala
```diff
@@ -0,0 +1,100 @@
+package kyo
+
+import kyo.Schedule.done
+import scala.language.implicitConversions
+
+/** Provides Text representation of a type. Needed for customizing how to display opaque types as alternative to toString
+  */
+abstract class Render[A]:
+    def asText(value: A): Text
+    def asString(value: A): String = asText(value).show
```
hearnadam, line 10:
```suggestion
    final def asString(value: A): String = asText(value).show
```

kyo-data/shared/src/main/scala/kyo/Render.scala
```diff
@@ -0,0 +1,100 @@
...
+
+sealed trait LowPriorityRenders:
+    given [A]: Render[A] with
+        def asText(value: A): Text = value.toString
+
+object Render extends LowPriorityRenders:
+    inline def apply[A](using sh: Render[A]): Render[A] = sh
```
hearnadam, line 17:
```suggestion
    inline def apply[A](using r: Render[A]): Render[A] = r
```

kyo-data/shared/src/main/scala/kyo/Render.scala
```diff
@@ -0,0 +1,100 @@
...
+
+    def asText[A](value: A)(using sh: Render[A]): Text = sh.asText(value)
```
hearnadam, line 19:
```suggestion
    def asText[A](value: A)(using r: Render[A]): Text = r.asText(value)
```

kyo-data/shared/src/main/scala/kyo/Render.scala
```diff
@@ -0,0 +1,100 @@
...
+                private[kyo] def textValue: Text = r.asText(a)
+    end given
+end Rendered
+
+extension (sc: StringContext)
+    def t(args: Rendered*): Text =
+        StringContext.checkLengths(args, sc.parts)
+        val pi         = sc.parts.iterator
+        val ai         = args.iterator
+        var text: Text = pi.next()
```
hearnadam, line 96: is it possible this initial value won't exist?

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -23,6 +23,13 @@ object Result:
     inline given [E, A: Flat]: Flat[Result[E, A]]                                              = Flat.unsafe.bypass
     inline given [E, A]: CanEqual[Result[E, A], Panic]                                         = CanEqual.derived
 
+    given [E, A, ResultEA <: Result[E, A]](using re: Render[E], ra: Render[A]): Render[ResultEA] with
+        def asText(value: ResultEA): String = value match
+            case Success(a) => s"Success(${ra.asText(a.asInstanceOf[A])})"
+            case f: Fail[?] => s"Fail(${re.asText(f.error.asInstanceOf[E])})"
+            case other      => other.toString()
```
hearnadam, line 30: I think we want to render `Panic/Error` separately?

kyo-core/shared/src/main/scala/kyo/Console.scala
```diff
@@ -19,28 +19,28 @@ final case class Console(unsafe: Console.Unsafe):
       * @param s
       *   The string to print.
       */
-    def print(s: String)(using Frame): Unit < (IO & Abort[IOException]) = IO.Unsafe(Abort.get(unsafe.print(s)))
+    def print(s: Text)(using Frame): Unit < (IO & Abort[IOException]) = IO.Unsafe(Abort.get(unsafe.print(s.show)))
```
hearnadam, line 22: Should these methods accept `print[A](a: A)(using Frame, Render[A])`? Or is that the value of the `Conversion`?

kyo-data/shared/src/main/scala/kyo/Render.scala
johnhungerford: Apparently not. (Just added a test for empty string.)

I basically copied this from the s-string interpolator implementation.

kyo-core/shared/src/main/scala/kyo/Console.scala
johnhungerford: It does not do any conversion. You can pass a string, since `Text` is a supertype of `String`, or you can convert to text using `Render` or t-string interpolation.

@fwbrasil  how would feel about having these methods doing conversions based on `Render`? We could do what @hearnadam suggests and `use` a `Render[A]`, or we can accepted a `Rendered` value which will be converted automatically. I prefer @hearnadam 's suggestion, since it's more transparent, but I'm on the fence about changing it at all.

fwbrasil: we already have the overloaded `Console` methods that take objects and use `pprint` to generate the string. The formatted output with coloring is quite helpful to inspect more complex objects. Perhaps we could migrate everything to `Render` and follow up later improving its output like in `pprint`? It'd be nice to avoid the dependency as well

kyo-data/shared/src/main/scala/kyo/AsText.scala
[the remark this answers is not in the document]
fwbrasil: I'd prefer the more explicit `text`, which should be helpful to newcomers, but I'd be ok with `t` if you think it's best. @hearnadam please let us know what you think as well

kyo-data/shared/src/main/scala/kyo/AsText.scala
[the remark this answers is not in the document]
fwbrasil: me too!

fwbrasil:
Are we good to merge this? We can follow up with improvements like replacing the pprint rendering in Console later

Issue #916, opened 2024-12-10
Channel: support inserting multiple elements 


fwbrasil:
Follow up of https://github.com/getkyo/kyo/pull/913. We could ensure that a batched insertion only notifies pending takes after the new items are in the queue, which would give an opportunity for batched takers to consume multiple elements at once.

hearnadam:
@johnhungerford can this be resolved?

johnhungerford:
@hearnadam yes I think this should be closed now

PR #918, opened 2024-12-11
Stream from Channel

johnhungerford:
- Adds `stream` and `streamUntilClosed` methods to `Channel`
- Optionally specify `maxChunkSize`
- Uses `drain` and `drainUpTo` under the hood


kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -158,6 +158,54 @@ object Channel:
           */
         def full(using Frame): Boolean < (Abort[Closed] & IO) = IO.Unsafe(Abort.get(self.full()))
 
+        private def emitChunks(maxChunkSize: Maybe[Int] = Maybe.Absent)(using
+            Tag[A],
```
fwbrasil, line 162: The `Tag` implementation is able to avoid allocations but only when the tag is fully known statically. In cases where it needs to be derived, an allocation is introduced. This should be a concern in user code but it's an important optimization for Kyo's primitives.
```suggestion
            Tag[Emit[Chunk[A]]],
```

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -158,6 +158,54 @@ object Channel:
...
+          * be limited only by channel capacity or the number of elements in the channel at a given time. (Chunks can still be larger than
+          * channel capacity.) Fails on channel closure.
+          *
+          * @param maxChunkSize
+          *   Maximum number of elements to take for each chunk
+          *
+          * @return
+          *   Asynchronous stream of elements in this channel
+          */
+        def stream(maxChunkSize: Maybe[Int] = Maybe.Absent)(using Tag[A], Frame): Stream[A, Abort[Closed] & Async] =
```
fwbrasil, line 191: I find it odd that the user needs to pass a `Maybe` to specify the max size. Could you use method overloads or an `Int.MaxValue` placeholder instead?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -158,6 +158,54 @@ object Channel:
...
+                Loop(()): _ =>
+                    Channel.take(self).map: v =>
+                        Emit.andMap(Chunk(v)):
+                            case Ack.Stop => Loop.done(Ack.Stop)
+                            case _        => Loop.continue(())
+            else
+                val drainEffect = maxChunkSize.fold(Channel.drain(self))(m => Channel.drainUpTo(self)(m - 1))
+                Loop[Unit, Ack, Abort[Closed] & Async & Emit[Chunk[A]]](()): _ =>
+                    Channel.take(self).map: a =>
+                        drainEffect.map: chunk =>
```
fwbrasil, line 176: I think this might not be very effective without batched inserts. Since channel writers are racing with this code, there's a risk that in most cases the emitted chunks will contain one or just a few values. I've created https://github.com/getkyo/kyo/issues/916 for batched inserts, we can address it as a follow up.

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -453,4 +453,88 @@ class ChannelTest extends Test:
...
+                rn <- sn.run
+                s  <- c.size
+            yield assert(r0 == Chunk.empty && rn == Chunk.empty && s == 4)
+            end for
+        }
+        "stream with maxChunkSize of 1 should stream in chunks of 1" in run {
+            for
+                c <- Channel.init[Int](4)
+                _ <- Kyo.foreach(1 to 4)(c.put)
+                stream = c.stream(Maybe(1)).mapChunk(Chunk(_)).take(4)
```
fwbrasil, line 481: We should probably add `stream.chunks`

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -453,4 +453,88 @@ class ChannelTest extends Test:
         }
     }
 
+    "stream" - {
```
fwbrasil, line 456: missing tests for streamUntilClosed?

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: When producing at <= rate of consumption, yes, but if the producer is faster than the consumer then chunking will be an optimization. If the producer is slower than the consumer, on the other hand, wouldn't we rather take small chunks quickly than wait for bigger chunks to be constructed?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -158,6 +158,54 @@ object Channel:
...
+                            Emit.andMap(Chunk(a).concat(chunk)):
+                                case Ack.Stop => Loop.done(Ack.Stop)
+                                case _        => Loop.continue(())
+
+        /** Stream elements from channel, optionally specifying a maximum chunk size. In the absence of [[maxChunkSize]], chunk sizes will
+          * be limited only by channel capacity or the number of elements in the channel at a given time. (Chunks can still be larger than
+          * channel capacity.) Fails on channel closure.
```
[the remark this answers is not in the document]
johnhungerford: Good point. (I can add a `Listener` integration next.)

kyo-core/shared/src/main/scala/kyo/Channel.scala
fwbrasil: Good point, it's a good behavior for low latency. My concern is more about the race when the consumer is waiting for new elements. If that's the case and the producer has 10 items to add to the channel for example, as soon as the first element arrives, the consumer will wake up and try drain immediately without waiting for the producer to finish adding the 10 elements to the queue. A batched insert operation in `Channel` could reduce the impact of this race by waking up the consumer only after the 10 items are already in the queue.

fwbrasil (approved):

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -158,6 +158,57 @@ object Channel:
...
+          *
+          * @return
+          *   Asynchronous stream of elements in this channel
+          */
+        def streamUntilClosed(maxChunkSize: Int = Int.MaxValue)(using Tag[Emit[Chunk[A]]], Frame): Stream[A, Async] =
+            Stream:
+                Abort.run[Closed](emitChunks(maxChunkSize)).map:
+                    case Result.Success(v) => v
+                    case Result.Fail(_)    => Ack.Stop
+                    case Result.Panic(e)   => Abort.panic(e)
```
hearnadam, line 210: We should reuse the panic class.

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -158,6 +158,57 @@ object Channel:
...
+            else
+                val drainEffect =
+                    if maxChunkSize == Int.MaxValue then Channel.drain(self)
+                    else Channel.drainUpTo(self)(maxChunkSize - 1)
+                Loop[Unit, Ack, Abort[Closed] & Async & Emit[Chunk[A]]](()): _ =>
+                    Channel.take(self).map: a =>
+                        drainEffect.map: chunk =>
+                            Emit.andMap(Chunk(a).concat(chunk)):
+                                case Ack.Stop => Loop.done(Ack.Stop)
+                                case _        => Loop.continue(())
```
hearnadam, line 182: We should make a version of the loop `Loop.stateless`

fwbrasil: Agreed! It's a common pattern

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: Is this what you mean?:
```scala
   case panic: Result.Panic => Abort.get(panic)
```

PR #919, opened 2024-12-11
[data] anchored and jittered schedules

fwbrasil:
I'm working on STM optimizations and want to use a jittered schedule to reduce the impact of retry storms. Since we avoid side effects in `kyo-data`, I'm passing the current `Instant` to `schedule.next` and calculating the jitter deterministically via hashing. It isn't random but, given that `Instant` has nanosecond precision, it seems reasonable.

We could pass a random number from `kyo-core` instead but we'd need to eventually pass the instant anyway to implement more advanced schedules like the new `Schedule.anchored` I've also added in this PR.

### DIFF
--- a/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
+++ b/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
@@ -77,10 +77,12 @@ extension [A, S](effect: A < S)
       */
     def repeat(schedule: Schedule)(using Flat[A], Frame): A < (S & Async) =
         Loop(schedule) { schedule =>
-            schedule.next.map { (delay, nextSchedule) =>
-                effect.delayed(delay).andThen(Loop.continue(nextSchedule))
-            }.getOrElse {
-                effect.map(Loop.done)
+            Clock.now.map { now =>
+                schedule.next(now).map { (delay, nextSchedule) =>
+                    effect.delayed(delay).andThen(Loop.continue(nextSchedule))
+                }.getOrElse {
+                    effect.map(Loop.done)
+                }
             }
         }
 
--- a/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
+++ b/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
@@ -76,12 +76,14 @@ extension [A, S](effect: A < S)
       *   A computation that produces the result of the last execution
       */
     def repeat(schedule: Schedule)(using Flat[A], Frame): A < (S & Async) =
-        Loop(schedule) { schedule =>
-            Clock.now.map { now =>
-                schedule.next(now).map { (delay, nextSchedule) =>
-                    effect.delayed(delay).andThen(Loop.continue(nextSchedule))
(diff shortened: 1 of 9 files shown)


kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -77,10 +77,12 @@ extension [A, S](effect: A < S)
       */
     def repeat(schedule: Schedule)(using Flat[A], Frame): A < (S & Async) =
         Loop(schedule) { schedule =>
-            schedule.next.map { (delay, nextSchedule) =>
-                effect.delayed(delay).andThen(Loop.continue(nextSchedule))
-            }.getOrElse {
-                effect.map(Loop.done)
+            Clock.now.map { now =>
```
hearnadam, line 80: Is it useful to fetch the clock once outside the loop to avoid pulling from the local multiple times?

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -401,10 +401,12 @@ object Clock:
         Async.run {
             Clock.use { clock =>
                 Loop(state, delaySchedule) { (state, schedule) =>
-                    schedule.next match
-                        case Absent => Loop.done(state)
-                        case Present((duration, nextSchedule)) =>
-                            clock.sleep(duration).map(_.use(_ => f(state).map(Loop.continue(_, nextSchedule))))
+                    Clock.now.map { now =>
```
hearnadam, line 404: can we use lowercase `clock`?

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -502,11 +504,13 @@ object Clock:
             Clock.use { clock =>
                 clock.now.map { now =>
                     Loop(now, state, intervalSchedule) { (lastExecution, state, period) =>
-                        period.next match
-                            case Absent => Loop.done(state)
-                            case Present((duration, nextSchedule)) =>
-                                val nextExecution = lastExecution + duration
-                                clock.sleep(duration).map(_.use(_ => f(state).map(Loop.continue(nextExecution, _, nextSchedule))))
+                        Clock.now.map { now =>
```
hearnadam, line 507: lowercase clock?

kyo-data/shared/src/main/scala/kyo/Schedule.scala
```diff
@@ -243,103 +263,168 @@ object Schedule:
...
+        end Anchored
 
         final case class Max(a: Schedule, b: Schedule) extends Schedule:
-            def next =
+            def next(now: Instant) =
                 for
-                    (d1, s1) <- a.next
-                    (d2, s2) <- b.next
+                    (d1, s1) <- a.next(now)
+                    (d2, s2) <- b.next(now)
```
hearnadam, line 356: pre-existing: I think this should be a `fold` - otherwise you may end up discarding when one schedule produces a next.

hearnadam (approved):

PR #923, opened 2024-12-11
[core][stm] fix compilation warning with Scala 3.6 + add AllowUnsafe to evalOrThrow

fwbrasil:


kyo-core/shared/src/test/scala/kyo/Test.scala
```diff
@@ -16,13 +16,16 @@ abstract class Test extends AsyncFreeSpec with BaseKyoTest[Abort[Any] & Async &
 
     def run(v: Future[Assertion] < (Abort[Any] & Async & Resource)): Future[Assertion] =
         import AllowUnsafe.embrace.danger
-        val a = Async.run(Abort.run(Resource.run(v)).map(_.fold {
-            _.getFailure match
+        v.pipe(
+            Resource.run,
+            Abort.recover[Any] {
                 case ex: Throwable => throw ex
                 case e             => throw new IllegalStateException(s"Test aborted with $e")
```
fwbrasil, line 23: The compiler started complaining of an unreachable case here but it's incorrect. I tried to reproduce in isolation but it seems bad interaction with inlining. As a workaround, I've rewrote the method to use `pipe`, which fixed the warning and seems more readable.

kyo-core/shared/src/main/scala/kyo/IO.scala
```diff
@@ -87,7 +87,7 @@ object IO:
           * @throws Throwable
           *   If the evaluation results in an error
           */
-        def evalOrThrow[A: Flat](v: A < (IO & Abort[Throwable]))(using Frame): A =
+        def evalOrThrow[A: Flat](v: A < (IO & Abort[Throwable]))(using Frame, AllowUnsafe): A =
```
fwbrasil, line 90: The method was missing `AllowUnsafe`

fwbrasil:
Merging since these are trivial fixes and the warnings keep drawing my attention while working on the codebase

PR #924, opened 2024-12-12
[stm] optimizations

fwbrasil:
Benchmark [results](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/28a2f0957f839310a85752634904a99d/raw/e414912bcd3b7d89d43663298a40a5e3dd5c560d/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/28a2f0957f839310a85752634904a99d/raw/e414912bcd3b7d89d43663298a40a5e3dd5c560d/jmh-result-candidate.json):

image

### DIFF
--- a/kyo-bench/src/main/scala/kyo/bench/TRefMultiBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/TRefMultiBench.scala
@@ -13,7 +13,7 @@ class TRefMultiBench(parallelism: Int) extends Bench.ForkOnly(parallelism):
 
         STM.runtime[IO].flatMap { stm =>
             for
-                refs   <- stm.commit(Seq.fill(parallelism)(stm.TVar.of(0)).sequence)
+                refs   <- Seq.fill(parallelism)(stm.commit(stm.TVar.of(0))).sequence
                 _      <- refs.map(ref => stm.commit(ref.modify(_ + 1))).parSequence_
                 result <- stm.commit(refs.traverse(_.get).map(_.sum))
             yield result
--- a/kyo-core/shared/src/main/scala/kyo/Retry.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Retry.scala
@@ -41,18 +41,17 @@ object Retry:
             SafeClassTag[E],
             Frame
         ): A < (Async & Abort[E] & S) =
-            Loop(schedule) { schedule =>
-                Abort.run[E](v).map(_.fold { r =>
+            Abort.run[E](v).map {
+                case Result.Success(r) => r
+                case error: Result.Error[?] =>
                     Clock.now.map { now =>
                         schedule.next(now).map { (delay, nextSchedule) =>
-                            Async.delay(delay)(Loop.continue(nextSchedule))
+                            Async.delay(delay)(Retry[E](nextSchedule)(v))
                         }.getOrElse {
-                            Abort.get(r)
+                            Abort.get(error)
                         }
                     }
-                }(Loop.done(_)))
             }
-        end apply
     end RetryOps
 
     /** Creates a RetryOps instance for the specified error type.
--- a/kyo-prelude/shared/src/main/scala/kyo/Local.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Local.scala
@@ -108,7 +108,7 @@ object Local:
                 ContextEffect.suspendAndMap(tag, Map.empty)(map => f(map.getOrElse(this, default).asInstanceOf[A]))
 
             def let[B, S](value: A)(v: B < S)(using Frame) =
-                ContextEffect.handle(tag, Map(this -> value), _.updated(this, value.asInstanceOf[AnyRef]))(v)
+                ContextEffect.handle(tag, Map.empty[Local[?], AnyRef].updated(this, value), _.updated(this, value.asInstanceOf[AnyRef]))(v)
 
             def update[B, S](f: A => A)(v: B < S)(using Frame) =
                 ContextEffect.handle(
--- a/kyo-prelude/shared/src/main/scala/kyo/Var.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Var.scala
@@ -167,7 +167,7 @@ object Var:
         runWith(state)(v)((state, result) => (state, result))
 
     object isolate:
-        abstract private[kyo] class Base[V: Tag] extends Isolate[Var[V]]:
+        abstract private[kyo] class Base[V](using Tag[Var[V]]) extends Isolate[Var[V]]:
             type State = V
             def use[A, S2](f: V => A < S2)(using Frame) = Var.use(f)
             def resume[A: Flat, S2](state: State, v: A < (Var[V] & S2))(using Frame) =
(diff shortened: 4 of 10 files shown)


kyo-bench/src/main/scala/kyo/bench/TRefMultiBench.scala
```diff
@@ -13,7 +13,7 @@ class TRefMultiBench(parallelism: Int) extends Bench.ForkOnly(parallelism):
 
         STM.runtime[IO].flatMap { stm =>
             for
-                refs   <- stm.commit(Seq.fill(parallelism)(stm.TVar.of(0)).sequence)
+                refs   <- Seq.fill(parallelism)(stm.commit(stm.TVar.of(0))).sequence
```
fwbrasil, line 16: I noticed the cats-effect implementation was different from kyo and zio, not committing each new var separately

kyo-core/shared/src/main/scala/kyo/Retry.scala
```diff
@@ -41,18 +41,17 @@ object Retry:
             SafeClassTag[E],
             Frame
         ): A < (Async & Abort[E] & S) =
-            Loop(schedule) { schedule =>
```
fwbrasil, line 44: Since in most cases the operation succeeds and doesn't retry, using regular recursion saves a few `Loop`-related allocations in the happy path. It's also why the code doesn't get the `Clock` first to reuse on each iteration.

kyo-prelude/shared/src/main/scala/kyo/Local.scala
```diff
@@ -108,7 +108,7 @@ object Local:
                 ContextEffect.suspendAndMap(tag, Map.empty)(map => f(map.getOrElse(this, default).asInstanceOf[A]))
 
             def let[B, S](value: A)(v: B < S)(using Frame) =
-                ContextEffect.handle(tag, Map(this -> value), _.updated(this, value.asInstanceOf[AnyRef]))(v)
+                ContextEffect.handle(tag, Map.empty[Local[?], AnyRef].updated(this, value), _.updated(this, value.asInstanceOf[AnyRef]))(v)
```
fwbrasil, line 111: avoids tuple allocation to insert into the map

kyo-prelude/shared/src/main/scala/kyo/Var.scala
```diff
@@ -167,7 +167,7 @@ object Var:
         runWith(state)(v)((state, result) => (state, result))
 
     object isolate:
-        abstract private[kyo] class Base[V: Tag] extends Isolate[Var[V]]:
+        abstract private[kyo] class Base[V](using Tag[Var[V]]) extends Isolate[Var[V]]:
```
fwbrasil, line 170: prevents allocations for the dynamic tag

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -108,34 +111,97 @@ object STM:
...
+                                            def fail = throw new FailedTransaction(frame)
+
+                                            var i = 0
+                                            // Pre-validate and dump the log to the flat array
+                                            logMap.foreachEntry { (ref, entry) =>
+                                                // This code uses exception throwing because
+                                                // foreachEntry is the only way to traverse the
+                                                // map without allocating tuples, so throwing
+                                                // is the workaround to short circuit
+                                                if !ref.validate(entry) then fail
```
fwbrasil, line 150: I've added an initial validation of the ref log first so it's possible to fail invalid transactions early without locking any refs. It has some performance cost but helps reduce the impact of retries.

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -108,34 +111,97 @@ object STM:
...
+                                    IO.Unsafe {
+                                        // Flattened representation of the log
+                                        val array = new Array[Any](size * 2)
+
+                                        try
+                                            def fail = throw new FailedTransaction(frame)
+
+                                            var i = 0
+                                            // Pre-validate and dump the log to the flat array
+                                            logMap.foreachEntry { (ref, entry) =>
```
fwbrasil, line 145: I thought Scala's `Map` didn't have a method to traverse the entries without allocating tuples but I've found `foreachEntry` that provides exactly that! Since there's no `forallEntry`, I'm using exception throwing as an efficient way to short circuit

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -145,4 +211,40 @@ object STM:
         }
 
     end run
+
+    private def quickSort(array: Array[Any], size: Int): Unit =
```
fwbrasil, line 215: Based on the implementation for `Async.gather` but specialized to the format of the flat ref log.

kyo-stm/shared/src/main/scala/kyo/TRefLog.scala
```diff
@@ -14,35 +17,26 @@ opaque type TRefLog = Map[TRef[Any], TRefLog.Entry[Any]]
...
         def get[A](ref: TRef[A]): Maybe[Entry[A]] =
             val refAny = ref.asInstanceOf[TRef[Any]]
             Maybe.when(self.contains(refAny))(self(refAny).asInstanceOf[Entry[A]])
 
-        def toSeq: Seq[(TRef[Any], Entry[Any])] =
-            self.toSeq
+        def toMap: Map[TRef[Any], TRefLog.Entry[Any]] = self
     end extension
 
-    def use[A, S](f: TRefLog => A < S)(using Frame): A < (S & Var[TRefLog]) =
```
fwbrasil, line 32: These methods were introducing a few allocations and triggering unnecessary boxing so I've changed `STM` to use `Var` directly and expose a `Tag[Var[TRefLog]]` (line 20)

kyo-stm/shared/src/main/scala/kyo/TID.scala
```diff
@@ -5,20 +5,20 @@ private[kyo] object TID:
     // Unique transaction ID generation
     private val nextTid = AtomicLong.Unsafe.init(0)(using AllowUnsafe.embrace.danger)
 
-    private val tidLocal = Local.initIsolated(-1L)
+    private val tidLocal = Local.initIsolated[java.lang.Long](-1L)
```
fwbrasil, line 8: The local was boxing on each use so I've made it hold a boxed long instead.

hearnadam (approved):

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -1,10 +1,13 @@
...
 
+import java.util.Arrays
+import kyo.Result.Fail
 import scala.annotation.tailrec
+import scala.util.control.NoStackTrace
 
 /** A FailedTransaction exception that is thrown when a transaction fails to commit. Contains the frame where the failure occurred.
   */
-case class FailedTransaction(frame: Frame) extends Exception(frame.position.show)
+case class FailedTransaction(frame: Frame) extends Exception(frame.position.show) with NoStackTrace
```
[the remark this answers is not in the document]
fwbrasil: yeah, I forgot about `getCause`. I'll follow up on that

PR #927, opened 2024-12-13
[prelude][core][stm] fix issue with preemption in nested async computations

### LINKED ISSUE #925: Infinite loop with STM + unhandled exception + toFuture/fromFuture
Version `0.15.1`.

```scala
package com.example

import scala.concurrent.Future
import kyo.{ <, Async, Abort, STM, TRef, Flat, IO, Duration }

final class MyException extends Exception

object Foo extends kyo.KyoApp {

  private def unsafeToFuture[A : Flat](a: => A < (Async & Abort[Throwable])): Future[A] = {
    import kyo.AllowUnsafe.embrace.danger
    IO.Unsafe.run(
      Async.run(a).map(_.toFuture)
    ).eval
  }

  private val faultyTransaction: Int < STM = TRef.init(42).map { r =>
    throw new MyException
    r.get
  }

  run[Int] {
    // this works fine, completes with 'Panic(exception = com.example.MyException)':
    // Async.runAndBlock(kyo.Duration.Infinity)(STM.run(faultyTransaction))

    // this hangs:
    Async.runAndBlock(Duration.Infinity)(Async.fromFuture(unsafeToFuture(STM.run(faultyTransaction))))
  }
}
```

This program never seems to finish, and one thread seems stuck in an infinite loop. Now, I realize I'm doing a lot of `Unsafe` things here, and that I shouldn't `throw` inside `map`, but still, I don't think I'm doing anything that should cause an infinite loop(?).

The thread in the loop looks like this:

```
"kyo-scheduler-worker-1" #19 daemon prio=5 os_prio=0 cpu=17018,64ms elapsed=17,26s tid=0x00007f6d685eba50 nid=0x3ad3 runnable  [0x00007f6d2fcfa000]
   java.lang.Thread.State: RUNNABLE
        at com.example.Foo$.evalLoop$1(Foo.scala:12)
        at com.example.Foo$.unsafeToFuture(Foo.scala:14)
        at com.example.Foo$.v$proxy8$1(Foo.scala:27)
        at com.example.Foo$.$init$$$anonfun$1$$anonfun$1(Foo.scala:27)
        at com.example.Foo$$$Lambda$11/0x00007f6ce40c1458.apply(Unknown Source)
        at kyo.Async$package$Async$$anon$7.apply(Async.scala:72)
        at kyo.Async$package$Async$$anon$7.apply(Async.scala:72)
        at kyo.Async$package$Async$$anon$1.apply(Async.scala:72)
        at kyo.Resource$$anon$14.apply(Resource.scala:120)
        at kyo.IO$package$IO$$anon$2.apply(IO.scala:64)
        at kyo.Resource$$anon$12.apply(Resource.scala:120)
        at kyo.scheduler.IOTask.partialLoop$1(IOTask.scala:39)
        at kyo.scheduler.IOTask.eval(IOTask.scala:62)
        at kyo.scheduler.IOTask.run(IOTask.scala:74)
        at kyo.scheduler.Worker.runTask(Worker.scala:259)
        at kyo.scheduler.Worker.run(Worker.scala:221)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(java.base@17.0.13/ThreadPoolExecutor.java:1136)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(java.base@17.0.13/ThreadPoolExecutor.java:635)
        at java.lang.Thread.run(java.base@17.0.13/Thread.java:840)
```

fwbrasil:
Fixes #925

The issue is more generic than the use with STM. Kyo's kernel uses a safepoint mechanism to preempt execution on any step of a computation. `kyo-core` extends this mechanism by setting the safepoint interceptor to the current fiber to handle preemption. For example, if the fiber is interrupted, the next safepoint will trigger preemption to stop the computation and clear finalizers. The other possibility is the scheduler itself triggering preemption due to the fiber taking more than the configured time slice (10ms by default) so it can return to the execution queue to provide fairness.

The cause of the bug is the interceptor propagating to nested async computations. This PR fixes `Safepoint.eval`, which is called from the pending type's `eval`, to clear the interceptor since the `eval` operation itself can't be delayed and has to return a concrete value. Even if the outer async computation is preempted, the nested one needs to execute to completion.


kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -87,6 +87,9 @@ sealed private[kyo] class IOTask[Ctx, E, A] private (
 
     private inline def nullResult = null.asInstanceOf[A < Ctx & Async & Abort[E]]
 
+    override def toString =
```
fwbrasil, line 90: to help with debugging

hearnadam (approved):

PR #928, opened 2024-12-13
[data] fix Render compilation warnings

fwbrasil:


kyo-data/shared/src/main/scala/kyo/Render.scala
```diff
@@ -81,11 +84,9 @@ sealed trait Rendered:
...
 
 object Rendered:
-    given [A](using r: Render[A]): Conversion[A, Rendered] with
-        def apply(a: A): Rendered =
-            new Rendered:
-                private[kyo] def textValue: Text = r.asText(a)
-    end given
+    implicit def apply[A](value: A)(using render: Render[A]): Rendered =
+        new Rendered:
+            private[kyo] def textValue: Text = render.asText(value)
```
fwbrasil, line 89: I'm cheating a bit here :) The compiler doesn't complain about the implicit conversion if we don't use `Conversion`. This is the warning I was seeing:

```
[warn] -- Feature Warning: /home/runner/work/kyo/kyo/kyo-data/shared/src/test/scala/kyo/ResultTest.scala:1007:22 
[warn] 1007 |            assert(t"$widened".show == "Success(Success(Success(23)))")
[warn]      |                      ^^^^^^^
[warn]      |Use of implicit conversion given instance given_Conversion_A_Rendered in object Rendered should be enabled
[warn]      |by adding the import clause 'import scala.language.implicitConversions'
[warn]      |or by setting the compiler option -language:implicitConversions.
[warn]      |See the Scala docs for value scala.language.implicitConversions for a discussion
[warn]      |why the feature should be explicitly enabled.
```

hearnadam (approved):

PR #331, opened 2024-05-06
fix Fibers.Interrupted to avoid lock contention in `Throwable::getCause`

fwbrasil:
I noticed this while optimizing `Requests`. When a large number of fibers is interrupted, sttp obtains the cause of the `Interrupted` exception, which is synchronized in `Throwable`. In this case, the cause is always `null` since it's a cached exception.

```
"kyo-scheduler-worker-33" #101 [56067] daemon prio=5 os_prio=31 cpu=59.03ms elapsed=0.65s tid=0x00000001018d0400 nid=56067 waiting for monitor entry  [0x00000008834d2000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at java.lang.Throwable.getCause(java.base@21.0.2/Throwable.java:422)
	- waiting to lock <0x00000002814c7760> (a kyo.fibers$package$Fibers$Interrupted$)
	at sttp.client3.SttpClientExceptionExtensions.defaultExceptionToSttpClientException(SttpClientExceptionExtensions.scala:29)
	at sttp.client3.SttpClientExceptionExtensions.defaultExceptionToSttpClientException$(SttpClientExceptionExtensions.scala:9)
	at sttp.client3.SttpClientException$.defaultExceptionToSttpClientException(SttpClientException.scala:24)
	at sttp.client3.HttpClientAsyncBackend.adjustExceptions$$anonfun$1(HttpClientAsyncBackend.scala:152)
	at sttp.client3.HttpClientAsyncBackend$$Lambda/0x000000b0012ccdd0.apply(Unknown Source)
	at sttp.client3.SttpClientException$$anon$1.applyOrElse(SttpClientException.scala:35)
	at sttp.client3.SttpClientException$$anon$1.applyOrElse(SttpClientException.scala:34)
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:35)
	at kyo.IOs$$anon$5.apply(ios.scala:53)
	at kyo.internal.KyoSttpMonad$$anon$3.apply(KyoSttpMonad.scala:19)
	at kyo.requests$package$Requests$$anon$2.apply(requests.scala:35)
	at kyo.scheduler.IOTask.eval(IOTask.scala:46)
	at kyo.scheduler.IOTask.run(IOTask.scala:83)
	at kyo.scheduler.Worker.runTask(Worker.scala:134)
	at kyo.scheduler.Worker.run(Worker.scala:112)
[...]

PR #929, opened 2024-12-13
introduce KyoException

fwbrasil:


kyo-data/shared/src/main/scala/kyo/KyoException.scala
```diff
@@ -0,0 +1,27 @@
+package kyo
+
+import kyo.*
+import kyo.Ansi.*
+import scala.util.control.NoStackTrace
+
+class KyoException private[kyo] (
+    message: Text = null,
+    cause: Text | Throwable = null
+)(using val frame: Frame) extends NoStackTrace:
```
sideeffffect, line 10:
```suggestion
)(using val frame: Frame) extends Exception(message.toString, cause match { case cause: Throwable => cause; case _ => null }) with NoStackTrace:
```
Would it make sense to make this a subclass of Exception? The docs says about Exception (and it's subclasses)

> indicates conditions that a reasonable application might want to catch

Is that the case here?

kyo-data/shared/src/main/scala/kyo/KyoException.scala
```diff
@@ -0,0 +1,27 @@
...
+            case _                => null
+
+    override def getMessage(): String =
+        val detail =
+            cause match
+                case _: Throwable           => Absent
+                case cause: Text @unchecked => Maybe(cause)
+
+        val msg = frame.render(("⚠️ KyoException".red.bold :: Maybe(message).toList ::: detail.toList).map(_.show)*)
+        s"\n$msg\n"
```
sideeffffect, line 24: Are the two `\n` really supposed to be there? Shouldn't formatting like this be potentially done by the users of `getMessage` instead of in `getMessage` itself?

kyo-data/shared/src/main/scala/kyo/KyoException.scala
hearnadam: I think you can also avoid the NoStackTrace mixin (though perhaps it's useful for users to see) via an argument to Exception/Throwable's constructor.

kyo-data/shared/src/main/scala/kyo/KyoException.scala
```diff
@@ -0,0 +1,27 @@
...
+            case cause: Throwable => cause
+            case _                => null
+
+    override def getMessage(): String =
+        val detail =
+            cause match
+                case _: Throwable           => Absent
+                case cause: Text @unchecked => Maybe(cause)
+
+        val msg = frame.render(("⚠️ KyoException".red.bold :: Maybe(message).toList ::: detail.toList).map(_.show)*)
```
hearnadam, line 23: it may be easier to use `.getOrElse("").show` to avoid the list?

kyo-data/shared/src/main/scala/kyo/KyoException.scala
fwbrasil: I've added `extends Exception` but kept the method overrides and `NoStackTrace`. We need to override `getCause` anyway to avoid synchronization (https://github.com/getkyo/kyo/pull/331) and the logic to handle non-`Throwable` causes seems more opaque as constructor parameters.

We also need `NoStackTrace` to signal to the kernel that it shouldn't [try to enrich](https://github.com/getkyo/kyo/blob/9cf49b271116fd51ec5303e644f670f048b44edb/kyo-kernel/shared/src/main/scala/kyo/kernel/Trace.scala#L83) the stack trace.

kyo-data/shared/src/main/scala/kyo/KyoException.scala
fwbrasil: The size of the list determines how many lines will be rendered

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
```diff
@@ -72,7 +72,7 @@ class AsyncTest extends Test:
                 .pipe(Async.runAndBlock(Duration.Infinity))
                 .pipe(Abort.run[Timeout](_))
                 .map {
-                    case Result.Fail(Timeout(_)) => succeed
+                    case Result.Fail(_: Timeout) => succeed
```
hearnadam, line 75: pre-existing: `_: Result.Fail[Timeout]`?

hearnadam (approved):

kyo-data/shared/src/main/scala/kyo/KyoException.scala
sideeffffect: > provide better rendering in the console

Yeah, totally. But my point was that it seems like a concert of the code which does the rendering/printing/formatting of an exception to standard output. `getMessage` is generally implemented in a way which is independent/oblivious whether the target is standard output, HTML, JSON or whatever else.

It's a very small thing, so I don't mind at all, but it caught my attention, so I've mentioned it. Feel free to ignore :slightly_smiling_face:

PR #933, opened 2024-12-15
[core] fix IOPromise exception handling

fwbrasil:
I noticed an odd scenario in a project I'm working on. The `IOPromise` flushing was throwing an exception due to a buggy `onInterrupt` function in the project. This PR ensures exception handling is present in all callback function evaluations.


kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -49,11 +49,7 @@ private[kyo] class IOPromise[+E, +A](init: State[E, A]) extends Safepoint.Interc
                 case l: Linked[E, A] @unchecked =>
                     interruptsLoop(l.p)
                 case _ =>
-                    try discard(other.interrupt(Result.Panic(Interrupt(frame))))
```
fwbrasil, line 52: This didn't require exception handling. It's probably copy-and-paste leftover

hearnadam (approved):

PR #935, opened 2024-12-16
KyoException initialize Exception

sideeffffect:
Maybe I'm being superstitious or overly cautious, but I feel like it's much safer if the Exception is initialized with these fields. For interoperability purposes and the like. With the Exception initialized like I'm suggesting, nothing can go wrong ever.

A very small, trivial thing, but I felt a strong urge to do it :sweat_smile:


fwbrasil (approved):
Better safe than sorry :) the build is failing because of formatting

sideeffffect:
formatted

sideeffffect:
It's green, finally :tada:

PR #937, opened 2024-12-17
reorganize kernel + remove unused imports

fwbrasil:
I've moved a few files to a new `internal` folder to make it easier to follow the main implementations in the `kyo.kernel` package. I also noticed several unused imports so I enabled the compiler warning and did a cleanup of the codebase.


fwbrasil:
merging to avoid conflicts since it's a simple refactoring

PR #939, opened 2024-12-17
[kyo-reactive-streams]: Remove un-necessary sleep time in test cases

### LINKED ISSUE #938: Reactive streams TCK tests are too slow
A quick test suite is an important design principle of Kyo's codebase. The new tests introduced by https://github.com/getkyo/kyo/pull/896 are taking much longer than all the other tests combined. We need to find a reliable way to execute them faster. cc/ @HollandDM

HollandDM:
Originally, `Async.sleep` was used to mimic the real-life delay of a publisher/subscriber, but in the `reactive-streams-tck` test cases, it provides little to no benefit while significantly slowing down the overall test. Additionally, a quick look at `fs2` and `zio` shows that they also do not use sleep in their test cases.

This PR takes it a step further by removing `Async.sleep` from all `kyo-reactive-stream` test cases. Since this feature is primarily about correctly combining effects, I believe that if individual components have been thoroughly tested, we can proceed with just the minimal testing.

Furthermore, because `StreamSubscriber` has two strategies, I initially duplicated the tests for both. However, I think it suffices to randomly choose one strategy for each test in the suite.

Overall, on my MacBook M3, the old tests took `1 minute and 37 seconds` to run, while the new tests complete in just `34 seconds`.

Aim to resolve #938

### DIFF
--- a/kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
+++ b/kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
@@ -190,12 +190,10 @@ abstract private class PublisherToSubscriberTest extends Test:
                 subscriber2 <- streamSubscriber
                 subscriber3 <- streamSubscriber
                 subscriber4 <- streamSubscriber
-                gatherFiber <- Fiber.gather[Nothing, Unit, Any](List(
-                    subscriber1.stream.run.unit,
-                    subscriber2.stream.run.unit,
-                    subscriber3.stream.run.unit,
-                    subscriber4.stream.run.unit
-                ))
+                fiber1      <- Async.run(subscriber1.stream.run.unit)
+                fiber2      <- Async.run(subscriber2.stream.run.unit)
+                fiber3      <- Async.run(subscriber3.stream.run.unit)
+                fiber4      <- Async.run(subscriber4.stream.run.unit)
                 publisherFiber <- Async.run(Resource.run(
                     Stream(Emit.andMap(Chunk.empty)(emit(_, counter)))
                         .toPublisher
@@ -209,8 +207,11 @@ abstract private class PublisherToSubscriberTest extends Test:
                 ))
                 _ <- Clock.sleep(10.millis).map { fiber =>
                     fiber.onComplete(_ => publisherFiber.interrupt.unit).andThen(fiber)
-                }.map(_.get)
-                _ <- gatherFiber.get
+                }.map(_.getResult)
+                _ <- fiber1.getResult
+                _ <- fiber2.getResult
+                _ <- fiber3.getResult
+                _ <- fiber4.getResult
             yield assert(true)
             end for
         }
--- a/kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
+++ b/kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
@@ -11,35 +11,25 @@ abstract private class PublisherToSubscriberTest extends Test:
     protected def streamSubscriber: StreamSubscriber[Int] < IO
 
     "should have the same output as input" in runJVM {
-        val stream = Stream
-            .range(0, MaxStreamLength, 1, BufferSize)
-            .map { int =>
-                Random
-                    .use(_.nextInt(10))
-                    .map(millis => Async.sleep(Duration.fromUnits(millis, Duration.Units.Millis)))
-                    .andThen(int)
-            }
+        val stream = Stream.range(0, MaxStreamLength, 1, BufferSize)
         for
             publisher  <- stream.toPublisher
             subscriber <- streamSubscriber
             _ = publisher.subscribe(subscriber)
             (isSame, _) <- subscriber.stream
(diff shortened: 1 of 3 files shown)


kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
```diff
@@ -225,17 +207,13 @@ abstract private class PublisherToSubscriberTest extends Test:
                             publisher.subscribe(subscriber3)
                             publisher.subscribe(subscriber4)
                         }
-                        .andThen(Async.sleep(2.seconds))
+                        .andThen(Async.sleep(1.seconds))
```
fwbrasil, line 210: could it be 100ms?

fwbrasil:
Another way we can try to speed it up is by running tests in parallel. If you override `fork := false` in the specific SBT module, the tests will run start running in parallel but without forking JVMs for execution. If I'm not mistaken, SBT will pick up the number of cores as the parallelism. Since these tests seem to use little CPU, we could probably override it to use a higher concurrency but I don't recall the SBT config for that.

kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
HollandDM: This is more like a keep alive effect, the whole fiber will be interrupted below in 10.millis. So i dont think duration here is that important

HollandDM:
After some playing around, I figured there are 2 solutions:
1. To set `fork := false` like you said, this cases it will fork all other test to another JVM and run them sequentially. The downside is that if the test failed, so do the sbt process.
2. Tweak the `concurrentRestrictions` and `parallelExecution`, e.g:
```scala
Global / parallelExecution := true
Global / concurrentRestrictions := Def.setting {
      val par = parallelExecution.value
      val max = EvaluateTask.SystemProcessors
      Tags.limitAll(if (par) max else 1) ::
        Tags.limit(Tags.ForkedTestGroup, if (par) max else 1) ::
        Tags.exclusiveGroup(Tags.Clean) ::
        Nil
    }.value
```
`concurrentRestrictions` value is the default value taken from sbt codebase, but the `Tags.ForkedTestGroup` will be free from 1. This will run **all the tests** in parallel, there for `kyo-reactive-streams` tests won't block other.
I saw that sbt allow us to group the test cases into group, then decide the fork strategy for each. It bests to group every other tests to one group and `kyo-reactive-streams` tests to another one and fork them separately, but this will required us to "un-aggregate" the test task and build them manually (as far as i can tell, could be wrong), so it's quite hard to do compare to the 2 solutions above, WDYT?

fwbrasil:
@HollandDM I tried the config you mentioned but it still executes sequentially. While debugging, I noticed the TCK tests end up waiting for the timeout in multiple scenarios to wait for potential new events. If I set `new TestEnvironment(10L)` in the constructor, the tests run quite quickly. Is that an option?

HollandDM:
@fwbrasil weird, maybe testing environment only has 1 core?
Anw, if we can speed up the test by limiting the waiting time, then it should be good enough, let me update this.

kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
```diff
@@ -208,12 +190,10 @@ abstract private class PublisherToSubscriberTest extends Test:
...
-                _ <- Fiber.parallelUnbounded[Nothing, Unit, Any](List(
-                    subscriber1.stream.run.unit,
-                    subscriber2.stream.run.unit,
-                    subscriber3.stream.run.unit,
-                    subscriber4.stream.run.unit
-                ))
+                fiber1      <- Async.run(subscriber1.stream.run.unit)
+                fiber2      <- Async.run(subscriber2.stream.run.unit)
+                fiber3      <- Async.run(subscriber3.stream.run.unit)
+                fiber4      <- Async.run(subscriber4.stream.run.unit)
```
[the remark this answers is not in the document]
fwbrasil: What's the goal here? Why not use `parallelUnbounded`?

fwbrasil (approved):
Can we merge this or are you're still planing to work on it? Thanks!

kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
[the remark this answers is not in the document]
HollandDM: The goal here is to prove that when publisher fiber got interrupted, all 4 subscribers stream will be interrupted. So ‘gather’ is more correct here.
I’ll try to reproduce the issue later

HollandDM:
Is good to merge now

HollandDM:
@fwbrasil This is good to merge now, we can change the ci runtime back to 15 mins after this

kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
sideeffffect: Is there a `Async.forever` or `Async.never` which will sleep forever or never return?

PR #941, opened 2024-12-17
[kernel] additional scaladocs

fwbrasil:
Scaladocs for some internal types to make it easier to follow the code.

### DIFF
--- a/kyo-kernel/shared/src/main/scala/kyo/kernel/internal/Kyo.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/kernel/internal/Kyo.scala
@@ -6,28 +6,77 @@ import kyo.Frame
 import kyo.Tag
 import kyo.kernel.ArrowEffect
 
-sealed private[kernel] trait Defer extends ArrowEffect[Const[Unit], Const[Unit]]
-
+/** The core representation of suspended computations in Kyo's effect system.
+  *
+  * A value of type Kyo[A, S] packages together the key pieces needed to represent and resume a suspended computation:
+  *
+  *   - An input value that will be transformed by an effect handler
+  *   - A tag identifying which handler should interpret the effect
+  *   - A continuation that resumes computation with the handler's result
+  *
+  * Continuations form the heart of composition in Kyo. Each continuation receives a result and returns a new Kyo instance, creating a chain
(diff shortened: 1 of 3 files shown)


kyo-kernel/shared/src/main/scala/kyo/kernel/internal/Kyo.scala
```diff
@@ -6,28 +6,77 @@ import kyo.Frame
...
     final override def toString =
         s"Kyo(${tag.show}, Input($input), ${frame.position.show}, ${frame.snippetShort})"
 end KyoSuspend
 
+/** Specialized effect type for pure computation deferral using Unit input/output.
+  */
 abstract private[kernel] class KyoContinue[I[_], O[_], E <: ArrowEffect[I, O], A, B, S](kyo: KyoSuspend[I, O, E, A, ?, ?])
     extends KyoSuspend[I, O, E, A, B, S]:
     val tag   = kyo.tag
     val input = kyo.input
```
hearnadam, line 66: is it actually better to use `val` here?

kyo-kernel/shared/src/main/scala/kyo/kernel/internal/Trace.scala
```diff
@@ -6,6 +6,19 @@ import scala.annotation.tailrec
...
+  * Traces are managed through a pooling system (see [[TracePool]]) to minimize allocation overhead. The pool maintains both thread-local
+  * and global sets of reusable Trace instances.
+  *
+  * @param frames
+  *   The circular buffer array storing Frame objects
+  * @param index
+  *   Current position in the circular buffer
+  */
 final private[kyo] class Trace(
     private[kernel] var frames: Array[Frame],
```
hearnadam, line 23:
```suggestion
    private[kernel] val frames: Array[Frame],
```

hearnadam (approved):

kyo-kernel/shared/src/main/scala/kyo/kernel/internal/Kyo.scala
fwbrasil: I haven't tried changing this in a long time but the vals avoid having to traverse accumulated transformations to obtain the tag and input during effect handling

PR #948, opened 2024-12-20
Introduce new Tag mechanism based on single opaque type

hearnadam:
Over the last few weeks I've been working on a new encoding of Tag to be able to represent intersections and unions in a single String. Primarily, I want to make sure that this grammar is sound. The subtype checking algorithm is almost fully written by claude. Once I validate the encoding, I will hand optimize an implementation.

This new encoding also removes the optimizations with compact tags, but that could probably be added back. I think we can support more compact values if we use non-alphabet (UTF8? 16?) characters.

I needed to temporarily disable TypeMap as I we need to produce an intersection from the individual Tags. I will followup to re-enable + the tests I removed.


hearnadam:
I don't have time to work on this, and we can always revive.

PR #951, opened 2024-12-23
Kyo's stream compressions

HollandDM:
The second step toward solving #764, this ticket introduce `Deflate`/`Inflate` and `Gzip`/`Gunzip` to Kyo's `Stream[Byte, S]`.

### DIFF
--- a/kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
+++ b/kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
@@ -1,0 +1,745 @@
+package kyo
+
+import java.util.zip.CRC32
+import java.util.zip.DataFormatException
+import java.util.zip.Deflater
+import java.util.zip.Inflater
+import kyo.*
+import kyo.kernel.ArrowEffect
+import kyo.kernel.Loop
+import scala.annotation.tailrec
+
+object StreamCompression:
+
+    final class StreamCompressionException(reason: String | Throwable)(using Frame) extends KyoException(
+            message = reason match
+                case string: String => Text(string)
+                case _: Throwable   => null,
+            cause = reason
+        )
+
+    enum CompressionLevel(val value: Int) derives CanEqual:
+        case Default         extends CompressionLevel(-1)
+        case NoCompression   extends CompressionLevel(0)
+        case BestSpeed       extends CompressionLevel(1)
+        case Level2          extends CompressionLevel(2)
+        case Level3          extends CompressionLevel(3)
+        case Level4          extends CompressionLevel(4)
+        case Level5          extends CompressionLevel(5)
+        case Level6          extends CompressionLevel(6)
+        case Level7          extends CompressionLevel(7)
+        case Level8          extends CompressionLevel(8)
+        case BestCompression extends CompressionLevel(9)
+    end CompressionLevel
+
+    enum CompressionStrategy(val value: Int) derives CanEqual:
+        case Default         extends CompressionStrategy(0)
+        case Filtered        extends CompressionStrategy(1)
+        case HuffmanOnly     extends CompressionStrategy(2)
+        case BestSpeed       extends CompressionStrategy(2)
+        case BestCompression extends CompressionStrategy(0)
+    end CompressionStrategy
+
+    enum FlushMode(val value: Int) derives CanEqual:
+        case Default         extends FlushMode(0)
+        case NoFlush         extends FlushMode(0)
+        case SyncFlush       extends FlushMode(2)
+        case FullFlush       extends FlushMode(3)
+        case BestSpeed       extends FlushMode(3)
+        case BestCompression extends FlushMode(0)
+    end FlushMode
+
+    private inline def toUnboxByteArray(byteChunk: Chunk[Byte]): Array[Byte] =
+        val builder = Array.newBuilder[Byte]
+        builder.addAll(byteChunk)
+        builder.result()
+    end toUnboxByteArray
+
+    private inline def fromUnboxByteArray(byteArray: Array[Byte], size: Int): Chunk[Byte] =
+        val builder = Chunk.newBuilder[Byte]
+        var i       = 0
+        var it      = byteArray.iterator
+        while it.hasNext && i < size do
+            builder.addOne(it.next())
+            i += 1
+        builder.result()
+    end fromUnboxByteArray
+
+    extension [A, Ctx](stream: Stream[A, Ctx])
+        inline def deflate(
+            bufferSize: Int = 1 << 15,
+            compressionLevel: CompressionLevel = CompressionLevel.Default,
+            strategy: CompressionStrategy = CompressionStrategy.Default,
+            flushMode: FlushMode = FlushMode.Default,
+            noWrap: Boolean = false
+        )(
+            using
+            tag: Tag[Emit[Chunk[Byte]]],
+            frame: Frame,
+            ev: A <:< Byte
+        ): Stream[Byte, Resource & IO & Ctx] =
+            deflateStream(stream.map(ev.apply), bufferSize, compressionLevel, strategy, flushMode, noWrap)
+        end deflate
+
+        inline def inflate(
+            bufferSize: Int = 1 << 15,
+            noWrap: Boolean = false
+        )(
+            using
+            tag: Tag[Emit[Chunk[Byte]]],
+            frame: Frame,
+            ev: A <:< Byte
+        ): Stream[Byte, IO & Resource & Ctx & Abort[StreamCompressionException]] =
+            inflateStream(stream.map(ev.apply), bufferSize, noWrap)
+        end inflate
+
+        inline def gzip(
+            bufferSize: Int = 1 << 15,
+            compressionLevel: CompressionLevel = CompressionLevel.Default,
+            strategy: CompressionStrategy = CompressionStrategy.Default,
+            flushMode: FlushMode = FlushMode.Default
+        )(
+            using
+            tag: Tag[Emit[Chunk[Byte]]],
+            frame: Frame,
+            ev: A <:< Byte
+        ): Stream[Byte, Resource & IO & Ctx] =
+            gzipStream(stream.map(ev.apply), bufferSize, compressionLevel, strategy, flushMode)
+        end gzip
+
+        inline def gunzip(
+            bufferSize: Int = 1 << 15
+        )(
+            using
+            tag: Tag[Emit[Chunk[Byte]]],
+            frame: Frame,
+            ev: A <:< Byte
+        ): Stream[Byte, IO & Resource & Ctx & Abort[StreamCompressionException]] =
+            gunzipStream(stream.map(ev.apply), bufferSize)
+        end gunzip
+    end extension
+
+    private def deflateStream[Ctx](
+        stream: Stream[Byte, Ctx],
+        bufferSize: Int,
+        compressionLevel: CompressionLevel,
+        strategy: CompressionStrategy,
+        flushMode: FlushMode,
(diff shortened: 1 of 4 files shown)


kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -42,9 +42,30 @@ sealed abstract class Chunk[A] extends Seq[A] derives CanEqual:
...
         else dropLeftAndRight(0, length - Math.min(Math.max(0, n), length))
 
+    /** Takes the last n elements of the Chunk.
+      *
+      * @param n
+      *   the number of elements to take
+      * @return
+      *   a new Chunk containing the last n elements
+      */
+    override def takeRight(n: Int): Chunk[A] =
```
HollandDM, line 65: Add this, so that `takeRight` can return `Chunk` instead of `Seq`

kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -328,7 +349,7 @@ sealed abstract class Chunk[A] extends Seq[A] derives CanEqual:
                 case c: Compact[A] =>
                     val l = c.array.length
                     if l > 0 then
-                        System.arraycopy(c.array, dropLeft, array, start, l - dropRight - dropLeft)
+                        Array.copy(c.array, dropLeft, array, start, l - dropRight - dropLeft)
```
HollandDM, line 352: `Array.copy` is more correct for Scala because it has fallback for dealing with `boxed`/`unboxed` primitive types. It does call `System.arraycopy` when possible.

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
```diff
@@ -0,0 +1,724 @@
...
+                        Emit.andMap(Chunk.empty): ack =>
+                            maybeEmitFn match
+                                case Present(emitFn) =>
+                                    Loop.continue(GunzipState.InflateInput(Chunk.empty, inflater, contentCrc32, emitFn(ack)))
+                                case Absent =>
+                                    Loop.continue(GunzipState.CheckTrailer(Chunk.empty, inflater, contentCrc32, Absent))
+                    else
+                        local.use: buffer =>
+                            Abort
+                                .catching[DataFormatException](inflater.inflate(buffer))
```
HollandDM, line 667: catching `DataFormatException` forces the signature to have `DataFormatException | StreamCompressionException`, I'd like a method that can `mapError` just like ZIO

HollandDM:
The code has some duplicate states between `deflate & gzip` and `inflate & unzip` because the logic can be reuse somewhat. But I want to make them distinct so that the logic can be more clear

kyo-data/shared/src/main/scala/kyo/Chunk.scala
fwbrasil: Did you observe issues related to this? The underlying arrays are always boxed in `Chunk`

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
```diff
@@ -0,0 +1,724 @@
...
+        end GZipState
+
+        def emit(
+            stream: Stream[Byte, Ctx],
+            bufferSize: Int,
+            compressionLevel: CompressionLevel,
+            strategy: CompressionStrategy,
+            flushMode: FlushMode
+        ): Ack < (Resource & IO & Emit[Chunk[Byte]] & Ctx) =
+            val local = Local.init(new Array[Byte](bufferSize))
```
fwbrasil, line 218: Why use a local here? Isn't it possible to use the Array directly? This also needs an `IO` suspension since it's allocating a mutable object.

kyo-data/shared/src/main/scala/kyo/Chunk.scala
HollandDM: When you need to provide `Chunk[Byte]` to a java method that requires `[]byte`. It'll throw a runtime exception about mismatch between `[]byte` and `[]array`.

HollandDM: One work around I used to do is to convert `Chunk` to `List` first, then `.toArray` that `List`. I can do it if you want

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
HollandDM: `Array[Byte]` here works only as a buffer for `Deflater`/`Inflater` to write data to. Buffer content then will be copied into a `Chunk[Byte]`, so the buffer can be reuse. Using `IO` can lead to unnecessary allocations.

hearnadam: Can you use a ChunkBuilder? It should be reusable.

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
```diff
@@ -0,0 +1,724 @@
...
+                            0,
+                            0,
+                            0,
+                            compressionLevel match
+                                case CompressionLevel.BestSpeed       => 0x4
+                                case CompressionLevel.BestCompression => 0x2
+                                case _                                => 0
+                            ,    // XFL: Extra flags
+                            0xff // OS: Operating System
+                        ).map(_.toByte))
```
hearnadam, line 253: Is the conversion necessary? or can we just use `Chunk[Byte](...)`?

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
```diff
@@ -0,0 +1,724 @@
...
+                case GZipState.EmitDeflated(deflater, crc32, maybeEmitFn, chunk) =>
+                    Emit.andMap(chunk): ack =>
+                        maybeEmitFn match
+                            case Present(emitFn) => Loop.continue(GZipState.DeflateInput(deflater, crc32, emitFn(ack)))
+                            case Absent          => Loop.continue(GZipState.SendTrailer(deflater, crc32))
+                case GZipState.SendTrailer(deflater, crc32) =>
+                    IO {
+                        val crcValue  = crc32.getValue
+                        val bytesRead = deflater.getBytesRead()
+                        val trailer = Chunk.from(Seq(
```
hearnadam, line 283: same comment as above

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
```diff
@@ -0,0 +1,724 @@
...
+                            0x8,  // CM: Compression Method
+                            0,
+                            0,
+                            0,
+                            0,
+                            0,
+                            compressionLevel match
+                                case CompressionLevel.BestSpeed       => 0x4
+                                case CompressionLevel.BestCompression => 0x2
+                                case _                                => 0
```
hearnadam, line 250: Do we need the other enum values if they aren't used?

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
HollandDM: Array[Byte] is required because deflate/inflate java functions take those as a buffer to write the output, this is the contract of those methods, so i don't think we can't provide other thing.

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
HollandDM: These are for the methods parameters, normally users don't have to input it, but I think it'll be nice to have the flexibility

HollandDM:
I pushed a new version to address some feedbacks

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
```diff
@@ -0,0 +1,733 @@
...
+            compressionLevel: CompressionLevel,
+            strategy: CompressionStrategy,
+            flushMode: FlushMode,
+            noWrap: Boolean
+        )(
+            using
+            Tag[Emit[Chunk[Byte]]],
+            Frame
+        ): Ack < (Resource & IO & Emit[Chunk[Byte]] & Ctx) =
+            val local = Local.init(new Array[Byte](bufferSize))
```
fwbrasil, line 143: I still don't understand why this can't be `val buffer = new Array[Byte](bufferSize)` and then use `buffer` directly in the loop. Also, an `IO` suspension is required for the initialization of the buffer since it's a mutable value.

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
fwbrasil: I still don't understand why this can't be `val buffer = new Array[Byte](bufferSize)` and then use buffer directly in the loop. Also, an IO suspension is required for the initialization of the buffer since it's a mutable value. It'll break referential transparency otherwise. Same for `deflateStream`.

kyo-data/shared/src/main/scala/kyo/Chunk.scala
fwbrasil: `Chunk` always uses boxed arrays. I think you can use `chunk.toArray` and then copy the contents to the unboxed array

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -329,6 +329,27 @@ object Abort:
...
+          * @param v
+          *   The computation to run and catch exceptions from
+          * @tparam A
+          *   The return type of the computation
+          * @tparam S
+          *   The effect type of the computation
+          * @return
+          *   A computation that may fail with an Abort[E] if an exception of type E is caught
+          */
+        def apply[A, S, E1](v: => A < S)(f: E => E1)(
```
fwbrasil, line 344: Can you make `f` the first parameter? It works better with `pipe`: 

```scala
someComputation.pipe(Abort.catching[SomeException](ex => someValue))
```

fwbrasil:
Sorry for the delay again @HollandDM. We should be more responsive now after the holidays

HollandDM:
Updated. I changed buffer from local to just `IO` as recommended

kyo-data/shared/src/main/scala/kyo/Chunk.scala
fwbrasil: can this be reverted?

fwbrasil (approved):
LGTM besides the `Chunk` change

kyo-data/shared/src/main/scala/kyo/Chunk.scala
HollandDM: Updated, although I had to do some conversion, but if it is by design then I'm fine with it

HollandDM: Reverted!

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
```diff
@@ -0,0 +1,745 @@
...
+    enum FlushMode(val value: Int) derives CanEqual:
+        case Default         extends FlushMode(0)
+        case NoFlush         extends FlushMode(0)
+        case SyncFlush       extends FlushMode(2)
+        case FullFlush       extends FlushMode(3)
+        case BestSpeed       extends FlushMode(3)
+        case BestCompression extends FlushMode(0)
+    end FlushMode
+
+    private inline def toUnboxByteArray(byteChunk: Chunk[Byte]): Array[Byte] =
```
HollandDM, line 52: Chunk <-> unbox byte array conversions are here

PR #955, opened 2024-12-27
Simple PR to check CI

sideeffffect:


kyo-data/shared/src/main/scala/kyo/KyoException.scala
```diff
@@ -7,7 +7,7 @@ import scala.util.control.NoStackTrace
 class KyoException private[kyo] (
     message: Text = null,
     cause: Text | Throwable = null
-)(using val frame: Frame) extends Exception with NoStackTrace:
+)(using val frame: Frame) extends Exception(message.toString) with NoStackTrace:
```
sideeffffect, line 10: @fwbrasil @hearnadam Hello guys, this seemingly innocent change causes tests to fail... something weird is going on, but I don't know what :scream_cat:

fwbrasil: It seems it's a null pointer exception. `FailedTransaction` extends `KyoException` without a message or cause and the default for them is `null`. I think we can change the default values to an empty string.

sideeffffect: Oh, yes, I didn't think of `null`s :facepalm: sorry...

PR #959, opened 2024-12-27
[data][stm] Record and TTable

### LINKED ISSUE #917: [data] records
I've explored an encoding of records using a type union to maintain the structure information. It's able to provide proper sub-type relationships between records but I couldn't figure out an issue. If a field is added multiple times with different types, they should be unified with a type union of the possible value types. Adding variance to `Value` doesn't work because then getting values from the record get inferred as `Nothing`.

Another option is macros but I wanted to see if it'd be possible to avoid them.

https://scastie.scala-lang.org/46EN1AHCQwOEpvI6YTXo9Q
```scala
object impl:

    infix opaque type ~>[Name <: String, Value] = (Name, Value)

    extension [Name <: String](self: Name)
        def ~>[Value](value: Value): self.type ~> Value = (self, value)

    opaque type Record[-Fields] = Map[String, Any]

    object Record:
        def init: Record[Nothing] = Map.empty

        extension [Fields](self: Record[Fields])
            def toMap: Map[String, Any] = self

            def apply[Name <: String, Value](field: Name ~> Value): Record[Fields | Name ~> Value] =
                self + field

            def apply[Name <: String, Value](name: Name)(using name.type ~> Value <:< Fields): Value =
                self(name).asInstanceOf[Value]

            def ++[Fields2](other: Record[Fields2]): Record[Fields | Fields2] =
                self ++ other
        end extension

    end Record
end impl

object Demo:
    import impl.*

    type Person   = Record["name" ~> String | "age" ~> Int]
    type Employee = Record["name" ~> String | "age" ~> Int | "department" ~> String]
    type Client   = Record["name" ~> String | "age" ~> Int | "company" ~> String]

    val employee: Employee       = Record.init("name" ~> "John")("age" ~> 33)("department" ~> "some dept")
    val employeeIsPerson: Person = employee

    val person: Person = Record.init("name" ~> "John")("age" ~> 33)
    // val personIsNotEmployee: Employee = person

    // val cantAddFieds: Record["name" ~> String | "inexistent" ~> Int] = person

    // This should be Record["field" ~> (String | Int)]
    val bug: Record["field" ~> String & "field" ~> Int] = Record.init("field" ~> "someString")("field" ~> 42)
end Demo
```

cc/ @johnhungerford in case you want to have some fun with a type-level puzzle :)

fwbrasil:
Fixes #917

This PR introduces a `Record` implementation that uses intersection types to track fields. This approach enables more advanced API patterns like the `TTable.Indexed` implementation that constrains queries to indexed fields only.


kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -0,0 +1,642 @@
...
+
+export Record.`~`
+
+object Record:
+    given [Fields]: Flat[Record[Fields]] = Flat.unsafe.bypass
+
+    inline given [Fields]: Tag[Record[Fields]] =
+        scala.compiletime.error(
+            "Cannot derive Tag for Record type. This commonly occurs when trying to nest Records, " +
+                "which is not currently supported by the Tag implementation."
```
fwbrasil, line 84: should be fixed by https://github.com/getkyo/kyo/pull/948

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -0,0 +1,642 @@
...
+    /** Type class for working with sets of Record fields.
+      *
+      * AsFields provides type-safe field set operations and is used primarily for the `compact` operation on Records.
+      *
+      * @tparam A
+      *   The combined type of all fields in the set
+      */
+    opaque type AsFields[+A] <: Set[Field[?, ?]] = Set[Field[?, ?]]
+
+    trait AsFields22:
```
fwbrasil, line 252: I tried a few approaches but I couldn't make the compiler destructure the type intersections so I decided to go with brute force

kyo-stm/shared/src/main/scala/kyo/TTable.scala
```diff
@@ -0,0 +1,280 @@
...
+object TTable:
+
+    /** Initializes a new basic table without indexing.
+      *
+      * @tparam Fields
+      *   The record structure for the table
+      * @return
+      *   A new TTable instance within the IO effect
+      */
+    def init[Fields: AsFields](using Frame): TTable[Fields] < IO =
```
fwbrasil, line 118: The distinction between `init` and `initNow` methods don't make much sense anymore after https://github.com/getkyo/kyo/pull/949. I'm adding only `init` here and will follow up to do the same for the other impls in the module

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -0,0 +1,469 @@
...
+        }
+
+        "doesn't limit the use of nested records" in pendingUntilFixed {
+            val inner = "x" ~ 1 & "y" ~ 2
+            assertCompiles(""""z" ~ inner""")
+        }
+    }
+
+    "Render" - {
+        "simple record" in pendingUntilFixed {
```
fwbrasil, line 463: I couldn't find a way to collect the `Render` evidences for each of the field types

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -0,0 +1,646 @@
...
+  * record[String]("value") // Returns "string"
+  * record[Int]("value")    // Returns 42
+  * }}}
+  *
+  * =Known Limitations=
+  *
+  *   - Nested records cannot be created directly
+  *   - Tag derivation for Records is not currently supported
+  *   - Records with more than 22 fields have limited type-level support, though runtime support exists for larger structures
+  *   - CanEqual and Render instances are not provided for Records
```
fwbrasil, line 73: I'll create tickets to follow up on these limitations

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -0,0 +1,643 @@
...
+        def compact(using AsFields[Fields]): Record[Fields] =
+            Record(self.toMap.view.filterKeys(AsFields[Fields].contains(_)).toMap)
+
+    extension (self: String)
+        /** Creates a single-field Record with the string as the field name.
+          *
+          * @param value
+          *   The value to associate with the field
+          */
+        def ~[Value](value: Value)(using tag: Tag[Value]): Record[self.type ~ Value] =
```
hearnadam, line 162: What would you think about `~>` for similarity to tuple?

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -0,0 +1,643 @@
...
+            Record(Map(Field(self, tag) -> value))
```
hearnadam, line 163:
```suggestion
            Record(Map.empty.updated(Field(self, tag), value))
```

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -0,0 +1,643 @@
...
+        Record[${Fields}]
+
+        Possible causes:
+          1. The field does not exist in this Record
+          2. The field exists but has a different type than expected
+        """)
+        ev: Fields <:< Name ~ Unspecified,
+        tag: Tag[Unspecified]
+    ): Unspecified =
+        toMap(Field(name, tag)).asInstanceOf[Unspecified]
```
hearnadam, line 96: Do you think there is any likelyhood of tag mismatches for subtypes? Given the hashcode will be a direct tag lookup? If not, we can just combine `name+tag` to avoid an object allocation?

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -0,0 +1,646 @@
...
+      */
+    case class Field[Name <: String, Value](name: Name, tag: Tag[Value])
+
+    extension (self: String)
+        /** Creates a single-field Record with the string as the field name.
+          *
+          * @param value
+          *   The value to associate with the field
+          */
+        def ~[Value](value: Value)(using tag: Tag[Value]): Record[self.type ~ Value] =
```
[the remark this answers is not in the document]
fwbrasil: The initial versions used `~>` but I thought the syntax was a bit heavyweight:

```scala
type Person = name ~> String & age ~> Int
```
vs
```scala
type Person = name ~ String & age ~ Int
```

It's easier to read to me

kyo-data/shared/src/main/scala/kyo/Record.scala
fwbrasil: Good point about avoiding the allocation. It isn't trivial, though. We'd need to concatenate in a macro otherwise it'd allocate at runtime. It seems a good follow-up optimization?

kyo-data/shared/src/main/scala/kyo/Record.scala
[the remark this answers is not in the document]
hearnadam: Sounds good.

fwbrasil:
@johnhungerford could you review this PR? It includes that solution you shared to produce records from case classes

kyo-stm/shared/src/main/scala/kyo/TTable.scala
```diff
@@ -0,0 +1,299 @@
...
+  * All operations are performed within STM transactions to ensure consistency. The table guarantees that:
+  *   - Records are uniquely identified by an auto-incrementing ID
+  *   - Updates are atomic - either all fields are updated or none
+  *   - Indexes (if configured) are automatically maintained in sync with record changes
+  *   - Concurrent access is safely handled through STM transactions
+  *
+  * @tparam Fields
+  *   The record structure defined as a type-level list of field definitions (e.g. "name" ~ String & "age" ~ Int)
+  */
+abstract class TTable[Fields]:
```
hearnadam, line 23:
```suggestion
sealed abstract class TTable[Fields]:
```

kyo-stm/shared/src/main/scala/kyo/TTable.scala
```diff
@@ -0,0 +1,299 @@
...
+        given Flat[Id] = Flat.derive
+
+    /** Converts a raw Int to the table's ID type. Should only be used internally by the implementation.
+      *
+      * @param id
+      *   The raw integer ID value
+      * @return
+      *   The typed ID value
+      */
+    def unsafeId(id: Int): Id
```
hearnadam, line 38:
```suggestion
    protected def unsafeId(id: Int): Id
```

fwbrasil: The idea here is to provide a way for the user to produce an id directly from `Int`. For example, if an update comes from an UI, this method enables converting the raw id value to the path-dependent type

kyo-data/shared/src/main/scala/kyo/Record.scala
johnhungerford: I'm not sure I understand what the tag is doing, insofar as it will be the same type for everything (`Unspecified`).

kyo-data/shared/src/main/scala/kyo/Record.scala
johnhungerford: Cool. This never occurred to me.

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -0,0 +1,643 @@
...
+  * val record: Record["name" ~ String & "age" ~ Int] = Record.fromProduct(Person("Alice", 30))
+  * }}}
+  *
+  * =Field Access=
+  * Fields are accessed with compile-time verification of both existence and type:
+  * {{{
+  * record.name // Returns "Alice" as String
+  * record.age  // Returns 30 as Int
+  * record.nonexistent // Won't compile
+  * (record.name: Int) // Won't compile
```
johnhungerford, line 35: What about `(record.name: String)`? I assume this won't compile either, and if so, is there a way to get properly typed values?

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -0,0 +1,470 @@
...
+            assert(merged.age == 31)
+            assert(merged.city == "Paris")
+        }
+
+        "merge with type conflicts" in {
+            val record1 = "value" ~ "string"
+            val record2 = "value" ~ 42
+            val merged  = record1 & record2
+            assert((merged.value: Int) == 42)
+            assert((merged.value: String) == "string")
```
johnhungerford, line 209: This seems like a foot-gun. Are you sure you want to support having multiple types? What's the behavior if a type annotation is not provided explicitly? Does it really make sense to support multiple values of different types for a given field name?

kyo-data/shared/src/main/scala/kyo/Record.scala
fwbrasil: I had used `Unspecified` to try to produce a better error messages including the expected type. It didn't work and I forgot to rename it back to `Value`. It's actually properly inferred based on the type-level field definitions.

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    "field access" - {
+        "type-safe access" in {
+            val record = "name" ~ "Bob" & "age" ~ 25
+            assert(record.name == "Bob")
+            assert(record.age == 25)
+        }
+
+        "incorrect type access should not compile" in {
```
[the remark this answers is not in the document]
fwbrasil: that'd be essentially all other scenarios since we have strict equality enabled :) There are a few tests that use explicit types as well

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
fwbrasil: I actually didn't want to have this behavior. Ideally, `"field1" ~ String & "field2" ~ Int` should be inferred as `"field1" ~ (String | Int)` but I tried a few times to play with the variance and couldn't make it work. The alternative I found was allowing duplicate names and using tags to disambiguate. It's still safe to have fields with the same name and different types since all access is based on both components.

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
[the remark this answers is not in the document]
fwbrasil: I've added explicit type checking via vals in the previous scenario

kyo-data/shared/src/main/scala/kyo/Record.scala
fwbrasil: This got me working if we couldn't generalize the transformation to extract individual types out of an intersection. The issues we have with `Boundary` are essentially because of having to split the pending set into individual elements to check if they're all `ContextEffect`s.

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
fwbrasil: About the behavior in case the compiler doesn’t infer a specific result type, the first matching field and its specific type is automatically selected. I agree it’s a rough edge. I wonder if it’d be possible to disallow name conflicts via macros.

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -0,0 +1,473 @@
...
+            assertCompiles("Record.AsFields[Fields23]")
+        }
+
+        "intersections with duplicate field names but different types" in {
+            type DupeFields = "value" ~ Int & "value" ~ String
+            val fields = Record.AsFields[DupeFields]
+            assert(fields.size == 2)
+            assert(fields.map(_.name).toSet == Set("value"))
+            assert(fields.map(_.tag).toSet.size == 2)
+        }
```
johnhungerford, line 329: What's the behavior for an invalid field intersection? E.g.,

`Record.AsFields[Int & String & "value" ~ Int]`

johnhungerford: (I assume it would be considered valid, and would be treated as a single field equivalent to `"value" ~ Int`)

fwbrasil: I've added a new [group of tests](https://github.com/getkyo/kyo/pull/959/commits/584d890c25f48a47c2fed7af37599c7a8400e7d6) for malformed record types. The gist:

1. It's not possible produce instances of them using the Record API. It's also not possible via type widening.
2. But users can declare methods with a malformed type since `Fields` doesn't have a type bound (maybe we could add `~` but I think we'd need variance)
3. It's possible use a record with a malformed type via unsafe type casts but not all operations work. Field selection, `toMap`, `size`, `fields`, `size`, and `&` work properly but `compact` and any other API using `AsFields` don't
4. It's not possible to materialize `AsFields` for a malformed record type

fwbrasil: I took a shot at adding a type bound to `Fields` but I couldn't make `AsRecord.FieldsOf` work with it

kyo-stm/shared/src/main/scala/kyo/TTable.scala
hearnadam: Sounds good.

kyo-stm/shared/src/main/scala/kyo/TTable.scala
```diff
@@ -0,0 +1,299 @@
...
+        def queryIds[A](filter: Record[A])(
+            using
+            @implicitNotFound(indexMismatch) ev: Indexes <:< A,
+            frame: Frame
+        ): Chunk[Id] < STM =
+            Kyo.foreach(filter.toMap.toSeq) { (field, value) =>
+                indexes(field).getOrElse(value, Set.empty)
+            }.map { r =>
+                if r.isEmpty then Chunk.empty
+                else Chunk.from(r.reduce(_ intersect _).toSeq.sorted.map(unsafeId(_)))
```
hearnadam, line 233:
```suggestion
                else Chunk.from(r.reduce(_ intersect _).sorted.map(unsafeId(_)))
```

Chunk is a Seq!

hearnadam (approved):

kyo-stm/shared/src/main/scala/kyo/TTable.scala
fwbrasil: I had to revert it. It's actually a `Set` at this point. It comes from the index:

```scala
indexes: Map[Field[?, ?], TMap[Any, Set[Int]]]
```

PR #960, opened 2024-12-30
Emit combinators

johnhungerford:
- Added `Kyo.emit`
- Added extensions on `B < Emit[A]` and `B < Emit[Chunk[A]]` to handle emissions in a variety of ways
  - Chunk non-chunk emits
  - Convert to stream (either discarding result or returning additionally a separate effect just for the result)
  - handle to chunk
  - handle foreach
  - handle to channel


kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -903,3 +904,175 @@ extension [A, E, Ctx](effect: A < (Abort[E] & Async & Ctx))
             a1      <- fiberA1.join
         yield (a, a1)
 end extension
+
```
fwbrasil, line 907: This file is getting large. It'd be nice to reorganize the module to make it easier to locate implementations and their tests

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -903,3 +904,175 @@ extension [A, E, Ctx](effect: A < (Abort[E] & Async & Ctx))
...
+    def emitChunked(
+        chunkSize: Int
+    )(
+        using
+        tag: Tag[Emit[A]],
+        fr: Frame,
+        at: Tag[A],
+        fl: Flat[B]
+    ): B < (Emit[Chunk[A]] & S) =
+        ArrowEffect.handleState(tag, Chunk.empty[A], effect)(
```
fwbrasil, line 979: This seems a good operation to promote to kyo-core cc/ @hearnadam

PR #961, opened 2024-12-30
Channel: batched put

johnhungerford:
Added `Channel#putBatch` method that adds a chunk of elements to a channel in a batch. It will break up the chunk into multiple batches if the chunk exceeds the channel capacity.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -77,25 +77,20 @@ object Channel:
             else
                 IO.Unsafe {
                     var i = 0
-                    val attempt = boundary[Maybe[Closed]]:
+                    boundary[Unit < (Abort[Closed] & Async)]:
                         values.foreach: value =>
-                            val attempt = self.offer(value)
-                            if attempt.contains(false) then boundary.break(Maybe.Absent)
-                            else if attempt.isFail then boundary.break(attempt.failure)
-                            i += 1
-                        Maybe.Absent
-                    attempt match
-                        case Maybe.Present(closed) =>
-                            Abort.fail(closed)
-                        case _ =>
-                            if i >= values.length then
-                                ()
-                            else if i == values.length - 1 then
-                                self.putFiber(values.last).safe.get
-                            else
-                                self.putBatchFiber(values.dropLeft(i)).safe.get
-                            end if
-                    end match
+                            self.offer(value) match
+                                case Result.Success(true) =>
+                                    i += 1
+                                case Result.Success(false) =>
+                                    boundary.break:
+                                        if i == values.length - 1 then
+                                            self.putFiber(values.last).safe.get
+                                        else
+                                            self.putBatchFiber(values.dropLeft(i)).safe.get
+                                        end if
+                                case Result.Fail(cl) =>
+                                    boundary.break(Abort.fail(cl))
                 }
 
         /** Takes an element from the channel, asynchronously blocking if necessary.
@@ -268,13 +263,12 @@ object Channel:
         type Put
         def apply(value: A): Put
         def chunk(values: Chunk[A]): Put
-        def unapplyChunk(value: Put): Option[Chunk[A]]
-        def unapplyValue(value: Put): Option[A]
+        def fold[B](put: Put)(isValue: A => B)(isChunk: Chunk[A] => B): B
 
         object IsChunk:
-            def unapply(value: Put): Option[Chunk[A]] = unapplyChunk(value)
+            def unapply(value: Put): Option[Chunk[A]] = fold(value)(_ => None)(Some(_))
         object IsValue:
-            def unapply(value: Put): Option[A] = unapplyValue(value)
+            def unapply(value: Put): Option[A] = fold(value)(Some(_))(_ => None)
     end PutType
 
     sealed trait LowPriorityPutTypes:
@@ -283,26 +277,22 @@ object Channel:
             opaque type Put = Wr[A] | Chunk[A]
             def apply(value: A): Put         = Wr(value)
             def chunk(values: Chunk[A]): Put = values
-            def unapplyChunk(value: Chunk[A] | Wr[A]): Option[Chunk[A]] = value match
-                case Wr(_)           => None
-                case chunk: Chunk[A] => Some(chunk)
-            def unapplyValue(value: Put): Option[A] = value match
-                case Wr(inner) => Some(inner)
-                case _         => None
+            def fold[B](put: Put)(isValue: A => B)(isChunk: Chunk[A] => B): B = put match
+                case Wr(a)           => isValue(a)
+                case chunk: Chunk[A] => isChunk(chunk)
         end nested
     end LowPriorityPutTypes
 
     object PutType extends LowPriorityPutTypes:
+        // For any type that could not be a Chunk, there is no need to wrap the non-chunk
+        // value. (Using Seq as bounds because Chunk is not covariant.)
         given notNested[A](using NotGiven[A <:< Seq[Any]], NotGiven[Seq[Any] <:< A]): PutType[A] with
             opaque type Put = A | Chunk[A]
             def apply(value: A): Put         = value
             def chunk(values: Chunk[A]): Put = values
-            def unapplyChunk(value: Put): Option[Chunk[A]] = value match
-                case result: Chunk[A] @unchecked => Some(result)
-                case _                           => None
-            def unapplyValue(value: Put): Option[A] = value match
-                case result: Chunk[A] @unchecked => None
-                case a: A @unchecked             => Some(a)
+            def fold[B](put: Put)(isValue: A => B)(isChunk: Chunk[A] => B): B = put match
+                case chunk: Chunk[A] @unchecked => isChunk(chunk)
+                case a: A @unchecked            => isValue(a)
         end notNested
     end PutType
 
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -402,7 +402,7 @@ object Channel:
                                 if remaining.isEmpty then
                                     discard(promise.complete(Result.unit))
                                 else
-                                    discard(puts.add(Put.Batch(chunk.dropLeft(i), promise)))
+                                    discard(puts.add(Put.Batch(remaining, promise)))
                                 end if
                             case put @ Put.Value(value, promise) =>
                                 if queue.offer(value).contains(true) then
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -398,10 +398,9 @@ object Channel:
                                     chunk.foreach: value =>
                                         if queue.offer(value).contains(false) then boundary.break(())
                                         else i += 1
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -227,9 +259,51 @@ object Channel:
       * @warning
       *   The actual capacity may be larger than the specified capacity due to rounding.
       */
-    def init[A](capacity: Int, access: Access = Access.MultiProducerMultiConsumer)(using Frame): Channel[A] < IO =
+    def init[A](capacity: Int, access: Access = Access.MultiProducerMultiConsumer)(using PutType[A], Frame): Channel[A] < IO =
         IO.Unsafe(Unsafe.init(capacity, access))
 
+    sealed trait PutType[A]:
```
johnhungerford, line 265: This is included to deal with the issue of `puts` being either a single value or a chunk of values. It's impossible to check whether `A | Chunk[A]` is a single value or a chunk in the generic case, because `A` might itself be a chunk. On the other hand, I want to avoid wrapping them if we don't need to, as this will add an allocation. The `PutType` trait provides ad-hoc polymorphism, so that the put type will be `A | Chunk[A]` for any `Channel[A]` where `A` is not a `Chunk` of something, and the put type will be `Wr[Chunk[A]] | Chunk[Chunk[A]]` for any `Channel[Chunk[A]]`. (`Wr` is just a simply wrapping case class that allows safely pattern matching.)

johnhungerford: (I'm not sure if all this is worth the trouble. Maybe there's a simpler approach?)

johnhungerford:
Closes #916

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -64,6 +67,32 @@ object Channel:
...
+                                    i += 1
+                                case Result.Success(false) =>
+                                    boundary.break:
+                                        if i == values.length - 1 then
+                                            self.putFiber(values.last).safe.get
+                                        else
+                                            self.putBatchFiber(values.dropLeft(i)).safe.get
+                                        end if
+                                case Result.Fail(cl) =>
+                                    boundary.break(Abort.fail(cl))
```
hearnadam, line 93: It's important to match on Panic as well. 

```suggestion
                                case e: Result.Error[?] =>
                                    boundary.break(Abort.get(e))
```

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -257,11 +324,13 @@ object Channel:
...
             access: Access = Access.MultiProducerMultiConsumer
-        )(using initFrame: Frame, allow: AllowUnsafe): Unsafe[A] =
+        )(using putType: PutType[A], initFrame: Frame, allow: AllowUnsafe): Unsafe[A] =
+            import putType.Put
+
             new Unsafe[A]:
                 val queue = Queue.Unsafe.init[A](_capacity, access)
                 val takes = new MpmcUnboundedXaddArrayQueue[Promise.Unsafe[Closed, A]](8)
-                val puts  = new MpmcUnboundedXaddArrayQueue[(A, Promise.Unsafe[Closed, Unit])](8)
+                val puts  = new MpmcUnboundedXaddArrayQueue[(Put, Promise.Unsafe[Closed, Unit])](8)
```
hearnadam, line 333: What if we keep the current type for unbatched puts, then swap the order for batched puts?

`(A, Promise.Unsafe[...]) | (Promise.Unsafe[...], Chunk[A])`

This should have the same cost as we need to match on the type anyways, but we can avoid a wrapper.

johnhungerford: Ah good point. Only potential problem is edge case where `A =:= Promise.Unsafe[?,?]`.

How about we replace the tuple with an `enum`?:
```scala
enum Put[A]:
  case Value(value: A, promise: Promise.Unsafe[Closed, Unit])
  case Batch(batch: Chunk[A], promise: PromiseUnsafe[Closed, Unit])
```

I take it case classes and tuples have the same overhead?

kyo-core/shared/src/main/scala/kyo/Channel.scala
hearnadam: Sorry, my comment was not clear. It's not necessary to match on both Panic and Fail, just match on the supertype `Error`. You can then use `Abort.get` to avoid re-allocating fail/panic classes.

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -64,6 +67,32 @@ object Channel:
...
+                IO.Unsafe {
+                    var i = 0
+                    boundary[Unit < (Abort[Closed] & Async)]:
+                        values.foreach: value =>
+                            self.offer(value) match
+                                case Result.Success(true) =>
+                                    i += 1
+                                case Result.Success(false) =>
+                                    boundary.break:
+                                        if i == values.length - 1 then
```
hearnadam, line 87: unfortunately `.length` will cause a full iteration for collections like `List` (as well as `last`). Can we always put a batch?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -350,30 +396,52 @@ object Channel:
...
+                            case Put.Batch(chunk, promise) =>
+                                var i = 0
+                                boundary[Unit]:
+                                    chunk.foreach: value =>
+                                        if queue.offer(value).contains(false) then boundary.break(())
+                                        else i += 1
+                                if i >= chunk.length then
+                                    discard(promise.complete(Result.unit))
+                                else if i == (chunk.length - 1) then
+                                    discard(puts.add(Put.Value(chunk.last, promise)))
```
hearnadam, line 409: I think we should avoid this branch. It's saving a single allocation, but will likely only hold true some of the time. The `else` case should handle all cases, right?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -350,30 +396,52 @@ object Channel:
...
+                                    boundary[Unit]:
+                                        var i: Int = 0
+                                        batch.foreach: value =>
+                                            Maybe(takes.poll()) match
+                                                case Present(takePromise) if takePromise.complete(Result.success(value)) =>
+                                                    promise.completeDiscard(Result.unit)
+                                                    i += 1
+                                                case _ =>
+                                                    discard(puts.add(Put.Batch(batch.dropLeft(i), promise)))
+                                                    boundary.break(())
```
hearnadam, line 444: do you need the `()`? 

https://github.com/scala/scala3/blob/3.6.2/library/src/scala/util/boundary.scala#L52-L53

Do you know the cost of using boundary vs using recursive function with `batch(i)`?

johnhungerford: I have to think this is faster than iterating using `batch(i)` simply because `batch(i)` isn't guaranteed to be O(1) (am I off on this? You've been working on `Chunk`.) We could recursively iterate using `batch.dropLeft(1)`, but I thought it would be better not to allocate unnecessarily. Finally, we could use an iterator and simply run `while (it.hasNext && queue.offer(it.next()).contains(true)`, but I thought a boundary would be more elegant than an iterator.

hearnadam: `foreach` is simply an iterator + `while`.
```scala
  def foreach[U](f: A => U): Unit = {
    val it = iterator
    while(it.hasNext) f(it.next())
  }
```

Direct indexing will be faster if the Chunk is Indexed. Since `Drop` is essentially an 'Indexed' operation (if the underlying Chunk is), I think it should be pretty fast.

Let's convert to `toIndexed` when putting a new Batch (`Chunk.from`) technically will already take care of this.

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -350,30 +391,51 @@ object Channel:
...
+                                            promise.completeDiscard(Result.unit)
+                                        case _ => discard(puts.add(put))
+
+                                case Put.Batch(batch, promise) =>
+                                    boundary[Unit]:
+                                        var i: Int = 0
+                                        batch.foreach: value =>
+                                            Maybe(takes.poll()) match
+                                                case Present(takePromise) if takePromise.complete(Result.success(value)) =>
+                                                    promise.completeDiscard(Result.unit)
```
hearnadam, line 434: does this promise want to be completed on the first take?

johnhungerford: good point, this should be after all elements have been taken

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -64,6 +67,28 @@ object Channel:
...
+          *
+          * @param values
+          *   Chunk of elements to put
+          */
+        def putBatch(values: Seq[A])(using Frame): Unit < (Abort[Closed] & Async) =
+            if values.isEmpty then ()
+            else
+                IO.Unsafe {
+                    var i = 0
+                    boundary[Unit < (Abort[Closed] & Async)]:
```
fwbrasil, line 80: `boundary`/`break` is a dangerous primitive to mix with value-based effect composition since it's based on exception throwing. The use seems correct but could we avoid it?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -64,6 +67,28 @@ object Channel:
...
+                        values.foreach: value =>
+                            self.offer(value) match
```
fwbrasil, line 82: I think we still won't leverage the full potential of batched puts and takes. The `offer` call here will immediately flush the channel so a batched take might pick up only one of a few elements. Could we "lower" this operation to the unsafe impl so the flush happens only after all elements are in the queue?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -254,14 +280,21 @@ object Channel:
 
     /** WARNING: Low-level API meant for integrations, libraries, and performance-sensitive code. See AllowUnsafe for more details. */
     object Unsafe:
+        private enum Put[A]:
+            val promise: Promise.Unsafe[Closed, Unit]
+            case Batch(batch: Chunk[A], override val promise: Promise.Unsafe[Closed, Unit])
+            case Value(value: A, override val promise: Promise.Unsafe[Closed, Unit])
+        end Put
```
fwbrasil, line 287: This is very hot code so minimizing the memory footprint as much as possible would be ideal. It's possible to avoid an object header + a field pointer by extending from `IOPromise` directly:

```scala
sealed abstract class Put[A] extends IOPromise[Closed, Unit]
object Put:
   class Batch(batch: Chunk[A]) extends Put[A]
   class Value(value: A) extends Put[A]
```

We can do that as a follow up, though.

johnhungerford: wow wild!

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: good point. I added an `offerAll` method to `Channel.Unsafe`

fwbrasil (approved):

PR #962, opened 2024-12-31
[data] Chunk improvements

hearnadam:
- Extend `StrictOptimizedSeqOps` with `Chunk` to optimize `Seq` based methods
- Improve APIs to correctly return `Chunk[A]` instead of `Seq[A]`
- Optimize `Chunk.Indexed#iterator`
- Optimize `Chunk#take` to avoid allocations
- Optimize `Chunk#appended` to use `Chunk#append`

### DIFF
--- a/kyo-data/shared/src/test/scala/kyo/ChunkTest.scala
+++ b/kyo-data/shared/src/test/scala/kyo/ChunkTest.scala
@@ -10,7 +10,7 @@ class ChunkTest extends Test:
                 val array = Array("a", "b", "c")
                 val chunk = Chunk.from(array)
                 assert(chunk.isInstanceOf[Chunk.Indexed[String]])
-                assert(chunk.toSeq == Seq("a", "b", "c"))
+                assert(chunk == Seq("a", "b", "c"))
             }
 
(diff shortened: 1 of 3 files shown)


kyo-data/shared/src/test/scala/kyo/ChunkTest.scala
```diff
@@ -334,7 +334,7 @@ class ChunkTest extends Test:
     "toSeq" - {
         "converts a Chunk to an IndexedSeq" in {
             val chunk = Chunk(1, 2, 3, 4, 5)
-            assert(chunk.toSeq == Chunk(1, 2, 3, 4, 5))
+            assert(chunk == IndexedSeq(1, 2, 3, 4, 5))
```
johnhungerford, line 337: Shouldn't this still be `chunk.toSeq`?

hearnadam: The test is slightly incorrect to begin with. `toSeq` is now a noop given Chunk extends Seq.

I can remove the test.

fwbrasil (approved):

PR #965, opened 2024-12-31
Hub: streaming from Listener

johnhungerford:
- `Listener#stream` (stream ends when listener is closed)
- `Listener#streamFailing` (aborts with `Closed` when listener is closed)
- `Hub#listen` is now a `Resource` effect, closing listener as finalizer

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Hub.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Hub.scala
@@ -100,7 +100,7 @@ class Hub[A] private[kyo] (
       * @return
       *   a new Listener
       */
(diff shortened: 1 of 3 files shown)


kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -126,7 +126,8 @@ class Hub[A] private[kyo] (
                                     fail
                                 }
                             case false =>
-                                new Listener[A](this, child)
+                                Resource.acquireRelease(new Listener[A](this, child)): listener =>
+                                    listener.close.map(_ => ())
```
hearnadam, line 130:
```suggestion
                                    listener.close.unit
```

hearnadam (approved):

PR #968, opened 2025-01-03
Chunk unboxed support

sideeffffect:
So as I was playing with Chunk, I realized that it always boxes, even primitives, including Byte.

Is more support for un-boxing something you guys would be interested?

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Chunk.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Chunk.scala
@@ -5,6 +5,7 @@ import scala.annotation.tailrec
 import scala.annotation.targetName
 import scala.collection.IterableFactoryDefaults
 import scala.collection.StrictOptimizedSeqFactory
+import scala.collection.immutable.SeqOps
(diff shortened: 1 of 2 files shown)


hearnadam:
Boxing was intentional, though it's unclear if it's the best. There are scenarios where boxing is unavoidable, and converting between boxed and unboxed will result in quite an overhead.

Does this PR go far enough? Can you try mixing boxed and unboxed collections? I remember ZIO hand some issues with that. Does `Chunk` need to expose a `ClassTag` for allocating new arrays internally?

fwbrasil:
Yes, the main reason was the Stream implementation. Without boxing, each step of the computation boxes because Scala 3 doesn't have specialization yet. @sideeffffect Can you check if there's an impact on the benchmarks with this change:

```
kyo.bench/jmh:run -wi 10 -i 5 -r 1 -w 1 -f 1 -t 1 -foe true -rf json -prof gc Stream.*Kyo
```

fwbrasil:
Closing for now, please feel free to reopen!

PR #969, opened 2025-01-05
[core] introduce Async.memoize

fwbrasil:


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -696,6 +696,54 @@ object Async:
...
+      *
+      * WARNING: If the initial computation never completes (e.g., hangs indefinitely), all subsequent calls will be permanently blocked
+      * waiting for the result. Ensure the computation can complete in a reasonable time or introduce a timeout via `Async.timeout`.
+      *
+      * @param v
+      *   The computation to memoize
+      * @return
+      *   A function that returns the memoized computation result
+      */
+    def memoize[A: Flat, S](v: A < S)(using Frame): (() => A < (S & Async)) < Async =
```
hearnadam, line 717: Can this become `A < (S & Async) < Async`?

fwbrasil: It's possible but the user would need to call `flatten` to use it. The anon function seems more ergonomic

PR #970, opened 2025-01-05
[core] introduce `initWith` convenience methods to initialize and use values

fwbrasil:
In the majority of the cases, `init` is immediately followed by a `map` to use the new value. This PR introduces a new pattern with `use` methods in the companion objects. It seems a good usability improvement and it'll also help with performance since it avoids a second allocation.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Adder.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Adder.scala
@@ -59,7 +59,16 @@ object LongAdder:
       * @return
       *   A new LongAdder
       */
-    def init(using frame: Frame): LongAdder < IO = IO.Unsafe(LongAdder(Unsafe.init()))
+    def init(using frame: Frame): LongAdder < IO = use(identity)
+
+    /** Uses a new LongAdder.
+      * @param f
+      *   The function to apply to the new LongAdder
+      * @return
+      *   The result of applying the function
+      */
+    inline def use[A, S](inline f: LongAdder => A < S)(using inline frame: Frame): A < (IO & S) =
+        IO.Unsafe(f(LongAdder(Unsafe.init())))
 
     /** WARNING: Low-level API meant for integrations, libraries, and performance-sensitive code. See AllowUnsafe for more details. */
     opaque type Unsafe = j.LongAdder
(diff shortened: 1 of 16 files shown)


kyo-core/shared/src/main/scala/kyo/Adder.scala
```diff
@@ -59,7 +59,16 @@ object LongAdder:
...
-    def init(using frame: Frame): LongAdder < IO = IO.Unsafe(LongAdder(Unsafe.init()))
+    def init(using frame: Frame): LongAdder < IO = use(identity)
+
+    /** Uses a new LongAdder.
+      * @param f
+      *   The function to apply to the new LongAdder
+      * @return
+      *   The result of applying the function
+      */
+    inline def use[A, S](inline f: LongAdder => A < S)(using inline frame: Frame): A < (IO & S) =
```
hearnadam, line 70: In ZIO, the convention is `*verb*With` for methods that accept a function. `ZIO.service` vs `ZIO.serviceWith`. It might be good to do something similar here to avoid confusion with users attempting to 'use' this (like `Clock.use`).

hearnadam: To be clear, `use` seems okay in the context of a class method, but less so when you are initializing something.

fwbrasil: Makes sense but would it be `initWith`? it gives me the impression that the function is part of the initialization

hearnadam: I think that's reasonable. In a sense, the `f` is a part of the initialization. If someone wanted to write `_ => x` they could...

fwbrasil: It doesn't sound great to me but `initWith` also helps with auto completion and better indicates that the method initializes a value. If that's a common pattern, maybe even `ArrowEffect.suspendAndMap` could be `ArrowEffect.suspendWith`?

hearnadam: That's shorter which is good, I think that's an improvement.

hearnadam (approved):

Issue #874, opened 2024-11-29
Avoid AtomicReference in IOPromise


fwbrasil:
We introduced an `AtomicReference` allocation in `IOPromise` to maintain the internal state because Scala Native doesn't seem to provide a solution to atomically update a field of a class. See https://github.com/getkyo/kyo/pull/873 for background. I saw some intrinsic implementations that might provide that but they don't seem user-facing. We need to find a solution to avoid the performance overhead of the atomic allocation and pointer chasing.

PR #971, opened 2025-01-05
[core] avoid AtomicReference allocation in IOPromise

fwbrasil:
Fixes https://github.com/getkyo/kyo/issues/874

### DIFF
--- a/kyo-core/native/src/main/scala/kyo/scheduler/IOPromisePlatformSpecific.scala
+++ b/kyo-core/native/src/main/scala/kyo/scheduler/IOPromisePlatformSpecific.scala
@@ -1,0 +1,18 @@
+package kyo.scheduler
+
+import kyo.Present
+import scala.scalanative.libc.stdatomic.AtomicRef
+import scala.scalanative.runtime.Intrinsics
+import scala.scalanative.runtime.fromRawPtr
+
+object IOPromisePlatformSpecific:
+
+    val stateHandle =
+        Present(
+            new IOPromise.StateHandle:
+                def compareAndSet[E, A](promise: IOPromise[E, A], curr: IOPromise.State[E, A], next: IOPromise.State[E, A]): Boolean =
+                    val ref = AtomicRef(fromRawPtr(Intrinsics.classFieldRawPtr(promise, "state")))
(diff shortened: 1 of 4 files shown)


kyo-core/native/src/main/scala/kyo/scheduler/IOPromisePlatformSpecific.scala
```diff
@@ -0,0 +1,18 @@
...
+import scala.scalanative.runtime.Intrinsics
+import scala.scalanative.runtime.fromRawPtr
+
+object IOPromisePlatformSpecific:
+
+    val stateHandle =
+        Present(
+            new IOPromise.StateHandle:
+                def compareAndSet[E, A](promise: IOPromise[E, A], curr: IOPromise.State[E, A], next: IOPromise.State[E, A]): Boolean =
+                    val ref = AtomicRef(fromRawPtr(Intrinsics.classFieldRawPtr(promise, "state")))
```
fwbrasil, line 14: Ideally, the field offset lookup should be separate like in the JVM version but I couldn't find an API for that. It seems ok since `AtomicRef` is an `AnyVal` and won't allocate.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -22,11 +22,18 @@ private[kyo] class IOPromise[+E, +A](init: State[E, A]) extends Safepoint.Interc
     def enter(frame: Frame, value: Any): Boolean = true
 
     private def compareAndSet[E2 >: E, A2 >: A](curr: State[E2, A2], next: State[E2, A2]): Boolean =
-        state.compareAndSet(curr, next)
+        IOPromisePlatformSpecific.stateHandle match
+            case Absent =>
```
hearnadam, line 26: if this method accepted the var (only necessary for JS), could all Platforms have a handle?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
fwbrasil: I tried exposing the `state` field but the compiler complains because of variance. I also tried to use dynamic field access via `js.Dynamic` but I couldn't make it work.

PR #972, opened 2025-01-06
[core] Channel fixes and additional tests

fwbrasil:


kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -303,15 +302,22 @@ object Channel:
 
                 def offerAll(values: Seq[A])(using AllowUnsafe): Result[Closed, Seq[A]] =
```
fwbrasil, line 303: A channel flush was missing here. If there were pending takes, they'd continue pending.

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -298,6 +347,34 @@ class ChannelTest extends Test:
                 s <- c.size
             yield assert(r == Seq(1, 2) && s == 2)
         }
+        "should consider pending puts" in pendingUntilFixed {
```
fwbrasil, line 350: @johnhungerford I noticed this behavior when trying to create concurrency tests with `drainUpTo`. Do you think the method should consider pending puts? It doesn't seem very useful without that.

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -303,15 +302,22 @@ object Channel:
...
+                                    loop(current.tail, true)
+                                case Result.Success(false) =>
+                                    if offered then flush()
+                                    Result.success(current)
+                                case result =>
+                                    if offered then flush()
+                                    result.map(_ => current)
+                        end if
+                    end loop
                     loop(values)
```
hearnadam, line 321: Can we do `Chunk.from` here?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -432,21 +440,32 @@ object Channel:
...
                                     // NB: this is only efficient if chunk is effectively indexed
                                     // (i.e. Chunk.Indexed or Chunk.Drop with Chunk.Indexed underlying)
                                     @tailrec
                                     def loop(i: Int): Unit =
                                         if i >= chunk.length then
+                                            // All items transfered, complete put
                                             promise.completeDiscard(Result.unit)
                                         else
                                             Maybe(takes.poll()) match
-                                                case Present(takePromise) if takePromise.complete(Result.success(chunk(i))) =>
```
fwbrasil, line 447: This was dropping the element in case the take is interrupted

kyo-core/shared/src/main/scala/kyo/Channel.scala
fwbrasil: makes sense, updated

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -301,18 +300,25 @@ object Channel:
...
-                    loop(values)
+                            queue.offer(current.head) match
+                                case Result.Success(true) =>
+                                    loop(current.tail, true)
+                                case Result.Success(false) =>
+                                    if offered then flush()
+                                    Result.success(current)
+                                case result =>
+                                    if offered then flush()
+                                    result.map(_ => current)
```
hearnadam, line 318: What would `result` be in this case? Does the `map` do anything if it's an `Error`?

fwbrasil: It's just a way to avoid an unsafe type cast but it shouldn't introduce overhead since `Result` will short circuit. The compiler can't narrow the types correctly via pattern matching

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
johnhungerford: It's still useful I think, but I don't see why we shouldn't drain the pending puts also. I take it regular `drain` should do so as well?

fwbrasil: I've created https://github.com/getkyo/kyo/issues/978 for this

PR #973, opened 2025-01-06
[kernel][core] IO and Async optimizations

fwbrasil:
I noticed a few regressions with recent changes while inspecting benchmarks

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/IO.scala
+++ b/kyo-core/shared/src/main/scala/kyo/IO.scala
@@ -42,7 +42,7 @@ object IO:
       *   The suspended computation wrapped in an IO effect.
       */
     inline def apply[A, S](inline f: Safepoint ?=> A < S)(using inline frame: Frame): A < (IO & S) =
-        Effect.defer(f)
+        Effect.deferInline(f)
 
     /** Ensures that a finalizer is run after the main computation, regardless of success or failure.
       *
@@ -69,7 +69,7 @@ object IO:
     object Unsafe:
 
         inline def apply[A, S](inline f: AllowUnsafe ?=> A < S)(using inline frame: Frame): A < (IO & S) =
-            Effect.defer {
+            Effect.deferInline {
                 import AllowUnsafe.embrace.danger
                 f
             }
--- a/kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
+++ b/kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
@@ -123,7 +123,7 @@ private[kyo] class IOPromise[+E, +A](init: State[E, A]) extends Safepoint.Interc
                 case l: Linked[E, A] @unchecked =>
                     onCompleteLoop(l.p)
                 case v =>
-                    IOPromise.eval(f, v.asInstanceOf[Result[E, A]])
+                    IOPromise.eval(f(v.asInstanceOf[Result[E, A]]))
         onCompleteLoop(this)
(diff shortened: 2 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/IO.scala
```diff
@@ -42,7 +42,7 @@ object IO:
       *   The suspended computation wrapped in an IO effect.
       */
     inline def apply[A, S](inline f: Safepoint ?=> A < S)(using inline frame: Frame): A < (IO & S) =
-        Effect.defer(f)
+        Effect.deferInline(f)
```
fwbrasil, line 45: This method used to inline the suspension as well when `IO` was still an `ArrowEffect`

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -123,7 +123,7 @@ private[kyo] class IOPromise[+E, +A](init: State[E, A]) extends Safepoint.Interc
                 case l: Linked[E, A] @unchecked =>
                     onCompleteLoop(l.p)
                 case v =>
-                    IOPromise.eval(f, v.asInstanceOf[Result[E, A]])
```
fwbrasil, line 126: The `eval` method was also introduced recently. I'm not sure why since it's inlined but it was introducing a function allocation

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -34,6 +34,8 @@ sealed private[kyo] class IOTask[Ctx, E, A] private (
 
     private inline def erasedAbortTag = Tag[Abort[Any]].asInstanceOf[Tag[Abort[E]]]
 
+    private inline def locally[A](inline f: A): A = f
```
fwbrasil, line 37: `locally` is used here just as a workaround for a scalafmt bug that makes the last line of handling functions be moved to the wrong indentation. It was also allocating a funciton.

kyo-kernel/shared/src/main/scala/kyo/kernel/Effect.scala
```diff
@@ -61,7 +61,11 @@ object Effect:
         end try
     end catching
 
-    private[kyo] def defer[A, S](f: Safepoint ?=> A < S)(using _frame: Frame): A < S =
+    private[kyo] def defer[A, S](f: Safepoint ?=> A < S)(using Frame): A < S =
```
fwbrasil, line 64: It's not possible to inline defer directly because it's used by `Safepoint` and it would increase the size of the kernel's inlined bytecode. I've created a separate method

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -309,8 +309,8 @@ private[kyo] object IOPromise:
         end Empty
     end Pending
 
-    private inline def eval[A, B](inline f: A => Unit, value: A): Unit =
```
hearnadam, line 312: Would it still allocate if `value` is inline?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
fwbrasil: yep. It seems some limitation because of the nested inlinings

PR #975, opened 2025-01-06
[build] use coursier/cache-action

fwbrasil:
The action caches dependencies across runs: https://github.com/coursier/cache-action

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -57,6 +57,7 @@ jobs:
       - uses: actions/checkout@v3.0.2
         with:
           fetch-depth: 0
+      - uses: coursier/cache-action@v6
       - uses: olafurpg/setup-scala@v13
         with:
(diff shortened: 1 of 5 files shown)


hearnadam (approved):

.github/workflows/bench.yml
```diff
@@ -57,6 +57,7 @@ jobs:
       - uses: actions/checkout@v3.0.2
         with:
           fetch-depth: 0
+      - uses: coursier/cache-action@v6
```
sideeffffect, line 60: You'll probably be better off with https://github.com/actions/setup-java?tab=readme-ov-file#caching-sbt-dependencies

I's literally just
```yaml
- uses: actions/setup-java@v4
  with:
    distribution: 'temurin'
    java-version: '21'
    cache: 'sbt'
```

Instead of needing coursier/cache-action and olafurpg/setup-scala

PR #976, opened 2025-01-06
[scheduler] don't preempt task if worker queue is empty

fwbrasil:
There's no need to preempt the current task if the worker queue is empty and there isn't another task waiting for execution.


kyo-scheduler/jvm-native/src/test/scala/kyo/scheduler/SchedulerTest.scala
```diff
@@ -80,47 +80,6 @@ class SchedulerTest extends AnyFreeSpec with NonImplicitAssertions {
         }
     }
 
-    "cycle" - {
```
fwbrasil, line 83: It's not easy to test this at the scheduler level because, even if we submit more tasks, the concurrency regulator can introduce new workers. The behavior is covered in `WorkerTest` so I'm removing the tests here.

hearnadam (approved):

PR #977, opened 2025-01-06
Remove ack from Emit and Poll

### LINKED ISSUE #966: Remove Ack from Emit/Stream
Currently, the output type of the `Emit` (and therefore `Stream`) effect is `Ack`, which specifies whether the emissions/streaming (henceforth I'll refer only to streams, but this should be understood to include emit) should continue or not. However, `Ack`'s effectiveness in ending a stream is by convention only. An `Ack.Stop` can be ignored:

```
val s1 = Emit(Chunk(1)).map(_ => Ack.Stop)
val s2 = s1.map(_ => Emit(Chunk(2))
Stream(s2).foreach(Console.printLine(_))
// Output:
// 1
// 2
```

Moreover, if users do respect the convention, it leads to strange results. For instance, if I make a stream that yields `Ack.Stop` and concatenate it with another stream, it will stop after the first stream and never run the second. It seems strange to give each stream this kind of control.

Finally, there is already a natural way for streams to terminate, which is simply for the effect to complete. This seems like the most intuitive way of determining stream termination. I propose getting rid of `Ack` and changing the output type of the `Emit` effect to `Unit`. This way users will not have to think about handling `Ack` when handling `Emit` effect and stream termination will be governed only by the lifetime of the streaming effect.

johnhungerford:
Closes #966


kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -135,12 +127,12 @@ sealed abstract class Stream[V, -S]:
...
                 [C] =>
                     (input, state, cont) =>
-                        if state == 0 then
-                            (0, cont(Stop))
-                        else
-                            val c   = input.take(state)
-                            val nst = state - c.size
-                            Emit.andMap(c)(ack => (nst, cont(ack.maxValues(nst))))
+                        debug.Debug((state, input)).andThen:
+                            if state == 0 then (0, Kyo.pure[Unit, Emit[Chunk[V]] & S](()))
```
johnhungerford: @fwbrasil does this make sense? When the state reaches zero, I don't call `cont` at all but just return unit. This seems right to me and works, but I haven't found examples elsewhere of short-circuiting this way. In fact, I can't figure out how the old version of `Emit` accomplished early termination via `Stop` (since all the handlers, as far as I can tell, always pass `Stop` to `cont`, which would presumably keep running the effect, even if it no longer emitted anything.

fwbrasil: Yes, the previous implementation would let the computation finish and the emitter was responsible for stopping. Short-circuiting by discarding the continuation is ok. It think it only happens in `Abort` and `Choice` for now (and derived effects like `Parse`). There's a mechanism that attaches finalizers to the fiber so they're cleared up even if the continuation isn't evaluated.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -85,14 +79,12 @@ sealed abstract class Stream[V, -S]:
         tagV2: Tag[Emit[Chunk[V2]]],
         frame: Frame
     ): Stream[V2, S & S2 & S3] =
-        Stream[V2, S & S2 & S3](ArrowEffect.handleState(tagV, (), emit)(
+        Stream[V2, S & S2 & S3](ArrowEffect.handle(tagV, emit)(
```
johnhungerford, line 82: @fwbrasil I'm noticing that replacing `handleState` with `handle` is changing the behavior unexpectedly. By using `handleState` in this (no-op) way, the handler will stop at the current suspension and execute anything prior to arriving at the next one. Why is this? (I've added `handleState` back in for this reason; I'm just confused about why it works this way.)

johnhungerford:
Might need some help with `DebugTest - trace - with stream`. I'm not quite sure how the fragments were chosen to assert on (e.g., `2147483647`).

This is the only test that's not passing.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
fwbrasil: Yeah, it a surprising behavior. I think the difference is the ordering of effects because `handleState` allows the handler to perform additional effects before returning the new tuple. In `handle`, the function returns a flat computation with all effects sequenced:

```scala
inline handle: [C] => (I[C], Safepoint ?=> O[C] => A < (E & S & S2)) => A < (E & S & S2)
```

but `handleState` allows an extra `S3`:

```scala
inline handle: [C] => (I[C], State, Safepoint ?=> O[C] => A < (E & S & S2)) => (State, A < (E & S & S2)) < S3,
```

Effects in `S3` will be handled before the computation returns the tuple with the new state the the resumption computation. All other handling methods follow the approach of `handle` and `handleState` is the exception.

This what enables stream handlers to emit values even of the same type before resuming the effect handling. It'd say both behaviors are valid but the naming isn't great since we might need the behavior of `handleState` without an actual state value.

kyo-prelude/shared/src/test/scala/kyo/debug/DebugTest.scala
```diff
@@ -157,10 +157,10 @@ class DebugTest extends Test:
         "with Stream" in
             testOutput(
                 "DebugTest.scala:53:36",
-                "2147483647",
```
johnhungerford, line 160: @fwbrasil I'm not sure where this number came from (I assume from `Ack.Continue(n)`?), but do these changes make sense?

fwbrasil: It’s probably Int.MaxValue for ack without a chunk limit. It’s fine to update the assertion

kyo-prelude/shared/src/main/scala/kyo/Poll.scala
```diff
@@ -249,17 +213,17 @@ object Poll:
                         // This creates the demand-driven cycle between emit and poll
                         ArrowEffect.handleFirst(pollTag, poll)(
                             handle = [C2] =>
-                                (ack, pollCont) =>
+                                (unit, pollCont) =>
```
fwbrasil, line 216: I think using `_` here and an explicit `()` bellow would be more clear. I wasn't sure if this was the regular unit value or something else initially

kyo-prelude/shared/src/main/scala/kyo/Emit.scala
```diff
@@ -96,29 +96,47 @@ object Emit:
...
+        /** Runs an Emit effect, allowing custom handling of each emitted value with a boolean result determining whether to continue.
+          *
+          * @param v
+          *   The computation with Emit effect
+          * @param f
+          *   A function to process each emitted value
+          * @return
+          *   The result of the computation
+          */
+        def apply[A: Flat, S, S2](v: A < (Emit[V] & S))(f: V => Boolean < S2)(using tag: Tag[Emit[V]], frame: Frame): A < (S & S2) =
```
fwbrasil, line 133: I've been placing the computation as the last parameter group to make it work well with `pipe`: `computation.pipe(Emit.runForeach(v => ...))`. I'm not very sure about this design decision but it's the common pattern across the code base.

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
```diff
@@ -373,27 +373,27 @@ object StreamCompression:
...
                             maybeEmitFn match
                                 case Present(emitFn) =>
                                     Resource
                                         .acquireRelease(IO(new Inflater(noWrap)))(inflater => IO(inflater.end()))
                                         .map: inflater =>
-                                            Loop.continue(InflateState.InflateInput(inflater, emitFn(ack)))
+                                            Loop.continue(InflateState.InflateInput(inflater, emitFn()))
                                 case Absent =>
-                                    Loop.done(Ack.Stop)
+                                    Loop.done(())
```
fwbrasil, line 384:
```suggestion
                                    Loop.done
```

fwbrasil, line 384: It's possible to use `Loop.done` directly for `Unit` completion

kyo-prelude/shared/src/main/scala/kyo/Ack.scala
```diff
@@ -1,66 +0,0 @@
-package kyo
```
fwbrasil, line 1: This is a nice cleanup!!

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -168,18 +159,18 @@ sealed abstract class Stream[V, -S]:
       *   A new stream containing at most n elements from the original stream
       */
     def take(n: Int)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Stream[V, S] =
-        if n <= 0 then discard
+        val kyoUnit = Kyo.pure[Unit, S & Emit[Chunk[V]]](())
```
fwbrasil, line 162: can this be moved to the `Emit.valueWith` function or is there an inference issue?

kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
```diff
@@ -45,37 +45,14 @@ abstract private class PublisherToSubscriberTest extends Test:
         end for
     }
 
-    "should cancel upstream if downstream completes" in runJVM {
```
fwbrasil, line 48: @HollandDM could you have a look at this changes?

fwbrasil:
This is looking great! It's interesting how the initial design ended up influencing the following iterations so much to the point that something that looks like a major building block can be removed

kyo-prelude/shared/src/main/scala/kyo/Emit.scala
johnhungerford: Isn't that how it is here?

fwbrasil: no, `v` would come after `f`

kyo-core/jvm/src/main/scala/kyo/StreamCompression.scala
johnhungerford: Good to know -- I had done that in over 30 places!

kyo-prelude/shared/src/main/scala/kyo/Emit.scala
johnhungerford: Done -- but this is not great for type inference in other use cases (see changes to test)

kyo-prelude/shared/src/test/scala/kyo/EmitTest.scala
```diff
@@ -69,162 +67,114 @@ class EmitTest extends Test:
...
-            }
+    "runForeach" - {
+        "with pure function" in {
+            var seen = List.empty[Int]
+            def emits(i: Int): Unit < Emit[Int] =
+                if i == 5 then ()
+                else Emit.valueWith(i)(emits(i + 1))
+            end emits
+
+            Emit.runForeach { (v: Int) =>
```
johnhungerford, line 78: @fwbrasil  Now that the order is switched you have to explicitly annotate the parameter type for it to compile.

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/StreamSubscriber.scala
```diff
@@ -243,21 +243,18 @@ final private[kyo] class StreamSubscriber[V](
         IO(handleInterupt())
     end interupt
 
-    private[interop] def emit(using Frame, Tag[Emit[Chunk[V]]]): Ack < (Emit[Chunk[V]] & Async) =
-        Emit.valueWith(Chunk.empty) { ack =>
-            Loop(ack) {
-                case Ack.Stop => interupt.andThen(Loop.done(Ack.Stop))
```
HollandDM, line 249: @fwbrasil @johnhungerford because now we don't use Ack anymore, the `StreamSubscriber` does not know when the downstream is completed, therefore it cannot call `interrupt` method
This is quite important because the `subscription` will not be canceled, meaning that Publisher cannot do it's cleanup when it's should be (because downstream is closed, we don't need Publisher anymore).
Is there a way for us to retain the stop signal here?

HollandDM: At the first glance, maybe if we can change this into `Poll` & `Emit` combo, maybe it can be done

johnhungerford: Before this rewrite, streams did not reliably communicate their state via `Ack`, so there was always the danger of a subscriber not interrupting the publisher. Now the stream's "ending" is determined only by the completion of the effect. However, since consumer stream may not (and probably will not) be run to completion because it will be transformed with `take`, the *original* stream generated by `StreamSubscriber` cannot communicate this.

In fact, the only thing that can *guarantee* that the stream is cleaned up is registering a finalizer at the beginning of the streaming effect. `StreamSubscriber#stream` should return a `Stream[V, Async & Resource]`. This will have to be handled when the stream is evaluated.

kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
johnhungerford: Actually I'm not sure that the upstream even should be canceled if the downstream is. Or at any rate, we should be clear what it means for the upstream to be canceled. As I see it, there are two options consistent with the reactive-streams specification:
1. The upstream is run each time a subscriber subscribes. This will produce different behaviors depending on on the underlying `emit`. For instance, a pure emit will produce the same values for every subscription.
2. The upstream is run once per publisher, and a subscription will receive the current values. In this arrangement the subscriptions can be multicast (we can use `Hub`/`Listener` for that) or unicast (single `Channel`).

For scenario (1), the upstream should stop running when the subscription is canceled. For scenario (2), the canceled subscription should not affect the upstream. Rather it should close the `Listener` (for multicast) or simply stop consuming from the shared `Channel` (for unicast).

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/StreamSubscriber.scala
johnhungerford: Alternatively, we could simply separate streaming from subscription interruption altogether. Why should the same subscription not stream multiple times? Who says the subscription has to end when the stream does? Auto-closing is a pattern that would be desired if we wanted to abstract streams from the `Subscriber` entirely and simply stream directly from a publisher. `publisher.stream` would have to ensure that the underlying subscription be canceled because the subscription is never exposed.

fwbrasil: I agree that `Resource` seems a better way to handle this. As mentioned, `Ack` isn't very reliable since it functions by convention. @HollandDM could we address this separate? I can create an issue for it.

HollandDM: Hm, ok, please create an issue, I'll have a follow up PR

johnhungerford: Done: #1022

fwbrasil (approved):

kyo-prelude/shared/src/main/scala/kyo/Emit.scala
```diff
@@ -96,29 +96,47 @@ object Emit:
           */
         def apply[A: Flat, S](v: A < (Emit[V] & S))(using tag: Tag[Emit[V]], frame: Frame): A < S =
             ArrowEffect.handle(tag, v)(
-                handle = [C] => (input, cont) => cont(Ack.Stop)
+                handle = [C] => (input, cont) => cont(())
```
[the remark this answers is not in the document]
johnhungerford: In this case there is no change in behavior. This code is for `runDiscard` which simply ignores emits (and did so before as well). The `Ack` is irrelevant because it is meant to run to completion regardless. (If it respected `Ack`, it would not not be able to return `A`, because it might not ever get to the end!)

PR #980, opened 2025-01-06
Channel: drain/drainUpTo include pending puts

### LINKED ISSUE #978: Channel drain and drainUpTo should consider pending puts
More context: https://github.com/getkyo/kyo/pull/972/files#r1903383610

johnhungerford:
closes #978


kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -357,7 +382,25 @@ object Channel:
...
 
                 def drain()(using AllowUnsafe) =
-                    val result = queue.drain()
+                    val result = queue.drain().map { chunk =>
+                        val builder = Chunk.newBuilder[A]
+                        @tailrec def loop(): Unit =
+                            puts.poll() match
+                                case Put.Value(v, promise) =>
+                                    builder.addOne(v)
+                                    promise.completeDiscard(Result.Success(()))
```
hearnadam, line 391:
```suggestion
                                    promise.completeDiscard(Result.unit)
```

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -357,7 +382,25 @@ object Channel:
...
+                                    loop()
+                                case Put.Batch(c, promise) =>
+                                    builder.addAll(c)
+                                    promise.completeDiscard(Result.Success(()))
+                                    loop()
+                                case _ => ()
```
hearnadam, line 397: Shouldn't this be unreachable? or is this a `null` case?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -328,7 +328,32 @@ object Channel:
...
+                                if i == 0 then ()
+                                else
+                                    puts.poll() match
+                                        case Put.Value(v, promise) =>
+                                            builder.addOne(v)
+                                            promise.completeDiscard(Result.Success(()))
+                                            loop(i - 1)
+                                        case Put.Batch(c, promise) =>
+                                            builder.addAll(c)
+                                            promise.completeDiscard(Result.Success(()))
```
hearnadam, line 346:
```suggestion
                                            promise.completeDiscard(Result.unit)
```

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -328,7 +328,32 @@ object Channel:
...
+                        if cl >= max then chunk
+                        else
+                            val builder = Chunk.newBuilder[A]
+                            @tailrec def loop(i: Int): Unit =
+                                if i == 0 then ()
+                                else
+                                    puts.poll() match
+                                        case Put.Value(v, promise) =>
+                                            builder.addOne(v)
+                                            promise.completeDiscard(Result.Success(()))
```
hearnadam, line 342:
```suggestion
                                            promise.completeDiscard(Result.unit)
```

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: null case. Just pushed a change to make it explicit.

fwbrasil:
Maintaining ordering would be ideal. What if we had a cycle of queue.drainUpTo/flush/queue.drainUpTo/flush/.. until drainUpTo returns empty?

johnhungerford:
Good idea, that works nicely.

fwbrasil:
More relaxed ordering when becomes queue is full and there are multiple concurrent puts seems more reasonable given that concurrent puts can arrive in any order. To ensure ordering even in this scenario, I think we'd need to have a single atomic operation instead of having to deal with both the main queue and the pending takes/puts queues, which I think wouldn't allow some of the fast paths we currently have in the implementation like `put` succeeding immediately if there's space in the queue. I'd say it's ok. The flushing logic can also change ordering under high contention (but it's very rare)

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -347,34 +360,34 @@ class ChannelTest extends Test:
...
-                    _         <- c.putFiber(2)
-                    _         <- c.putFiber(3)
-                    result    <- c.drainUpTo(3)
-                    finalSize <- c.size
-                yield assert(result == Chunk(1, 2, 3) && finalSize == 0)
+                yield assert(result == Chunk(1, 2, 3) && finalSize == 1)
             }
-            ()
         }
+        // "should consider pending puts - zero capacity" in run {
```
fwbrasil, line 377: can you keep this test pending and open an issue to handle channels with zero capacity separately? It'd be possible to provide better performance with a dedicated impl as well

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -328,9 +330,22 @@ object Channel:
...
 
                 def drainUpTo(max: Int)(using AllowUnsafe) =
-                    val result = queue.drainUpTo(max)
-                    if result.exists(_.nonEmpty) then flush()
-                    result
+                    def loop(current: Chunk[A], i: Int): Result[Closed, Chunk[A]] =
+                        if i == 0 then Result.Success(current)
+                        else
+                            val next = queue.drainUpTo(i)
+                            next match
```
fwbrasil, line 337: `next.flatMap` should work with `@tailrec` as well

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -357,9 +372,19 @@ object Channel:
                 end takeFiber
 
                 def drain()(using AllowUnsafe) =
-                    val result = queue.drain()
-                    if result.exists(_.nonEmpty) then flush()
-                    result
+                    def loop(current: Chunk[A]): Result[Closed, Chunk[A]] =
```
fwbrasil, line 375: `@tailrec`

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: When I try this:
```scala
@tailrec
def loop(current: Chunk[A], i: Int): Result[Closed, Chunk[A]] =
    if i == 0 then Result.Success(current)
    else
        queue.drainUpTo(i).flatMap { chunk =>
            if chunk.isEmpty then Result.Success(current)
            else
                flush()
                loop(current.concat(chunk), i - chunk.length)
        }
    end if
end loop
```

I get a tailrec error:

```
Cannot rewrite recursive call: it is not in tail position
```

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -105,16 +105,18 @@ object Channel:
         def takeExactly(n: Int)(using Frame): Chunk[A] < (Abort[Closed] & Async) =
             if n <= 0 then Chunk.empty
             else
-                Loop(Chunk.empty[A]): lastChunk =>
-                    val nextN = n - lastChunk.size
+                Loop(Chunk.empty[A], 0): (lastChunk, lastSize) =>
+                    val nextN = n - lastSize
                     Channel.drainUpTo(self)(nextN).map: chunk =>
                         val chunk1 = lastChunk.concat(chunk)
+                        val size1  = chunk1.size
```
fwbrasil, line 112: `Chunk.size` returns in constant time so it's best to get it on each iteration instead of capturing `size1`, which requires an additional field in the generated function

kyo-core/shared/src/main/scala/kyo/Channel.scala
fwbrasil: Thanks for checking! I saw it working in some code using `Maybe` but it seems there's something different here

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -670,6 +698,7 @@ class ChannelTest extends Test:
         }
 
         "putBatch and takeExactly" in run {
+            val repeats = 1
```
fwbrasil, line 701: Is this intentional?

johnhungerford: My bad -- leftover from when I was trying to narrow down an issue.

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: Yeah looks like `Maybe#flatMap` can call the mapping function in tail position since it just has to check whether the current `Maybe` is absent or present before running it, whereas `Result` has to run the mapping function in a try block before returning the result.

fwbrasil (approved):

PR #982, opened 2025-01-07
[scheduler] tuning + cats improvement + scaladocs

fwbrasil:
I did some more testing using Kyo's scheduler with cats-effect and identified tunings to offset the regression in some benchmarks. [Results](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/fwbrasil/396006030e3dae7e157982d04c1f7235/raw/921394ba87a1ffb8f9c09c5a17488c02e23597cf/jmh-result-baseline.json,https://gist.githubusercontent.com/fwbrasil/396006030e3dae7e157982d04c1f7235/raw/921394ba87a1ffb8f9c09c5a17488c02e23597cf/jmh-result-candidate.json) with/without the tuning using Kyo's scheduler for both ZIO and cats-effect:

image

PR #983, opened 2025-01-08
[scheduler] improve XSRandom distribution

fwbrasil:
I noticed this opportunity to improve the distribution of `XSRandom` while working on https://github.com/getkyo/kyo/pull/982. By mixing in the thread ID during the xorshift operations, we can prevent threads that land on the same seed slot from generating identical sequences, which could skew scheduling decisions.


kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -780,13 +780,15 @@ class ChannelTest extends Test:
         }
 
         "should stream concurrently with ingest via putBatch, yielding consistent chunk sizes" in run {
+            pending
```
fwbrasil, line 783: @johnhungerford This test is flaky. I was able to reproduce the failure locally with and without this PR but it's rare. It seems the scenario you were worried about. Since `channel.stream` takes a `maxChunkSize`, is it reasonable that chunks of smaller size may arrive? It's difficult to debug but it seems `putBatch` is still adding items to the queue while `drainUpTo` is executing and it finalizes earlier before `putBatch` is able to place all items in the queue. It seems a reasonable behavior to me, maybe we can remove the test

johnhungerford: This should probably just be like the other concurrent streaming test, which only asserts that every batch size is `<=` to `maxChunkSize` and that the flattened result contains all expected values ordered correctly. As you say it's reasonable that chunk sizes are not guaranteed, but we should confirm everything else works when concurrently streaming in and out.

kyo-scheduler/jvm-native/src/main/scala/kyo/scheduler/util/XSRandom.scala
```diff
@@ -5,21 +5,33 @@ import java.util.Random
...
-  * do not affect correctness of the scheduling algorithms.
+  * The implementation includes the thread index in the xorshift operations to better handle thread slot conflicts. While multiple threads
+  * may still share the same seed slot, their random sequences will differ due to the thread-specific mixing in the xorshift steps. This
+  * provides better statistical properties while maintaining the performance characteristics needed for scheduling operations.
   */
 private[kyo] object XSRandom extends Random {
-    private val seeds = List.fill(32)(31L).toArray
+
+    // Size is core * 32 to reduce false sharing by spacing seeds across cache lines
+    private val seeds = List.fill(Runtime.getRuntime().availableProcessors() * 32)(31L).toArray
```
hearnadam, line 17: Why not use `Array.fill`?

hearnadam (approved):

PR #984, opened 2025-01-08
adopt `*With` naming strategy + avoid some `apply` methods

fwbrasil:
Following up on https://github.com/getkyo/kyo/pull/970#discussion_r1903664424. I've also removed some `apply` methods where I think having an explicit method name can help with readability. For example, `Emit.value(1)` seems to read more naturally than `Emit(1)`


fwbrasil:
merging to avoid conflicts given that this PR touches many files

PR #985, opened 2025-01-08
[direct] improve error messages + test coverage

fwbrasil:


kyo-direct/shared/src/main/scala/kyo/Validate.scala
```diff
@@ -47,97 +82,52 @@ private[kyo] object Validate:
                        |""".stripMargin
                 )
 
-            case tree @ Apply(TypeApply(Ident("defer"), _), _) =>
```
fwbrasil, line 50: This case was unnecessary. The error message for a missing `now` triggers first

kyo-direct/shared/src/main/scala/kyo/Validate.scala
```diff
@@ -47,97 +82,52 @@ private[kyo] object Validate:
...
                     s"""${"`var`".yellow} declarations are not allowed inside a ${"`defer`".yellow} block.
                        |
                        |Mutable state can lead to unexpected behavior with effects. Instead, use proper state management tools:
                        |
-                       |• Var (kyo-prelude)
-                       |• Atomic* classes (kyo-core)
-                       |• TRef and derivatives (kyo-stm)""".stripMargin
-                )
-
-            case Return(_, _) =>
```
fwbrasil, line 86: Also unnecessary, it's not possible to have a `return` outside of `def`

kyo-direct/shared/src/main/scala/kyo/Validate.scala
```diff
@@ -47,97 +82,52 @@ private[kyo] object Validate:
...
-                       |}
-                       |
-                       |// Use regular val:
-                       |defer {
-                       |  val x = IO(1).now       // OK - eager
-                       |  x + 1
-                       |}""".stripMargin)}""".stripMargin
-                )
-
-            case Lambda(_, body) if !pure(body) =>
```
fwbrasil, line 127: This pattern match is wrong and I couldn't make it work. I've decided to remove because `dotty-cps-async` already produces an error message for this.

kyo-direct/shared/src/test/scala/kyo/CoreTest.scala
```diff
@@ -0,0 +1,195 @@
+package kyo
```
fwbrasil, line 1: I've removed `DirectTest` and replaced it with more complete `PreludeTest` and `CoreTest` suites

kyo-direct/shared/src/test/scala/kyo/CoreTest.scala
```diff
@@ -0,0 +1,195 @@
...
+        "deadline" in run {
+            defer {
+                val deadline = Clock.deadline(1.second).now
+                assert(!deadline.isOverdue.now)
+                assert(deadline.timeLeft.now <= 1.second)
+            }
+        }
+    }
+
+    // TODO Compiler crash because `Queue` is an opaque type without a type bound
```
fwbrasil, line 44: I can't use `pendingUntilFixed` because the compiler crashes. I'll create an issue for this

kyo-direct/shared/src/test/scala/kyo/CoreTest.scala
```diff
@@ -0,0 +1,195 @@
...
+                    defer {
+                        assert(mutex.availablePermits.now == 0)
+                    }
+                }.now
+                assert(mutex.availablePermits.now == 1)
+            }
+        }
+    }
+
+    // TODO Compiler crash because `Queue` is an opaque type without a type bound
```
fwbrasil, line 139: Same issue

hearnadam (approved):

PR #990, opened 2025-01-09
[kernel] TracePool flaky test

fwbrasil:
I noticed a [flaky test in CI](https://github.com/getkyo/kyo/actions/runs/12665683690/job/35295758155?pr=984#step:6:5797) and decided to investigate. `TracePool.Local.borrow` returned a `null`, which shouldn't be possible. I wasn't able to reproduce it locally but I've introduced new tests and implemented a preventive measure.

I think there's a possibility that `MpmcArrayQueue.drain` may pass a `null` to the consumer in rare race conditions. I've added a check to ignore values if that's the case. I believe we don't use this method in other places.

### DIFF
--- a/kyo-kernel/jvm/src/main/scala/kyo/kernel/internal/TracePool.scala
+++ b/kyo-kernel/jvm/src/main/scala/kyo/kernel/internal/TracePool.scala
@@ -6,6 +6,23 @@ import kyo.kernel.internal.*
 import org.jctools.queues.MessagePassingQueue.Consumer
 import org.jctools.queues.MpmcArrayQueue
 
+/** A concurrent two-level object pooling system for managing Trace objects used in execution tracking.
+  *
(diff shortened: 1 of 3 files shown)


kyo-kernel/jvm/src/main/scala/kyo/kernel/internal/TracePool.scala
```diff
@@ -17,28 +18,33 @@ private[kernel] object TracePool:
 
         final def borrow(): Trace =
             if size == 0 then
-                if global.drain(add, localCapacity - size) == 0 then
-                    return Trace.init
-            size -= 1
-            val buffer = pool(size)
-            pool(size) = null
-            buffer
+                discard(global.drain(add, localCapacity))
```
fwbrasil, line 21: I've also made minor optimizations to avoid reading the `size` field multiple times

PR #991, opened 2025-01-09
[build] rename kyo-test to kyo-zio-test + sbt build cleanups

fwbrasil:
I'd like to reserve `kyo-test` for an eventual kyo-native test module. We should be able to provide a better experience with an approach more integrated with Kyo's effects. In addition, I've made a cleanup of the build removing unnecessary dependencies.


hearnadam (commented):
I was thinking this as well. Would `kyo-test-zio` be better? If we end up sharing anything with `kyo-test` it would make things clearer?

fwbrasil:
I think `kyo-test-*` could be reserved for extensions of the module like in `kyo-scheduler`. `kyo-zio-test` also seems to better indicate that it's an integration

build.sbt
```diff
@@ -427,9 +423,7 @@ lazy val `kyo-zio` =
         .dependsOn(`kyo-core`)
         .settings(
             `kyo-settings`,
-            libraryDependencies += "dev.zio" %%% "zio"          % zioVersion,
-            libraryDependencies += "dev.zio" %%% "zio-test"     % zioVersion,
-            libraryDependencies += "dev.zio" %%% "zio-test-sbt" % zioVersion % Test
```
hearnadam, line 432: I think this needs `zio-test`?

fwbrasil: The module uses regular scalatest only

fwbrasil:
I had renamed `kyo-zio` to `kyo-zio-test` 🙈 Fixed

hearnadam (approved):

PR #992, opened 2025-01-09
[scheduler] ensure the scheduler is a singleton across classloaders

fwbrasil:
I had issues working on similar adaptive schedulers in the past because the scheduler was loaded multiple times due to class loader isolation. The interference between the multiple schedulers can defeat the adaptive behavior. This PR introduces a work around to ensure the scheduler is a singleton even if multiple class loaders are used.


He-Pin:
Cool

hearnadam (approved):

PR #994, opened 2025-01-10
[scheduler] finagle integration

fwbrasil:
Uses extension points I introduced when working on a fiber-based scheduler at Twitter. The best part is that the admission control is automatically wired in the server stack to shed load.


.github/workflows/build.yml
```diff
@@ -30,6 +30,10 @@ jobs:
       run: |
         sbt '+kyoJVM/test'
 
+    - name: Build kyo-scheduler-finagle
```
fwbrasil, line 33: Finagle doesn't publish Scala 3 artifacts so the module is set only to Scala 2.13. I spent a few hours trying to make the SBT work but it always compiles the module even if on Scala 3. As a workaround, the new module isn't aggregated in `kyoJVM` and the workflows build the module individually.

kyo-scheduler-finagle/jvm/src/main/resources/META-INF/services/com.twitter.finagle.exp.FinagleSchedulerService
```diff
@@ -0,0 +1 @@
+kyo.scheduler.KyoFinagleSchedulerService
```
fwbrasil, line 1: Finagle automatically finds the integration via service loading.

kyo-scheduler-finagle/jvm/src/main/scala/kyo/scheduler/KyoFinagleSchedulerService.scala
```diff
@@ -0,0 +1,119 @@
...
+                    }
+
+                    def tryFork[T](f: => Future[T]): Future[Option[T]] =
+                        if (scheduler.reject())
+                            Future.None
+                        else
+                            fork(f.map(Some(_)))
+
+                    def fork[T](executor: Executor)(f: => Future[T]): Future[T] =
+                        fork(f)
```
hearnadam, line 98: This avoids shifting work to a passed executor?

kyo-scheduler-finagle/jvm/src/main/scala/kyo/scheduler/KyoFinagleSchedulerService.scala
```diff
@@ -0,0 +1,119 @@
...
+      * @return
+      *   Some(Scheduler) containing the configured scheduler instance
+      */
+    def create(params: List[String]): Option[FinagleScheduler with ForkingScheduler] = {
+        if (params != List("kyo"))
+            None
+        else {
+            val scheduler =
+                new FinagleScheduler with ForkingScheduler {
+                    val original        = FinagleScheduler()
```
hearnadam, line 51: Without checking the source, I assume this is Finagle's equivalent of `Scheduler.get`? Is there a reason to flush that scheduler?

kyo-scheduler/jvm-native/src/main/scala/kyo/scheduler/Scheduler.scala
```diff
@@ -223,6 +227,30 @@ final class Scheduler(
...
+      * @return
+      *   An ExecutorService that submits Runnables as scheduler tasks
+      */
+    def asExecutorService: ExecutorService =
+        new AbstractExecutorService {
+            def awaitTermination(timeout: Long, unit: TimeUnit): Boolean = false
+            def execute(command: Runnable): Unit                         = schedule(Task(command))
+            def isShutdown(): Boolean                                    = false
+            def isTerminated(): Boolean                                  = false
+            def shutdown(): Unit                                         = {}
```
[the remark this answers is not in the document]
fwbrasil: yeah but in this case it should be ok. Finagle also has a few thread pools and it doesn't shut them down

kyo-scheduler-finagle/jvm/src/main/scala/kyo/scheduler/KyoFinagleSchedulerService.scala
fwbrasil: It redirects all workload from future pools to Kyo's scheduler. The API passes the executor for flexibility since it'd be possible to have custom logic based on it but it doesn't need to be used

fwbrasil: Future pools are essentially the standard way to fork computations in Finagle

kyo-scheduler-finagle/jvm/src/main/scala/kyo/scheduler/KyoFinagleSchedulerService.scala
fwbrasil: Yes, this will return the default [LocalScheduler](https://github.com/twitter/util/blob/81bab63ae10cec593915754d59666f90202dfa0a/util-core/concurrent-extra/src/main/scala/com/twitter/concurrent/Scheduler.scala#L58), which uses a thread-local trampolining mechanism without forking. When performing a blocking operation, the tasks pending in the trampolining queue need to be executed before getting to the task that will block execution otherwise they'd stay hanging until the blocking task finishes.

fwbrasil: Note that `submit` uses the original `LocalScheduler` to handle `Future` execution. Only actual forking via future pools get moved to Kyo's scheduler

hearnadam (approved):

PR #995, opened 2025-01-10
[combinators] fix flaky test

fwbrasil:
I noticed this flaky test in CI: https://github.com/getkyo/kyo/actions/runs/12703763768/job/35412083650?pr=959#step:8:2644

### DIFF
--- a/kyo-combinators/shared/src/test/scala/kyo/ResourceCombinatorTest.scala
+++ b/kyo-combinators/shared/src/test/scala/kyo/ResourceCombinatorTest.scala
@@ -17,14 +17,13 @@ class ResourceCombinatorTest extends Test:
                     _      <- setter(50)
                     result <- IO(state)
                 yield result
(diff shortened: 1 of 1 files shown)


kyo-combinators/shared/src/test/scala/kyo/ResourceCombinatorTest.scala
```diff
@@ -17,14 +17,13 @@ class ResourceCombinatorTest extends Test:
                     _      <- setter(50)
                     result <- IO(state)
                 yield result
-            val beforeResources               = scala.concurrent.Future(assert(state == 0))
```
fwbrasil, line 20: @johnhungerford I'm not sure what the goal was here but, given that `Future` will fork the computation, there isn't a guarantee that it will run before `Resource.run`.

hearnadam (approved):

PR #998, opened 2025-01-10
[core] make IO.ensure private[kyo]

fwbrasil:
I think `IO.ensure` method exposes too much of the current underlying implementation, which might block potential kernel improvements and optimizations in the future. Since `Resource` provides the same functionality in a more high-level API, `IO.ensure` can be kept internal to the library.


hearnadam:
Not sure I agree. ZIO has `.ensure` exposed. Perhaps we can have a better API for public use?

fwbrasil:
I see. Have you used it yet? I thought it'd be something rare in user code

fwbrasil:
Closing this since it seems an important limitation

PR #1002, opened 2025-01-12
[core] Async.never

hearnadam:


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -154,6 +154,8 @@ object Async:
             _mask(isolate.resume(state, v)).map(isolate.restore(_, _))
         }
 
+    def never(using Frame): Unit < Async = Fiber.never.get
```
fwbrasil, line 157: can you add scaladoc?

fwbrasil (approved):

fwbrasil (approved):

PR #1003, opened 2025-01-13
[offheap] introduce module

fwbrasil:
I've put it in a new module because I have plans to introduce new off-heap data structures.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -93,6 +93,7 @@ lazy val kyoJVM = project
         `kyo-kernel`.jvm,
         `kyo-prelude`.jvm,
         `kyo-core`.jvm,
+        `kyo-offheap`.jvm,
         `kyo-direct`.jvm,
         `kyo-stm`.jvm,
         `kyo-stats-registry`.jvm,
@@ -285,6 +286,15 @@ lazy val `kyo-core` =
             `js-settings`,
             libraryDependencies += ("org.scala-js" %%% "scalajs-java-logging" % "1.0.0").cross(CrossVersion.for3Use2_13)
         )
+
+lazy val `kyo-offheap` =
+    crossProject(JVMPlatform)
+        .withoutSuffixFor(JVMPlatform)
+        .crossType(CrossType.Full)
+        .in(file("kyo-offheap"))
+        .dependsOn(`kyo-core`)
+        .settings(`kyo-settings`)
+        .jvmSettings(mimaCheck(false))
 
 lazy val `kyo-direct` =
     crossProject(JSPlatform, JVMPlatform, NativePlatform)
--- a/build.sbt
+++ b/build.sbt
@@ -65,6 +65,15 @@ lazy val `kyo-settings` = Seq(
 )
 
 Global / onLoad := {
+
+    val javaVersion  = System.getProperty("java.version")
+    val majorVersion = javaVersion.split("\\.")(0).toInt
+    if (majorVersion < 21) {
+        throw new IllegalStateException(
+            s"Java version $javaVersion is not supported. Please use Java 21 or higher."
+        )
+    }
+
     val project =
         System.getProperty("platform", "JVM").toUpperCase match {
             case "JVM"    => kyoJVM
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -65,6 +65,15 @@ lazy val `kyo-settings` = Seq(
...
 
 Global / onLoad := {
+
+    val javaVersion  = System.getProperty("java.version")
+    val majorVersion = javaVersion.split("\\.")(0).toInt
+    if (majorVersion < 21) {
+        throw new IllegalStateException(
+            s"Java version $javaVersion is not supported. Please use Java 21 or higher."
+        )
+    }
```
fwbrasil, line 75: Fail-fast to provide a better experience

sideeffffect (commented):
Just a curious question: Is it thinkable to implement this also in Scala Native?

kyo-offheap/jvm/src/main/scala/kyo/Memory.scala
```diff
@@ -0,0 +1,329 @@
...
+  * memory operations while ensuring they're properly contained within Arena contexts. When maximum performance is required, the unsafe API
+  * allows dropping down to direct memory operations while keeping such operations explicitly marked.
+  *
+  * Note that memory allocation operations (init, initWith) are intentionally not provided in the Unsafe API. This ensures that all memory
+  * allocations are tracked by the Arena effect, which is necessary for automatic resource cleanup.
+  *
+  * @tparam A
+  *   The type of elements stored in this memory segment
+  */
+opaque type Memory[A] = MemorySegment
```
hearnadam, line 35: Is there any concern with using opaque types over the java type if we want to port this to Scala native?

kyo-offheap/jvm/src/main/scala/kyo/Memory.scala
```diff
@@ -0,0 +1,329 @@
...
+                MemorySegment.copy(self, srcPos * l.size, target, targetPos * l.size, len * l.size)
+
+            def size: Long = self.byteSize / l.size
+
+            def safe: Memory[A] = self
+        end extension
+    end Unsafe
+
+    /** Defines how values of type A are laid out in memory. */
+    abstract class Layout[A]:
```
hearnadam, line 271: Should this be sealed for now? or do you expect users will create them?

Perhaps it would be safer to expose `def transform[A, B](f: Layout[A] => Layout[B])` somehow.

kyo-offheap/jvm/src/main/scala/kyo/Memory.scala
```diff
@@ -0,0 +1,329 @@
...
+          * @param index
+          *   The position to read from
+          * @return
+          *   The element at the specified index
+          */
+        inline def get(index: Int)(using inline frame: Frame): A < Arena =
+            IO.Unsafe {
+                val x = Unsafe.get(self)(index)
+                x
+            }
```
hearnadam, line 84: Is this leftover/inference issue?
```suggestion
            IO.Unsafe(Unsafe.get(self)(index))
```

hearnadam (commented):
Nice!

kyo-offheap/jvm/src/main/scala/kyo/Memory.scala
fwbrasil: I think it's ok. Ideally, SN should provide stubs for these types but we can also add them to the codebase. Also, given that the underlying type isn't exposed, it seems we could swap the implementation without breaking binary compatibility

hearnadam: Technically opaque types changing will not be binary compatible (they won't exist in the bytecode), but they will be source compatible.

kyo-offheap/jvm/src/main/scala/kyo/Memory.scala
fwbrasil: The goal was to allow users to define new layouts since the default ones are limited to simple primitive types. It's possible to compose layouts to represent a complex object for example. I'm planning to explore ways to provide automatic derivation of layouts as well but I think user-defined layouts will still be an important feature given these APIs are designed for performance-sensitive code.

kyo-offheap/jvm/src/main/scala/kyo/Memory.scala
fwbrasil: ah, I forgot to mention this. Yes, I get the missing `AllowUnsafe` compilation failure without the `val`. I've added a comment about it

hearnadam (approved):

PR #1004, opened 2025-01-13
[build] fix kyo-scheduler-finagle release

fwbrasil:
I tried for a couple of hours to avoid `kyoJVM` aggregating `kyo-scheduler-finagle` in Scala 3 but didn't have success. As a work around, I've kept the cross to Scala 3 but disabled all source code + dependency + release in the module's config.


fwbrasil:
merging to check if it works on main

PR #1012, opened 2025-01-14
[stm] fix TTable index initialization

fwbrasil:
The wrong implicit was being used to initialize the indexes, which caused all fields to be indexed.

### DIFF
--- a/kyo-stm/shared/src/main/scala/kyo/TTable.scala
+++ b/kyo-stm/shared/src/main/scala/kyo/TTable.scala
@@ -204,6 +204,8 @@ object TTable:
         def isEmpty(using Frame)  = store.isEmpty
         def snapshot(using Frame) = store.snapshot
 
+        def indexFields: Set[Field[?, ?]] = indexes.keySet
+
         private inline val indexMismatch = """
             Cannot query on fields that are not indexed.
             The filter contains fields that are not part of the table's index configuration.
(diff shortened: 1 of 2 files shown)


kyo-stm/shared/src/main/scala/kyo/TTable.scala
```diff
@@ -287,11 +289,11 @@ object TTable:
           * @return
           *   A new Indexed table instance within the IO effect
           */
-        def init[Fields: AsFields as fields, Indexes >: Fields: AsFields as indexes](using Frame): Indexed[Fields, Indexes] < IO =
+        def init[Fields: AsFields as fields, Indexes >: Fields: AsFields as indexFields](using Frame): Indexed[Fields, Indexes] < IO =
             for
                 table <- TTable.init[Fields]
                 indexes <-
-                    Kyo.foreach(fields.toSeq) { field =>
+                    Kyo.foreach(indexFields.toSeq) { field =>
```
hearnadam, line 296: We may want `Kyo.foreach` to traverse more than `Seq`...

hearnadam (approved):

kyo-stm/shared/src/main/scala/kyo/TTable.scala
fwbrasil: good point, can you make an issue for it?

Issue #1006, opened 2025-01-13
Record: support more than 22 fields in `AsFields`


fwbrasil:
`Record` uses `Record.AsFields` in some of its methods to reflect on the type-level set of fields. `AsFields` is currently using several implicit evidences, one for each arity, for up to 22 elements. We need to find a way to accept an arbitrary number of elements, preferably using the type system itself but with a macro if needed.

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/1013

PR #1013, opened 2025-01-14
[data] records more than 22 AsFields (#1006)

road21:
I haven't found a way to avoid macro (intersection types don't play well with implicits). So I have created typeclass `FieldsLike` (evidence) that splits `Fields` type into tuple of `~` types, and then derives instance for every tuple member.

This typeclass `FieldsLike` could potentially be used for other features like removing fields by keys.


fwbrasil (approved):

fwbrasil:
thank you @road21!

Issue #1008, opened 2025-01-13
Record: Provide Render evidence


fwbrasil:
This has a similar challenge to https://github.com/getkyo/kyo/issues/1007. We need to find a way to collect the correct `Render` instances for each field in the record.

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/1019/files

PR #1019, opened 2025-01-15
[data] add Render instance for Record (#1008)

road21:
I had to extend `FieldsLike` evidence to derive `Render` instances, and now it also contains tuples (types) of names and values.
Also I have noticed that an actual type of `~` is never used, so I removed it and now `~` is an abstract type member.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Record.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Record.scala
@@ -1,9 +1,11 @@
 package kyo
 
 import Record.*
+import kyo.internal.Inliner
 import kyo.internal.TypeIntersection
 import scala.annotation.implicitNotFound
 import scala.compiletime.constValue
+import scala.compiletime.erasedValue
 import scala.compiletime.summonInline
 import scala.deriving.Mirror
 import scala.language.dynamics
@@ -73,7 +75,7 @@ import scala.util.NotGiven
   *   - Tag derivation for Records is not currently supported
   *   - CanEqual and Render instances are not provided for Records
   */
-class Record[+Fields](val toMap: Map[Field[?, ?], Any]) extends AnyVal with Dynamic:
+final class Record[+Fields] private (val toMap: Map[Field[?, ?], Any]) extends AnyVal with Dynamic:
 
     /** Retrieves a value from the Record by field name.
       *
@@ -115,12 +117,15 @@ class Record[+Fields](val toMap: Map[Field[?, ?], Any]) extends AnyVal with Dyna
     /** Returns the number of fields in this Record.
       */
     def size: Int = toMap.size
-
 end Record
 
 export Record.`~`
 
 object Record:
+    /** Creates an empty Record
+      */
+    val empty: Record[Any] = Record[Any](Map())
+
     given [Fields]: Flat[Record[Fields]] = Flat.unsafe.bypass
 
     final infix class ~[Name <: String, Value] private ()
@@ -257,7 +262,7 @@ object Record:
         Field match
             case name ~ value => CanEqual[value, value]
 
-    transparent inline given [Fields]: CanEqual[Record[Fields], Record[Fields]] =
+    inline given [Fields: TypeIntersection as ts]: CanEqual[Record[Fields], Record[Fields]] =
         discard(TypeIntersection.summonAll[Fields, HasCanEqual])
         CanEqual.derived
     end given
@@ -267,6 +272,23 @@ object Record:
             "Cannot derive Tag for Record type. This commonly occurs when trying to nest Records, " +
                 "which is not currently supported by the Tag implementation."
         )
+
+    private object RenderInliner extends Inliner[(String, Render[?])]:
+        inline def apply[T]: (String, Render[?]) =
+            inline erasedValue[T] match
+                case _: (n ~ v) =>
+                    val ev   = summonInline[n <:< String]
+                    val inst = summonInline[Render[v]]
+                    ev(constValue[n]) -> inst
+    end RenderInliner
+
+    inline given [Fields: TypeIntersection]: Render[Record[Fields]] =
+        val insts = TypeIntersection.inlineAll[Fields, (String, Render[?])](RenderInliner).toMap
+        Render.from: (value: Record[Fields]) =>
+            value.toMap.map((field, value) => (field, value, insts.get(field.name))).collect {
+                case (field, value, Some(r: Render[x])) => field.name + " ~ " + r.asText(value.asInstanceOf[x])
+            }.mkString(" & ")
+    end given
 end Record
 
 object AsFieldsInternal:
--- a/kyo-data/shared/src/main/scala/kyo/Record.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Record.scala
@@ -262,7 +262,7 @@ object Record:
         Field match
             case name ~ value => CanEqual[value, value]
 
-    inline given [Fields: TypeIntersection as ts]: CanEqual[Record[Fields], Record[Fields]] =
+    inline given [Fields: TypeIntersection]: CanEqual[Record[Fields], Record[Fields]] =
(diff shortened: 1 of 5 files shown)


kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -464,10 +465,31 @@ class RecordTest extends Test:
     }
 
     "Render" - {
-        "simple record" in pendingUntilFixed {
+        "simple record" in {
             val record = "name" ~ "John" & "age" ~ 30
-            assert(Render.asText(record).show == """name ~ "John" & age ~ 30""")
-            ()
+            assert(Render.asText(record).show == """name ~ John & age ~ 30""")
```
road21, line 470: In the original test there was `name ~ "John" & age ~ 30`

But now the logic is to use Render instances of fields, so there should be one format for all values. I can do something like `name ~ [John] & age ~ [30]` (apply brackets/quotes/... to each field)

Do we really want `name ~ "John" & age ~ 30` ?

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -74,7 +75,7 @@ import scala.util.NotGiven
   *   - Records with more than 22 fields have limited type-level support, though runtime support exists for larger structures
   *   - CanEqual and Render instances are not provided for Records
   */
-class Record[+Fields](val toMap: Map[Field[?, ?], Any]) extends AnyVal with Dynamic:
+class Record[+Fields] private (val toMap: Map[Field[?, ?], Any]) extends AnyVal with Dynamic:
```
hearnadam, line 78:
```suggestion
final class Record[+Fields] private (val toMap: Map[Field[?, ?], Any]) extends AnyVal with Dynamic:
```

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -116,12 +117,15 @@ class Record[+Fields](val toMap: Map[Field[?, ?], Any]) extends AnyVal with Dyna
...
     def size: Int = toMap.size
-
 end Record
 
 export Record.`~`
 
 object Record:
+    /** Creates an empty Record
+      */
+    def empty: Record[Any] = Record[Any](Map())
```
hearnadam, line 127:
```suggestion
    val empty: Record[Any] = Record[Any](Map.empty)
```

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -254,39 +260,67 @@ object Record:
...
 
         transparent inline given derive[A]: FieldsLike[A] =
             ${ FieldsLikeMacro.fieldsLikeImpl[A] }
     end FieldsLike
+
+    inline given [Fields](using ev: FieldsLike[Fields]): Render[Record[Fields]] =
+        val insts = RecordInternal.summonRender[ev.Names, ev.Values]
+        Render.instance: (value: Record[Fields]) =>
+            value.toMap.map((field, value) =>
+                insts(field.name) match
```
hearnadam, line 301: Will `insts` always contain all values as the map? Is `apply` safe here?

kyo-data/shared/src/main/scala/kyo/Render.scala
```diff
@@ -16,63 +16,56 @@ sealed trait LowPriorityRenders:
 object Render extends LowPriorityRenders:
     inline def apply[A](using r: Render[A]): Render[A] = r
 
+    def instance[A](impl: A => Text): Render[A] =
```
hearnadam, line 19:
```suggestion
    def from[A](impl: A => Text): Render[A] =
```

I would like to avoid `instance` for non-singleton style values.

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
hearnadam: I think your change is fine.

hearnadam (commented):
This is awesome!

fwbrasil:
I was about to post a PR generalizing `FieldsLike` to a `TypeSet` utility :) We have an issue with the kernel's `Boundary` that I think we could solved with it: https://github.com/getkyo/kyo/pull/new/typeset. @road21 let me know what you think, I'll pause working on `Record` for now since you're working on it

kyo-data/shared/src/main/scala/kyo/Render.scala
```diff
@@ -16,63 +16,56 @@ sealed trait LowPriorityRenders:
...
-                while valIter.hasNext do
-                    builder.append(",")
-                    builder.append(showIter.next().asText(valIter.next()))
-                    ()
-                end while
-                builder.append(")")
-                builder.toString()
-            end asText
-        end new
+        Render.instance: (value: A) =>
```
fwbrasil, line 29: Do we need `Render.instance`? I think the compiler can materialize `Render` directly from an anonymous function

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
fwbrasil: agree

road21:
> I was about to post a PR generalizing `FieldsLike` to a `TypeSet` utility :) We have an issue with the kernel's `Boundary` that I think we could solved with it: https://github.com/getkyo/kyo/pull/new/typeset. @road21 let me know what you think, I'll pause working on `Record` for now since you're working on it

The main reason I've added `Names` and `Values` type members to `FieldsLike` is that `erasedValue` and `match types` work bad with abstract types (or opaque types). As I see, you solved this problem with `final infix class ~[Name <: String, Value] private ()`. 

I think your approach is much better and `TypeSet` really can help in many other cases (I haven't seen yet issue with `Boundary`, but I will check).

So, let's finish your PR first, and after that I will fix this PR.

fwbrasil:
Cool! I've just posted https://github.com/getkyo/kyo/pull/1021

kyo-data/shared/src/main/scala/kyo/Record.scala
road21: It's not safe and there was a bug here because of record subtyping (we may want to print reach record as record with only one field)

I have fixed that and added test

kyo-data/shared/src/main/scala/kyo/Render.scala
road21: When I try to to make `Render` in inline methods via anonymous function I get compiler warn:

```
An inline given alias with a function value as right-hand side can significantly increase
generated code size. You should either drop the `inline` or rewrite the given with an
explicit `apply` method.
```

May be we can just ignore this warn, not sure.

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -432,10 +432,12 @@ class RecordTest extends Test:
             """)
         }
 
-        "not compile when fields lack CanEqual" in {
+        "not compile when fields lack CanEqual" in pendingUntilFixed { // looks like scala3 bug
```
road21, line 435: I don't know why this test works in main

When I was fighting with it, I minimized my case to
```scala
import scala.compiletime.*
import scala.language.strictEquality

class Lol
object Lol:
  inline given CanEqual[Lol, Lol] = error("lol")

val x = new Lol
val y = new Lol

// summon[CanEqual[Lol, Lol]] // doesn't compile
x == y // compile, and this is strange
```

https://scastie.scala-lang.org/road21/B8hWnSQNTriBcOYndUduRg

For me this looks like a bug in scala3 compiler, what do you think?

fwbrasil: yeah, that's odd. I can follow up on this

road21: https://github.com/scala/scala3/issues/22420

PR #1021, opened 2025-01-15
[data] Record: Generalize FieldsLike to TypeSet + provide CanEqual evidence

fwbrasil:
This PR generalizes the approach introduced by @road21 in https://github.com/getkyo/kyo/pull/1013 and introduces `CanEqual` for `Record`. I gave a shot at providing `Render` with but couldn't make it work properly.

I'm planning to try `TypeSet` to resolve the inference issues we have with `Boundary`. We currently have to use an `inline` [method indirection](https://github.com/getkyo/kyo/blob/5674fab6a54feae3d8936182c18f73082910810f/kyo-core/shared/src/main/scala/kyo/Async.scala#L76) for boundaries to infer correctly.


kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -423,6 +415,63 @@ class RecordTest extends Test:
...
+            val record2 = "name" ~ "Alice" & "age" ~ 30 & "scores" ~ List(1, 2, 3)
+
+            assert(record1 == record2)
+        }
+
+        "not compile with different field names" in {
+            val record1 = "name" ~ "Alice" & "age" ~ 30
+            val record2 = "name" ~ "Alice" & "years" ~ 30
+
+            assertDoesNotCompile("""
```
hearnadam, line 430: Is the `typeCheck` function available here? Would be good to check the error.

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -423,6 +415,63 @@ class RecordTest extends Test:
 
             assert(record1 != record2)
         }
+
+        "compile when all fields have CanEqual" in {
```
hearnadam, line 419: Can you add a test when a CanEqual doesn't exist? Is the error message clear?

kyo-data/shared/src/main/scala/kyo/internal/TypeSet.scala
```diff
@@ -0,0 +1,122 @@
...
+/** A type-level utility for decomposing intersection types into their constituent parts.
+  *
+  * TypeSet addresses the challenge of working with intersection types in a type-safe manner, particularly when implementing type-level
+  * operations like equality checking and type class derivation. It provides a way to:
+  *
+  *   - Break down complex intersection types into simpler components
+  *   - Apply type constructors uniformly across all components
+  *   - Collect type class instances for all component types
+  */
+trait TypeSet[A]:
```
hearnadam, line 19: Can this be sealed?
```suggestion
sealed abstract class TypeSet[A]:
```

kyo-data/shared/src/main/scala/kyo/internal/TypeSet.scala
```diff
@@ -0,0 +1,122 @@
...
+      *   the intersection type to decompose
+      * @tparam F
+      *   the type class to summon instances for
+      * @return
+      *   a List of type class instances
+      */
+    transparent inline def summonAll[A: TypeSet as ts, F[_]]: List[F[Any]] =
+        summonAllLoop[ts.AsTuple, F]
+
+    private type Join[A <: Tuple] = Tuple.Fold[A, Any, [B, C] =>> B & C]
```
hearnadam, line 64: As a reminder we have the `Join` type in Async/Fiber... seems okay.

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -262,31 +252,27 @@ object Record:
         class All[T <: Tuple](val fields: Set[Field[?, ?]]) extends AnyVal
```
hearnadam, line 252:
```suggestion
        final class All[T <: Tuple](val fields: Set[Field[?, ?]]) extends AnyVal
```

kyo-data/shared/src/main/scala/kyo/internal/TypeSet.scala
```diff
@@ -0,0 +1,122 @@
...
+end TypeSet
+
+object TypeSet:
+
+    // Cached instance used for optimization.
+    private val cached: TypeSet[Any] =
+        new TypeSet[Any]:
+            type AsTuple = EmptyTuple
+
+    private transparent inline def summonAllLoop[T <: Tuple, F[_]]: List[F[Any]] =
```
hearnadam, line 46: Will this be impacted by max inlines?

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
fwbrasil: I had forgotten about that. It's available in the module but let me follow up on that separately and do an overall cleanup of the codebase. I'll also try to avoid the `Test` file copies in each module

kyo-data/shared/src/main/scala/kyo/internal/TypeSet.scala
fwbrasil: Good point. I've unrolled the loop. It should support 16 * 32 elements with the default inlining limit of 32: https://github.com/getkyo/kyo/pull/1021/commits/0922281d73aac7a1379cdd3e4a9cbe61c83f46ad

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
fwbrasil: It gives the regular `CanEqual` error message. There a few tests below

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
hearnadam: Sounds good.

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -259,34 +249,30 @@ object Record:
...
+        final class All[T <: Tuple](val fields: Set[Field[?, ?]]) extends AnyVal
         object All:
             given All[EmptyTuple] = All(Set())
-            given [H, T <: Tuple](using a: AsField[H], f: All[T]): All[H *: T] =
+            given [Name <: String, Value, T <: Tuple](using a: AsField[Name, Value], f: All[T]): All[Name ~ Value *: T] =
                 All(f.fields + a)
         end All
 
-        given [A](using ev: FieldsLike[A], a: All[ev.T]): AsFields[A] = a.fields
+        given [A](using ev: TypeSet[A], a: All[ev.AsTuple]): AsFields[A] = a.fields
```
road21, line 259: With `TypeSet` you can drop `AsFields.All` and replace it with similar to `CanEquals` derivation. Like:

```scala
object AsFields:
        def apply[A](using af: AsFields[A]): Set[Field[?, ?]] = af

        inline given [Fields](using ev: TypeSet[Fields]): AsFields[Fields] =
            AsFieldsInternal.summonAsField
end AsFields
    
 object AsFieldsInternal:
    private type HasAsField[Field] =
        Field match
            case name ~ value => AsField[name, value]

    inline def summonAsField[Fields](using ev: TypeSet[Fields]): Set[Field[?, ?]] =
        TypeSet.summonAll[Fields, HasAsField].map(Record.AsField.toField).toSet
end AsFieldsInternal
```

External object is needed here because of [this compiler bug](https://github.com/scala/scala3/issues/22370)

kyo-data/shared/src/main/scala/kyo/internal/TypeSet.scala
```diff
@@ -0,0 +1,140 @@
...
+/** A type-level utility for decomposing intersection types into their constituent parts.
+  *
+  * TypeSet addresses the challenge of working with intersection types in a type-safe manner, particularly when implementing type-level
+  * operations like equality checking and type class derivation. It provides a way to:
+  *
+  *   - Break down complex intersection types into simpler components
+  *   - Apply type constructors uniformly across all components
+  *   - Collect type class instances for all component types
+  */
+sealed abstract class TypeSet[A]:
```
road21, line 19: I afraid that this naming can confuse because it looks like `TypeMap`

May be call it `AsTuple` or smth similar?

But, on the other hand, this is internal package

kyo-data/shared/src/main/scala/kyo/Record.scala
fwbrasil: I've tried applying this change and a few approaches like changing `AsField` to take a single param to make it work but I always end up with the compiler not being able to summon `AsField`: 

```scala
[error] -- [E172] Type Error: /Users/fwbrasil/workspace/kyo/kyo-data/shared/src/test/scala/kyo/RecordTest.scala:271:42
[error] 271 |            val compacted = record.compact
[error]     |                                          ^
[error]     |No given instance of type Record$_this.AsField[("name" : String) ~ String] was found
[error]     |
[error]     |The following import might fix the problem:
[error]     |
[error]     |  import kyo.Record.AsField.given_AsField_~
[error]     |
[error]     |---------------------------------------------------------------------------
[error]     |Inline stack trace
[error]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]     |This location contains code that was inlined from TypeSet.scala:139
[error] 139 |                summonInline[F[t]].asInstanceOf[F[Any]] :: summonAllLoop[ts, F]
[error]     |                ^^^^^^^^^^^^^^^^^^
[error]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]     |This location contains code that was inlined from TypeSet.scala:139
[error]  69 |        summonAllLoop[ts.AsTuple, F]
[error]     |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]     |This location contains code that was inlined from TypeSet.scala:139
[error] 256 |            TypeSet.summonAll[Fields, AsField].toSet
[error]     |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[error]     |This location contains code that was inlined from TypeSet.scala:139
[error] 250 |            AsFieldsInternal.summonAsField
[error]     |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]      ---------------------------------------------------------------------------
[error] one error found
```

can we address this as a follow up?

kyo-data/shared/src/main/scala/kyo/internal/TypeSet.scala
fwbrasil: Good point, I think even in an `internal` package it can be confusing since IDEs can automatically import it. I think `AsTuple` doesn't express well that it's meant for manipulating type sets. Maybe `IntersectionOps`?

hearnadam: How about `TypeIntersection`?

kyo-data/shared/src/main/scala/kyo/Record.scala
road21: https://github.com/getkyo/kyo/pull/1024

road21: > can we address this as a follow up?

Yes, of course, we can

kyo-data/shared/src/main/scala/kyo/internal/TypeSet.scala
road21: `IntersectionOps`, `TypeIntersection` both look good to me

fwbrasil: renamed to `TypeIntersection`

fwbrasil:
merging to unblock follow-up changes

PR #1024, opened 2025-01-16
drop All trait

road21:


kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -276,3 +268,12 @@ object Record:
                 "which is not currently supported by the Tag implementation."
         )
 end Record
+
+object AsFieldsInternal:
```
fwbrasil, line 272: hmmm I hadn't tried outside of the `Record` companion. Thanks! :)

PR #1025, opened 2025-01-17
reorganize base test classes

fwbrasil:
First step to replace the use of `assertDoesNotCompile` with `typeCheckFailure`. With this change, the new method is available in all modules. I've researched using symbolic links to avoid the file copies but it's tricky in Windows environments. I've reduced the code duplication by creating base test classes instead.


kyo-data/shared/src/main/scala/kyo/internal/BaseKyoDataTest.scala
```diff
@@ -0,0 +1,35 @@
+package kyo.internal
+
+import kyo.Result
+import scala.compiletime.testing.typeCheckErrors
+import scala.util.Try
+
+private[kyo] trait BaseKyoDataTest:
```
hearnadam, line 7: How about just calling this `BaseTest`?

hearnadam (approved):

PR #1027, opened 2025-01-17
[prelude] tests for Choice with collection operations

fwbrasil:
Kyo provides more strict purity in its implementations due to its support for effects that may invoke the continuation multiple times. Other effect systems typically optimize collection operations by using under-the-hood mutation via builders and iterators but we can't do the same because of the purity restriction. This PR introduces tests to ensure that remains the case or, if we find another solution, multi-shot continuations continue working properly.

### DIFF
--- a/kyo-prelude/shared/src/test/scala/kyo/ChoiceTest.scala
+++ b/kyo-prelude/shared/src/test/scala/kyo/ChoiceTest.scala
@@ -97,4 +97,51 @@ class ChoiceTest extends Test:
         end try
         ()
     }
+
+    "interaction with collection operations" - {
+        "foreach" in {
+            val result = Choice.run(
+                Kyo.foreach(List("x", "y")) { str =>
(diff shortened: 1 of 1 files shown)


kyo-prelude/shared/src/test/scala/kyo/ChoiceTest.scala
```diff
@@ -97,4 +97,51 @@ class ChoiceTest extends Test:
         end try
         ()
     }
+
+    "interaction with collection operations" - {
+        "foreach" in {
+            val result = Choice.run(
+                Kyo.foreach(List("x", "y")) { str =>
+                    Choice.get(List(true, false)).map(b =>
```
fwbrasil, line 105: Note how `Choice` will branch the computation while the collection operation is still being processed. If `foreach` used a mutable buffer, `Choice` calling the continuation multiple times would produce inconsistent results.

hearnadam (approved):

Issue #1022, opened 2025-01-16
Reactive-streams: StreamSubscriber#stream needs to cancel subscription


johnhungerford:
With the pending removal of `Ack` from `Emit`/`Stream`, the existing means of closing subscriptions via streams integration does not work. As discussed in #977 , there are two ways to manage subscriptions for streaming integrations:
1. `StreamSubscriber#stream` produces a `Stream[V, Async & Resource]`, with a finalizer that cancels the subscription. This would best approximate the current behavior.
2. `StreamSubscriber#stream` does not have responsibility for closing the subscription, so that the stream can be run multiple times. In this case, `StreamPublisher` should probably have its own `stream` method which creates a new subscription and cleans it up with a `Resource` finalizer.

HollandDM:
/attempt #1022
I'll try to fix this, aiming to resolve in the first way

johnhungerford:
Closed with #1029

PR #1029, opened 2025-01-18
Update for kyo-reactive-stream

HollandDM:
/claim #1022
This is an update for post-ack era of `Emit` and `Poll` for `kyo-reactive-stream`

### DIFF
--- a/kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/StreamSubscription.scala
+++ b/kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/StreamSubscription.scala
@@ -31,9 +31,9 @@ final private[kyo] class StreamSubscription[V, Ctx](
 
     private[interop] inline def subscribe(using Frame): Unit < IO = IO(subscriber.onSubscribe(this))
 
-    private[interop] def poll(using Tag[Poll[Chunk[V]]], Frame): StreamFinishState < (Async & Poll[Chunk[V]]) =
-        inline def loopPoll(requesting: Long): (Chunk[V] | StreamFinishState) < (IO & Poll[Chunk[V]]) =
-            Loop[Long, Chunk[V] | StreamFinishState, IO & Poll[Chunk[V]]](requesting): requesting =>
+    private[interop] def poll(using Tag[Poll[Chunk[V]]], Frame): StreamComplete < (Async & Poll[Chunk[V]] & Abort[StreamCanceled]) =
+        inline def loopPoll(requesting: Long): (Chunk[V] | StreamComplete) < (IO & Poll[Chunk[V]]) =
+            Loop[Long, Chunk[V] | StreamComplete, IO & Poll[Chunk[V]]](requesting): requesting =>
                 Poll.andMap:
                     case Present(values) =>
                         if values.size <= requesting then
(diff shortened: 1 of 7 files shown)


HollandDM:
About the upstream confusion in the original PR #977, it was my bad to not clarify this at the moment.
`StreamSubscriber`, when subscribed to a `flow.Pulisher`, will receive a `flow.Subscription` as the mechanism to back-pressure the publisher. When a subscriber need `n` elements, it would `request(n)` via subscription, and it is guaranteed to receive `n` `onNext()` message, or the finalize message `onComplete/onError`. It's also the job of Subscriber to call `cancel` on Subscription when it's not needed anymore. So the upstream I was talking about was the subscription, not the publisher. Publisher remain intact whether subscriber done/fail or not (there is a testcase for this).
Anw, I've re-added the removed testcase, and change it description to be clearer in it's meaning

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/StreamSubscription.scala
```diff
@@ -55,27 +55,29 @@ final private[kyo] class StreamSubscription[V, Ctx](
...
                             .andThen(loopPoll(requesting - leftOver.size))
                             .map {
-                                case nextLeftOver: Chunk[V]   => Loop.continue(nextLeftOver)
-                                case state: StreamFinishState => Loop.done(state)
+                                case nextLeftOver: Chunk[V] => Loop.continue(nextLeftOver)
+                                case _: StreamComplete      => Loop.done(StreamComplete)
                             }
-                case Result.Failure(_)       => IO(Loop.done(StreamFinishState.StreamCanceled))
-                case Result.Panic(exception) => Abort.panic(exception).andThen(Loop.done(StreamFinishState.StreamCanceled))
+                case Result.Failure(_) =>
```
fwbrasil, line 61: It seems these two cases could be simplified to:

```scala
case result => Abort.get(result)
```

kyo-reactive-streams/shared/src/main/scala/kyo/interop/flow/StreamSubscription.scala
```diff
@@ -55,27 +55,29 @@ final private[kyo] class StreamSubscription[V, Ctx](
...
-            ._run[Nothing, StreamFinishState, Ctx](Poll.run(stream.emit)(poll).map(_._2))
+            ._run[StreamCanceled, StreamComplete, Ctx](Poll.run(stream.emit)(poll).map(_._2))
             .map { fiber =>
                 fiber.onComplete {
-                    case Result.Success(StreamFinishState.StreamComplete) => IO(subscriber.onComplete())
-                    case Result.Panic(e)                                  => IO(subscriber.onError(e))
-                    case _                                                => IO.unit
+                    case Result.Success(StreamComplete) => IO(subscriber.onComplete())
+                    case Result.Panic(e)                => IO(subscriber.onError(e))
+                    case _                              => IO.unit
```
fwbrasil, line 80: This case catches `Result.Faiure(StreamCanceled)`, is that the intent? Doesn't the subscriber need to be notified?

HollandDM: This is intended, cancel can only be called by subscriber, so they should known already. Notify here will actually fail a test in `reactive-stream-tck`

HollandDM: I'll change this into a concrete branch to be clearer

HollandDM:
Updated!

fwbrasil (approved):

PR #1038, opened 2025-01-21
[core] hub improvements

fwbrasil:


kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -74,74 +83,108 @@ class Hub[A] private[kyo] (
...
+      *   - All subsequent operations will fail with Closed
       *
       * @return
-      *   a Maybe containing any remaining elements in the Hub
+      *   a Maybe containing any messages that were in the Hub's buffer at the time of closing. Returns Absent if the close operation fails
+      *   (e.g., if the Hub was already closed).
       */
     def close(using frame: Frame): Maybe[Seq[A]] < IO =
-        fiber.interruptDiscard(Result.Panic(Closed("Hub", initFrame))).andThen {
+        fiber.interruptDiscard(Result.Failure(Closed("Hub", initFrame))).andThen {
```
fwbrasil, line 99: This was wrong, probably because we didn't support interrupting with a `Failure` in the past

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -74,74 +83,108 @@ class Hub[A] private[kyo] (
...
-    /** Creates a new listener for this Hub with default buffer size.
+    /** Creates a new listener for this Hub with the default buffer size.
       *
       * @return
-      *   a new Listener
+      *   a new Listener that will receive messages from the Hub
       */
     def listen(using Frame): Listener[A] < (IO & Abort[Closed] & Resource) =
-        listen(0)
+        listen(DefaultBufferSize)
```
fwbrasil, line 115: Use a default buffer size of `4096` instead of `0`. Hub's approach with buffers for listeners is designed to reduce contention in high-throughput scenarios. A default of `0` goes against this design.

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -74,74 +83,108 @@ class Hub[A] private[kyo] (
...
+        listen(bufferSize, _ => true)
+
+    /** Creates a new listener for this Hub with a custom filter predicate.
+      *
+      * @param filter
+      *   a predicate function that determines which messages this listener receives
+      * @return
+      *   a new Listener that will only receive messages matching the filter
+      */
+    def listen(filter: A => Boolean)(using frame: Frame): Listener[A] < (IO & Abort[Closed] & Resource) =
```
fwbrasil, line 134: new feature: listeners can provide a function to select messages they're interested in

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -165,33 +209,45 @@ object Hub:
...
-                    Loop.foreach {
-                        ch.take.map { v =>
-                            IO {
-                                val puts =
-                                    listeners.toArray()
-                                        .toList.asInstanceOf[List[Channel[A]]]
-                                        .map(child => Abort.run[Throwable](child.put(v)))
-                                Async.parallelUnbounded(puts).map(_ => Loop.continue)
-                            }
+        IO.Unsafe {
```
fwbrasil, line 212: Optimized to use unsafe operations to avoid allocations + improved error handling

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -74,74 +83,132 @@ class Hub[A] private[kyo] (
...
-            discard(listeners.remove(child))
+            discard(listeners.remove(listener))
         }
+
+    /** Puts multiple elements into the Hub as a batch.
+      *
+      * @param values
+      *   The sequence of elements to put
+      */
+    def putBatch(values: Seq[A])(using Frame): Unit < (Abort[Closed] & Async) = ch.putBatch(values)
```
fwbrasil, line 186: new batch operations

fwbrasil:
@johnhungerford in case you have some time to review

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -3,30 +3,36 @@ package kyo
...
-    listeners: CopyOnWriteArraySet[Channel[A]]
+    listeners: CopyOnWriteArraySet[Listener[A]]
 )(using initFrame: Frame):
 
-    /** Returns the current size of the Hub.
-      *
-      * @return
-      *   the number of elements currently in the Hub
-      */
-    def size(using Frame): Int < (IO & Abort[Closed]) = ch.size
```
fwbrasil, line 22: I've removed `size` because it has poor semantics. Since a hub has both the main buffer and the listener buffers, this method can be confusing given that it can indicate it's the total number of messages in the hub. We'll likely need to expose more information but it's probably better to do that via stats/metrics.

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -165,33 +249,45 @@ object Hub:
...
-                        ch.take.map { v =>
-                            IO {
-                                val puts =
-                                    listeners.toArray()
-                                        .toList.asInstanceOf[List[Channel[A]]]
-                                        .map(child => Abort.run[Throwable](child.put(v)))
-                                Async.parallelUnbounded(puts).map(_ => Loop.continue)
-                            }
+        IO.Unsafe {
+            val channel          = Channel.Unsafe.init[A](capacity, Access.MultiProducerSingleConsumer).safe
```
fwbrasil, line 253: Using relaxed access policy for both the main and listener buffers since it provides better performance. Since there's only one fiber handling the distribution, the global buffer is `MultiProducerSingleConsumer` and the listener buffer is `SingleProducerMultiConsumer`

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -60,10 +66,13 @@ class Hub[A] private[kyo] (
...
 
-    /** Puts an element into the Hub, potentially blocking if the Hub is full.
+    /** Puts an element into the Hub, blocking if necessary until space is available.
+      *
+      * This operation will block when both a listener's buffer is full and preventing the Hub from processing, and the Hub's buffer is
+      * full. The element will be delivered to all active listeners once accepted.
       *
       * @param v
-      *   the element to put
+      *   the element to put if the Hub has been closed
```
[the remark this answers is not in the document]
fwbrasil: yeah, it doesn't make sense. Reverting

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -165,33 +249,45 @@ object Hub:
...
-                                val puts =
-                                    listeners.toArray()
-                                        .toList.asInstanceOf[List[Channel[A]]]
-                                        .map(child => Abort.run[Throwable](child.put(v)))
-                                Async.parallelUnbounded(puts).map(_ => Loop.continue)
-                            }
+        IO.Unsafe {
+            val channel          = Channel.Unsafe.init[A](capacity, Access.MultiProducerSingleConsumer).safe
+            val listeners        = new CopyOnWriteArraySet[Listener[A]]
+            def currentListeners = Chunk.from(listeners.toArray()).asInstanceOf[Chunk[Listener[A]]]
```
hearnadam, line 255: Can you avoid copying here as well?

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -257,22 +363,26 @@ object Hub:
...
+          *   - Respects backpressure from downstream consumers
           *
+          * @param maxChunkSize
+          *   Maximum number of elements to include in each chunk. If not specified, chunks will be limited only by buffer capacity or
+          *   available elements
           * @return
-          *   Asynchronous stream of elements from listener
+          *   An asynchronous stream of elements from the listener that completes when closed
           */
         def stream(maxChunkSize: Int = Int.MaxValue)(using Tag[Emit[Chunk[A]]], Frame): Stream[A, Async] =
```
hearnadam, line 382: Should we use `Stream.DefaultChunkSize`?

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -74,74 +83,148 @@ class Hub[A] private[kyo] (
...
-                        }
+                IO.Unsafe {
+                    val child    = Channel.Unsafe.init[A](bufferSize, Access.SingleProducerMultiConsumer).safe
+                    val listener = new Listener[A](this, child, filter)
+                    discard(listeners.add(listener))
+                    closed.map {
+                        case true =>
+                            // race condition
+                            IO {
+                                discard(listeners.remove(listener))
```
hearnadam, line 170: should the `child` be closed? I guess it won't leak the scope here.

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -74,74 +83,148 @@ class Hub[A] private[kyo] (
...
+                                fail
+                            }
+                        case false =>
+                            Resource.acquireRelease(listener)(_.close.unit)
```
hearnadam, line 174: If `acquireRelease` accepts another argument we can avoid the `.unit`, right?

https://github.com/getkyo/kyo/blob/main/kyo-core/shared/src/main/scala/kyo/Resource.scala#L54-L57

johnhungerford: The `release` parameter (which here is `_.close.unit`) has to return a `Unit < ...` effect.

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -74,74 +83,148 @@ class Hub[A] private[kyo] (
...
-                    discard(listeners.removeIf(_ => true))
-                    Loop.indexed { idx =>
-                        if idx == array.length then Loop.done
-                        else
-                            array(idx).asInstanceOf[Channel[A]].close
-                                .map(_ => Loop.continue)
-                    }.andThen(r)
+                    val l = Chunk.from(listeners.toArray()).asInstanceOf[Chunk[Listener[A]]]
+                    discard(listeners.removeIf(_ => true)) // clear is not available in Scala Native
+                    Kyo.foreachDiscard(l)(_.child.close.unit).andThen(r)
```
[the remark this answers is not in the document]
fwbrasil: No, `close` returns a `Chunk` with the remaining elements. I think we should change `foreachDiscard` to take a function that returns `Any` instead of `Unit` like we did for `andThen`

kyo-core/shared/src/main/scala/kyo/Hub.scala
fwbrasil: yeah, there's no need since the object is just discarded. This branch handles a race condition that should be rare as well.

kyo-core/shared/src/main/scala/kyo/Hub.scala
fwbrasil: we could make ` acquireRelease` take a computation `Any < ...` to avoid it

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -74,74 +83,148 @@ class Hub[A] private[kyo] (
...
 
-    private[kyo] def remove(child: Channel[A])(using Frame): Unit < IO =
+    /** Puts multiple elements into the Hub as a batch.
+      *
+      * @param values
+      *   The sequence of elements to put
+      */
+    def putBatch(values: Seq[A])(using Frame): Unit < (Abort[Closed] & Async) = ch.putBatch(values)
+
+    /** Takes exactly n elements from the Hub's buffer, blocking until enough elements are available.
```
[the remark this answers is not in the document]
fwbrasil: Good point. I think it's better to remove, I'll update

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -165,33 +249,45 @@ object Hub:
       *   The result of applying the function
       */
     def initWith[A](capacity: Int)[B, S](f: Hub[A] => B < S)(using Frame): B < (S & IO) =
```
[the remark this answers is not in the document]
fwbrasil: yeah, the API currently requires users to explicitly close resources. Can we follow up on that separately?

kyo-core/shared/src/test/scala/kyo/HubTest.scala
```diff
@@ -1,241 +1,514 @@
...
-            c  <- l.close
-            v1 <- h.offer(2)
-            v2 <- Abort.run(l.poll)
-        yield assert(c == Maybe(Seq()) && v1 && v2.isFailure)
+    "basic operations" - {
+
+        "empty/full state" in run {
+            for
+                h  <- Hub.init[Int](2)
+                _  <- h.listen(0)
```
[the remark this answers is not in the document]
fwbrasil: If there are no listeners, the hub processes and discards messages. I've added this so the listener immediately backpressures the hub. One item stays pending in the fiber execution loop waiting for space in the listener and the other two populate the hub's buffer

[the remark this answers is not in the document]
johnhungerford: Oh ok so it does work the way I expected. In that case (if I understand correctly), the comment below is a little misleading, as it's the first element that is going to be "held by the fiber", not the third.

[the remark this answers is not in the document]
fwbrasil: ah, good point! Updating

hearnadam (approved):

PR #1040, opened 2025-01-21
github issue and PR templates

fwbrasil:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,30 @@
+<!--
+PRs require an approval from any of the core contributors, other than the PR author.
```
fwbrasil, line 2: Formalizing that one approval from any core maintainer is enough to merge a PR. This is a more relaxed control but we can introduce more process if we see it's necessary. I'd like to still make occasional pushes without approval in current phase while we have ground to cover until 1.0 but should tight the controls after 1.0.

Please voice if you have concerns or suggestions!

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
+<!--
+This template isn't a strict requirement to open issues, but please try to provide as much information as possible.
+-->
+
+**Version**: (e.g. `0.16.0`)
+**Module**: (e.g. `kyo-core`)
+**Scala Version**: (e.g. `2.13.10`)
```
hearnadam, line 7:
```suggestion
**Scala Version**: (e.g. `3.6.2`)
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
+<!--
```
hearnadam, line 1: Noting that this doesn't cover feature requests. Might be worth mentioning so people know they are 'accepted'.

hearnadam (approved):

.github/ISSUE_TEMPLATE.md
fwbrasil: Good point. I did some research and found out that github has a way to define forms to create issues. I've added separate forms for bugs and features. When opening an issue, the user will have to select one of them. This is the first time I see this github feature so I don't know if it works well but it seems worth trying.

fwbrasil: I've marked most fields as not required to avoid introducing too much process. I think we can adjust it as we go

PR #1041, opened 2025-01-21
remove IO.unit

fwbrasil:
This method was introduced before we had the `Kyo` companion object. It doesn't make much sense since I'd expect it to return a suspended computation given the pending `IO`.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/IO.scala
+++ b/kyo-core/shared/src/main/scala/kyo/IO.scala
@@ -24,13 +24,6 @@ opaque type IO <: Abort[Nothing] = Abort[Nothing]
 
 object IO:
 
-    /** Creates a unit IO effect, representing a no-op side effect.
-      *
-      * @return
-      *   A unit value wrapped in an IO effect.
-      */
-    inline def unit: Unit < IO = ()
-
     /** Suspends a potentially side-effecting computation in an IO effect.
       *
       * This method allows you to lift any computation (including those with side effects) into the IO context, deferring its execution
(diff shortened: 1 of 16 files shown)


kyo-core/shared/src/main/scala/kyo/IO.scala
```diff
@@ -24,13 +24,6 @@ opaque type IO <: Abort[Nothing] = Abort[Nothing]
 
 object IO:
 
-    /** Creates a unit IO effect, representing a no-op side effect.
-      *
-      * @return
-      *   A unit value wrapped in an IO effect.
-      */
-    inline def unit: Unit < IO = ()
```
hearnadam, line 32: Do you see any value in keeping this with a suspension?  `val unit = IO(())`? I agree the current value is perhaps not there.

fwbrasil: Not really. All the current uses don't require suspension for example. I think if we keep it with suspension, we'll likely see code that doesn't require suspension being suspended.

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -23,6 +23,8 @@ object Kyo:
       */
     inline def pure[A, S](inline v: A): A < S = v
 
+    inline def unit[S]: Unit < S = ()
```
hearnadam, line 26: Can you add a scaladoc and note this doesn't suspend?

hearnadam (approved):

Issue #1014, opened 2025-01-14
Error handling without panic


johnhungerford:
Like ZIO, Kyo effectively has two separate error channels, one precisely typed and the other constrained to `Throwable`. Like ZIO, I tend to use the latter only for unexpected errors. As such, I generally do not handle them, but let them crash the program. All other errors -- even when they are `Throwable`, I like to represent as an `Abort[T]` effect within the pending effect intersection.

The problem is that when I want to handle an error via `Abort.run` (or `.result` in kyo-combinators), I am forced to handle all three cases: success, expected failures, and unexpected (`Throwable`) failures. In my codebase this means a lot of boilerplate where I'm simply passing the panic case back to `Abort.panic`.

Ideally, Kyo would have a narrower version of `Result` that includes only `Success` and `Fail`. We could then have separate methods on `Abort` abort to fully handle errors or only handle the `Fail` case.

hearnadam:
Perhaps we can just use a union type. `Abort.run` returning the subset, then `Abort.runAll` exposing the full Result?

johnhungerford:
Closed with #1042

PR #1042, opened 2025-01-21
Error handle without panic

johnhungerford:
- `Result.Partial[E, A]` type alias for `Success[A] | Failure[E]`
- `Abort.runPartial` to run to `Result.Partial` and not handle `Panic` cases
- `Abort.fold` to handle either all three cases or (overloaded) just `Success` and `Failure`.
- `partialResult` combinator wrapping `Abort.runPartial`
- `foldAbort` combinator wrapping `Abort.fold`
- `fold` and `partialResult` on `ForAbortOps` to handle a subset of the `Abort` union.


johnhungerford:
Closes #1014

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -195,6 +195,62 @@ object Abort:
...
+                        input =>
+                            input.isPanic ||
+                                input.asInstanceOf[Error[Any]].failure.exists(ct.accepts),
+                    handle = [C] =>
+                        (input, _) => input,
+                    done =
+                        case Panic(ct(fail))                        => Result.Failure(fail)
+                        case Panic(e)                               => throw e
+                        case error: Result.Partial[E, A] @unchecked => error,
+                    recover = e => throw e
```
johnhungerford, line 243: I don't love how this means the `Panic` value will be thrown, caught, and then thrown again (`done` -> `recover`), but I couldn't figure out another way to handle this. Would it be better to just `run` and then call `Abort.panic` on the `Panic` case?

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -44,10 +44,16 @@ import scala.util.control.NonFatal
...
-object Result:
+private trait LowPriorityImplicits:
+    given flatSuccess[A](using Flat[A]): Flat[Success[A]] =
+        Flat.unsafe.bypass
+
+object Result extends LowPriorityImplicits:
 
     import internal.*
 
+    type Partial[+E, +A] = Success[A] | Failure[E]
```
fwbrasil, line 55: An opaque type would give us more flexibility to introduce extension methods and potentially change the internal type (not sure that'd ever happen, though). 
```suggestion
    opaque type Partial[+E, +A] >: (Success[A] | Failure[E]) = Success[A] | Failure[E]
```

johnhungerford:
Need to add tests for `Result.Partial` API now that it's an opaque type

fwbrasil:
FYI there's something going on in CI making builds fail. I thought it could be the build caching I had introduced but I reverted it and it's still happening.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
fwbrasil: Couldn't it be `recover = Abort.panic(_)`?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -195,6 +195,62 @@ object Abort:
...
+                    Abort[ER] & S,
+                    Abort[ER] & S,
+                    Any
+                ](
+                    erasedTag[E],
+                    v.map(Result.Success(_))
+                )(
+                    accept = [C] =>
+                        input =>
+                            input.isPanic ||
```
fwbrasil, line 235: I think this condition can be removed since we don't want to handle panics

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -300,6 +356,115 @@ object Abort:
       */
     inline def recover[E]: RecoverOps[E] = RecoverOps(())
 
+    final case class FoldOps[E](dummy: Unit) extends AnyVal:
```
fwbrasil, line 359: Could you make these methods resemble their counterparts in `Result`? It'd provide a more consistent experience. Feel free to make improvements to `Result` if needed as well.

fwbrasil: It seems we could avoid code duplication by introducing `runWith` taking a continuation function and then using the regular `Result` APIs to transform it

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -694,4 +701,39 @@ object Result:
             case other         => other.toString()
     end given
 
+    opaque type Partial[+E, +A] >: Success[A] | Failure[E] <: Result[E, A] = Success[A] | Failure[E]
```
fwbrasil, line 704: Can you add scaladocs here and update the top `Result` scaladoc?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: I've added `runWith` and am using it for `run`, `runPartial`, and `fold`. Shall I update other methods to use it as well? Is there a performance concern with building other `Abort` handlers on top of `runWith`?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -183,18 +188,78 @@ object Abort:
...
+          */
+        def apply[A: Flat, S, ER](v: => A < (Abort[E | ER] & S))(
+            using
+            frame: Frame,
+            ct: SafeClassTag[E],
+            f2: Flat[Result.Partial[E, A]],
+            reduce: Reducible[Abort[ER]]
+        ): Result.Partial[E, A] < (S & reduce.SReduced) =
+            Abort.runWith[E](v):
+                case Panic(thr)                      => throw thr
```
fwbrasil, line 253: This doesn't seem correct. The returned computation should have a pending `Abort[Nothing]` given that panics aren't handled. I think we can remove the `Reducible` evidence and return `Result.Partial[E, A] < (S & Abort[ER])`, which will infer to `Abort[Nothing]` if it's the only failure pending.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
fwbrasil: There shouldn't be a significant performance penalty. Yes, feel free to refactor other methods! :) Do you think we should keep `runWith` public? If not, we could make it `private` and `inline` to reduce potential overhead. We can't inline directly in user code, though. Effect handling generates large trees.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: I've added some additional `...OrThrow` methods to handle completely and throwing in the case of `Panic`.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: My attempt to make `Abort` and `Result` more consistent has led to a pretty massive change. Since I want to overload `fold` in `Abort`, this means un-currying and changing the parameter order in `Result`. Let me know if you have an issue with this.

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
```diff
@@ -267,7 +267,7 @@ class AbortsTest extends Test:
                 "converts matching Panic to Fail" in {
                     val ex     = new RuntimeException("Test exception")
                     val result = Abort.run[RuntimeException](Abort.panic(ex)).eval
-                    assert(result == Result.panic(ex))
+                    assert(result == Result.Failure(ex))
```
johnhungerford, line 270: This test did not assert what it purported to be testing. Note that I made changes to `Abort.run` (now `Abort.runWith`) to produce the behavior this test seems to want. This is now producing a bunch of failed kyo-core tests, however.

Do we actually want `Panic` to be lifted to `Failure` or not? If yes, I'll go and fix the other tests.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -370,6 +383,61 @@ extension [A, S, E](effect: A < (Abort[E] & S))
...
+        onSuccess: A => B < S1,
+        onFail: E => B < S1
+    )(
+        using
+        ct: SafeClassTag[E],
+        fl1: Flat[A],
+        fl2: Flat[B],
+        fr: Frame
+    ): B < (S & S1) =
+        Abort.foldOrThrow(onSuccess, onFail)(effect)
```
johnhungerford, line 410: Note that in kyo-combinators, I'm throwing `Panic` exceptions by default for all error-handling methods that ignore panic *when handling for the full `Abort` type*. When using `forAbort[E]`, these methods will leave `Panic` unhandled and eventually reduce to `Abort[Nothing]`.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
fwbrasil: Sounds good. My concern is more uniformity between the APIs so the user can form a mental model that works for both

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
fwbrasil: I think I had introduced this behavior initially and changed it later based on @hearnadam's feedback but I forgot to update the test. We should probably remove it.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
fwbrasil: Is it a usability decision? It'd be nice to keep a similar behavior to the Abort APIs

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
johnhungerford: Ok -- that's where I was leaning after making the changes.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
johnhungerford: Yeah usability.  The point of this PR is to have a way to deal with `Abort` where you don't have to think about panic. Leaving an `Abort[Nothing]` in the pending intersection when you handle success and failure means you have to handle `Abort` a second time which is not great.

On the other hand, if you are only handling a single subtype within the `Abort`, you probably don't want to throw `Panic` exceptions. For this reason, I broke the `Abort` methods into two forms -- one that leaves `Abort[Nothing]` and one that throws (`Abort.fold` vs `Abort.foldOrThrow`, and `Abort.runPartial` and `Abort.runPartialOrThrow`). The `...OrThrow` methods operate on only the whole `Abort` type rather than allowing you to specify a type. This is essentially the same behavior as kyo-combinators.

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
johnhungerford: Do we want to continue to convert *caught exceptions* to `Failure` when they are subtypes? This doesn't seem to be tested, but it's implemented in `Abort.run` (now `Abort.runWith`). This behavior seems similarly problematic in that it means potentially handling failures from some other part of your application unexpectedly:

```scala
// Say you are given a program that gets some text, and it throws an unexpected (i.e., no Abort) `IOException` 
val effect1: String < Async = ???

// Now say you have an method to persist text, which expects to sometimes throw an `IOException`.
// Accordingly, it tracks these failures with `Abort`.
def persistValue(int: Int): Unit < (Async & Abort[IOException]) = ???

// Then you combine them retrieve and persist text
val effect2: Int < (Async & Abort[IOException]) = effect1.map(someMethod)

// Finally, you handle the error, assuming it's from persistence, not retrieval
val effect3 = Abort.fold[IOException[(effect2):
   case Result.Success(_) => Console.printLine("Successfully retrieved and persisted text")
   case Result.Failure(_) => Console.printLine("Failed to persist text")
   case Result.Panic(e) => Console.printLine(s"Failed to retrieve or persist text with unexpected exception: $e")
```

In the above scenario, `effect3`, when run, would report that it failed to persist the text when in cases where it in fact failed to retrieve the text.

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
fwbrasil: Sounds reasonable. Something to keep in mind is that `Abort[Nothing]` is included in `IO` and `Async` but I think the compiler doesn't automatically de-duplicate the pending effect.

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
fwbrasil: I feel this is a reasonable behavior. If we don't catch exceptions in `Abort` operations, we would need to provide method variations that include catching since it's a common need. I normally use it to handle failures coming from other libraries for example. @hearnadam do you mind sharing your thoughts on this?

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
johnhungerford: Hmm I didn't realize that. In that case, it probably makes more sense to have all of these methods *not* throw and just handle to `Abort[Nothing]`, and to add separate `...OrThrow` methods. Most cases where a user will want to handle `Abort` will be in conjunction with `Async` or `IO` anyways, and this could result in repeatedly throwing and catching the same exceptions, which seems bad.

I'll make the combinators work the same way as `Abort`: `fold`/`foldOrCatch` and `partialResult`/`partialResultOrThrow`.

I'll make `orPanic` reduce `Abort[X]` to `Abort[Nothing]`, and I'll add an `orThrow` method to eliminate `Abort` entirely by throwing exceptions.

johnhungerford: Done. I feel happy with the current API.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -142,7 +142,7 @@ object Abort:
       */
     inline def get[E >: Nothing]: GetOps[E] = GetOps(())
 
-    final class RunOps[E >: Nothing](dummy: Unit) extends AnyVal:
+    final class RunWithOps[E >: Nothing](dummy: Unit) extends AnyVal:
```
fwbrasil, line 145: missing `private` here

fwbrasil (approved):
This is a great improvement! Thanks for making the changes across the modules :)

fwbrasil:
Let's wait for @hearnadam's feedback to merge

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -405,7 +512,7 @@ extension [A, S, E](effect: A < (Abort[E] & S))
         handled.map((v: Result[E, A]) => Abort.get(v.swap))
     end swapAbort
 
-    /** Catches any Aborts and panics instead
+    /** Converts any Aborts to Panic, wrapping non-Throwable Failures in PanicException
```
hearnadam, line 515:
```suggestion
    /** Converts all Aborts to Panic, wrapping non-Throwable Failures in PanicException
```

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -632,17 +827,14 @@ class ForAbortOps[A, S, E, E1 <: E](effect: A < (Abort[E] & S)) extends AnyVal:
...
         fl: Flat[A],
         frame: Frame
-    ): A < (S & reduce.SReduced) =
-        Abort.run[E1](effect.asInstanceOf[A < (Abort[E1 | ER] & S)]).map:
+    ): A < (S & Abort[ER]) =
+        Abort.runPartial[E1](effect.asInstanceOf[A < (Abort[E1 | ER] & S)]).map:
             case Result.Success(v)              => v
-            case Result.Failure(thr: Throwable) => Abort.panic(thr).asInstanceOf[Nothing < Any]
+            case Result.Failure(thr: Throwable) => Abort.panic(thr)
             case Result.Failure(other)          => Abort.panic(PanicException(other)).asInstanceOf[Nothing < Any]
```
hearnadam, line 837:
```suggestion
            case Result.Failure(other)          => Abort.panic(PanicException(other))
```

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
hearnadam: I think we should avoid over-catching exceptions the user does not wrap. Obviously the machinery for effect handling may need to catch this for enriched traces, but I doubt we should be putting that into the 'expected' error channel.

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -694,4 +717,56 @@ object Result:
...
+                self match
+                    case failure: Failure[E] @unchecked => onFailure(failure.failure)
+                    case other                          => onSuccess(other.asInstanceOf[Success[A]].successValue)
+
+            /** Converts the Partial to an Either.
+              *
+              * @return
+              *   An Either with Left containing the error or exception, and Right containing the successful value
+              */
+            def toEitherPartial: Either[E, A] =
```
hearnadam, line 755: Do the methods need to have distinct names?

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -694,4 +717,56 @@ object Result:
...
+      *   - foldPartial
+      *   - toEitherPartial
+      *   - flattenPartial
+      *
+      * @tparam E
+      *   The type of expected errors that can occur
+      * @tparam A
+      *   The type of the successful value
+      */
+    opaque type Partial[+E, +A] >: Success[A] | Failure[E] <: Result[E, A] = Success[A] | Failure[E]
```
hearnadam, line 737: My only real hesitation is that a `Partial` result sounds almost the opposite of what this is. This is the `Expected` result, but I don't think that's great either...

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -183,18 +187,101 @@ object Abort:
...
+            ct: SafeClassTag[E],
+            reduce: Reducible[Abort[ER]]
+        ): Result[E, A] < (S & reduce.SReduced) =
+            runWith[E](v)(identity)
         end apply
     end RunOps
 
     /** Runs an Abort effect. This operation handles the Abort effect, converting it into a Result type.
       */
     inline def run[E]: RunOps[E] = RunOps(())
```
hearnadam, line 228: I'm tempted to suggest that `run` should return `Partial` by default. If we don't expect users to manipulate panics that is...

`run` & `runAll`?

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -269,6 +271,13 @@ object Result:
                 throw exception
     end Panic
 
+    /** Provides extension methods for Result type */
+    extension [A](self: Success[A])
+        def successValue: A = self match
```
hearnadam, line 276: is `value` taken?

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -305,75 +314,86 @@ object Result:
...
           * @param ifError
           *   Function to apply if the Result is an Error (either Failure or Panic)
-          * @param ifSuccess
+          * @param onSuccess
           *   Function to apply if the Result is a Success
           * @return
           *   The result of applying the appropriate function
           */
-        inline def foldError[B](ifError: Error[E] => B)(inline ifSuccess: A => B): B =
+        inline def foldError[B](inline onSuccess: A => B, ifError: Error[E] => B): B =
```
hearnadam, line 356: Perhaps I missed a comment, but doesn't this make the inference worse?

hearnadam (commented):
I'm very happy to see these methods being added, though I'm not in love with much of the naming. We can always change them later... but I don't like doing so 😅

kyo-data/shared/src/main/scala/kyo/Result.scala
fwbrasil: `Partial` made sense to me given that it partially represents a `Result` but I agree it doesn't communicate much meaning. I can't think of good suggestions. Maybe something like `NoPanic` or `WithoutPanic`?

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
fwbrasil: The internal implementation could not catch, that's possible. My concern is the semantics for the user. I think this code not catching the exception would be surprising and inconvenient:

```scala
Abort.run[SomeException](someCodeMayThrowTheException)
```

kyo-data/shared/src/main/scala/kyo/Result.scala
johnhungerford: Yes, because `Partial` is a subtype of `Result` so the methods conflict otherwise. `fold` ought to work without renaming it since the signature is different, but I couldn't get it to work.

kyo-data/shared/src/main/scala/kyo/Result.scala
johnhungerford: Yes, it's an extension on `Result` that returns `Maybe[A]`.

kyo-data/shared/src/main/scala/kyo/Result.scala
johnhungerford: This change was made to make all the `fold...` methods more consistent. `fold` is not curried now and begins with `onSuccess` so that it can be overloaded to include or exclude `onPanic`. I haven't noticed any issues with type inference in the tests at least, but maybe I'm missing something?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: If that's the case, then really what is now called `Partial` should be called `Result`, and what is now called `Result` should be `Result.OrPanic` or something like that. (That would be my preference, FWIW, given how I tend to use `Abort`.)

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
johnhungerford: I think the scenario I gave above is an telling case. While some users would probably want to lift thrown exceptions to failures, those who do not would not have any way to distinguish between expected exceptions and those of the same type thrown unexpectedly by some other part of their code base.

For the former user, we could add a method `runCatching`.

fwbrasil: I still have trouble to see the value of such behavior considering the potential drawback in usability but I might be missing something. Can we follow up on in a new issue?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
fwbrasil: What would be the behavior of `Result.flatMap` in case there's a failure during transformation? It seems the only option is throwing if panics aren't represented

johnhungerford: Yeah good point. Presumably we'd have a `flatMapAll` for `Result.OrPanic` and restrict `flatMap` to the narrower `Result`. This is all on the premise that users generally won't be dealing with the `Panic` case.

fwbrasil:
@johnhungerford @hearnadam are we ok to merge this?

Issue #1034, opened 2025-01-20
Poll docs still mentions Ack


steinybot:
The docs for `Poll` still mentions the use of `Ack`:
https://github.com/getkyo/kyo/blob/e1a2a97be6f5d4cdd412a4be09b64d3b63ab0736/kyo-prelude/shared/src/main/scala/kyo/Poll.scala#L5-L26

johnhungerford:
Weird -- could have sworn I got all of them. Maybe brought a couple back in with a merge. (I notice there a couple others, too.)

I'll fix this.

johnhungerford:
@steinybot this can be closed (#1043)

PR #1043, opened 2025-01-21
Removed all references to Ack in docs

johnhungerford:


johnhungerford:
Closes #1034

kyo-prelude/shared/src/main/scala/kyo/Poll.scala
```diff
@@ -4,21 +4,11 @@ import kyo.kernel.ArrowEffect
...
-  *   - Each poll operation requires an Ack value that signals the consumer's readiness to receive more data
-  *   - Poll returns Maybe[V], where:
-  *     - Present(v) indicates a successful poll with value v
-  *     - Absent indicates the end of the stream (no more values will be available)
-  *   - Once Absent is received, the consumer should stop polling as the stream has terminated
-  *   - Backpressure is maintained through the Ack responses:
-  *     - Continue signals readiness to receive more values
-  *     - Stop indicates the consumer wants to pause receiving values
+  * Poll is used to consume values. Each poll operation signals readiness to receive data, and returns Maybe[V] indicating whether a value
+  * was available.
```
fwbrasil, line 8: should we keep this part of the scaladoc? It seems still valid

```
  *   - Poll returns Maybe[V], where:
  *     - Present(v) indicates a successful poll with value v
  *     - Absent indicates the end of the stream (no more values will be available)
  *   - Once Absent is received, the consumer should stop polling as the stream has terminated
  ```

PR #1045, opened 2025-01-22
Overloaded Stream methods for pure/effectful transformations

### LINKED ISSUE #1026: StreamBench is not fair
I was taking a look at `StreamBench` to see where we are with the recent changes and noticed something important. The stream operations in Kyo all take effectful functions while both ZIO and cat-effect are taking functions that return plain values. To be a fair comparison, the benchmarks for the other libraries should use methods like `evalFilter` and `filterZIO`. I see two main paths:

1. Provide a similar set of paired effectful and non-effectful methods in `Stream`
2. Optimize the handling of effectful functions to fast-path the case where the function doesn't introduce effect suspensions

I think we could try 2 first since it provides a better user experience.

johnhungerford:
Fixes #1026 

### Problem

Stream transformation methods like `map`, `filter`, etc. currently use Kyo's effectful traverse utilities to support effectful transformations across chunks. This approach is not optimal for pure transformations, however, where each chunk could be more performantly transformed using Chunk's own methods.

### Solution

This PR overloads a number of Stream methods to provide separate implementations for pure and effectful transformations. The performance improvement is reflected in `StreamBench`.

### Notes

- In one case, `runFold`, a separate `runFoldKyo` method was added because overloading `runFold` interfered with type inference on the folding function.
- In other cases, overloading required adding implicit evidence. In several of these cases (all except `map`), this evidence is (as far as I can tell) trivial.

### Checklist

- [x] Unit test all changes
- [x] Update scaladocs if needed
- [x] Update the README if needed


johnhungerford:
With the changes to `Stream`, Kyo is faster than ZStream and close to fs2 in performance for both sync and fork in the existing `StreamBench` when running on my laptop. I added a new `StreamIOBench` benchmark that uses suspended `IO` effects when filtering and mapping. For the new benchmark, Kyo is faster than fs2 and ZIO by more than a factor of two on my laptop.

fwbrasil:
Exciting! Thanks for adding the new benchmark as well 🙏

kyo-bench/src/main/scala/kyo/bench/StreamBench.scala
```diff
@@ -21,7 +20,6 @@ class StreamBench extends Bench.SyncAndFork(25000000):
             .filter(_ % 2 == 0)
             .map(_ + 1)
             .runFold(0)(_ + _)
-            .eval
```
fwbrasil, line 24: I hadn't noticed that the `eval` is not needed here  👍

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -43,7 +65,33 @@ sealed abstract class Stream[V, -S]:
...
 
-    /** Transforms each chunk in the stream using the given function.
+    /** Transforms each chunk in the stream using the given pure function.
+      *
+      * @param f
+      *   The function to apply to each chunk
+      * @return
+      *   A new stream with transformed chunks
+      */
+    def mapChunk[V2](f: Chunk[V] => Seq[V2])(
```
fwbrasil, line 75: isn't the `NotGiven` needed here?

johnhungerford: No because since the output is wrapped in `Seq` there's no danger that it will accidentally be a pending value. I include an otherwise unnecessary `Flat[V]` implicit evidence to distinguish the two methods.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -371,7 +475,7 @@ sealed abstract class Stream[V, -S]:
       * @return
       *   The final accumulated value
       */
-    def runFold[A, S2](acc: A)(f: (A, V) => A < S2)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): A < (S & S2) =
+    def runFoldKyo[A, S2](acc: A)(f: (A, V) => A < S2)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): A < (S & S2) =
```
fwbrasil, line 478: I wonder if clause interleaving or a `RunFoldOps` class could avoid a different name here

johnhungerford: I tried both and couldn't make it work. Something about having two parameters makes inference stop working as expected.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -240,16 +311,29 @@ sealed abstract class Stream[V, -S]:
       * @return
       *   A new stream containing only elements that satisfy the predicate
       */
-    def filter[S2](f: V => Boolean < S2)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Stream[V, S & S2] =
+    def filter[B, S2](f: V => Boolean < S2)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Stream[V, S & S2] =
```
hearnadam, line 314: What's the `B` for?

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -240,16 +311,29 @@ sealed abstract class Stream[V, -S]:
...
 
+    def filter(f: V => Boolean)(using
+        tag: Tag[Emit[Chunk[V]]],
+        discr: Flat[Boolean],
+        frame: Frame
+    ): Stream[V, S] =
+        Stream[V, S](ArrowEffect.handleState(tag, (), emit)(
+            [C] =>
+                (input, _, cont) =>
+                    val c = input.filter(f.asInstanceOf[V => Boolean])
```
hearnadam, line 332: Why do you need `asInstanceOf`?

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -194,10 +242,33 @@ sealed abstract class Stream[V, -S]:
...
+      */
+    def takeWhile(f: V => Boolean)(using
+        tag: Tag[Emit[Chunk[V]]],
+        discr: Flat[Boolean],
+        frame: Frame
+    ): Stream[V, S] =
+        Stream[V, S](ArrowEffect.handleState(tag, true, emit)(
+            [C] =>
+                (input, state, cont) =>
+                    if !state then (false, Kyo.pure[Unit, Emit[Chunk[V]] & S](()))
```
hearnadam, line 260: This will Box/unbox the State

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -43,7 +65,33 @@ sealed abstract class Stream[V, -S]:
...
+    ): Stream[V2, S] =
+        Stream[V2, S](ArrowEffect.handleState(tagV, (), emit)(
+            [C] =>
+                (input, _, cont) =>
+                    if input.isEmpty then ((), cont(()))
+                    else
+                        val s = Chunk.from(f(input))
+                        if s.isEmpty then ((), cont(()))
+                        else
+                            Emit.valueWith(Chunk.from(s))(((), cont(())))
```
hearnadam, line 90: Can you reuse the `Chunk` on line 87?

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -33,7 +34,28 @@ sealed abstract class Stream[V, -S]:
...
+    def map[V2](f: V => V2)(using
+        t1: Tag[Emit[Chunk[V]]],
+        t2: Tag[Emit[Chunk[V2]]],
+        ev: NotGiven[V2 <:< (Any < Nothing)],
+        fr: Frame
+    ): Stream[V2, S] =
+        Stream[V2, S](ArrowEffect.handleState(t1, (), emit)(
+            [C] =>
+                (input, _, cont) =>
+                    val c = input.map(f)
```
hearnadam, line 53: I wonder if we can specialize Stream with some inlining, then use specialized Chunk methods for a potentially fully unboxed Stream encoding...

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: Leftover from other attempts at getting the overload to work properly. I'll remove it.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: Oh I meant for line 87 to just be `val s = f(input)` and not convert to chunk until line 90. I'll update.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
fwbrasil: The challenge is the size of the inlined tree but it might be possible do something like:

```scala
inline def map[V2](inline f: V => V2)(using ...) =
    _map(new Mapper(f))

def _map[V2](mapper: Mapper[V, V2]) = ...         
```

Essentially stage only the portion of the code that needs specialization. Let's follow up on this separately, though.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
fwbrasil: I don't think it will. `handleState` is inlined so it gets specialized

README.md
```diff
@@ -1340,6 +1340,8 @@ val result: Chunk[String] < (Env[Config] & Async) =
 
 The `Stream` effect is useful for processing large amounts of data in a memory-efficient manner, as it allows for lazy evaluation and only keeps a small portion of the data in memory at any given time. It's also composable, allowing you to build complex data processing pipelines by chaining stream operations.
 
+Note that a number of `Stream` methods (e.g., `map`, `filter`, `mapChunk`) are overloaded to provide different implementations for pure vs effectful transformations. This can make a big difference for performance, so take care that the functions you pass to these methods are typed to return pure values if they do not include effects. Unncecessarily lifting them to return `A < Any` will result in perfomance loss.
```
hearnadam, line 1343: I suspect this isn't necessary to mention, though I could be wrong.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -43,7 +65,33 @@ sealed abstract class Stream[V, -S]:
...
+      * @param f
+      *   The function to apply to each chunk
+      * @return
+      *   A new stream with transformed chunks
+      */
+    def mapChunk[V2](f: Chunk[V] => Seq[V2])(
+        using
+        tagV: Tag[Emit[Chunk[V]]],
+        tagV2: Tag[Emit[Chunk[V2]]],
+        discr: Flat[Boolean],
```
hearnadam, line 79: It might be worth adding a dummy evidence specifically for this purpose in the future.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -244,12 +315,25 @@ sealed abstract class Stream[V, -S]:
         Stream[V, S & S2](ArrowEffect.handleState(tag, (), emit)(
             [C] =>
                 (input, _, cont) =>
-                    Kyo.filter(input)(f).map { c =>
+                    Kyo.filter(input)(f.asInstanceOf[V => Boolean < S2]).map { c =>
```
hearnadam, line 318:
```suggestion
                    Kyo.filter(input)(f).map { c =>
```

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: Added to `object Stream` to make it clearer to developers that this `Flat` isn't needed. If we need to use it elsewhere we can move it to kyo-data

README.md
johnhungerford: It makes a substantial difference for performance and would be easy to miss, since it seems like the same function. @fwbrasil what do you think?

hearnadam: My general point is not that this doesn't matter (it does). I just suspect few people expose methods on streams that accept functions. If they do, they are likely going to be effectful or not...

johnhungerford: My concern is that Kyo generally permits you to treat `A` and `A < Any` interchangeably without having to worry about performance like you do you in `ZIO`. Users are likely, therefore, to treat the following three scenarios as equivalent and might very easily miss the overload:

```scala
val values1 = stream.map(v => Var.run[Int](0)(someVarMethod(v)).eval).run
val values2 = stream.map(v => Var.run[Int](0)(someVarMethod(v))).run
val values3 = Var.run[Int](0)(stream.map(v => someVarMethod(v)))
```

In this case, I presume, `values1` will be evaluated more performantly than `values2` and `values3`.

fwbrasil: I think it's good to mention it since the performance different can be surprising

fwbrasil:
Apologies for the radio silence the past few days. I was down with flu but I'm getting better now. I hope the project will continue moving forward without me eventually :) We only need an approval of one core contributor to merge

Issue #1051, opened 2025-01-23
[BUG]: Render's CanEqual can't validate all fields


fwbrasil:
### Version

0.16.1

### Scala Version

3.6.2

### Expected Behavior

Records with field values that don't have a `CanEqual` evidence shouldn't allow equality tests under strict equality.

### Actual Behavior

This [pending test](https://github.com/getkyo/kyo/blob/96f5af6836c75f4404cde49756b9446486ac0867/kyo-data/shared/src/test/scala/kyo/RecordTest.scala#L435) reproduces this issue. The compiler currently considers a `CanEqual` evidence is valid even if its inlined body ends up failing to compile. @road21 created compiler [a ticket](https://github.com/scala/scala3/issues/22420) but it seems the developers consider it a [reasonable limitation](https://github.com/scala/scala3/issues/13725).

fwbrasil:
I tried a few things like having a separate evidence that then gets used as an implicit by the `CanEqual` given and using opaque types/wrapper classes but I couldn't make progress.

PR #1052, opened 2025-01-24
fix: bug CanEqual derivation for records

road21:
Closes https://github.com/getkyo/kyo/issues/1051


kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -262,10 +266,8 @@ object Record:
         Field match
```
fwbrasil, line 266: Do we still need `HasCanEqual`?

road21: dropped it

hearnadam (approved):

PR #1055, opened 2025-01-27
migrate tests to use typeCheckFailure

fwbrasil:
### Problem

The `assertDoesNotEqual` implementation provided by scalatest doesn't allow matching the error message, which is a source of tech debt as we change the code base. Tests might be continuing passing if we rename methods or types for example

### Solution

Migrate all tests from `assertDoesNotEqual` to `typeCheckFailure`

### DIFF
--- a/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
@@ -42,7 +42,9 @@ class AsyncTest extends Test:
             }
         }
         "non IO-based effect" in run {
-            assertDoesNotCompile("Async.run(Vars[Int].get)")
+            typeCheckFailure("Async.run(Var.get[Int])")(
+                "The computation you're trying to fork with Async has pending effects"
+            )
         }
     }
 
@@ -140,7 +142,9 @@ class AsyncTest extends Test:
 
     "race" - {
         "zero" in runNotJS {
-            assertDoesNotCompile("Async.race()")
+            typeCheckFailure("Async.race()")(
+                "None of the overloaded alternatives of method race in object Async"
+            )
         }
         "one" in runNotJS {
             Async.race(1).map { r =>
--- a/kyo-core/shared/src/test/scala/kyo/IOTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/IOTest.scala
@@ -109,9 +109,6 @@ class IOTest extends Test:
             }
             succeed
         }
-        "doesn't accept other pending effects" in {
-            assertDoesNotCompile("IO.Unsafe.run[Int < Options](Options.get(Some(1)))")
-        }
     }
 
     "ensure" - {
@@ -175,18 +172,22 @@ class IOTest extends Test:
         }
 
         "does not include wider Abort types" in {
-            assertDoesNotCompile("""
+            typeCheckFailure("""
                 val a: Int < Abort[String] = 1
                 val b: Int < IO            = a
-            """)
+            """)(
+                "Required: Int < kyo.IO"
+            )
         }
 
         "preserves Nothing as most specific error type" in {
-            assertDoesNotCompile("""
+            typeCheckFailure("""
                 val io: Int < IO = IO {
                     Abort.fail[String]("error")
                 }
-            """)
+            """)(
+                "Required: Int < kyo.IO"
+            )
         }
     }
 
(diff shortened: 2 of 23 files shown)


kyo-core/shared/src/test/scala/kyo/IOTest.scala
```diff
@@ -109,9 +109,6 @@ class IOTest extends Test:
             }
             succeed
         }
-        "doesn't accept other pending effects" in {
-            assertDoesNotCompile("IO.Unsafe.run[Int < Options](Options.get(Some(1)))")
```
fwbrasil, line 113: This is a good example of how `assertDoesNotCompile` generates tech debt. `Options` doesn't exist for a long time now.

kyo-data/shared/src/main/scala/kyo/internal/BaseKyoDataTest.scala
```diff
@@ -24,11 +26,11 @@ private[kyo] trait BaseKyoDataTest:
...
     end typeCheck
 
-    transparent inline def typeCheckFailure(inline code: String)(inline error: String): Assertion =
+    transparent inline def typeCheckFailure(inline code: String)(inline expected: String)(using frame: Frame): Assertion =
         typeCheck(code) match
             case Result.Failure(errors) =>
-                if errors.contains(error) && !error.isEmpty() then assertionSuccess
-                else assertionFailure(s"Predicate not satisfied by $errors")
+                if errors.contains(expected) && !expected.isEmpty() then assertionSuccess
+                else assertionFailure(frame.render(Map("expected" -> expected, "actual" -> errors)))
```
fwbrasil, line 33: added frame rendering to make it easier to read failures and present the correct source position. It was reporting `Test.scala` as the source of failures.

kyo-data/shared/src/test/scala/kyo/ResultTest.scala
```diff
@@ -197,18 +197,6 @@ class ResultTest extends Test:
         }
     }
 
-    "get" - {
```
fwbrasil, line 200: this method was removed in favor of `getOrThrow`

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
```diff
@@ -140,7 +142,9 @@ class AsyncTest extends Test:
 
     "race" - {
         "zero" in runNotJS {
-            assertDoesNotCompile("Async.race()")
+            typeCheckFailure("Async.race()")(
+                "None of the overloaded alternatives of method race in object Async"
```
hearnadam, line 146: seems like this error can be improved?

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -611,30 +611,32 @@ class RecordTest extends Test:
         "AsFields behavior" - {
             import Record.AsFields
 
+            val error = "No given instance of type kyo.AsFieldsInternal.HasAsField"
```
hearnadam, line 614: Can this error be improved as well? Implicit not found annotation?

kyo-data/shared/src/test/scala/kyo/ResultTest.scala
```diff
@@ -453,7 +441,7 @@ class ResultTest extends Test:
             "fails to compile for non-Throwable error" in {
                 val failure: Result[String, Int] = Failure("Something went wrong")
                 val _                            = failure
-                assertDoesNotCompile("failure.toTry")
+                typeCheckFailure("failure.toTry")("value toTry is not a member of kyo.Result")
```
hearnadam, line 444: I think we should require `ev: E <:< Throwable` so it has a better error.

hearnadam (approved):

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
fwbrasil: We'd need to introduce a method overload without params. It doesn't seem worth it.

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
fwbrasil: I gave it a try but the compiler doesn't recognize `@implicitNotFound` in opaque types. We'd need to restructure the code to use classes.

PR #1057, opened 2025-01-28
[data] introduce Environment and improve KyoException

fwbrasil:
### Problem

`KyoException` is too verbose in production environments. It's designed to improve the development experience by showing errors with the `Frame` information but, in a production environment, the overhead to log failures can significantly contribute to performance degradation.

### Solution

Introduce `Environment` that enables the verbose mode only if the user explicitly enables the development mode or sbt entries are present in the classpath, which should be the case for Kyo development environments since we recommend both sbt and metals. The risk of false positives seems low since the check is strict.

### Notes

I've also improved `KyoException` because I noticed a `NullPointerException`.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/KyoException.scala
+++ b/kyo-data/shared/src/main/scala/kyo/KyoException.scala
@@ -24,7 +24,7 @@ class KyoException private[kyo] (message: Text = "", cause: Text | Throwable = "
                 case _: Throwable           => Absent
                 case cause: Text @unchecked => Maybe(cause)
 
-        if Environment.isDevelopment() then
+        if Environment.isDevelopment then
             val msg = frame.render(("⚠️ KyoException".red.bold :: Maybe(message).toList ::: detail.toList).map(_.show)*)
             s"\n$msg\n"
         else
--- a/kyo-data/shared/src/main/scala/kyo/KyoException.scala
+++ b/kyo-data/shared/src/main/scala/kyo/KyoException.scala
@@ -2,17 +2,14 @@ package kyo
 
 import kyo.*
 import kyo.Ansi.*
+import kyo.internal.Environment
 import scala.util.control.NoStackTrace
 
-class KyoException private[kyo] (
-    message: Text | Null = null,
-    cause: Text | Throwable | Null = null
-)(using val frame: Frame) extends Exception(
-        Option(message) match
-            case Some(message) => message.toString;
-            case _             => null,
+class KyoException private[kyo] (message: Text = "", cause: Text | Throwable = "")(using val frame: Frame)
+    extends Exception(
+        message.toString,
         cause match
-            case cause: Throwable => cause;
+            case cause: Throwable => cause
             case _                => null
     ) with NoStackTrace:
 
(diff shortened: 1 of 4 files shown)


kyo-data/shared/src/main/scala/kyo/KyoException.scala
```diff
@@ -2,17 +2,14 @@ package kyo
...
 import scala.util.control.NoStackTrace
 
-class KyoException private[kyo] (
-    message: Text | Null = null,
-    cause: Text | Throwable | Null = null
-)(using val frame: Frame) extends Exception(
-        Option(message) match
-            case Some(message) => message.toString;
-            case _             => null,
+class KyoException private[kyo] (message: Text = "", cause: Text | Throwable = "")(using val frame: Frame)
```
fwbrasil, line 8: fall back to empty string instead of null

kyo-data/shared/src/main/scala/kyo/internal/Environment.scala
```diff
@@ -0,0 +1,11 @@
+package kyo.internal
+
+object Environment:
+
+    def isDevelopment(): Boolean =
+        sys.props.get("kyo.development-mode.enable").map(_.toLowerCase) match
```
fwbrasil, line 6: development can also be enabled explicitly via the `kyo.development-mode.enable` system property

kyo-data/shared/src/main/scala/kyo/internal/Environment.scala
```diff
@@ -0,0 +1,37 @@
...
+      *      - Returns true if SBT is detected in the classpath (contains "org.scala-sbt")
+      *      - Returns false otherwise
+      *
+      * @return
+      *   true if running in a development environment, false otherwise
+      */
+    val isDevelopment: Boolean = inferIsDevelopment()
+
+    private[internal] def inferIsDevelopment(): Boolean =
+        sys.props.get("kyo.development-mode.enable").map(_.toLowerCase) match
```
sideeffffect, line 32:
```suggestion
        sys.props.get("kyo.development-mode.enable").flatMap(_.toBooleanOption) match
```
?
Trivial improvement, but it may be better to not dealing with strings where possible.

kyo-data/shared/src/main/scala/kyo/KyoException.scala
```diff
@@ -2,17 +2,14 @@ package kyo
...
+    extends Exception(
+        message.toString,
         cause match
-            case cause: Throwable => cause;
```
sideeffffect, line 15: :scream_cat: is the `;` the cause of the trouble? I'm so sorry :crying_cat_face:

fwbrasil: No, it's harmless :)

kyo-data/shared/src/main/scala/kyo/internal/Environment.scala
fwbrasil: I had to revert this. If the property is set to an incorrect value, I wanted to disable the development mode. It'd be possible to implement the same logic with `toBooleanOption` but it doesn't seem it'd be more clear.

hearnadam (approved):

PR #1059, opened 2025-01-28
[core] attempt to fix flaky Meter test

fwbrasil:
### Problem

I've been debugging some CI timeouts with the Scala Native build. After adding a timeout too all tests yesterday, these two scenarios in `MeterTest` [started failing](https://github.com/getkyo/kyo/actions/runs/13015501108/job/36303570145?pr=1057#step:8:8757) instead just making the build hang. 

### Solution

I was able to reproduce locally by increasing the number of repeats of the concurrency tests but it's quite rare. Inspecting the code, I couldn't find race conditions that could lead to inconsistent state but I noticed that `release` had an undesirable behavior for semaphores since they should always release a permit with the only exception of when the meter is closed. I honestly don't know why but I wasn't able to reproduce the timeouts again locally after addressing this behavior.

I think we could merge this for now and observe if we still have issues in CI.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Meter.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Meter.scala
@@ -113,7 +113,7 @@ object Meter:
             new Base(concurrency, reentrant):
                 def dispatch[A, S](v: => A < S) =
                     // Release the permit right after the computation
-                    IO.ensure(discard(release()))(v)
+                    IO.ensure(release())(v)
                 def onClose(): Unit = ()
         }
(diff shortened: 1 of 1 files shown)


fwbrasil:
build is still hanging, closing this. I'm starting to suspect issues with Scala Native's test runner. I sometimes [see it throwing](https://github.com/getkyo/kyo/actions/runs/13015501108/job/36303570145?pr=1057#step:8:8922) `ConcurrentModificationException`s as well. Closing this.

PR #1061, opened 2025-01-29
[kernel] Loop optimizations

fwbrasil:
### Problem

I'm taking a look at profiling sessions for Kyo's benchmarks and noticed that `Loop` is getting the `Safepoint` thread local on each iteration because of the `run` call being in a default parameter. The compiler generates a method without the `Safepoint` evidence.

### Solution

Restructure methods to avoid the default parameter.


kyo-kernel/shared/src/main/scala/kyo/kernel/Loop.scala
```diff
@@ -575,21 +574,20 @@ object Loop:
       */
     inline def repeat[S](n: Int)(inline run: Safepoint ?=> Unit < S)(using inline _frame: Frame, safepoint: Safepoint): Unit < S =
         @nowarn("msg=anonymous")
-        @tailrec def loop(i: Int)(v: Outcome[Unit, Unit] < S = run)(using Safepoint): Unit < S =
-            if i == n then ()
+        @tailrec def loop(i: Int)(v: Unit < S)(using Safepoint): Unit < S =
```
fwbrasil, line 577: the behavior was correct but the type of the computation here was incorrect

kyo-kernel/shared/src/main/scala/kyo/kernel/Loop.scala
```diff
@@ -602,14 +600,18 @@ object Loop:
       * @return
       *   Nothing, as this loop runs forever unless interrupted
       */
-    inline def forever[S](inline run: Safepoint ?=> Unit < S)(using Frame, Safepoint): Unit < S =
-        def _loop(): Unit < S = loop()
-        @tailrec def loop(): Unit < S =
-            run match
-                case kyo: Kyo[Unit, S] @unchecked =>
-                    kyo.andThen(_loop())
```
fwbrasil, line 610: Avoid `andThen` since it has more overhead than instantiating `KyoSuspend` directly, which is also the pattern in the other methods

kyo-kernel/shared/src/main/scala/kyo/kernel/Loop.scala
```diff
@@ -357,18 +357,18 @@ object Loop:
...
+        @tailrec def loop(v: Outcome4[A, B, C, D, O] < S)(using Safepoint): O < S =
             v match
                 case next: Continue4[A, B, C, D] @unchecked =>
-                    loop(next._1, next._2, next._3, next._4)()
+                    loop(run(next._1, next._2, next._3, next._4))
                 case kyo: KyoSuspend[IX, OX, EX, Any, Outcome4[A, B, C, D, O], S] @unchecked =>
                     new KyoContinue[IX, OX, EX, Any, O, S](kyo):
                         def frame = _frame
                         def apply(v: OX[Any], context: Context)(using Safepoint) =
-                            loop(i1, i2, i3, i4)(kyo(v, context))
```
fwbrasil, line 368: The change also reduces the memory footprint. For example, here it avoids four fields in the `KyoSuspend` closure since the inputs don't need to be captured anymore.

kyo-kernel/shared/src/main/scala/kyo/kernel/Loop.scala
```diff
@@ -602,14 +600,18 @@ object Loop:
       * @return
       *   Nothing, as this loop runs forever unless interrupted
       */
-    inline def forever[S](inline run: Safepoint ?=> Unit < S)(using Frame, Safepoint): Unit < S =
-        def _loop(): Unit < S = loop()
-        @tailrec def loop(): Unit < S =
-            run match
-                case kyo: Kyo[Unit, S] @unchecked =>
-                    kyo.andThen(_loop())
+    inline def forever[S](inline run: Safepoint ?=> Unit < S)(using inline _frame: Frame, safepoint: Safepoint): Unit < S =
```
hearnadam, line 603: I think it might be useful to have an `until` variant (`while` is taken...). We can accept a predicate `A => Boolean`. I prototyped this a while back, but never PR'd it.

hearnadam (approved):

kyo-kernel/shared/src/main/scala/kyo/kernel/Loop.scala
fwbrasil: Good point! I thought `Loop.until` could be confusing to users so I added `Loop.whileTrue` instead

PR #1062, opened 2025-01-31
rename Stream methods prefixed with run

### LINKED ISSUE #1047: Rename Stream.run* methods
Some of the methods in `Stream` have the `run` prefix, which was introduced to indicate that the stream will be fully handled and return a computation instead of a `Stream` instance. I think the prefix doesn't have much value and renaming methods like `Stream.runForeach` to just `Stream.foreach` should make it easier for users to find these methods.

c0d33ngr:
Fixes #1047

### Problem

Some of the methods in `Stream` have the `run` prefix, which was introduced to indicate that the `stream` will be fully handled and return a computation instead of a `Stream` instance. Renaming the `Stream` methods prefixed with `run` makes it easier to be found

### Solution

Rename the `Stream` methods

### DIFF
--- a/README.md
+++ b/README.md
@@ -1290,7 +1290,7 @@ val h: Chunk[Int] < Any =
 
 // Process stream elements without collecting results
 val i: Unit < Any =
-    a.runDiscard
+    a.discard
 
 // Fold over stream elements
 val j: Int < Any =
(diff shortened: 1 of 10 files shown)


fwbrasil:
Thanks for the contribution :) LGTM besides the merge conflicts

fwbrasil:
@c0d33ngr The repository is configured to require an approval to execute CI jobs if it's the user's first contribution. Once the first PR is merged, the following ones won't require approval and execute immediately.

fwbrasil:
I'm running the CI job again since it failed with an unrelated infra issue.

fwbrasil:
I see. The `build-main` and `build-pr` workflows use are configured to use custom dedicated runners. You'd need to create runners with the same names or change the configs: https://github.com/getkyo/kyo/tree/main/.github/workflows

fwbrasil:
awesome! Congrats on your first contribution  🙌

PR #1065, opened 2025-02-01
[data][prelude] Debug.trace fixes and improvements

fwbrasil:
### Problem

The output of `Debug.trace` is confusing because it logs the intermediate steps within a Kyo method as coming from the user-facing frame. The rendering of `Frame` is also presenting the incorrect line number.

### Solution

Change the interceptor to discard intermediate values within a frame and fix the line number rendering in `Frame`.


kyo-data/shared/src/main/scala/kyo/Frame.scala
```diff
@@ -72,7 +72,7 @@ object Frame:
         def show: String = s"Frame(${Position.show(position)}, $className, $methodName, $snippetShort)"
 
         def render: String =
-            Ansi.highlight(s"// ${Position.show(position)} $className $methodName", snippet.toString(), s"", Position.lineNumber(self))
+            Ansi.highlight(s"// ${Position.show(position)} $className $methodName", snippet.toString(), s"", Position.lineNumber(self) - 2)
```
fwbrasil, line 75: The start line was incorrect here because the macro captures code around the actual line

kyo-prelude/shared/src/main/scala/kyo/debug/Debug.scala
```diff
@@ -46,9 +46,15 @@ object Debug:
...
+            var lastValue = Maybe.empty[Any]
             def enter(frame: Frame, value: Any): Boolean =
-                printValue(value)
-                println(frame.render)
+                if frame ne lastFrame then
+                    lastValue.foreach(printValue)
+                    println(frame.render)
+                    lastFrame = frame
+                end if
+                lastValue = Present(value)
```
fwbrasil, line 57: This discards any intermediary values and only logs the final result of a frame

hearnadam (approved):

PR #1066, opened 2025-02-03
Hub.init/initWith is Resource

### LINKED ISSUE #1044: Hub.init/initWith should be Resource
When a `Hub` is initialized using `Hub.init` or `Hub.initWith` it forks a long-running fiber in the background that continually publishes the `Hub` elements to all subscribed listeners. Currently, users are expected to clean these up manually by calling `.close`. I think it makes more sense for `init` and `initWith` to add a `Resource` finalizer closing the `Hub` since an application could end up adding unlimited suspended fibers to the runtime that never get cleaned up.

johnhungerford:
Fixes #1044 

### Problem

`Hub`s are initialized with a long-running fiber that should be cleaned up when the hub is no longer being used. This PR makes the initializing methods `init`, `initWith`, and all overloads `Resource` effects, where the finalizer calls `Hub#close`.

### Solution

`Hub#initWith` uses `Resource.acquireRelease` to construct the `Hub` and calls `_.close.unit` as the finalizer.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Hub.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Hub.scala
@@ -204,7 +204,7 @@ object Hub:
       * @see
       *   [[Hub.DefaultBufferSize]] for the default capacity value used by this method
       */
-    def init[A](using Frame): Hub[A] < IO =
+    def init[A](using Frame): Hub[A] < (IO & Resource) =
         init(DefaultBufferSize)
 
     /** Initializes a new Hub with the specified capacity.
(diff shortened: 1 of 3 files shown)


fwbrasil (approved):

kyo-core/shared/src/main/scala/kyo/Hub.scala
```diff
@@ -251,7 +251,9 @@ object Hub:
                     }
                 }
             }.map { fiber =>
-                f(new Hub(channel, fiber, listeners))
+                Resource
+                    .acquireRelease(new Hub(channel, fiber, listeners))(_.close.unit)
```
hearnadam, line 255: https://github.com/getkyo/kyo/issues/1067

johnhungerford: I think that should be fixed in a separate PR. This means some substantial changes to `Resource`

hearnadam: Yep, just making sure you knew!

hearnadam (approved):

PR #1069, opened 2025-02-05
Print methods of the Console effect should not depend on Abort[IOException]

### LINKED ISSUE #1037: Print methods of the `Console` effect should not depend on `Abort[IOException]`
The Scala `Console.println` doesn't throw an `IOException`. The method uses the Java `PrintStream` type, which fails silently in case of errors. So, the `kyo.Console` effect should not depend upon an `Abort[IOException]` for printing text on the console.

We should rewrite the code such that the examples in the documentation will change accordingly:

```scala
val b: Unit < IO = Console.print("ok")

// Print to stdout with a new line
val c: Unit < IO = Console.printLine("ok")

// Print to stderr
val d: Unit < IO = Console.printErr("fail")

// Print to stderr with a new line
val e: Unit < IO = Console.printLineErr("fail")

// Explicitly specifying the 'Console' implementation
val f: Unit < IO = Console.let(Console.live)(e)
```

See [PrintWriter and PrintStream never throw IOExceptions](https://stackoverflow.com/questions/297303/printwriter-and-printstream-never-throw-ioexceptions) for further details.

rcardin:
Fixes #1037 

### Problem
The Scala `Console.println` doesn't throw an `IOException`. The method uses the Java `PrintStream` type, which fails silently in case of errors. So, the `kyo.Console` effect should not depend upon an `Abort[IOException]` for printing text on the console.

We should rewrite the code such that the examples in the documentation will change accordingly:

```scala
val b: Unit < IO = Console.print("ok")

// Print to stdout with a new line
val c: Unit < IO = Console.printLine("ok")

// Print to stderr
val d: Unit < IO = Console.printErr("fail")

// Print to stderr with a new line
val e: Unit < IO = Console.printLineErr("fail")

// Explicitly specifying the 'Console' implementation
val f: Unit < IO = Console.let(Console.live)(e)
```

See [PrintWriter and PrintStream never throw IOExceptions](https://stackoverflow.com/questions/297303/printwriter-and-printstream-never-throw-ioexceptions) for further details.

### Solution
I removed the `Abort[IOException]` effect from the return type of the `kyo.Console` methods concerning printing. Moreover, I added the following method, which returns whether the output or the error stream behind the `scala.Console` encountered an `IOException` error:

```scala 3
def checkErrors(using Frame): Boolean < IO
```


kyo-core/shared/src/main/scala/kyo/Console.scala
```diff
@@ -19,34 +19,41 @@ final case class Console(unsafe: Console.Unsafe):
       * @param s
```
fwbrasil, line 19: Can you also update the top `Console` scaladoc with the rationale for not having a pending abort + the role of `checkErrors`? I think it's important to clearly flag this behavior to the user.

rcardin: Done. Let me know if the text works for you.

fwbrasil: perfect! thanks

PR #1077, opened 2025-02-18
[kernel] unified Isolate mechanism

fwbrasil:
### Problem

Kyo provides two similar mechanisms to isolate the execution of computations while forking fibers and other cases that require isolation like `STM`. `Boundary` is in the kernel and provides isolation for `ContextEffect` and `Isolate` is designed in prelude to enable forking with effects that handle state via `ArrowEffect`. The current approach has a few major issues:

- Two separate mechanisms make it more difficult to understand the their purpose.
- `Boundary` has limitations with nested invocations like `Async.run(Async.run(..))`, which we work around by providing a user-facing `inline`d method + internal implementation deriving the `Boundary`.
- The use of inlined user-facing methods and the need to support both `Boundary` and `Isolate` in different methods introduces significant code complexity, with `Async` being the most symptomatic scenario.
- I hadn't noticed before working on this change but the `Boundary` macro was failing to disallow forking with composite effects defined via `opaque type`s like `Parse`. It allows forking in those cases, which typically makes the computation hang indefinitely due to the missing handlers.
- `Isolate` requires "manual" use without derivation, which increases the complexity to users.

### Solution

This PR introduces a new `Isolate` implementation in `kyo-kernel`. It still provides two different isolation mechanisms. `Boundary` becomes `Isolate.Contextual` and `Isolate` becomes `Isolate.Stateful`. Important aspects:

- The two versions are still necessary because of methods like `Async.run` and `Fiber.*` can't perform the necessary transformations to isolate and later restore state due to the expectation that the forked `Fiber`s with these methods will run to completion and not suspend with some effect. This expectation isn't present in other methods like `Async.parallel`, which uses `Isolate.Stateful` by wiring the necessary transformations and the final continuation to restore the necessary state.
- Automatic derivation is now provided for both cases but with an important restriction: it's only able to derive for intersections and requires each individual effect to provide their own evidence. This makes it so `opaque type` effects like `Parse` can choose to support isolation or not, even if all the effects they use provide isolation. This approach is more sensible since the ability to provide isolation is a characteristic of the specific effect. For example, running computations in parallel with `Parse` doesn't make much sense since the parsing consumes from a continuous stream of text that couldn't be supported with parallel execution.
- Both `Contextual` and `Stateful` now provide a second type parameter `Passthrough`. With this approach, `opaque type` effects like `Topic` can isolate specific effects and leave other effects like `Async` and `Abort` pending. The type parameter is contravariant so the evidences can be used in contexts where a super set of `Passthrough` effects is used.
- I've added implicit evidences only to effects that can provide isolation without ambiguity. For example, `Var` and `Emit` don't offer evidences because, depending on the specific use, values may need to be merged when restoring the state. In those cases, I still kept the `isolate` companion object like `Var.isolate.update[Int]`.
- The methods of `Stateful` were updated in comparison to the original `Isolate`. There's a new `use` method designed to introduce an isolate as an implicit evidence in a scope like `Var.isolate.update[Int].use(Async.parallel(..))` and the isolation step methods were renamed to `capture`/`isolate`/`restore`. There's also a new `Transform` type that enables effects to have more control over the packing and unpacking of state during isolation. For example, `Stateful.noop` is able to avoid a tuple allocation now by using `type Transform[A] = A`.
- Effects that short circuit would be problematic with the automatic derivation since the ordering of the effect handling could change the semantics of execution. For this reason `Abort` and `Choice` don't provide isolates although it'd be possible to implement `Stateful` for them with the new `Transform` mechanism. I've also added a warning regarding this to the top scaladoc in `Isolate`.
- The issue with nested isolation like in `Async.run(Async.run(..))` is worked around in this new encoding by taking the isolate evidence as the first parameter. With this approach, the compiler is able to infer the correct types. This has an important limitation: if there's an isolate instance in the immediate scope of the method call, it can be automatically selected even if it doesn't make sense. This shouldn't be a problem in user code unless they're defining new isolates, which is an advanced usage, but it's something we need to keep in mind when working on Kyo's effects. For example, I had to add explicit type parameters in `Resource` due to this aspect.

### Notes

- This PR addresses one of the two main issues we have to fix before 1.0. The other main one is the limitations of `Tag` and effect handling with variance.
- I'll follow up with another PR changing the `Async` effect to provide methods that resemble the collection operations in `Kyo.*`. It's something I've been wanting to do for some time but I wanted to address the boundary/isolate issue first.
- Adding support for forking with algebraic effects is actually an open problem in the field. OCaml's algebraic effects for example don't support proper parallelism with effects and provide only concurrency. Haskell libraries don't seem to provide async primitives like Kyo's, and Unison only allows a pending `IO` when forking. It seems Kyo is the first solution to encode this behavior.


kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -598,10 +574,10 @@ object Fiber extends FiberPlatformSpecific:
             def onComplete(f: Result[E, A] => Unit)(using AllowUnsafe): Unit                             = self.onComplete(f)
             def onInterrupt(f: Result.Error[E] => Unit)(using Frame): Unit                               = self.onInterrupt(f)
             def block(deadline: Clock.Deadline.Unsafe)(using AllowUnsafe, Frame): Result[E | Timeout, A] = self.block(deadline)
-            def interrupt()(using frame: Frame, allow: AllowUnsafe): Boolean = self.interrupt(Panic(Interrupted(frame)))
-            def interrupt(error: Panic)(using AllowUnsafe): Boolean          = self.interrupt(error)
-            def interruptDiscard(error: Panic)(using AllowUnsafe): Unit      = discard(self.interrupt(error))
-            def mask()(using AllowUnsafe): Unsafe[E, A]                      = self.mask()
+            def interrupt()(using frame: Frame, allow: AllowUnsafe): Boolean      = self.interrupt(Panic(Interrupted(frame)))
+            def interrupt(error: Result.Error[E])(using AllowUnsafe): Boolean     = self.interrupt(error)
+            def interruptDiscard(error: Result.Error[E])(using AllowUnsafe): Unit = discard(self.interrupt(error))
```
fwbrasil, line 579: This is an unrelated change. `IOPromise` was changed some time ago to support `Error` interruptions instead of just `Panic` but the change wasn't reflected here.

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -114,7 +114,7 @@ object Resource:
                                 }
                             }
                                 .unit
-                                .pipe(Async.run)
+                                .pipe(Async.run[Nothing, Unit, Any])
```
fwbrasil, line 117: This is the example I mentioned in the description where the compiler picks up the isolate in the enclosing context, which requires the workaround with explicit type params

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -1034,29 +1025,21 @@ extension [A, E, Ctx](effect: A < (Abort[E] & Async & Ctx))
       *   A computation that produces the result of this computation
       */
     @targetName("zipLeftPar")
-    inline def <&[A1, E1, Ctx1](next: A1 < (Abort[E1] & Async & Ctx1))(
+    def <&[A1, E1, S1](next: A1 < (Abort[E1] & Async & S1))(
```
fwbrasil, line 1028: We should migrate these methods to use `Async.parallel` so `Isolate.Stateful` is supported. I can work on this with the `Async` changes I'm planning

kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
```diff
@@ -0,0 +1,370 @@
...
+        self =>
+
+        /** The type of state being managed */
+        type State
+
+        /** How state is transformed during isolated execution */
+        type Transform[_]
+
+        /** Evidence that transformed state can be properly handled */
+        given flatTransform[A: Flat]: Flat[Transform[A]]
```
fwbrasil, line 190: I've explored other encodings to have a less intrusive way to introduce the `Flat` evidence for `Transform` but I couldn't find a good solution so I've added it as a regular member that needs to be implemented by sub classes.

kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
```diff
@@ -0,0 +1,370 @@
...
+        /** No-op isolate with no state. */
+        val noop: Stateful[Any, Any] =
+            new Stateful[Any, Any]:
+                type State        = Unit
+                type Transform[A] = A
+                given flatTransform[A: Flat]: Flat[A]                                 = Flat.derive[A]
+                def capture[A: Flat, S2](f: State => A < S2)(using Frame)             = f(())
+                def isolate[A: Flat, S2](state: Unit, v: A < (Any & S2))(using Frame) = v
+                def restore[A: Flat, S2](v: A < S2)(using Frame)                      = v
+                override def andThen[R2, P2](next: Stateful[R2, P2])                  = next
```
fwbrasil, line 271: The new encoding makes `noop` much more efficient since it avoids allocations in all its methods.

kyo-prelude/shared/src/main/scala/kyo/Check.scala
```diff
@@ -88,51 +88,32 @@ object Check:
...
-                    v
-            end new
-        end discard
+    /** Default isolate that accumulates and re-emits failures.
+      *
+      * When the isolation ends, accumulates any check failures that occurred during the isolated computation with failures from the outer
+      * context. This allows building up a complete set of failed checks.
+      *
+      * Important: Note that `Check.runAbort(Async.parallel(computation1, computation2))` will only short circuit once both computations
+      * finish and the isolate re-emits values to restore its state.
```
fwbrasil, line 97: I thought providing more fine-grained isolation strategies wasn't worth it for `Check`. Since it's designed to accumulate errors, collecting all failures and just propagating them seems a sensible behavior but I've added this warning regarding `Check.runAbort`.

kyo-prelude/shared/src/main/scala/kyo/Env.scala
```diff
@@ -102,4 +99,10 @@ object Env:
     /** Creates a UseOps instance for a given type. */
     inline def use[R >: Nothing]: UseOps[R] = UseOps(())
 
+    private val cachedIsolate                         = Isolate.Contextual.derive[Env[Any], Any]
+    given isolate[V]: Isolate.Contextual[Env[V], Any] = cachedIsolate.asInstanceOf[Isolate.Contextual[Env[V], Any]]
+
+    given eliminateEnv: Reducible.Eliminable[Env[Any]] with {}
+    private inline def erasedTag[R] = Tag[Env[Any]].asInstanceOf[Tag[Env[R]]]
```
fwbrasil, line 106: I'll follow up moving this kind of implicit evidences to the end of files. I think keeping at the top isn't great because it's the first thing people will read in the source file and takes away the focus on the provided features.

kyo-prelude/shared/src/main/scala/kyo/Memo.scala
```diff
@@ -58,39 +58,11 @@ object Memo:
...
-          *   An isolate that updates the cache with isolated results
-          */
-        val update: Isolate[Memo] =
-            Var.isolate.update[Memo.Cache]
+    /** Default isolate that combines memoization caches.
+      *
+      * When the isolation ends, merges any cached results from the isolated computation with the outer cache using the later result on
+      * conflicts. This allows memoized results computed in isolation to be reused later.
+      */
+    given isolate: Isolate.Stateful[Memo, Any] = Var.isolate.merge[Cache]((a, b) => Cache(a.map ++ b.map))
```
fwbrasil, line 66: I've also decided to provide an evidence for `Memo` that always merges results. The fine-grained isolation options didn't seem very valuable.

fwbrasil:
I was discussing this change with @hearnadam and `Local.initIsolated` came up. It's actually a different mechanism, similar to Java's non-inheritable thread locals where forked fibers don't inherit values from their parent. I've renamed it to `Local.initNoninheritable` to avoid confusion with `Isolate`.

kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
```diff
@@ -0,0 +1,370 @@
...
+      *   1. Capture initial state before forking
+      *   2. Transform state during isolated execution
+      *   3. Restore transformed state when the fork completes
+      *
+      * @tparam Retain
+      *   The effects that will be satisfied by this isolation
+      * @tparam Passthrough
+      *   Additional effects that will remain pending after isolation
+      */
+    abstract class Stateful[Retain, -Passthrough]:
```
hearnadam, line 180: As discussed offline, I think it would be ideal of `Stateful` could extend `Contextual`, given it's superset of functionality.

kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
```diff
@@ -0,0 +1,370 @@
...
+        type Transform[_]
+
+        /** Evidence that transformed state can be properly handled */
+        given flatTransform[A: Flat]: Flat[Transform[A]]
+
+        /** Runs a computation with full state lifecycle management.
+          *
+          * Convenience method that composes capture, isolate and restore to handle the complete state lifecycle.
+          */
+        def run[A: Flat, S](v: A < (S & Retain))(using Frame): A < (S & Retain & Passthrough) =
```
hearnadam, line 196: Should this be final?
```suggestion
        final def run[A: Flat, S](v: A < (S & Retain))(using Frame): A < (S & Retain & Passthrough) =
```

hearnadam (approved):

PR #1081, opened 2025-02-22
Kyo test

nightscape:


hearnadam:
@nightscape thanks for your contribution. Do you have some time to discuss the approach? While zio-test gets a lot of things right, I think we can do even better, especially considering we only have to support Scala 3.

nightscape:
@hearnadam I agree. I thought I could semi-mechanically apply an LLM-assisted `ZIO[R, E, A]` to `A < Abort[R] & Env[E] & IO & Async` transformation, comment out some less important things and get to a green state along which I could then refactor to a more Kyo-suitable solution.
This was also an experiment with how far you can get with an agentic LLM workflow on a realistic task in a non-trivial Scala codebase.
Unfortunately that didn't work out that well, I'm still stuck with ~600 compile errors...
A better approach would probably be to start with a small feature set and grow that into a more powerful solution, taking inspiration from ZIO.
I burnt most of my available spare time for the first experiment, but I could collaborate with someone on this.

hearnadam:
Sounds good. Are you alright if we close this PR for now then? I have some ideas I can share later in the issue.

PR #1082, opened 2025-02-23
[core] introduce reactive Signal

fwbrasil:
This new effect is in honor of the 600th person to start the repository. Congrats and thank you @diegobernardes!! 🎉

### Problem

Kyo doesn't offer a solution to handle reactive signals, which is a common need in UI solutions for example.

### Solution

Introduce `Signal` to `kyo-core` with mutable and constant versions and methods to stream the state.

### Notes

- This is inspired by fs2's `Signal`/`SignallingRef` but the implementation is more optimized with two separate `AtomicRef`s to manage the state and listeners, which reduces allocations.
- Please see the scaladocs for an explanation of the functionality and implementation.
- I had to add a method to allow masking unsafe promises.
- I've also changed the atomic classes to provide `updateAndGet` + `getAndUpdate` for consistency.


kyo-core/jvm/src/main/scala/kyo/internal/OSSignalPlatformSpecific.scala
```diff
@@ -13,7 +13,7 @@ import kyo.Result
 /** This class provides a jvm-specific implementation of signal handling. It uses reflection to install signal handlers as they may not be
   * available on all implementations of the JVM.
   */
-private[internal] class SignalPlatformSpecific:
+private[internal] class OSSignalPlatformSpecific:
```
hearnadam, line 16: How about calling this `SignalHandler`? `OSSIgnal` is a bit hard to read.

kyo-core/shared/src/main/scala/kyo/Signal.scala
```diff
@@ -0,0 +1,494 @@
...
+  * The companion object provides these creation methods:
+  *   - `Signal.initRef[A]`: creates a mutable `Signal.Ref[A]` initialized with a starting value
+  *   - `Signal.initConst[A]`: creates an immutable `Signal[A]` that always returns the same value
+  *   - `Signal.initRaw[A]`: (low-level API) creates a custom `Signal[A]` by directly implementing its fundamental operations, primarily
+  *     intended for implementing signal combinators and custom signal types
+  *
+  * @tparam A
+  *   The type of value contained in the signal. Must have an instance of `CanEqual[A, A]`
+  */
+abstract class Signal[A](using CanEqual[A, A]):
```
hearnadam, line 31: Should this be `sealed` since we have `initRaw`?

kyo-core/shared/src/main/scala/kyo/Signal.scala
```diff
@@ -0,0 +1,494 @@
...
+
+    /** A mutable reference implementation of Signal that allows modification of its value over time.
+      *
+      * This class provides methods to get, set, and modify the contained value atomically. All operations are thread-safe and will properly
+      * notify observers of changes.
+      *
+      * @tparam A
+      *   The type of value contained in the reference. Must have an instance of `CanEqual[A, A]`
+      */
+    final class Ref[A] private[Signal] (_unsafe: Ref.Unsafe[A])(using CanEqual[A, A]) extends Signal[A]:
```
hearnadam, line 319: Should we avoid the `Ref` class if we plan to have a similar class for an `AtomicRef`?

kyo-core/shared/src/main/scala/kyo/Atomic.scala
```diff
@@ -91,6 +91,22 @@ final case class AtomicInt private (unsafe: AtomicInt.Unsafe):
       */
     inline def addAndGet(v: Int)(using inline frame: Frame): Int < IO = IO.Unsafe(unsafe.addAndGet(v))
 
+    /** Atomically updates the current value using the given function and returns the old value.
+      * @param f
+      *   The function to apply to the current value
+      * @return
+      *   The previous value
+      */
+    inline def getAndUpdate(inline f: Int => Int)(using inline frame: Frame): Int < IO = IO.Unsafe(unsafe.getAndUpdate(f(_)))
```
hearnadam, line 100: `f(_)` application is necessary to convert to the interface?

kyo-core/jvm/src/main/scala/kyo/internal/OSSignalPlatformSpecific.scala
fwbrasil: I think `SignalHandler` could still be confusing with the new `Signal` impl. Maybe it's ok to be more opaque since it's an internal API? Would `OsSignal` work better?

kyo-core/shared/src/main/scala/kyo/Signal.scala
fwbrasil: Good point, I think it can be confusing especially given that other effect systems use `Ref` for the equivalent of Kyo's `AtomicRef`. FS2 uses `SignallingRef` but sounds a bit heavyweight to me. Maybe `SignalRef` exported to the `kyo` package?

kyo-core/shared/src/main/scala/kyo/Atomic.scala
fwbrasil: It's actually unnecessary. Updated.

kyo-core/shared/src/main/scala/kyo/Signal.scala
hearnadam: `SignalRef` is good with me.

kyo-core/jvm/src/main/scala/kyo/internal/OSSignalPlatformSpecific.scala
hearnadam: `Os` might be better

hearnadam (approved):

PR #1083, opened 2025-03-02
[data] optimize Maybe.collect

matteobilardi:
### Problem
The current implementation of `Maybe.collect` checks if the partial function is defined on the input value before calling it. This can be sub-optimal, as it causes a double evaluation of pattern matchers and guards in the partial function (once on the call to `pf.isDefinedAt(value)` and once on the application `pf(value)`. 

### Solution

Using the `applyOrElse` method allows optimized implementations of `PartialFuncion`, such as partial function literals (mentioned in https://github.com/scala/scala/blob/780d80c714354871bc5972045926b7ccd042e102/src/library/scala/PartialFunction.scala#L194)  to avoid double evaluations of guards.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Maybe.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Maybe.scala
@@ -324,14 +324,8 @@ object Maybe:
           *   a new Maybe containing the result of the partial function if defined and applicable, or Absent otherwise
           */
         inline def collect[B](pf: PartialFunction[A, B]): Maybe[B] =
-            if !isEmpty then
-                val value = get
-                if pf.isDefinedAt(value) then
-                    pf(value)
-                else
-                    Absent
-                end if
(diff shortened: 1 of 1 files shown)


kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -445,5 +439,7 @@ object Maybe:
                 else
                     new PresentAbsent(depth)
         end PresentAbsent
+
+        def constAbsent(ignored: Any): Absent = Absent
```
fwbrasil, line 443: We can also avoid the function allocation to invoke `applyOrElse`:
```suggestion
        private val constAbsent: Any => Absent = _ => Absent
```

fwbrasil:
Welcome, and thanks for the contribution! :)

hearnadam (approved):

Issue #389, opened 2024-05-19
Implementing a Techempower benchmark


UtilFunction:
[Techempower](https://github.com/TechEmpower/FrameworkBenchmarks) ABC is a popular benchmark for web frameworks. Including Kyo in the benchmark could be beneficial both in terms of raising awareness of Kyo and Scala, but also in terms of identifying further optimisations.

Omar-Elrefaei:
Hi fwbrasil,
I'll give this a go soon /attempt #389

alphaho:
Hi @fwbrasil,

I would like to have a try. /attempt #389

PR #1085, opened 2025-03-11
Add a link to the TechEmpower's benchmark example

alphaho:
/claim #389

### DIFF
--- a/README.md
+++ b/README.md
@@ -3178,7 +3178,7 @@ val a: Unit < Routes =
     }
 ```
 
-For further examples, Kyo's [example ledger service](https://github.com/getkyo/kyo/tree/main/kyo-examples/jvm/src/main/scala/examples/ledger) provides practical applications of these concepts.
+For further examples, refer to TechEmpower's benchmark [subproject](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Scala/kyo-tapir) for a simple runnable demonstration, and  Kyo's [example ledger service](https://github.com/getkyo/kyo/tree/main/kyo-examples/jvm/src/main/scala/examples/ledger) for practical applications of these concepts.
 
 ### ZIOs: Integration with ZIO
 


alphaho:
And please feel free to let me know is there any improvements I can make on the Kyo Tapir Project under TechEmpower.

fwbrasil:
hi @alphaho, thank you for the contribution! I'm sorry, the bounty description wasn't very clear. Could you address the following as well:

- Have you checked the results of the benchmark or done some analysis of its performance? Our main focus should be the Scala-based alternatives.
- Can you add the code to the `kyo-examples` project as well so we can detect breaking changes?
- As a stretch, it'd nice to have a benchmark in `kyo-bench` reproducing the techempower execution

alphaho:
Aha, let me follow up and address these points!

fwbrasil:
@alphaho any updates on this? I tried looking for Kyo's in the periodic run but couldn't locate it. Do you have initial results comparing to other Scala stacks?

fwbrasil:
@alphaho could you share your analysis of the result here? With it, I think we can merge the change and consider https://github.com/getkyo/kyo/issues/389 done

alphaho:
@fwbrasil sorry for the late reply. As TechEmpower Framework Benchmark is publishing the results to the official website once in a while, the new benchmark numbers are not yet available officially. 

But we can now find the latest test results from the continuous runs: https://tfb-status.techempower.com/

Here's the latest one that have all the kyo test succeeded:
https://www.techempower.com/benchmarks/#section=test&runid=2d00902c-f610-4891-bb36-68372ea27408&test=json

alphaho:
> @alphaho could you share your analysis of the results here? With it, I think we can merge the change and consider #389 done

Here's some findings based on the latest continuous test result from [TechEmpower Framework Benchmark](https://www.techempower.com/benchmarks/#section=test&runid=2d00902c-f610-4891-bb36-68372ea27408&test=json&l=zik0zj-cm7) 

The result now contains:
- the different server backends from tapir (to compare how `kyo-tapir` performs to similar alternatives)
- some other http framework that can run with `kyo-scheduler` (to compare how other frameworks perform with different scheduler)

To minimize the perfromance difference introduced by the JVM, the different tests using the Scala frameworks are now also aligned to use the latest Java LTS version (Java 21 as of today) as much as possible. While some are stuck with Java 17 due to the compatiiblity with Java 21.

Similar to what I've found on my machine, Here's what we can tell from the results:

- It turns out that `tapir as netty(kyo) server` (which is running with the `kyo-tapir` module) is not very performant comparing to other Scala-based alternatives. It's only beating `http4s`, `akka-http` and `pekko-http`.

- Comparing the different tests that leverage KYO, we can see that:
  1. `tapir`'s netty backend is pretty slow, no matter we are using `KYO`, `ZIO` or `Cats` as the effect type. While in the json serialization test, `KYO`'s version is hehind Cats' and ZIO's. 
  2. `zio-http with kyo-scheduler` is the fastest combination out of all the effect system based implementations, a little faster than the `zio-http` native implementation.
  3. Both `zio-http with kyo-scheduler` and `http4s with kyo-scheduler` are faster than their counterparts from this result.

fwbrasil:
Thank you for the additional tests @alphaho  🙏

PR #1086, opened 2025-03-14
[kernel][core] more consistent sync/async collection operations

fwbrasil:
### Problem

Kyo's `Async` method aren't very convenient to deal with collections. 

### Solution

Provide a new set of methods resembling the collection operations in the kernel's `Kyo` companion object.

### Notes

- There was another `Kyo.scala` method in the internal package of the kernel. I renamed it to `KyoInternal.scala` to avoid confusion
- I'll provide more details on the changes in code comments.


kyo-bench/src/main/scala/kyo/bench/arena/TestHttpServer.scala
```diff
@@ -59,7 +59,7 @@ object TestHttpServer:
     def start(concurrency: Int): String =
         val javaBin   = System.getProperty("java.home") + "/bin/java"
         val classpath = System.getProperty("java.class.path")
-        val command   = List(javaBin, "-cp", classpath, "kyo.bench.TestHttpServer", concurrency.toString)
+        val command   = List(javaBin, "-cp", classpath, "kyo.bench.arena.TestHttpServer", concurrency.toString)
```
fwbrasil, line 62: this was broken since the package rename

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -20,25 +20,57 @@ import scala.util.control.NonFatal
...
+      *   - Available system resources
+      *   - Specific performance requirements
+      *
+      * Note: This only affects collection processing methods. Operations like race and gather run with unbounded concurrency.
+      */
+    val defaultConcurrency =
+        import AllowUnsafe.embrace.danger
+        given Frame = Frame.internal
+        IO.Unsafe.evalOrThrow(System.property[Int]("kyo.async.concurrency.default", Runtime.getRuntime().availableProcessors() * 2))
+    end defaultConcurrency
```
fwbrasil, line 72: I'm following a more explicit approach where the user has to override the behavior in each method call otherwise the execution falls back to `cores * 2`. The most problematic cases I've seen in production in similar scenarios are collections with hundreds of elements so this bound should be enough to prevent issues.

I'm not 100% happy with the approach but it seems a good improvement that we can iterate on later.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -298,101 +333,287 @@ object Async:
...
+      * @param seq
+      *   The sequence of elements to process
+      * @param concurrency
+      *   Maximum number of concurrent computations (defaults to [[defaultConcurrency]])
+      * @param f
+      *   Function that takes an index and element, returning a computation
+      * @return
+      *   Chunk containing results in the original sequence order
+      */
+    def foreachIndexed[E, A, B: Flat, S](
```
fwbrasil, line 350: The new collection methods similar to the `Kyo` companion object start here. I've decided to make all of them rely only `foreachIndexed` but there's potential for optimization leveraging the characteristics of each method.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -298,101 +333,287 @@ object Async:
...
-      *   The sequence of computations to run in parallel
+      *   The sequence of computations to execute
+      * @param concurrency
+      *   Maximum number of concurrent computations (defaults to [[defaultConcurrency]])
       * @return
-      *   A sequence containing the results of all computations in their original order
+      *   Chunk containing results in the original sequence order
       */
-    def parallel[E, A: Flat, S](
+    def collectAll[E, A: Flat, S](
```
fwbrasil, line 453: Renamed following the naming pattern in ZIO.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -298,101 +333,287 @@ object Async:
...
+    def repeat[E, A: Flat, S](
+        using isolate: Isolate.Stateful[S, Abort[E] & Async]
+    )(n: Int, concurrency: Int = defaultConcurrency)(
+        f: => A < (Abort[E] & Async & S)
+    )(using Frame): Chunk[A] < (Abort[E] & Async & S) =
+        foreach(0 until n, concurrency)(_ => f)
+
+    /** Executes two computations in parallel and returns their results as a tuple.
+      */
+    inline def zip[E, A1: Flat, A2: Flat, S](
```
fwbrasil, line 490: I've renamed the `parallel` methods to `zip` like in the `Kyo` companion

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -260,7 +281,7 @@ object Fiber extends FiberPlatformSpecific:
       * @return
       *   A Fiber that completes with the result of the first Fiber to complete
       */
-    def race[E, A: Flat, S](
+    private[kyo] def race[E, A: Flat, S](
```
fwbrasil, line 284: I think having the same methods in `Async` and `Fiber` can be confusing so I've decided to remove some and only keep the ones needed to implement the `Async` methods as `private[kyo]`. It limits the functionality but it's also easy to go back to a `Fiber` via `Async.run`.

kyo-data/shared/src/main/scala/kyo/Result.scala
```diff
@@ -346,20 +346,20 @@ object Result:
 
         /** Folds a Result into a value, with separate handling for errors and successes.
           *
-          * @param ifError
+          * @param onError
```
fwbrasil, line 349: minor unrelated fix

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -23,7 +28,7 @@ object Kyo:
       * @return
       *   A computation that directly produces the given value without suspension
       */
-    inline def pure[A, S](inline v: A): A < S = v
+    inline def lift[A, S](inline v: A): A < S = v
```
fwbrasil, line 31: I think `pure` was a bit opaque for people not used to FP.

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -38,47 +43,41 @@ object Kyo:
     inline def unit[S]: Unit < S = ()
 
     /** Zips two effects into a tuple.
-      *
```
fwbrasil, line 41: some cleanup to avoid the file getting too large with the new versions with up to 8 elements

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -422,7 +466,7 @@ object Kyo:
       * @return
       *   A new effect that produces a Chunk of repeated values
       */
-    def fill[A, S](n: Int)(v: Safepoint ?=> A < S)(using Frame, Safepoint): Chunk[A] < S =
+    def repeat[A, S](n: Int)(v: Safepoint ?=> A < S)(using Frame, Safepoint): Chunk[A] < S =
```
fwbrasil, line 469: `Kyo.repeat`/`Async.repeat` seem to be more clear

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -298,101 +333,287 @@ object Async:
...
+                case 1 => f(0, seq(0)).map(Chunk(_))
+                case size if size <= concurrency =>
+                    isolate.capture { state =>
+                        Fiber.foreachIndexed(seq)((idx, v) => isolate.isolate(state, f(idx, v)))
+                            .map(_.use(r => Kyo.foreach(r)(isolate.restore)))
+                    }
+                case size =>
+                    isolate.capture { state =>
+                        val groupSize = Math.ceil(size.toDouble / Math.max(1, concurrency)).toInt
+                        Fiber.foreachIndexed(seq.grouped(groupSize).toSeq)((idx, seq) =>
```
hearnadam, line 369: I think we need to add support for non-seq collections. We may need a typeclass if we want to support only immutable collections for compatibility with multi-shot continuations.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -298,101 +333,287 @@ object Async:
...
+      * @param concurrency
+      *   Maximum number of concurrent computations (defaults to [[defaultConcurrency]])
+      * @param f
+      *   Function that takes an index and element, returning a computation
+      * @return
+      *   Chunk containing results in the original sequence order
+      */
+    def foreachIndexed[E, A, B: Flat, S](
+        using isolate: Isolate.Stateful[S, Abort[E] & Async]
+    )(seq: Seq[A], concurrency: Int = defaultConcurrency)(f: (Int, A) => B < (Abort[E] & Async & S))(using
```
hearnadam, line 352: Should we have an `IndexedFunction` type to avoid boxing?

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -298,101 +333,287 @@ object Async:
...
+      * @param n
+      *   Number of times to repeat the computation
+      * @param concurrency
+      *   Maximum number of concurrent computations (defaults to defaultConcurrency)
+      * @param f
+      *   The computation to repeat
+      * @return
+      *   Chunk containing results of all iterations
+      */
+    def repeat[E, A: Flat, S](
```
hearnadam, line 481: This seems useful. Should we call it `fill`?

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
hearnadam: If we are choosing similar naming to ZIO, what do you think about `succeed`?

kyo-core/shared/src/main/scala/kyo/Async.scala
fwbrasil: Multi-shot continuations aren't a concern here since there's `IO` suspension. I'm not a fan of supporting mutable collections but I think other effect systems allow them, right? It might be important for usability.

kyo-core/shared/src/main/scala/kyo/Async.scala
fwbrasil: This operation was actually called `fill` in the `Kyo` companion. I've renamed it `repeat` because I thought it was a better name indicating that the computation will be executed multiple times. I'd be ok with changing it back to `fill` here and in the `Kyo` companion.

kyo-core/shared/src/main/scala/kyo/Async.scala
fwbrasil: I think we can wait the compiler to implement specialization unless we see it introducing overhead. The JIT can sometimes avoid the allocation.

kyo-core/shared/src/main/scala/kyo/Async.scala
hearnadam: Yes, we need to support more collections for usability.

kyo-core/shared/src/main/scala/kyo/Async.scala
hearnadam: Sounds good

kyo-core/shared/src/main/scala/kyo/Async.scala
hearnadam: How about `replicate`?

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
fwbrasil: I think `succeed` makes more sense in a monad that includes error handling by default. In this case, `lift` seems more appropriate.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -188,14 +221,14 @@ object Async:
       */
     def race[E, A: Flat, S](
         using isolate: Isolate.Stateful[S, Abort[E] & Async]
-    )(seq: Seq[A < (Abort[E] & Async & S)])(
+    )(seq: Iterable[A < (Abort[E] & Async & S)])(
```
hearnadam, line 224: same as elsewhere, could use rename of `seq`

kyo-core/shared/src/main/scala/kyo/Async.scala
hearnadam: `foreachDiscard` seems like the one case this is very true.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -293,106 +328,292 @@ object Async:
...
+                case 1 => f(0, seq.head).map(Chunk(_))
+                case size if size <= concurrency =>
+                    isolate.capture { state =>
+                        Fiber.foreachIndexed(seq)((idx, v) => isolate.isolate(state, f(idx, v)))
+                            .map(_.use(r => Kyo.foreach(r)(isolate.restore)))
+                    }
+                case size =>
+                    isolate.capture { state =>
+                        val groupSize = Math.ceil(size.toDouble / Math.max(1, concurrency)).toInt
+                        Fiber.foreachIndexed(seq.grouped(groupSize).toSeq)((idx, group) =>
```
hearnadam, line 369: does this need to be converted to a Seq now?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/Async.scala
fwbrasil: yep, `group` returns an `Iterator`.  I tried accepting `IterableOnce` instead of `Iterable`, which would support `Iterator`, but then it's not possible to get the size to initialize data structures in `Fiber.*` methods since it allows traversing only once.

kyo-core/shared/src/main/scala/kyo/Async.scala
fwbrasil: I ended up renaming it back to `fill`. The similarity with Scala collection APIs should be helpful too

PR #1088, opened 2025-03-14
[build] attempt to enable Native tests again

fwbrasil:
### Problem

We disabled the Scala Native tests some time ago. It's a problem since we might not catch potential issues with changes.

### Solution

I couldn't identify why we were seeing timeouts before but I did a quick test in main and the build passed. I think we can try again and see how it goes.

### DIFF
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -30,4 +30,4 @@ jobs:
       run: sbt '+kyoJS/test'
 
     - name: Build Native
-      run: sbt '+kyoNative/Test/compile' # test
+      run: sbt '+kyoNative/test'
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -34,4 +34,4 @@ jobs:
       run: sbt '+kyoJS/testQuick'
 
     - name: Build Native
-      run: sbt '+kyoNative/Test/compile' # testQuick
+      run: sbt '+kyoNative/testQuick' 


hearnadam (approved):

fwbrasil:
The build failed again like before :( 

@WojciechMazur sorry for pinging you directly, I hope it's ok. I'm not sure how to isolate this issue and my attempts at debugging it in the past weren't successful. The tests pass but the SBT build just keeps hanging until build times out. Have you seen this issue before? It happens only in Native.

WojciechMazur:
I can see in logs that some thread crashed 
```
Exception in thread "test-scheduler-worker-1"java.lang.RuntimeException: Test exception.
	at kyo.scheduler.SchedulerTest.$anonfun$3(Unknown Source)
	at kyo.scheduler.SchedulerTest$$Lambda$11.apply(Unknown Source)
```

I suspect that we might incorrectly wait for non-deamon threads that crashed for some reason. It's also possible that there might be some bug in port of `java.util.concurrent`. Anyway I'll try to debug it locally. Lately I'm a bit overwhelmed with other responsibilities, but I'll try to fix it asap 
You can observe the [dedicated issue](https://github.com/scala-native/scala-native/issues/4265) for future updates, but I'll also let you know with progress

fwbrasil:
Thank you @WojciechMazur! It's nothing urgent since we currently don't have changes specific to Native in the pipeline. The exception log is actually expected. It checks this code path that delegates to the default uncaught exception handler of the thread: https://github.com/getkyo/kyo/blob/c277a4a19dbe98f3355e7caa7337dd6aa3025bbe/kyo-scheduler/jvm-native/src/main/scala/kyo/scheduler/Worker.scala#L259

fwbrasil:
I'll close the PR for now and track the progress in https://github.com/scala-native/scala-native/issues/4265

Issue #1084, opened 2025-03-04
[BUG]: Path.readAll returns the wrong file name


fwbrasil:
### Version

0.16.2

### Scala Version

3.6.2

### Expected Behavior

`Path.readAll` should return a map with the file name as the key.

### Actual Behavior

The method currently returns the first segment of the path, not the file name. It should use the last part of the path:

https://github.com/getkyo/kyo/blob/b083b22646bf874d1dbd54f70865495b163f13bb/kyo-core/jvm/src/main/scala/kyo/Path.scala#L39

PR #1090, opened 2025-03-14
fix: Path#readAll

hearnadam:
### Problem
#1084 

This will only capture files farthest from the root. We may want to produce a tree structure of some kind later.


fwbrasil:
java version issues?

```
[error] -- [E008] Not Found Error: /home/runner/work/kyo/kyo/kyo-core/jvm/src/main/scala/kyo/Process.scala:3:16 
[error] 3 |import java.io.{IO as _, *}
[error]   |                ^^
[error]   |                value IO is not a member of java.io
[error] one error found
```

hearnadam:
Yeah, will fix today. Need to explicitly import to avoid java 23 issues.

PR #1091, opened 2025-03-14
chore: update scala to 3.6.4

road21:
Fixes problem from https://github.com/scala/scala3/issues/22661#issuecomment-2723119684

Also I have got new compiler warnings that inner (inside `def`) types `cannot be checked at runtime because it's a local class`, so I moved It to internal object.


fwbrasil:
awesome! thank you

Issue #1009, opened 2025-01-14
Record: staging


fwbrasil:
I'd like to explore using `Record` as a way to build EDSLs. For example, let's say we want to build a SQL module where tables are defined via records:

```scala
type Person = Record["name" ~ String & "age" ~ Int]
val people = Table[Person]
```

The idea is to provide a way to materialize a set of fields with a parametrized function, which would essentially map the field domain to EDSL types:

```scala
trait Column[A]: ...
val personColumns: Record["name" ~ Column[String] & "age" ~ Column[Int]] = 
   Record.stage[Person]([Value] => (field: Field[?, Value]) => Column[Value](field))
```

Each field would still maintain its specific type, preserved via the polymorphic function's `Value` parameter and the `Column` type parameter. With this building block, it'd be easy to build a SQL EDSL by providing methods that handle the staged `Column` type. In Scala 3, it's even possible to override the regular `==`:

```scala
people.where(p => p.name == "John" && p.age < 10).select(_.age)
```

PR #1094, opened 2025-03-17
[kyo-data] records: stage method

road21:
Closes https://github.com/getkyo/kyo/issues/1009


kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -125,6 +126,23 @@ object Record:
       */
     val empty: Record[Any] = Record[Any](Map())
 
+    private def unsafeFrom[Fields: AsFields](map: Map[Field[?, ?], Any]): Record[Fields] = Record(map)
```
fwbrasil, line 129: is this necessary to make type inference work?

kyo-data/shared/src/main/scala/kyo/Record.scala
```diff
@@ -290,15 +318,36 @@ object Record:
                 end match
             }.mkString(" & ")
     end given
+
+    trait StageAs[F[_, _]] extends Inliner[(ForSome2[AsFieldAny], ForSome2[F])]:
+        inline def fieldApply[Name <: String, Value](field: Field[Name, Value]): F[Name, Value]
```
fwbrasil, line 323: how about `stage` instead of `fieldApply`?

kyo-data/shared/src/main/scala/kyo/internal/ForSome.scala
```diff
@@ -0,0 +1,52 @@
+package kyo.internal
+
+/** Existentinal types encoding for type constructors `F[_]`. `ForSome[F]` is semantically equiavalent to `F[?]`
+  */
+type ForSome[F[_]] = ForSome.Type[F]
```
fwbrasil, line 5: Given that existential types were remove due to unsoundness, I wonder if we could have issues with this mechanism. I think it's ok to merge and see how it goes, though

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -283,6 +284,41 @@ class RecordTest extends Test:
         }
     }
 
+    "stage" - {
```
fwbrasil, line 287: I'm looking forward to build some DSLs with this! :)

kyo-data/shared/src/main/scala/kyo/internal/ForSome.scala
road21: I have fixed the encoding, now it doesn't use `asInstanceOf`.
However, now there is no guarantee that no extra allocation (of `Unwrap`) can happen.

kyo-data/shared/src/main/scala/kyo/Record.scala
road21: This is because `stage` is `inline` method. When I try to call private constructor of `Record` inside it, I get the error: 
`Implementation restriction: cannot use private constructors in inline methods`

fwbrasil (approved):

PR #1095, opened 2025-03-18
[core] fix safepoint leak in Fiber

fwbrasil:
### Problem

I noticed a few builds failing after https://github.com/getkyo/kyo/pull/1086. This error should only happen if a `Safepoint` from one thread leaks to the execution in another thread via implicits, introducing concurrency issues since the `Trace` management is meant to used by its owning thread only and isn't synchronized:

```
[info] - no buffer *** FAILED *** (6 milliseconds)
[info]   java.lang.NullPointerException: Cannot invoke "kyo.kernel.internal.Trace.frames()" because "newTrace" is null
[info]   at kyo.kernel.internal.Trace$Owner.copyTrace(Trace.scala:60)
[info]   at kyo.Fiber$package$Fiber$$anon$33$$anon$34.loop$8(Fiber.scala:497)
[info]   at kyo.Fiber$package$Fiber$$anon$33$$anon$34.apply(Fiber.scala:661)
[info]   at kyo.Fiber$package$Fiber$$anon$33$$anon$34.apply(Fiber.scala:495)
[info]   at kyo.Async$package$Async$$anon$15.apply(Async.scala:363)
[info]   at kyo.scheduler.IOTask.partialLoop$1(IOTask.scala:42)
[info]   at kyo.scheduler.IOTask.eval(IOTask.scala:66)
[info]   at kyo.scheduler.IOTask.run(IOTask.scala:78)
[info]   at kyo.scheduler.Worker.runTask(Worker.scala:257)
[info]   at kyo.scheduler.Worker.run(Worker.scala:219)
```
https://github.com/getkyo/kyo/actions/runs/13888866130/job/38857394040#step:6:5899

### Solution

I believe the cause of this new issue is that I removed an `IO` suspension that seemed to be unnecessary in `Fiber.foreachIndexed`. The suspension was actually hiding potential safepoint leaks since suspensions automatically restore the correct safepoint. This PR keeps `Fiber.foreachIndexed` without the additional suspension, removes similar suspensions from other methods, and ensures the methods can't use a leaked safepoint by aways getting the instance from the thread local via `Safepoint.get`.

I can't reproduce the issue locally to validate and writing a unit test is challenging without exposing new `Safepoint` APIs but this seems a solid attempt at fixing the issue.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -283,7 +283,7 @@ object Fiber extends FiberPlatformSpecific:
       */
     private[kyo] def race[E, A: Flat, S](
         using isolate: Isolate.Contextual[S, IO]
-    )(iterable: Iterable[A < (Abort[E] & Async & S)])(using frame: Frame, safepoint: Safepoint): Fiber[E, A] < (IO & S) =
+    )(iterable: Iterable[A < (Abort[E] & Async & S)])(using frame: Frame): Fiber[E, A] < (IO & S) =
         IO.Unsafe {
             class State extends IOPromise[E, A] with Function1[Result[E, A], Unit]:
                 val pending = AtomicInt.Unsafe.init(iterable.size)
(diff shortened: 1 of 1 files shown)


fwbrasil:
I'm merging to unblock other changes

PR #1096, opened 2025-03-18
Expand Layer.from overloads 

johnhungerford:
### Problem

I frequently surpass the 4-input limit of `Layer.from` in my applications

### Solution

Add six overloads of `Layer.from` using the same pattern to support up to 10 inputs. This requires adding two overloads to `Kyo.zip` as well.


kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -167,9 +167,252 @@ object Layer:
...
         apply {
-            Kyo.zip(Env.get[A], Env.get[B], Env.get[C], Env.get[D]).map { case (a, b, c, d) => f(a, b, c, d) }
+            Kyo.zip(Env.get[A], Env.get[B], Env.get[C], Env.get[D]).map { case (a, b, c, d) => fn(a, b, c, d) }
+        }
+
+    /** Creates a layer from a function that takes five inputs and produces an effect.
+      *
+      * @param fn
+      *   The function to wrap in a layer
+      * @tparam A
```
hearnadam, line 179: I think Flavio mentioned keeping the files smaller in his recent change to `Async.zip`. Can we remove the type parameter comments?

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -89,7 +89,7 @@ object Layer:
 
     /** Creates a layer from a function that takes one input and produces an effect.
       *
-      * @param f
+      * @param fn
```
hearnadam, line 92: It looks like you forgot to change to `fn` in the actual signature... though I prefer `f`...

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -120,14 +120,14 @@ object Layer:
       * @return
       *   A new layer that requires an environment with A and B and produces C
       */
-    def from[A: Tag, B: Tag, C: Tag, S](f: (A, B) => C < S)(using Frame): Layer[C, Env[A & B] & S] =
+    def from[A: Tag, B: Tag, C: Tag, S](fn: (A, B) => C < S)(using Frame): Layer[C, Env[A & B] & S] =
         apply {
-            Kyo.zip(Env.get[A], Env.get[B]).map { case (a, b) => f(a, b) }
+            Kyo.zip(Env.get[A], Env.get[B]).map { case (a, b) => fn(a, b) }
```
hearnadam, line 125: How would you feel about converting these blocks to use `Env.use`? it should reduce the overhead.

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
fwbrasil: yeah, keeping the complete scaladoc in this kind of scenario feels too heavyweight

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
johnhungerford: I changed to `f` only because subsequent overloads include parameter `f: F`, so there's a name clash. I changed it everywhere for consistency.

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
fwbrasil: We actually don't even need multiple suspensions. I've posted https://github.com/getkyo/kyo/pull/1099 with APIs that allow getting multiple values at once. We can update this code to use the new APIs as a follow up.

hearnadam: great point

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
johnhungerford: reverted to `f` since now that we're using `Env.useAll` we don't need to actually refer to the inputs by name anywhere

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
johnhungerford: I went ahead and migrated to `useAll` now. Figured there's no reason for separate PRs.

johnhungerford:
I went ahead and kept the `Kyo.zip` overloads even though I'm not using them anymore.

hearnadam (approved):

PR #1097, opened 2025-03-18
[kernel] pipe with up to 10 transformations

fwbrasil:
### Problem

I've been using `pipe` more and I've reached a case where I wanted to have more than 6 transformations.

### Solution

Introduce transformations for up to 10 transformations.


hearnadam (approved):

hearnadam:
We might run out of letters for type parameters soon 🙂

fwbrasil:
yeah, I wonder if there's a better way to provide this kind of API with the new type-level tuple transfomations

PR #1098, opened 2025-03-18
[offheap] disable scaladoc generation to fix build

fwbrasil:
### Problem

The `release` workflow in main is failing because scaladoc generation of the `kyo-offheapNative` project fails with a compiler crash.

### Solution

I've tried to fix the issue again without success so, as a workaround, this PR disables scaladoc generation for the module but only for Native. Given that we direct users to the JVM scaladocs, it shouldn't be an issue.

### DIFF
--- a/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
@@ -756,7 +756,7 @@ class AsyncTest extends Test:
                             f(1),
                             f(1)
                         ),
-                        2
+                        3
                     ),
                     f(1)
                 )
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
```diff
@@ -756,7 +756,7 @@ class AsyncTest extends Test:
                             f(1),
                             f(1)
                         ),
-                        2
+                        3
```
fwbrasil, line 759: unrelated flaky test fix. The computations should always run in parallel in order for the memo count call to always be 3.

fwbrasil:
merging to fix the main build

PR #1101, opened 2025-03-19
[core] preserve Async.zip concurrency + support up to 10 computations

fwbrasil:
### Problem

The new `Async.zip` methods don't preserve the concurrency of the number of computations being zipped. I think that's not an ideal behavior since the goal of these APIs is to execute computations concurrently.

### Solution

Ensure the concurrency is preserved by specifying it when calling `collectAll`.

### Notes

I've also added methods for up to 10 computations.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -495,7 +495,7 @@ object Async:
     )(
         using frame: Frame
     ): (A1, A2) < (Abort[E] & Async & S) =
-        collectAll(Seq(v1, v2))(using Flat.unsafe.bypass).map { s =>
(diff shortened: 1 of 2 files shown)


hearnadam (approved):
change LGTM - if we had `collect` and other methods use `BuildFrom`, would it be possible to pass tuples directly?

https://github.com/zio/zio/blob/series/2.x/core/shared/src/main/scala/zio/ZIO.scala#L2922

fwbrasil:
> change LGTM - if we had `collect` and other methods use `BuildFrom`, would it be possible to pass tuples directly?
> 
> https://github.com/zio/zio/blob/series/2.x/core/shared/src/main/scala/zio/ZIO.scala#L2922

No idea, I think not since `BuildFrom` builds a homogeneous collection and tuples have different types for elements.

PR #1102, opened 2025-03-20
Cleanup combinators

johnhungerford:
### Problem

1. Kyo-combinators is getting a bit bloated
2. Retries currently only work for `Abort[Throwable]`
3. Some of the repeat logic can be improved (using `Loop`, e.g.)

### Solution

1. Break apart `Combinators.scala` into separate files according to effect type
2. Integrate retry methods into abort-combinators
   - `retry(s: Schedule)`, `retry(i: Int)`, and `retryForever` for the entire `Abort` union
   - `forAbort[E].retry(s: Schedule)`, `forAbort[E].retry(i: Int)`, and `forAbort[E].retryForever` for a specific error type
   - `retryForever` (new) handles `Abort` completely
3. Using Loop for all repeat logic

### Notes

- Also added `Kyo.sleepForever` which I've found useful in a couple of cases


kyo-combinators/shared/src/main/scala/kyo/AbortCombinators.scala
```diff
@@ -0,0 +1,637 @@
...
+        fr: Frame
+    ): Result[E, A] < S =
+        Abort.run[E](effect)
+
+    /** Handles the Abort effect and returns its result as a `Result.Partial[E, A]`, not handling Panic exceptions.
+      *
+      * @return
+      *   A computation that produces a partial result of this computation with the Abort[E] effect handled
+      */
+    def partialResult(
```
fwbrasil, line 29: if one of the main goals of combinators is to help with auto completion, wouldn't `resultPartial` and `resultPartialOrThrow` work better?

kyo-combinators/shared/src/main/scala/kyo/AbortCombinators.scala
```diff
@@ -0,0 +1,637 @@
...
+        effect.catching(e => fn(e).map(Kyo.fail))
+
+    def forAbort[E1 <: E]: ForAbortOps[A, S, E, E1] = ForAbortOps(effect)
+
+    /** Translates the Abort effect to a Choice effect by handling failures as empty choices.
+      *
+      * @return
+      *   A computation that produces the result of this computation with the Abort[E] effect translated to an empty Choice
+      */
+    def abortToEmpty(
```
fwbrasil, line 73: From the name, I was expecting this to leave a pending `Abort[Empty]`. Maybe a more explicit name like `abortToChoiceDrop`?

kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
```diff
@@ -0,0 +1,137 @@
...
+        Async.run(effect)
+
+    /** Forks this computation using the Async effect and returns its result as a `Fiber[E, A]`, managed by the Resource effect. Unlike
+      * `fork`, which creates an unmanaged fiber, `forkScoped` ensures that the fiber is properly cleaned up when the enclosing scope is
+      * closed, preventing resource leaks.
+      *
+      * @return
+      *   A computation that produces the result of this computation with Async and Resource effects
+      */
+    inline def forkScoped(
```
fwbrasil, line 30: We should have a similar method in `Async`

kyo-combinators/shared/src/main/scala/kyo/EmitCombinators.scala
```diff
@@ -0,0 +1,180 @@
...
+    )(
+        using
+        tag: Tag[Emit[A]],
+        fl: Flat[B],
+        f: Frame
+    ): B < (S & S1) =
+        ArrowEffect.handle(tag, effect):
+            [C] =>
+                (a, cont) =>
+                    fn(a).map(_ => cont(()))
```
fwbrasil, line 53: `andThen` would work better here

kyo-combinators/shared/src/main/scala/kyo/EmitCombinators.scala
```diff
@@ -0,0 +1,180 @@
...
+
+    /** Handle Emit[A] by passing emitted values to [[channel]]. Fails with Abort[Closed] on channel closure
+      *
+      * @param channel
+      *   Channel in which to put emitted values
+      * @return
+      *   Asynchronous computation with handled Emit that can fail with Abort[Closed]
+      */
+    def emitToChannel(channel: Channel[A])(using Tag[A], Flat[B], Frame): B < (S & Async & Abort[Closed]) =
+        effect.foreachEmit(a => channel.put(a))
```
fwbrasil, line 63: this could use a batched channel put

kyo-combinators/shared/src/main/scala/kyo/KyoCombinators.scala
```diff
@@ -0,0 +1,245 @@
+package kyo
+
+import kyo.debug.Debug
+import kyo.kernel.ArrowEffect
+import scala.annotation.tailrec
+import scala.annotation.targetName
+import scala.util.NotGiven
+
+sealed case class PanicException[A] private[kyo] (val error: A) extends Exception(s"Uncaught error: $error")
```
fwbrasil, line 9: Could you migrate to `KyoException`?

kyo-combinators/shared/src/main/scala/kyo/KyoCombinators.scala
```diff
@@ -0,0 +1,245 @@
...
+    /** Performs this computation and then the next one, discarding the result of this computation.
+      *
+      * @param next
+      *   The computation to perform after this one
+      * @return
+      *   A computation that produces the result of `next`
+      */
+    @targetName("zipRight")
+    def *>[A1, S1](next: => A1 < S1)(using Frame): A1 < (S & S1) =
+        effect.map(_ => next)
```
fwbrasil, line 22: `andThen`

kyo-combinators/shared/src/main/scala/kyo/KyoCombinators.scala
```diff
@@ -0,0 +1,245 @@
...
+      *   A computation that produces the result of this computation
+      */
+    def debugValue(using Frame): A < S = Debug(effect)
+
+    /** Performs this computation and prints its result to the console with a detailed execution trace.
+      *
+      * @return
+      *   A computation that produces the result of this computation
+      */
+    def debugTrace(using Frame): A < S = Debug.trace(effect)
```
fwbrasil, line 58: cool! :)

kyo-combinators/shared/src/main/scala/kyo/KyoCombinators.scala
```diff
@@ -0,0 +1,245 @@
...
+    /** Performs this computation repeatedly with a backoff policy and a limit.
+      *
+      * @param backoff
+      *   The backoff policy to use
+      * @param limit
+      *   The limit to use
+      * @return
+      *   A computation that produces the result of this computation with Async effect
+      */
+    def repeat(backoff: Int => Duration, limit: Int)(using Flat[A], Frame): A < (S & Async) =
```
fwbrasil, line 112: we have similar methods in `Clock.repeat*` now. It'd be nice to follow the same semantics and naming.

kyo-combinators/shared/src/main/scala/kyo/AbortCombinators.scala
johnhungerford: yes -- and in fact now that you mention it, I had trouble finding this method recently. I'll update.

kyo-combinators/shared/src/main/scala/kyo/AbortCombinators.scala
johnhungerford: Yeah absent vs empty is confusing. I'm not sure about `...ChoiceDrop`, but I can't think of anything better at the moment.

kyo-combinators/shared/src/main/scala/kyo/KyoCombinators.scala
johnhungerford: I'll take a look

kyo-combinators/shared/src/main/scala/kyo/AbortCombinators.scala
```diff
@@ -0,0 +1,637 @@
...
+            case Result.Failure(err)            => Abort.fail(PanicException(err))
+            case p: Result.Panic                => Abort.get(p)
+        }
+
+    /** Handles the Abort effect and applies a recovery function to the error.
+      *
+      * @return
+      *   A computation that produces the result of this computation with the Abort[E] effect handled
+      */
+    def catching[A1 >: A, S1](fn: E => A1 < S1)(
```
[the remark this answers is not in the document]
johnhungerford: Yeah good point -- I'll use recover

kyo-combinators/shared/src/main/scala/kyo/KyoCombinators.scala
```diff
@@ -0,0 +1,245 @@
...
+
+    /** Performs this computation unless the given condition holds, in which case it returns an Abort[Absent] effect.
+      *
+      * @param condition
+      *   The condition to check
+      * @return
+      *   A computation that produces the result of this computation with Abort[Absent] effect
+      */
+    def unless[S1](condition: Boolean < S1)(using Frame): A < (S & S1 & Abort[Absent]) =
+        condition.map(c => if c then Abort.fail(Absent) else effect)
```
hearnadam, line 233: Should this signature be the same as `when`? Why split between Abort[Absent] and Maybe[A]?

kyo-combinators/shared/src/main/scala/kyo/KyoCombinators.scala
```diff
@@ -0,0 +1,245 @@
...
+    /** Ensures that the specified finalizer is executed after this effect, whether it succeeds or fails. The finalizer will execute when
+      * the Resource effect is handled.
+      *
+      * @param finalizer
+      *   The finalizer to execute after this effect
+      * @return
+      *   An effect that executes the finalizer after this effect
+      */
+    def ensuring(finalizer: => Any < Async)(using Frame): A < (S & Resource & IO) =
+        Resource.ensure(finalizer).andThen(effect)
```
hearnadam, line 244: I wonder if `ensure` should have an `ensureWith` as well?

kyo-combinators/shared/src/main/scala/kyo/KyoCombinators.scala
```diff
@@ -0,0 +1,245 @@
...
+    @targetName("zip")
+    def <*>[A1, S1](next: => A1 < S1)(using Frame): (A, A1) < (S & S1) =
+        effect.map(e => next.map(n => (e, n)))
+
+    /** Performs this computation and prints its result to the console.
+      *
+      * @return
+      *   A computation that produces the result of this computation
+      */
+    def debugValue(using Frame): A < S = Debug(effect)
```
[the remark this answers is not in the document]
johnhungerford: That creates a name class with the `debug` package, which I'd like to avoid.

kyo-combinators/shared/src/main/scala/kyo/Constructors.scala
```diff
@@ -394,6 +384,14 @@ extension (kyoObject: Kyo.type)
     def sleep(duration: Duration)(using Frame): Unit < Async =
         Async.sleep(duration)
 
+    /** Sleeps for all eternity.
+      *
+      * @return
+      *   An effect that sleeps forever and accomplishes nothing
+      */
+    def sleepForever(using Frame): Nothing < Async =
+        sleep(Duration.Infinity).forever
```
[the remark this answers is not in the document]
johnhungerford: Ah didn't see that. Thanks

kyo-combinators/shared/src/main/scala/kyo/EmitCombinators.scala
```diff
@@ -0,0 +1,180 @@
...
+import scala.annotation.targetName
+import scala.util.NotGiven
+
+extension [A, S](effect: Unit < (Emit[Chunk[A]] & S))
+    /** Convert streaming Emit effect to a [[Stream[A, S]]]
+      *
+      * @return
+      *   Stream representation of original effect
+      */
+    def emitToStream: Stream[A, S] = Stream(effect)
```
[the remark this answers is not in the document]
fwbrasil: `Stream.apply` is inlined and takes a by-name function. That'd require a separate group of extension methods where `effect` is also declared as `inline`. I'm not sure it's worth it, especially given that "someValue.doSomething" typically implies the value being evaluated before the method call. We have `pipe` that breaks this assumption but I think it's more justifiable.

kyo-combinators/shared/src/main/scala/kyo/KyoCombinators.scala
fwbrasil: I've added a similar method in Abort in the actors PR: https://github.com/getkyo/kyo/pull/1107/files#diff-64f89377b66b892a1b92f9dc391454bd378df56a20952fff17ae30b3a6079784R81

kyo-combinators/shared/src/main/scala/kyo/EmitCombinators.scala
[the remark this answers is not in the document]
hearnadam: Ah you're correct. Extension methods do break this assumption.

kyo-combinators/shared/src/main/scala/kyo/KyoCombinators.scala
johnhungerford: I'll add an issue for this

johnhungerford:
I think latest updates address all PR review comments

PR #1104, opened 2025-03-21
[core] improve scaladocs

fwbrasil:


kyo-core/shared/src/main/scala/kyo/Atomic.scala
```diff
@@ -1,8 +1,15 @@
 package kyo
 
-import java.util.concurrent.atomic as j
```
fwbrasil, line 3: I don't know why Metals started complaining about this import after I added the scaladocs. I've used the full name instead.

fwbrasil:
merging to avoid conflicts

PR #1107, opened 2025-03-24
[actor] Introduce module

fwbrasil:
### Problem

Dealing with effects that can't be trivially fork like local `Var`s isn't trivial and, in some cases, `Isolate`s can't even be provided due to the specific semantics of effects.

### Solution

Actors provide a straightforward way to isolate a computation with arbitrary effects. Elements can be serially consumed via `Poll` while keeping the complete state of the computation, including its effect handlers. 

`Stream` provides a similar solution via `fold` but the execution model isn't very flexible. Unlike streaming, `Actor` references can be freely passed around and receive messages in multiple code paths. The implementation also provides supervision hierarchy, which is a more convenient way to handle resources.

### Notes

This is based on the collaboration with @DamianReeves for Functional Scala last year. Thank you!


kyo-prelude/shared/src/main/scala/kyo/Poll.scala
```diff
@@ -177,16 +177,16 @@ object Poll:
             using
             tag: Tag[Poll[V]],
             frame: Frame
-        ): (Maybe[V] => A < (Poll[V] & S)) < S =
+        ): Either[A, Maybe[V] => A < (Poll[V] & S)] < S =
```
fwbrasil, line 180: The API was hiding the information if the poll had finished or not

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -0,0 +1,354 @@
...
+
+    /** Default mailbox capacity for actors.
+      *
+      * This value can be configured through the system property "kyo.actor.capacity.default". If not specified, it defaults to 100
+      * messages.
+      */
+    val defaultCapacity =
+        import AllowUnsafe.embrace.danger
+        given Frame = Frame.internal
+        IO.Unsafe.evalOrThrow(System.property[Int]("kyo.actor.capacity.default", 100))
```
hearnadam, line 104: How about a power of 2?

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -0,0 +1,354 @@
...
+    )(capacity: Int)(behavior: B < (Context[A] & Abort[E] & S))(
+        using
+        Tag[Poll[A]],
+        Tag[Emit[A]],
+        Frame
+    ): Actor[E, A, B] < (Resource & Async & S) =
+        for
+            mailbox <-
+                // Create an unbounded channel to serve as the actor's mailbox
+                // Unbounded capacity in this initial prototype
```
hearnadam, line 329: Is this still true?

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -0,0 +1,354 @@
...
+                        case Right(cont) =>
+                            mailbox.take.map(v => Loop.continue(cont(Maybe(v))))
+                    }
+                }.pipe(
+                    IO.ensure(mailbox.close), // Ensure mailbox cleanup by closing it when the actor completes or fails
+                    Env.run(_subject),        // Provide the actor's Subject to the environment so it can be accessed via Actor.self
+                    Resource.run,             // Close used resources
+                    Async.run                 // Start the actor's processing loop in an async context
+                )
+            _ <- Resource.ensure(mailbox.close)
```
hearnadam, line 349: Double ensured with line 344?

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -0,0 +1,354 @@
...
+  * `result`.
+  *
+  * @tparam E
+  *   The type of errors that can terminate the actor if not handled internally
+  * @tparam A
+  *   The type of messages this actor can receive
+  * @tparam B
+  *   The type of result this actor produces upon completion
+  */
+abstract class Actor[+E, A, B]:
```
hearnadam, line 47: sealed?

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -0,0 +1,354 @@
...
+
+    /** Retrieves the final result of this actor.
+      *
+      * Waits for the actor to complete processing all messages and return its final value. Will fail with error E if an unhandled error
+      * occurs during message processing, or with Closed if the actor is closed prematurely.
+      *
+      * @return
+      *   The actor's final result of type B
+      */
+    def result(using Frame): B < (Async & Abort[Closed | E]) = fiber.get
```
hearnadam, line 76: Maybe this should be await?

kyo-actor/shared/src/main/scala/kyo/Actor.scala
fwbrasil: no, I've removed it

kyo-actor/shared/src/main/scala/kyo/Actor.scala
fwbrasil: The one in line 344 closes the mailbox when the actor itself finishes. This other one attaches a finalizer to the outer scope, which provides the actor hierarchy behavior

kyo-actor/shared/src/main/scala/kyo/Subject.scala
```diff
@@ -0,0 +1,169 @@
...
+      * @tparam E
+      *   The error type of the Promise
+      * @tparam A
+      *   The type of messages this Subject can receive
+      * @return
+      *   A Subject[A] that completes the Promise with the first received message and aborts with Closed for subsequent messages
+      */
+    def init[E, A](promise: Promise[E, A])(using frame: Frame): Subject[A] =
+        def tryComplete(r: A) =
+            Abort.whenNot(promise.complete(Result.succeed(r)))(Closed("Subject", frame))
```
fwbrasil, line 115: I've updated `Abort.when` to allow effects in any of its parameters. Not allowing effects in the condition makes the API not very useful. I've also added `Abort.whenNot`.

kyo-actor/shared/src/main/scala/kyo/Subject.scala
```diff
@@ -0,0 +1,187 @@
...
+      * Queue's `add` method, while `trySend` also uses `add` and always returns true since unbounded queues never reject messages.
+      *
+      * @param queue
+      *   The Unbounded Queue to add messages into
+      * @tparam A
+      *   The type of messages this Subject can receive
+      * @return
+      *   A Subject[A] that adds received messages into the Unbounded Queue
+      */
+    def init[A](queue: Queue.Unbounded[A]): Subject[A] =
```
hearnadam, line 152: Is there a reason this is Unbounded only? I assume because the add can't fail?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -78,8 +78,20 @@ object Abort:
...
+    /** Fails the computation if the condition is false.
+      *
+      * @param b
+      *   The condition to check
+      * @param value
+      *   The failure value to use if the condition is false
+      * @return
+      *   A unit computation that may fail if the condition is false
+      */
+    inline def whenNot[E, S](b: Boolean < S)(inline value: => E < S)(using inline frame: Frame): Unit < (Abort[E] & S) =
```
hearnadam, line 93: How about calling this `unless`?

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -0,0 +1,320 @@
...
+    )(capacity: Int)(behavior: B < (Context[A] & Abort[E] & S))(
+        using
+        Tag[Poll[A]],
+        Tag[Emit[A]],
+        Frame
+    ): Actor[E, A, B] < (Resource & Async & S) =
+        for
+            mailbox <-
+                // Create a bounded channel to serve as the actor's mailbox
+                Channel.init[A](Int.MaxValue, Access.MultiProducerSingleConsumer)
```
hearnadam, line 296:
```suggestion
                Channel.init[A](capacity, Access.MultiProducerSingleConsumer)
```

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -0,0 +1,320 @@
...
+sealed abstract class Actor[+E, A, B]:
+
+    /** Returns the message subject interface for sending messages to this actor.
+      *
+      * Messages sent through this subject will be queued in the actor's mailbox and processed sequentially in FIFO order.
+      *
+      * @return
+      *   A Subject[A] that can be used to send messages to this actor
+      */
+    val subject: Subject[A]
```
hearnadam, line 56: Any reason these are `val` in the class definition?

kyo-actor/shared/src/main/scala/kyo/Subject.scala
fwbrasil: Yes, the regular queue only has `offer`.

kyo-actor/shared/src/main/scala/kyo/Actor.scala
fwbrasil: It was a lazy way to allow `export subject.*` so the methods are available in the actor instance directly. I've changed it to use constructor params + avoid `val`.

fwbrasil:
@hearnadam can we merge this?

Issue #987, opened 2025-01-09
Channel implementation with zero capacity


johnhungerford:
Streaming from channels currently uses `drainUpTo` on the underlying `Queue` to grab chunks from the channel. Since there is no queue for zero-capacity channels, they will necessarily stream one element at a time (that is, chunks of size 1 -- essentially a loop over `take`), even though there may be multiple pending puts that could be drained.

It probably makes sense to provide a special implementation for channels of capacity zero that has no main queue and simply passes values directly from the pending `puts` to the pending `takes`. `drain` and `drainUpTo` could be specialized to drain the puts. Since there is no queue for these to pass to, there is no issue with ordering to worry about (which is why regular channels only drain the main queue and not the `puts`).

johnhungerford:
@fwbrasil @hearnadam Question for this: should the synchronous `Channel` methods like `offer` and `poll` be no-op, since there is no intermediate queue, or should they diverge from standard `Channel` implementation in interacting directly with the `puts` queue?

Approach 1: `Channel#offer` always returns `false` and `Channel.poll` always returns `Absent`

Approach 2: `Channel#offer` adds a `Put.Value` to the `puts` queue, with an unused `Promise`. `Channel#poll` polls the `puts` queue, completes the promise of any `Put` that it takes, and returns the value.

PR #1108, opened 2025-03-24
Zero-capacity channel

johnhungerford:
Addresses #987 

### Problem

Zero-capacity channels did not function properly since the default implementation expected to be able to offer to and pull from the intermediate queue directly.

### Solution

Updated `Channel.Unsafe.init` to return a separate implementation of `Channel.Unsafe` when `_capacity == 0`. This new implementation has no `queue`. It's behavior is subtly different in the following ways:
1. Channel closure is tracked by an atomic boolean instead of being tied to the underlying queue (which doesn't exist)
2. Offer and poll interacts with the `puts` and `takes` queues directly
3. flush() handles only two cases: (a) channel is closed, and (b) `puts` and `takes` are both non-empty

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -2,6 +2,7 @@ package kyo
 
 import org.jctools.queues.MpmcUnboundedXaddArrayQueue
 import scala.annotation.tailrec
+import scala.collection.mutable.Builder
 import scala.util.NotGiven
 
 /** A channel for communicating between fibers.
@@ -75,14 +76,19 @@ object Channel:
           */
         def putBatch(values: Seq[A])(using Frame): Unit < (Abort[Closed] & Async) =
             if values.isEmpty then ()
+            else if self.capacity == 0 then
+                IO.Unsafe(self.putBatchFiber(values).safe.get)
             else
                 IO.Unsafe {
                     self.offerAll(values) match
                         case Result.Success(remaining) =>
                             if remaining.isEmpty then ()
-                            else self.putBatchFiber(remaining).safe.get
+                            else
+                                self.putBatchFiber(remaining).safe.get
                         case err @ Result.Error(_) => Abort.get(err).unit
                 }
+            end if
+        end putBatch
 
         /** Takes an element from the channel, asynchronously blocking if necessary.
           *
@@ -99,6 +105,7 @@ object Channel:
                     Abort.error
                 )
             }
+        end take
 
         /** Takes [[n]] elements from the channel, semantically blocking until enough elements are present. Note that if enough elements are
           * not added to the channel it can block indefinitely.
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -296,224 +303,400 @@ object Channel:
             _capacity: Int,
             access: Access = Access.MultiProducerMultiConsumer
         )(using initFrame: Frame, allow: AllowUnsafe): Unsafe[A] =
+            if _capacity <= 0 then
+                new Unsafe[A]:
```
hearnadam, line 307: Similar to `Meter`, it may be useful to have a `Base` implementation sharing some functionality.

I don't have much time to follow the algorithm right now, so not sure how much can be shared.

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -296,224 +303,400 @@ object Channel:
...
-                def offerAll(values: Seq[A])(using AllowUnsafe): Result[Closed, Chunk[A]] =
-                    @tailrec
-                    def loop(current: Chunk[A], offered: Boolean = false): Result[Closed, Chunk[A]] =
-                        if current.isEmpty then
-                            if offered then flush()
-                            Result.Success(Chunk.empty)
+                    def offerAll(values: Seq[A])(using AllowUnsafe): Result[Closed, Chunk[A]] = result(Chunk.from(values))
+
+                    def poll()(using AllowUnsafe) =
+                        result:
```
hearnadam, line 324: Can you use curlys here? it seems more readable.

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -296,224 +303,400 @@ object Channel:
...
-                                    Result.succeed(current)
-                                case result =>
-                                    if offered then flush()
-                                    result.map(_ => current)
+                            flush()
+                            Present(Chunk.empty)
+                    end close
+
+                    def empty()(using AllowUnsafe)  = result(true)
+                    def full()(using AllowUnsafe)   = result(true)
```
hearnadam, line 424: 🫨

kyo-core/shared/src/main/scala/kyo/Channel.scala
hearnadam, line 307: Perhaps it's best to have a named class to improve stack traces/profiles. It might be useful to distinguish between Queues with capacity. 

`ZeroCapacity`?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -75,14 +76,19 @@ object Channel:
...
+                IO.Unsafe(self.putBatchFiber(values).safe.get)
             else
                 IO.Unsafe {
                     self.offerAll(values) match
                         case Result.Success(remaining) =>
                             if remaining.isEmpty then ()
-                            else self.putBatchFiber(remaining).safe.get
+                            else
+                                self.putBatchFiber(remaining).safe.get
                         case err @ Result.Error(_) => Abort.get(err).unit
```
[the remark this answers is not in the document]
fwbrasil: it seems required but it could also be `Abort.get(err.unit)`, which avoids an allocation in case of a failure

kyo-core/shared/src/main/scala/kyo/Channel.scala
fwbrasil: Yeah, this was confusing to me. Can we rename to `asResult` as well?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -296,224 +303,400 @@ object Channel:
...
 
-                def size()(using AllowUnsafe) = queue.size()
+                    def size()(using AllowUnsafe) = result(0)
 
-                def offer(value: A)(using AllowUnsafe) =
-                    val result = queue.offer(value)
-                    if result.contains(true) then flush()
-                    result
-                end offer
+                    def offer(value: A)(using AllowUnsafe) = result(false)
```
fwbrasil, line 319: I think this behavior isn't ideal because there can be a pending take that could be satisfied. It seems we could just attempt to get a pending take from the queue?

kyo-core/shared/src/main/scala/kyo/Channel.scala
fwbrasil: It seems some of the implementations could be shared. It'd be nice to avoid duplication where possible

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: That's correct though, isn't it?

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: Ok I added this update (along with others -- see resolved conversations). This is now breaking a bunch of tests. Do we want to revert or change tests to expect different behavior for the zero-capacity case?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -286,234 +293,418 @@ object Channel:
...
+                            discard(promise.complete(Result.unit))
+                            flush()
+                            Present(value)
+                        case Present(Put.Batch(batch, promise)) =>
+                            val result = batch.headMaybe match
+                                case Absent =>
+                                    discard(promise.complete(Result.unit))
+                                    Absent
+                                case Present(value) =>
+                                    if batch.tail.nonEmpty then discard(puts.offer(Put.Batch(batch.tail, promise)))
```
fwbrasil, line 384: This can break fairness since the batched put goes back to the end of the queue. It's pre-existing so we could follow up separately

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -286,234 +293,418 @@ object Channel:
...
+            def takeFiber()(using AllowUnsafe): Fiber.Unsafe[Closed, A] =
+                val promise = Promise.Unsafe.init[Closed, A]()
+                takes.add(promise)
+                flush()
+                promise
+            end takeFiber
+        end BaseUnsafe
+
+        final class ZeroCapacityUnsafe[A](using initFrame: Frame, allow: AllowUnsafe) extends BaseUnsafe[A]:
+            val isClosed = new java.util.concurrent.atomic.AtomicBoolean()
```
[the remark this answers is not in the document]
johnhungerford: Sure -- didn't know that was a thing!

kyo-core/shared/src/main/scala/kyo/Channel.scala
hearnadam: I think it is, yes. Just a bit ironic, as I generally would expect empty != full

kyo-core/shared/src/main/scala/kyo/Channel.scala
fwbrasil: Could you elaborate on the behavior changes that break the tests?

johnhungerford: Current implementation of `ZeroCapacityChannel#close` always returns an empty backlog, since there's no `queue`, but since it's `offer` behavior is different, the accounting that's used in tests for comparing the backlog to what's been `offer`ed doesn't translate.

For instance, the test "offer and close" `offer`s a bunch of values and expects to get everything that was successfully offered back when calling `close()`. A zero-capacity channel can successfully offer them, but returns none of them on `close`.

Basically any test that asserts on backlogs from `close()` either fail or are flaky. Some `Hub` tests are failing now too.

fwbrasil: It seems just a test issue? The behavior sounds ok to me

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -286,234 +293,418 @@ object Channel:
...
+
+            def putBatchFiber(values: Seq[A])(using AllowUnsafe): Fiber.Unsafe[Closed, Unit] =
+                val promise = Promise.Unsafe.init[Closed, Unit]()
+                val put     = Put.Batch(Chunk.from(values), promise)
+                puts.add(put)
+                flush()
+                promise
+            end putBatchFiber
+
+            def takeFiber()(using AllowUnsafe): Fiber.Unsafe[Closed, A] =
```
hearnadam, line 324:
```suggestion
            final def takeFiber()(using AllowUnsafe): Fiber.Unsafe[Closed, A] =
```

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -286,234 +293,418 @@ object Channel:
...
 
                                 loop(0)
+                        end match
+                    }
+                    flush()
+                end if
+            end flush
+        end NonZeroCapacityUnsafe
+
+        def init[A](
```
hearnadam, line 702: Can you move this above the impls? Then users can more easily see the the function exists?

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: Made backlog assertions conditional on channel size.

johnhungerford:
I think the latest updates address all PR review comments

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -684,7 +698,7 @@ class ChannelTest extends Test:
         "putBatch and take" in run {
             (for
                 size    <- Choice.get(Seq(0, 1, 2, 10, 100))
-                channel <- Channel.init[Int](size)
+                channel <- Channel.init[Int](0)
```
fwbrasil, line 701: is this intended?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -2,6 +2,7 @@ package kyo
 
 import org.jctools.queues.MpmcUnboundedXaddArrayQueue
 import scala.annotation.tailrec
+import scala.collection.mutable.Builder
```
fwbrasil, line 5: seems unused

fwbrasil (approved):
LGTM besides the couple of comments

PR #1110, opened 2025-03-25
[core] Async.fillIndexed

fwbrasil:
### Problem

`Async.foreach(0 until n)(idx => ...)` is a common pattern to repeat a computation with an index.

### Solution

Add `Async.fillIndexed`.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -485,7 +485,28 @@ object Async:
     )(n: Int, concurrency: Int = defaultConcurrency)(
         f: => A < (Abort[E] & Async & S)
     )(using Frame): Chunk[A] < (Abort[E] & Async & S) =
-        foreach(0 until n, concurrency)(_ => f)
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -485,7 +485,28 @@ object Async:
...
+      * @param n
+      *   Number of times to repeat the computation
+      * @param concurrency
+      *   Maximum number of concurrent computations (defaults to [[defaultConcurrency]])
+      * @param f
+      *   Function that takes the current index (0 to n-1) and returns a computation
+      * @return
+      *   Chunk containing results of all iterations in index order
+      */
+    def fillIndexed[E, A: Flat, S](
```
hearnadam, line 504: How about `tabulate`?

fwbrasil: hmm I think the pattern with the `Indexed` suffix is more intuitive since it appears in other APIs

hearnadam (approved):

PR #1113, opened 2025-03-27
[playwright] introduce Browser integration

fwbrasil:
### Problem

Kyo doesn't offer an effect to interact with browsers, which is a common feature in AI solutions these days.

### Solution

Introduce an integration with [Playwright](https://playwright.dev/) that provides programmatic browser access. The effect has two APIs: the regular methods in the `Browser` companion and the `Op` classes, which are convenient for LLM access via tools.


build.sbt
```diff
@@ -120,6 +120,7 @@ lazy val kyoJVM = project
         `kyo-zio`.jvm,
         `kyo-cats`.jvm,
         `kyo-combinators`.jvm,
+        `kyo-playwright`.jvm,
```
fwbrasil, line 123: It seems possible to also support ScalaJS via Playwright's JS interfaces

kyo-playwright/shared/src/main/scala/kyo/Browser.scala
```diff
@@ -0,0 +1,962 @@
...
+
+        case class Count(selector: String)
+            extends Op(page => page.locator(selector).count())
+
+        // Content
+
+        case object GetStatus
+            extends Op(page => Status(page.url(), page.title()))
+
+        case object ReadableContent
```
fwbrasil, line 837: This op uses two JS libraries to extract the readable content and then transform it into markdown.

kyo-scheduler-pekko/jvm/src/test/scala/kyo/scheduler/KyoExecutorServiceConfiguratorTest.scala
```diff
@@ -58,7 +58,7 @@ class KyoExecutorServiceConfiguratorTest
         val threadNames = Await.result(Future.sequence(futures), 5.seconds)
 
         assert(threadNames.forall(_.contains("kyo")))
-        assert(threadNames.toSet.size > 1)
+        assert(threadNames.toSet.size >= 1)
```
fwbrasil, line 61: unrelated flaky test fix

kyo-playwright/shared/src/main/scala/kyo/Image.scala
```diff
@@ -0,0 +1,130 @@
...
+      */
+    def writeFileBase64(path: Path)(using Frame): Unit < IO =
+        path.write(base64)
+
+    /** Converts the image data to an immutable array of bytes.
+      *
+      * @return
+      *   The image data as an IArray[Byte]
+      */
+    def binary: IArray[Byte] = IArray(data*)
```
hearnadam, line 61:
```suggestion
    def binary: IArray[Byte] = IArray.unsafeFromArray(data)
```
https://www.scala-lang.org/api/current/scala/IArray$.html#unsafeFromArray-957

kyo-playwright/shared/src/main/scala/kyo/Image.scala
```diff
@@ -0,0 +1,130 @@
...
+      * @param charsHeight
+      *   Optional height in character cells (default: 0, uses terminal default)
+      * @return
+      *   A Maybe containing the terminal control sequence if rendering is supported, or Absent if the terminal doesn't support inline
+      *   images
+      */
+    def renderToConsole(charsWidth: Int = 0, charsHeight: Int = 0)(using Frame): Maybe[String] =
+        val base64Image = base64
+        val termProgram = sys.env.getOrElse("TERM_PROGRAM", "").toLowerCase
+        val term        = sys.env.getOrElse("TERM", "").toLowerCase
```
hearnadam, line 85: How about making a method `renderFor(console: Console, w, h): String`, then it should be trivial to make the 'sys' based variant?

hearnadam (commented):
This looks awesome!

kyo-playwright/shared/src/main/scala/kyo/Browser.scala
fwbrasil: Yes, but it seems the only option. The injection JS disappears once there's a page refresh or navigation. I haven't noticed issues with multiple readablecontent calls as well

kyo-playwright/shared/src/main/scala/kyo/Image.scala
fwbrasil: I've added `Image.ConsoleType`

kyo-playwright/shared/src/main/scala/kyo/Image.scala
```diff
@@ -0,0 +1,147 @@
...
+      *   Optional width in character cells (default: 0, uses terminal default)
+      * @param charsHeight
+      *   Optional height in character cells (default: 0, uses terminal default)
+      * @return
+      *   A Maybe containing the terminal control sequence if rendering is supported, or Absent if the terminal doesn't support inline
+      *   images
+      */
+    def renderToConsole(charsWidth: Int = 0, charsHeight: Int = 0, consoleType: Maybe[ConsoleType] = ConsoleType.get)(
+        using Frame
+    ): Maybe[String] =
```
hearnadam, line 85: If this accepted a `ConsoleType` directly, it could return a `String`, then you can just `map` over the `ConsoleType.get`

```
def renderToConsole(charsWidth: Int, charsHeight: Int, consoleType: ConsoleType): String
def renderToConsole(charsWidth: Int = 0, charsHeight: Int = 0): Maybe[String] =
  ConsoleType.get.map(rendertoConsole(charsWidth, charsHeight))
```

kyo-playwright/shared/src/main/scala/kyo/Browser.scala
```diff
@@ -0,0 +1,964 @@
...
+      *
+      * @param selector
+      *   The selector to find the element
+      * @param clickCount
+      *   Number of clicks (default: 1)
+      * @return
+      *   Unit
+      */
+    def click(selector: String, clickCount: Int = 1)(using Frame): Unit < Browser =
+        Click(selector, true, clickCount).suspend.unit
```
hearnadam, line 141: I think we can use `suspendWith` throughout, but otherwise lgtm.

kyo-playwright/shared/src/main/scala/kyo/Browser.scala
```diff
@@ -0,0 +1,962 @@
...
+      */
+    def goto(url: String)(using Frame): Unit < (Browser & Abort[Nothing]) =
+        Goto(url).suspend.map(handleResponse)
+
+    /** Navigates back in the browser history.
+      *
+      * @return
+      *   Unit
+      */
+    def back(using Frame): Unit < (Browser & Abort[Nothing]) =
```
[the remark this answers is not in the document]
hearnadam: Sounds good!

kyo-playwright/shared/src/main/scala/kyo/Browser.scala
fwbrasil: makes sense. I've also added `suspendDiscard` since it's a common pattern in the code

PR #1114, opened 2025-03-28
remove Ops classes

fwbrasil:
### Problem

We used to use `Ops` classes to encode methods with multiple type parameter groups. The latest versions of the compiler include the new clause interleaving feature that provides a similar mechanism. I was initially reluctant to adopt it because I thought it could make it harder for people reading the codebase to understand the method signatures but there are important drawbacks with the current `Ops` classes:

- The scaladoc generation isn't great because there's the op builder method indirection
- Use of `pipe` with methods using `Ops` classes require an explicit function. For instance, `a.pipe(Abort.run)` doesn't compile but `a.pipe(Abort.run(_))` does.
- The noise of the `Ops` classes makes the code less readable

### Solution

Migrate the all `Ops` classes to clause interleaving. To have some consistency, I used the `Frame` as the implicit to separate type parameter groups since the compiler currently requires parameters between type parameter groups.

### Notes

- There are no changes to the method bodies nor scaladocs.

### DIFF
--- a/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
+++ b/kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
@@ -419,7 +419,7 @@ extension [A, S, E](effect: A < (Abort[E] & S))
         fl2: Flat[B],
         fr: Frame
     ): B < (Abort[Nothing] & S & S1) =
-        Abort.fold(onSuccess, onFail)(effect)
+        Abort.fold[E](onSuccess, onFail)(effect)
 
     /** Recovers from an Abort failure by applying the provided function.
       *
--- a/kyo-core/shared/src/main/scala/kyo/Retry.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Retry.scala
@@ -28,57 +28,39 @@ object Retry:
         Schedule.exponentialBackoff(initial = 100.millis, factor = 2, maxBackoff = 5.seconds)
             .jitter(0.2).take(3)
 
-    /** Provides retry operations for a specific error type. */
-    final class RetryOps[E >: Nothing](dummy: Unit) extends AnyVal:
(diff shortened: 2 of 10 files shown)


kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
```diff
@@ -419,7 +419,7 @@ extension [A, S, E](effect: A < (Abort[E] & S))
         fl2: Flat[B],
         fr: Frame
     ): B < (Abort[Nothing] & S & S1) =
-        Abort.fold(onSuccess, onFail)(effect)
+        Abort.fold[E](onSuccess, onFail)(effect)
```
fwbrasil, line 422: The explicit type parameter here shouldn't be necessary. It seems the issue is the use of `Reducible`. Considering that `Abort[Nothing]` is included in `IO` and derived effects, I think we can remove `Reducible`. I'm planning to follow up on that. Let me know if you have concerns @hearnadam @johnhungerford

johnhungerford: When you say we can remove it, do you mean entirely or just from `Abort`? `Env` still uses `Reducible`, correct?

fwbrasil: Good point! I forgot about `Env`. It'd be only the `Abort`-related ones

hearnadam (approved):

kyo-combinators/shared/src/main/scala/kyo/Combinators.scala
fwbrasil: No luck removing `Reducible` from `Abort`. We need to be able to fully handle `Abort` in the APIs that handle panics anyway

PR #1115, opened 2025-03-29
[kernel] rename `pipe` to `handle`

fwbrasil:
### Problem

The `pipe` method is convenient for effect handling but I think using it for arbitrary transformations isn't very intuitive. 

### Solution

Rename `pipe` to `handle` to more clearly indicate that the method is meant to be used for effect handling. It still supports arbitrary transformations but the scaladocs are also updated to indicate that it's meant primarily for effect handling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ val b: String < IO =
     a.andThen("test")
 ```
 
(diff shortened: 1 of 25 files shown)


hearnadam (approved):

hearnadam:
@fwbrasil looks like you missed some readme usages

PR #1118, opened 2025-03-29
[prelude] Abort with literal types

### LINKED ISSUE #739: Abort with literal strings

hearnadam:
Fixes #739 

### Problem
With `Abort.fail`, manual type annotations are required to write expressions that fail with literal types.

### Solution
Add an object `literal` to `Abort` to enable failing with literal types, namely Strings.

### Notes
The functions will degrade to non-literal behavior in a generic scope.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Abort.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Abort.scala
@@ -487,4 +487,72 @@ object Abort:
             case ex     => Abort.panic(ex)
         }
 
+    /** Provides methods for working with literal error values in Abort effects.
+      *
+      * The literal namespace ensures that error types are preserved as singleton types rather than being widened to their base types. This
+      * is particularly useful when you want to maintain precise type information about specific error values.
+      *
+      * For example:
+      * ```scala
+      * val error1 = Abort.fail("error")         // Nothing < Abort[String]
+      * val error2 = Abort.literal.fail("error") // Nothing < Abort["error"]
+      * ```
+      */
+    object literal:
+        /** Fails the computation with a literal value, preserving its singleton type.
+          *
+          * Unlike [[kyo.Abort.fail]], this method preserves the exact literal type of the failure value.
+          *
+          * @param value
+          *   The literal failure value
+          * @return
+          *   A computation that immediately fails with the given literal value, preserving its exact type
+          */
+        inline def fail[V <: Singleton](inline value: V)(using inline frame: Frame): Nothing < Abort[V] =
+            Abort.fail(value)
+
+        /** Fails the computation if the condition is true, using a literal failure value.
+          *
+          * @param b
+          *   The condition to check
+          * @param value
+          *   The literal failure value to use if the condition is true
+          * @return
+          *   A unit computation that may fail with the literal value if the condition is true
+          */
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
```diff
@@ -1214,4 +1214,70 @@ class AbortTest extends Test:
         }
     }
 
+    "literal" - {
```
fwbrasil, line 1217: can you also test inference? 

```scala
val result = ...
val _: ExpectedType = result
```

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
```diff
@@ -1214,4 +1214,70 @@ class AbortTest extends Test:
...
+            def when(b: Boolean): Unit < Abort["TOO_BIG"] =
+                Abort.literal.when(b)("TOO_BIG")
+
+            val result: Unit < Abort["TOO_BIG"] = when(false)
+            assert(Abort.run(result).eval == Result.unit)
+            assert(Abort.run(when(true)).eval == Result.fail("TOO_BIG"))
+        }
+
+        "effects" in {
+            val result = Env.run(15) {
```
fwbrasil, line 1257: what's the type of `result` here?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -487,4 +487,72 @@ object Abort:
...
+      * The literal namespace ensures that error types are preserved as singleton types rather than being widened to their base types. This
+      * is particularly useful when you want to maintain precise type information about specific error values.
+      *
+      * For example:
+      * ```scala
+      * val error1 = Abort.fail("error")         // Nothing < Abort[String]
+      * val error2 = Abort.literal.fail("error") // Nothing < Abort["error"]
+      * ```
+      */
+    object literal:
```
fwbrasil, line 501: don't we need `run` methods for literals as well?

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
hearnadam: Good catch. `Abort.run` wasn't preserving the correct type in a generic context. I added `literal.run`

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
hearnadam: Not in all contexts, but it will help preserve the type information in some cases (mixing effects). Added.

fwbrasil (approved):

PR #1119, opened 2025-03-29
merge contextual and stateful

hearnadam:
@fwbrasil WIP

### DIFF
--- a/kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
@@ -68,14 +68,26 @@ import scala.quoted.*
   */
 object Isolate:
 
-    /** Controls effect propagation during forking by copying effect state.
+    /** Controls effect propagation during forking by capturing and copying effect state.
       *
       * @tparam Retain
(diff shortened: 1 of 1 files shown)


kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
```diff
@@ -173,67 +208,37 @@ object Isolate:
       * @tparam Passthrough
       *   Additional effects that will remain pending after isolation
       */
-    abstract class Stateful[Retain, -Passthrough]:
+    abstract class Stateful[Retain, -Passthrough] extends Contextual[Retain, Passthrough]:
```
fwbrasil, line 211: I'm not sure this is valid. A contextual isolate can be used in more constrained scenarios where a stateful isolate can't be used. For example, we can't receive a stateful isolate in `Async.run` since there isn't a way to restore the state at the end. If `Stateful` extends from `Contextual`, I think the restriction wouldn't be properly enforced.

hearnadam: Yeah, I think this is wrong.

Issue #1111, opened 2025-03-25
[scheduler]: Java fascade


hearnadam:
### Problem Statement

Applications using primarily Java cannot easily access the Kyo scheduler. This ticket should be used to scope the work required for a Java application to use `kyo-scheduler` standalone, or with interfaces such as `CompletableFuture`

### Proposed Solution

Rewrite a minimum number of files to Java, limiting the scope of the files touched to those that Java applications will need to use.

PR #1121, opened 2025-03-30
[scheduler] add example of using Scheduler from Java

hearnadam:
### Problem
The Scheduler is currently published as a Scala module, so I suspect users may not realize that it's available for other JVM applications.

#1111 

### Solution
Add an example test in Java to make sure our APIs are always accessible from Java - not depending on Scala specific features.

### DIFF
--- a/kyo-scheduler/jvm/src/test/java/examples/JavaSchedulerTest.java
+++ b/kyo-scheduler/jvm/src/test/java/examples/JavaSchedulerTest.java
@@ -1,0 +1,38 @@
+package examples;
+
+import kyo.scheduler.Scheduler;
+import kyo.scheduler.Task;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class JavaSchedulerTest {
+
+    public static void main(String[] args) throws Exception {
+        forks();
+    }
+
+    private static void forks() throws Exception {
+        Scheduler scheduler = Scheduler.get();
+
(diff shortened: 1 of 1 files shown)


kyo-scheduler/jvm/src/test/java/examples/JavaSchedulerTest.java
```diff
@@ -0,0 +1,38 @@
...
+    private static void forks() throws Exception {
+        Scheduler scheduler = Scheduler.get();
+
+        AtomicReference<String> threadName = new AtomicReference<>("");
+        CountDownLatch latch = new CountDownLatch(1);
+
+        scheduler.schedule(Task.apply(() -> {
+            threadName.set(Thread.currentThread().getName());
+            latch.countDown();
+        }));
```
hearnadam, line 26: Perhaps we should add a `Function1` overload to `Task.apply`. Currently this is lifting to `Runnable`.
We could also add some overloads for task scheduling to the Scheduler itself.

kyo-scheduler/jvm/src/test/java/examples/JavaSchedulerTest.java
```diff
@@ -0,0 +1,38 @@
...
+        forks();
+    }
+
+    private static void forks() throws Exception {
+        Scheduler scheduler = Scheduler.get();
+
+        AtomicReference<String> threadName = new AtomicReference<>("");
+        CountDownLatch latch = new CountDownLatch(1);
+
+        scheduler.schedule(Task.apply(() -> {
```
fwbrasil, line 23: cool! I wasn't expecting the Scala API to work this well from Java

fwbrasil (approved):

PR #1123, opened 2025-04-02
Stream merge

johnhungerford:
Addresses #679 

### Problem

We need to be able to merge concurrent streams. This is probably the last crucial stream feature that's missing. 

### Solution

- My approach is fairly naive: create a channel, run source streams into it in concurrent background fibers, and stream from the channel.
- The trickiest part of this is to signal when the stream should complete. The simplest way to do so is to wrap the streamed elements in `Maybe` and signal an end to the stream by putting `Absent` in the channel. Since there is no `Tag` for `Maybe`, however, I am using `Option` which is not ideal for performance.
- Merging requires `Async`, which is not in `kyo-prelude`, so I also had to work out how to provide an async streaming API without breaking the project structure. To do this, I've added extension methods in `kyo-core` both on the `Stream` companion object (factory methods) and on `Stream[V, S]` instances (combinators).
- The API I've provided allows you to merge two or more streams giving you the option of whether to stop streaming when all streams finish, when any of them finishes, or (when merging only two streams) when the first or second stream finishes. Unlike ZIO, I haven't made `HaltingStrategy` a parameter of any method, since I never used this, but just made different methods for the different behaviors.

### Notes

- Since unions are unsupported, this PR only supports merging streams of the same exact type.
- To help with the implementation, I added `collect` and `collectWhile` methods to `Stream`, which I think are valuable additions to the API independent of their utility to this PR


kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -352,6 +354,88 @@ sealed abstract class Stream[V, -S]:
...
+
+    /** Transform the stream with a partial function, terminating the stream when the first element is encountered for which the partial
+      * function is undefined. Combines the functionality of map and takeWhile.
+      *
+      * @param f
+      *   Partial function transforming V to V2
+      * @return
+      *   A new stream containing transformed elements
+      */
+    def collectWhile[V2, S2](f: PartialFunction[V, V2 < S2])(using
```
HollandDM, line 402: Maybe just me, but `collectWhile`'s signature does look weird.
I think a more mainstream signature would be `def collectWhile[V2, S2](f: V => Maybe[V2] < S2])` and the stream stop when Absent is encountered, wdyt?

johnhungerford: Yeah and we're using `Maybe` elsewhere for collect too, so just for consistently I should change it.

kyo-core/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -0,0 +1,132 @@
...
+      * @note
+      *   Merges chunks and not individual elements. Rechunk source streams to size 1 to interleave individual elements.
+      * @note
+      *   Resulting stream does not preserve chunking from sources
+      * @param other
+      *   Stream to merge with
+      * @param bufferSize
+      *   Size of the buffer that source streams will write to and outputs stream will read from
+      */
+    inline def merge[S2](other: Stream[V, S2], bufferSize: Int = 1024): Stream[V, S & S2 & Async] =
```
hearnadam, line 75: Can we use a constant value for the buffer size?

hearnadam (commented):
Will take a more detailed look tomorrow.

kyo-core/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -0,0 +1,68 @@
+package kyo
```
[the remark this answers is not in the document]
fwbrasil: I think this kind of extension will become more common over time so it'd be important to have more structure. If we put the extensions directly in the package, users that want to be more mindful of imports and avoid `import kyo.*` would have to import each extension method individually. Having the same file name could also be confusing for people browsing the codebase. 

In the reactive streams module we introduced a companion object with the extensions and then exported them to the package so the user can choose to import from the companion. It seems a good direction but it'd be nice to also have a standard for naming. The module seems the most determining factor for the introduction of these extensions but what comes to mind is `StreamCoreExtensions`, which doesn't sound very good to me. Please feel free to share you thoughts on how we could best structure this!

kyo-core/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -0,0 +1,68 @@
...
+
+extension (streamObj: Stream.type)
+    inline def mergeAll[V, S](streams: Seq[Stream[V, S]], bufferSize: Int = 1024)(using Frame): Stream[V, S & Async] =
```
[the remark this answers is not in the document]
fwbrasil: In the new naming for `Kyo` and `Async`, the equivalent would be `collectAll`. It'd be nice if we could follow a similar naming. 

> Would it work to have internal structure of Chunk[Chunk[A]]? Then you can avoid re-allocating all the internal chunks. You can use the regular put?

Good point. The behavior would be more predictable without chunks ever interleaving. Or was the potential interleaving a goal @johnhungerford? I can imagine the current behavior being useful in certain scenarios. I wonder if we're missing something more generic here. For example, in other scenarios time-based chunking could be useful as well.

kyo-core/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -0,0 +1,68 @@
...
+        Stream:
+            Resource.run:
+                for
+                    channel <- Resource.acquireRelease(Channel.init[Option[V]](bufferSize, Access.MultiProducerMultiConsumer))(_.close)
```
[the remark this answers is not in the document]
fwbrasil: I imagine you mean `IO.ensure`? Yeah, `Resource` has more overhead since it allocates a few objects, including a queue. For contained scopes like here, `IO.ensure` is more appropriate.

kyo-core/shared/src/main/scala/kyo/Stream.scala
fwbrasil: I'm not sure I understand. Do you mean not allowing customization? The pattern in recent related PR has been allowing customization via a parameter + a system property for the default (see `Async.defaultConcurrency`, `Actor.defaultCapacity`. By the way, we need a better way to obtain user-configurable default values

kyo-core/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -0,0 +1,132 @@
...
+      * @param streams
+      *   Sequence of streams to be merged
+      * @param bufferSize
+      *   Size of the buffer that source streams will write to and outputs stream will read from
+      */
+    inline def mergeAllHalting[V, S](streams: Seq[Stream[V, S]], bufferSize: Int = 1024)(using Frame): Stream[V, S & Async] =
+        Stream:
+            Resource.run:
+                for
+                    channel <- Resource.acquireRelease(Channel.init[Option[V]](bufferSize, Access.MultiProducerMultiConsumer))(_.close)
```
fwbrasil, line 48: Why not `Maybe[V]`?

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -272,14 +272,16 @@ sealed abstract class Stream[V, -S]:
         discr: Stream.Dummy,
         frame: Frame
     ): Stream[V, S] =
-        Stream[V, S](ArrowEffect.handleState(tag, true, emit)(
```
fwbrasil, line 275: is this an unrelated refactoring? It's ok if it is, just to confirm

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -352,6 +354,85 @@ sealed abstract class Stream[V, -S]:
...
+      *   A new stream containing transformed elements
+      */
+    def collect[V2, S2](f: V => Maybe[V2] < S2)(using
+        tag: Tag[Emit[Chunk[V]]],
+        t2: Tag[Emit[Chunk[V2]]],
+        frame: Frame
+    ): Stream[V2, S & S2] =
+        Stream[V2, S & S2](ArrowEffect.handleState(tag, (), emit)(
+            [C] =>
+                (input, _, cont) =>
```
fwbrasil, line 372: This pattern of using `handleState` without a state so it's possible to suspend other effects during handling is common. I'll follow up on a kernel improvement.

kyo-core/shared/src/main/scala/kyo/Stream.scala
hearnadam: I just meant a `val DefaultBufferSize` or something

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -352,6 +354,85 @@ sealed abstract class Stream[V, -S]:
...
+    def collect[V2](f: V => Maybe[V2])(using
+        tag: Tag[Emit[Chunk[V]]],
+        t2: Tag[Emit[Chunk[V2]]],
+        discr: Stream.Dummy,
+        frame: Frame
+    ): Stream[V2, S] =
+        Stream[V2, S](ArrowEffect.handleState(tag, (), emit)(
+            [C] =>
+                (input, _, cont) =>
+                    val c = input.map(f).collect({ case Present(v) => v })
```
hearnadam, line 386: Can the transformation and collection happen in a single iteration?

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -352,6 +354,85 @@ sealed abstract class Stream[V, -S]:
...
+        tag: Tag[Emit[Chunk[V]]],
+        t2: Tag[Emit[Chunk[V2]]],
+        frame: Frame
+    ): Stream[V2, S & S2] =
+        Stream[V2, S & S2](ArrowEffect.handleState(tag, (), emit)(
+            [C] =>
+                (input, _, cont) =>
+                    Kyo.foreach(input)(f)
+                        .map(_.takeWhile(_.isDefined)
+                            .collect({ case Present(v) => v }))
```
hearnadam, line 409: Perhaps we can followup with a `collectWhile` in Chunk?

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -272,14 +272,16 @@ sealed abstract class Stream[V, -S]:
...
-                    if !state then (false, Kyo.lift[Unit, Emit[Chunk[V]] & S](()))
-                    else if input.isEmpty then (state, cont(()))
+                (input, _, cont) =>
+                    if input.isEmpty then ((), cont(()))
                     else
                         val c = input.takeWhile(f)
-                        Emit.valueWith(c)((c.size == input.size, cont(())))
+                        Emit.valueWith(c):
+                            if c.size == input.size then ((), cont(()))
+                            else ((), Kyo.lift[Unit, Emit[Chunk[V]] & S](()))
```
hearnadam, line 283: How about `Kyo.unit`? Might reduce some complexity throughout.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -352,6 +354,85 @@ sealed abstract class Stream[V, -S]:
...
+        tag: Tag[Emit[Chunk[V]]],
+        t2: Tag[Emit[Chunk[V2]]],
+        discr: Stream.Dummy,
+        frame: Frame
+    ): Stream[V2, S] =
+        Stream[V2, S](ArrowEffect.handleState(tag, (), emit)(
+            [C] =>
+                (input, _, cont) =>
+                    val c = input.map(f).takeWhile(_.isDefined).collect({ case Present(v) => v })
+                    if c.isEmpty && c.size != input.size then ((), Kyo.lift(()))
```
hearnadam, line 428: Does this need to be a separate branch? Do we need to avoid emitting empty chunks? I've noticed this pattern a bit and I wonder if it actually helps anything.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: Oh right, forgot about that

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: Yeah that would be better

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: Yeah I just didn't want to mess with `Chunk` for this PR

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
fwbrasil: Avoiding effect suspensions where possible is a good practice to reduce overhead, it makes sense to me. Even if it currently doesn't happen, it might be the case as the code evolves

hearnadam: While I understand that perspective, I think it's questionable if Streams should modify (empty) chunks where not necessary.

kyo-core/shared/src/main/scala/kyo/Stream.scala
johnhungerford: I can't stream `Maybe` since there is no tag for it

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: Wouldn't this be very likely to make a performance difference in cases where there's a lot of filtered chunks?  What's the downside? Can you provide an example where it would be important or useful to emit empty chunks?

kyo-core/shared/src/main/scala/kyo/Stream.scala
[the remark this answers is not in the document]
johnhungerford: I've put the extensions in `object AsyncStreamExtensions` and exported like in the reactive-streams module.

kyo-core/shared/src/main/scala/kyo/Stream.scala
johnhungerford: Now that I'm passing `Chunk` directly, I'm just using empty chunks to signal end-of-stream (empty chunks are not passed to the `Channel` from source streams)

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: Now that I'm passing `Chunk`s directly, I'm using empty `Chunk` to signal end-of-stream. This won't work if we need to preserve empty chunks from the source streams. If we decide that we need to preserve empty chunks, however, it's possible to use `Maybe` in the `Channel` while not streaming `Maybe`s (just a bit more work).

fwbrasil: I think considering empty chunks as end of stream can be problematic. We should probably avoid emitting empty chunks if possible for performance but it's a weak contract. We could end up having trouble with streams terminating unexpectedly. Regarding `Maybe`, I think it's fine to use `Option` instead given the tag limitation.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: Now that we're passing chunks directly I'm not using `takeWhile` or `collectWhile` anymore, and when I went back to make tests I wasn't able to reproduce the problematic behavior I ran into. I decided to revert to the old implementation.

I'm still keeping `Stream#collectWhile` in this PR, though, because it's useful to have in the API otherwise.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: I was able to implement this more cleanly using `Channel[Maybe[Chunk[V]]]`.

kyo-core/shared/src/main/scala/kyo/AsyncStreamExtensions.scala
```diff
@@ -0,0 +1,198 @@
...
+          * @note
+          *   Merges chunks and not individual elements. Rechunk source streams to size 1 to interleave individual elements.
+          * @note
+          *   Resulting stream does not preserve chunking from sources
+          * @param other
+          *   Stream to merge with
+          * @param bufferSize
+          *   Size of the buffer that source streams will write to and outputs stream will read from
+          */
+        def merge[S2](
```
johnhungerford, line 99: I can't call this `collect`, since that's already a method on `Stream`.

johnhungerford:
@fwbrasil why are a bunch of the channel tests marked `runNotJS`? I'm finding one of the stream tests is failing when testing in for javascript and can't figure out why. Are channels not fully supported in JS env?

fwbrasil:
@johnhungerford We need to review the tests disabled for JS but, in general, scenarios where more than one thread is required are disabled. What's the test that is failing? I only see a build timeout in the logs

kyo-core/shared/src/main/scala/kyo/AsyncStreamExtensions.scala
```diff
@@ -0,0 +1,198 @@
...
+                    IO.ensure(channel.close):
+                        for
+                            _ <- Async.run[E, Unit, S](Abort.run {
+                                val somat = Async
+                                    .foreachDiscard(streams)(
+                                        _.foreachChunk(c => Abort.run[Closed](channel.put(Present(c))))
+                                    )
+                                somat
+                            }
+                                .andThen(Abort.run(channel.put(Absent)).unit))
```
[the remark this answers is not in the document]
fwbrasil: `Abort.runDiscard` would be more consistent with other APIs

kyo-core/shared/src/main/scala/kyo/AsyncStreamExtensions.scala
```diff
@@ -0,0 +1,198 @@
+package kyo
+
+import kyo.kernel.ArrowEffect
+
+object AsyncStreamExtensions:
```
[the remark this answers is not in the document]
fwbrasil: I'd prefer to put all extensions of a type in a module under the same companion object to help people navigate the codebase. We could have `Stream` extensions not related to `Async`. How about `StreamCoreExtensions`?

kyo-core/shared/src/main/scala/kyo/AsyncStreamExtensions.scala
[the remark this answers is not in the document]
hearnadam: Sounds good

fwbrasil:
> @johnhungerford We need to review the tests disabled for JS but, in general, scenarios where more than one thread is required are disabled. What's the test that is failing? I only see a build timeout in the logs

https://github.com/getkyo/kyo/pull/1141

kyo-core/shared/src/test/scala/kyo/StreamCoreExtensionsTest.scala
```diff
@@ -0,0 +1,109 @@
...
+                    s1     = Stream.init(0 to 99 by 3)
+                    s2     = Stream.init(1 to 99 by 3)
+                    s3     = Stream.init(2 to 99 by 3)
+                    merged = Stream.collectAll(Seq(s1, s2, s3), size)
+                    res <- merged.run
+                yield assert(res.sorted == (0 to 99))
+            }.andThen(succeed)
+        }
+
+        "collectAllHalting" in runNotJS {
```
johnhungerford, line 19: @fwbrasil this is the test that's failing in JS. I don't understand why this one fails, but `mergeHaltingRight/Left` does not.

PR #1125, opened 2025-04-03
[core] fix interrupts leak

### LINKED ISSUE #1122: [BUG]: Memory leak when creating a large number of Promise instances
### Version

0.17.0

### Scala Version

3.6.2

### Expected Behavior

_No response_

### Actual Behavior

Hello. While playing around with kyo streams, I observed an `OutOfMemoryError`. My program was also making use of the kyo `Promise`. After investigating the OOM, I managed to come up with a minimal program which reproduces the issue.

Image

### Steps to Reproduce

Run the following program with `-Xmx64m`.  Note that if the range of values to emit is smaller (eg 100), then the program will not run out of memory. But with large values it runs out of memory.

```
package lukekafka.example

import kyo.*
import kyo.KyoApp

import java.io.IOException

case class Ack(asyncAck: String < Async)

object MemoryLeak extends KyoApp {
  val myStream: Stream[Int, Any] = Stream.init(0 to 10_000_000, chunkSize = 1)
//  val myStream: Stream[Int, Any] = Stream.init(0 to 100, chunkSize = 1) // This is small enough to not generate an OOM

  private val stringStream: Stream[String, IO & Abort[IOException]] = myStream.map(i => s"hello $i")

  val program: Stream[String, IO & Async & Abort[IOException]] = stringStream
    .map { string =>
      for {
        serverAck <- Promise.init[Nothing, String]
        _ <- serverAck.complete(Result.succeed(string))
      } yield Ack(serverAck.get)
    }
    .map { ack =>
      ack.asyncAck
    }

  run(program.discard.map(_ => println("Done")))
}
```

### Current Workaround

_No response_

fwbrasil:
Fixes #1122

### Problem

When a parent `Fiber` is executing and encounters a new child `Fiber` to be waited on, the parent needs to register a callback to propagate interrupts to the child in case there's an interrupt during the waiting. The issue is that, once the child fiber finishes and the parent resumes, the interrupt callback for the child isn't necessary anymore but it stays in the pending list until the parent fully completes execution. This leads to a memory leak when a parent fiber handles many children during its execution.

### Solution

When the parent fiber resumes execution, remove the interrupt callback for the completed child.

### Notes

I've also included a fast-path optimization: if the child fiber is already completed, the parent resumes execution immeditely without suspending. There's a performance penalty with this change due to the removal of a few `inline`s in `IOPromise` that should be easily offset with this additional optimization.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -273,6 +273,13 @@ object Fiber extends FiberPlatformSpecific:
           */
         def waiters(using Frame): Int < IO = IO(self.waiters())
 
+        /** Polls the Fiber for a result without blocking.
+          *
+          * @return
+          *   Maybe containing the Result if the Fiber is done, or Absent if still pending
+          */
+        def poll(using Frame): Maybe[Result[E, A]] < IO = IO(self.poll())
+
     end extension
 
     case class Interrupted(at: Frame)
@@ -571,6 +578,8 @@ object Fiber extends FiberPlatformSpecific:
             def safe: Fiber[E, A] = self
 
             def waiters()(using AllowUnsafe): Int = self.waiters()
+
+            def poll()(using AllowUnsafe): Maybe[Result[E, A]] = self.poll()
         end extension
     end Unsafe
 
@@ -642,6 +651,13 @@ object Fiber extends FiberPlatformSpecific:
               */
             def waiters(using Frame): Int < IO = IO(self.waiters())
 
+            /** Polls the Promise for a result without blocking.
+              *
+              * @return
+              *   Maybe containing the Result if the Promise is done, or Absent if still pending
+              */
+            def poll(using Frame): Maybe[Result[E, A]] < IO = IO(self.poll())
+
         end extension
 
         opaque type Unsafe[+E, +A] <: Fiber.Unsafe[E, A] = IOPromise[E, A]
@@ -665,6 +681,7 @@ object Fiber extends FiberPlatformSpecific:
                 def becomeDiscard[E2 <: E, A2 <: A](other: Fiber[E2, A2])(using AllowUnsafe): Unit = discard(self.become(other))
                 def safe: Promise[E, A]                                                            = self
                 def waiters()(using AllowUnsafe): Int                                              = self.waiters()
+                def poll()(using AllowUnsafe): Maybe[Result[E, A]]                                 = self.poll()
             end extension
         end Unsafe
     end Promise
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -263,6 +263,16 @@ object Fiber extends FiberPlatformSpecific:
 
         def unsafe: Fiber.Unsafe[E, A] = self
 
+        /** Gets the number of waiters on this Fiber.
+          *
+          * This method returns the count of callbacks and other fibers waiting for this fiber to complete. Primarily useful for debugging
+          * and monitoring purposes.
+          *
+          * @return
+          *   The number of waiters on this Fiber
+          */
+        def waiters(using Frame): Int < IO = IO(self.waiters())
+
     end extension
 
     case class Interrupted(at: Frame)
@@ -559,6 +569,8 @@ object Fiber extends FiberPlatformSpecific:
             end mapResult
 
             def safe: Fiber[E, A] = self
+
+            def waiters()(using AllowUnsafe): Int = self.waiters()
         end extension
     end Unsafe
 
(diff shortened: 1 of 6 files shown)


kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -263,6 +263,16 @@ object Fiber extends FiberPlatformSpecific:
...
 
+        /** Gets the number of waiters on this Fiber.
+          *
+          * This method returns the count of callbacks and other fibers waiting for this fiber to complete. Primarily useful for debugging
+          * and monitoring purposes.
+          *
+          * @return
+          *   The number of waiters on this Fiber
+          */
+        def waiters(using Frame): Int < IO = IO(self.waiters())
```
fwbrasil, line 274: I needed it for testing and thought it'd be useful to expose to users

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -237,15 +273,18 @@ private[kyo] object IOPromise:
...
         @nowarn("msg=anonymous")
-        inline def onComplete(inline f: Result[E, A] => Any): Pending[E, A] =
+        final def onComplete(f: Result[E, A] => Any): Pending[E, A] =
             new Pending[E, A]:
                 def waiters: Int = self.waiters + 1
                 def interrupt[E2 >: E](error: Error[E2]) =
                     eval(discard(f(error.asInstanceOf[Error[E]])))
                     self
+                def removeInterrupt(other: IOPromise[?, ?]) =
+                    self.removeInterrupt(other).onComplete(f)
```
fwbrasil, line 287: I had to remove `inline` since it'd recurse the inlining here

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -237,15 +273,18 @@ private[kyo] object IOPromise:
 
         def waiters: Int
         def interrupt[E2 >: E](v: Error[E2]): Pending[E, A]
+        def removeInterrupt(other: IOPromise[?, ?]): Pending[E, A]
```
fwbrasil, line 276: This new method is relatively expensive since it'll rebuild the pending list. It should be ok since, in the majority of cases, the list is quite small with typically two elements: one for interrupt and another for completion. The exception is the collection handling methods in `Fiber`, where `State` objects link to all forked fibers but it becomes available for GC right after the collection operation finishes. I think we don't need an early cleanup there.

kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
```diff
@@ -178,12 +178,13 @@ abstract private class PublisherToSubscriberTest extends Test:
                 subStream3  <- subscriber3.stream
                 subscriber4 <- streamSubscriber
                 subStream4  <- subscriber4.stream
```
fwbrasil, line 180: this test was flaky

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -256,16 +295,21 @@ private[kyo] object IOPromise:
...
                     discard(p.interrupt(error.asInstanceOf[Error[E]]))
                     self
+                def removeInterrupt(other: IOPromise[?, ?]) =
+                    if p eq other then self
+                    else self.removeInterrupt(other).interrupts(p)
                 def waiters: Int = self.waiters + 1
                 def run[E2 >: E, A2 >: A](v: Result[E2, A2]) =
                     self
 
         @nowarn("msg=anonymous")
```
hearnadam, line 305: unused?

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -237,15 +273,18 @@ private[kyo] object IOPromise:
 
         def waiters: Int
         def interrupt[E2 >: E](v: Error[E2]): Pending[E, A]
+        def removeInterrupt(other: IOPromise[?, ?]): Pending[E, A]
         def run[E2 >: E, A2 >: A](v: Result[E2, A2]): Pending[E2, A2]
 
         @nowarn("msg=anonymous")
```
hearnadam, line 279: unused?

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -51,17 +51,25 @@ sealed private[kyo] class IOTask[Ctx, E, A] private (
...
-                                val trace = this.trace
-                                this.trace = null.asInstanceOf[Trace]
-                                input.onComplete { r =>
-                                    val task = IOTask(IO(cont(r.asInstanceOf[Result[Nothing, C]])), trace, context, finalizers, runtime)
-                                    this.becomeDiscard(task)
-                                }
-                                nullResult
+                                input.poll() match
+                                    case null =>
+                                        cont(null)
```
hearnadam, line 56: When is this possible?

fwbrasil: It's an edge case when the promise is completed with `null`. I added a couple of test scenarios for that a some time ago after having a complex issue due to it not being handled well.

hearnadam (approved):

PR #1127, opened 2025-04-04
[data] Flat scaladoc improvement

fwbrasil:
I noticed the missing scaladoc today and it's an important piece of Kyo's design

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Flat.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Flat.scala
@@ -2,12 +2,43 @@ package kyo
 
 import scala.quoted.*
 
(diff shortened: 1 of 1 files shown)


kyo-data/shared/src/main/scala/kyo/Flat.scala
```diff
@@ -2,12 +2,43 @@ package kyo
 
 import scala.quoted.*
 
+/** A compile-time guarantee that a parametrized opaque type with internal flat encoding doesn't contain instances of itself.
+  *
+  * When designing opaque types with flat internal encodings like `A | Container[A]` (union types), runtime discrimination between the cases
```
hearnadam, line 7: I think a one liner on why you do this is helpful. But otherwise LGTM.

hearnadam (approved):

PR #1134, opened 2025-04-08
[build] disable test fork output to try fix Scala Steward

fwbrasil:
### Problem

Scala Steward isn't working for the repo since Jan.

### Solution

I don't know if it's related but this PR introduces a config that is suggested in https://github.com/VirtusLab/scala-steward-repos/issues/98#issuecomment-2288919616. I'm not sure what the config supposed to do since I see the regular test logs locally but it seems worth trying.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -61,7 +61,11 @@ lazy val `kyo-settings` = Seq(
     Test / testOptions += Tests.Argument("-oDG"),
     ThisBuild / versionScheme               := Some("early-semver"),
     libraryDependencies += "org.scalatest" %%% "scalatest" % scalaTestVersion % Test,
-    Test / javaOptions += "--add-opens=java.base/java.lang=ALL-UNNAMED"
+    Test / javaOptions += "--add-opens=java.base/java.lang=ALL-UNNAMED",
+    Test / forkOptions := (Test / forkOptions).value
+        .withOutputStrategy(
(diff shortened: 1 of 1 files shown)


sageserpent-open:
@fwbrasil I'm the author of that comment you referred to above. The idea was to limit the amount of noise from running forked tests by piping the *forked* test output to a null output stream - otherwise the build output buffer used by Scala Steward is overwhelmed by output from parameterised test runs.

I looked at this PR, and noticed this: https://github.com/getkyo/kyo/blob/743090f40948b5df235921d18319d97e930706ca/build.sbt#L631

```scala
                    Tests.Group(
                        name = test.name,
                        tests = Seq(test),
                        runPolicy = Tests.SubProcess(
                            ForkOptions(
                                javaHome = javaHome.value,
                                outputStrategy = outputStrategy.value,
                                bootJars = Vector.empty,
                                workingDirectory = Some(baseDirectory.value),
                                runJVMOptions = javaOptionsValue,
                                connectInput = connectInput.value,
                                envVars = envsVarsValue
                            )
                        )
                    )
```

If I read that correctly, that will set the wrong value of `outputStrategy`, because that binding comes from `ThisBuild` rather than from the implicit settings definition derived from `kyo-settings` that refers to `Test / forkOptions`.

Having said that, I have no idea whether Scala Steward failed when running the benchmark project tests or elsewhere, so this could be a complete red herring...

PR #1135, opened 2025-04-08
[prelude] Introduce Var.updateWith

fwbrasil:
It'd probably be a good improvement for the effect zoo bench

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Var.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Var.scala
@@ -117,9 +117,16 @@ object Var:
       * @return
       *   The new value after applying the update function
       */
-    @nowarn("msg=anonymous")
     inline def update[V](inline f: V => V)(using inline tag: Tag[Var[V]], inline frame: Frame): V < Var[V] =
         ArrowEffect.suspend[V](tag, (v => f(v)): Update[V])
+
+    @nowarn("msg=anonymous")
+    inline def updateWith[V](inline update: V => V)[A, S](inline f: V => A < S)(
+        using
+        inline tag: Tag[Var[V]],
+        inline frame: Frame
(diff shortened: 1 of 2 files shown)


fwbrasil:
I've also pushed a change to make `Var.setWith` public

kyo-prelude/shared/src/main/scala/kyo/Var.scala
```diff
@@ -117,9 +117,16 @@ object Var:
...
+
     @nowarn("msg=anonymous")
-    inline def update[V](inline f: V => V)(using inline tag: Tag[Var[V]], inline frame: Frame): V < Var[V] =
-        ArrowEffect.suspend[V](tag, (v => f(v)): Update[V])
+    inline def updateWith[V](inline update: V => V)[A, S](inline f: V => A < S)(
+        using
+        inline tag: Tag[Var[V]],
+        inline frame: Frame
+    ): A < (Var[V] & S) =
+        ArrowEffect.suspendWith[V](tag, (v => update(v)): Update[V])(f)
```
hearnadam, line 129: Does using an abstract class here actually help? Or could we use `Function1[A, A]` in the opaque `Op` type?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/Var.scala
fwbrasil: Yes, I need it to generate the json schema for all `Op` types to implement the LLM tool

fwbrasil: ouch, sorry. I got the threads mixed up. Since we don't inline handlers, calls of the update functions will typically be megamorphic. An abstract class enables a regular virtual table dispatch of the method while `Function1` would require an interface dispatch, which is more expensive.

PR #1136, opened 2025-04-08
Mark abstract classes as sealed

hearnadam:
### Problem
Unsealed abstract classes enable for anyone to extend, potentially leading to unsafe/incorrect implementations.

### Solution
Mark all internal/Unsafe classes as sealed where possible. Note: I think `Meter` should also be sealed, but we had a test abusing the extension - not sure how often people will be creating custom meters.


hearnadam:
I will followup for the `trait` related changes

Issue #1131, opened 2025-04-08
[feature]: Heirarchical Resource scopes


hearnadam:
### Problem Statement

Heirarchical scopes allow for early closure of Resources when completed, but can also defer to the parent scope.

### Proposed Solution

Attach Child `Resource.Finalizer` to parent Finalizer/Scope

johnhungerford:
Can you give an example of a use case for this?

fwbrasil:
If I understand correctly, a good use case would be the [nested resource handling](https://github.com/getkyo/kyo/blob/8fec68c83a2f234abf7025277c257701e05e6bcd/kyo-actor/shared/src/main/scala/kyo/Actor.scala#L427) in the new actor implementation.  An actor can have its mailbox closed either when it finishes or when the parent resource scope closes.

calvinlfer:
Also streams would be a good use case for this too right?

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/1988

PR #1137, opened 2025-04-11
Resource: avoid leaking `Queue` based Finalizer details

hearnadam:
### Problem
`Finalizer` could be implemented a number of ways. Exposing the `Queue` makes the implementation hard to change. Additionally, this makes it more complex when we support hierarchical scopes in #1131.

### Solution
Use an `abstract class`, exposing only the necessary behaviors. In a followup, I will work towards a rename to `Scope` (optional?), then adding support for `Scope.fork` or some similar naming.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Resource.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Resource.scala
@@ -40,20 +40,20 @@ object Resource:
         def ensure(v: => Any < (Async & Abort[Throwable]))(using Frame): Unit < IO
 
     object Finalizer:
-        sealed abstract class Closeable extends Finalizer:
+        sealed abstract class Awaitable extends Finalizer:
             def close(using Frame): Unit < IO
             def await(using Frame): Unit < Async
 
-        object Closeable:
+        object Awaitable:
             object Unsafe:
-                def init(parallelism: Int)(using frame: Frame, u: AllowUnsafe): Closeable =
-                    new Closeable:
+                def init(parallelism: Int)(using frame: Frame, u: AllowUnsafe): Awaitable =
+                    new Awaitable:
                         val queue   = Queue.Unbounded.Unsafe.init[Unit < (Async & Abort[Throwable])](Access.MultiProducerSingleConsumer)
                         val promise = Promise.Unsafe.init[Nothing, Unit]().safe
 
                         def ensure(v: => Any < (Async & Abort[Throwable]))(using Frame): Unit < IO =
                             IO.Unsafe {
-                                if queue.offer(IO(v.unit)).isFailure then
+                                if queue.offer(IO(v.unit)).isError then
                                     Abort.panic(new Closed(
                                         "Finalizer",
                                         frame,
@@ -80,7 +80,7 @@ object Resource:
                         def await(using Frame): Unit < Async = promise.get
                 end init
             end Unsafe
-        end Closeable
+        end Awaitable
     end Finalizer
 
     /** Ensures that the given effect is executed when the resource is released.
@@ -155,10 +155,10 @@ object Resource:
       */
     def run[A, S](closeParallelism: Int)(v: A < (Resource & S))(using frame: Frame): A < (Async & S) =
         IO.Unsafe {
-            val closeable = Finalizer.Closeable.Unsafe.init(closeParallelism)
-            ContextEffect.handle(Tag[Resource], closeable, _ => closeable)(v)
-                .handle(IO.ensure(closeable.close))
-                .map(result => closeable.await.andThen(result))
+            val finalizer = Finalizer.Awaitable.Unsafe.init(closeParallelism)
+            ContextEffect.handle(Tag[Resource], finalizer, _ => finalizer)(v)
+                .handle(IO.ensure(finalizer.close))
+                .map(result => finalizer.await.andThen(result))
         }
 
     given Isolate.Contextual[Resource, Any] = Isolate.Contextual.derive[Resource, Any]
--- a/kyo-core/shared/src/main/scala/kyo/Resource.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Resource.scala
@@ -1,6 +1,5 @@
 package kyo
 
-import java.io.Closeable
 import kyo.Tag
 import kyo.kernel.ContextEffect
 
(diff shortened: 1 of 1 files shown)


kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -36,9 +35,53 @@ import kyo.kernel.ContextEffect
 sealed trait Resource extends ContextEffect[Resource.Finalizer]
 
 object Resource:
-
     /** Represents a finalizer for a resource. */
-    case class Finalizer(createdAt: Frame, queue: Queue.Unbounded[Unit < (Async & Abort[Throwable])])
+    sealed abstract class Finalizer:
+        def ensure(v: => Any < (Async & Abort[Throwable]))(using Frame): Unit < IO
```
hearnadam, line 40: Ideally `ensure` should immediately run the finalizer if the queue is closed, but there is no mechanism for back-pressure if the `Promise` is already completed. What do you think @fwbrasil

hearnadam:
```
[info]   - parallel *** FAILED *** (404 milliseconds)
[info]     done.await(millis, java.util.concurrent.TimeUnit.MILLISECONDS) was false Latch had count 2 after 200.millis (MonixTest.scala:177)
```

Flaky test...

kyo-core/shared/src/main/scala/kyo/Resource.scala
fwbrasil: I think back pressure isn't very relevant here. The only case where the queue would already be closed should be when a resource escapes the scope of `Resource.run`. It's undefined behavior territory that might be covered by capture checking in the future. If that happens, I think throwing makes more sense to signal to the user that there's an issue with resource management in the code.

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -36,9 +35,53 @@ import kyo.kernel.ContextEffect
...
+        object Closeable:
+            object Unsafe:
+                def init(parallelism: Int)(using frame: Frame, u: AllowUnsafe): Closeable =
+                    new Closeable:
+                        val queue   = Queue.Unbounded.Unsafe.init[Unit < (Async & Abort[Throwable])](Access.MultiProducerSingleConsumer)
+                        val promise = Promise.Unsafe.init[Nothing, Unit]().safe
+
+                        def ensure(v: => Any < (Async & Abort[Throwable]))(using Frame): Unit < IO =
+                            IO.Unsafe {
+                                if queue.offer(IO(v.unit)).isFailure then
```
fwbrasil, line 56: This seems to ignore panics. I think JCTools can't panic for an `offer` but it'd be nice to cover the scenario since we might change the underlying impl at some point

hearnadam: Ah good point, I will update.

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -36,9 +35,53 @@ import kyo.kernel.ContextEffect
...
+                                    case Present(tasks) =>
+                                        Async.foreachDiscard(tasks, parallelism) { task =>
+                                            Abort.run[Throwable](task)
+                                                .map(_.foldError(_ => (), ex => Log.error("Resource finalizer failed", ex.exception)))
+                                        }
+                                            .handle(Async.run[Nothing, Unit, Any])
+                                            .map(promise.becomeDiscard)
+                            }
+
+                        def await(using Frame): Unit < Async = promise.get
```
fwbrasil, line 80: nice! I think this essentially implements proper back pressure?

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -36,9 +35,53 @@ import kyo.kernel.ContextEffect
...
 
 object Resource:
-
     /** Represents a finalizer for a resource. */
-    case class Finalizer(createdAt: Frame, queue: Queue.Unbounded[Unit < (Async & Abort[Throwable])])
+    sealed abstract class Finalizer:
+        def ensure(v: => Any < (Async & Abort[Throwable]))(using Frame): Unit < IO
+
+    object Finalizer:
+        sealed abstract class Closeable extends Finalizer:
```
fwbrasil, line 43: how about `Awaitable`?

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -122,26 +154,11 @@ object Resource:
...
-                    }
-                ContextEffect.handle(Tag[Resource], finalizer, _ => finalizer)(v)
-                    .handle(IO.ensure(close))
-                    .map(result => p.get.andThen(result))
-            }
+        IO.Unsafe {
+            val closeable = Finalizer.Closeable.Unsafe.init(closeParallelism)
+            ContextEffect.handle(Tag[Resource], closeable, _ => closeable)(v)
+                .handle(IO.ensure(closeable.close))
+                .map(result => closeable.await.andThen(result))
```
fwbrasil, line 161: another option is returning a `Fiber` so `use` can be used. Not sure it's worth it

fwbrasil:
> ```
> 
> [info]   - parallel *** FAILED *** (404 milliseconds)
> 
> [info]     done.await(millis, java.util.concurrent.TimeUnit.MILLISECONDS) was false Latch had count 2 after 200.millis (MonixTest.scala:177)
> 
> ```
> 
> 
> 
> Flaky test...

Yeah, I've looked into these failures a couple of times. I suspect Monix has some race condition with interrupts

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -36,9 +35,53 @@ import kyo.kernel.ContextEffect
...
+        object Awaitable:
+            object Unsafe:
+                def init(parallelism: Int)(using frame: Frame, u: AllowUnsafe): Awaitable =
+                    new Awaitable:
+                        val queue   = Queue.Unbounded.Unsafe.init[Unit < (Async & Abort[Throwable])](Access.MultiProducerSingleConsumer)
+                        val promise = Promise.Unsafe.init[Nothing, Unit]().safe
+
+                        def ensure(v: => Any < (Async & Abort[Throwable]))(using Frame): Unit < IO =
+                            IO.Unsafe {
+                                if queue.offer(IO(v.unit)).isError then
```
fwbrasil, line 56:
```suggestion
                                if !queue.offer(IO(v.unit)).contains(true) then
```
Would be safer

PR #1139, opened 2025-04-12
Creates a stream by repeatedly calling a lazily evaluated function

vladpo:
Creates a stream by repeatedly calling a lazily evaluated function, until the return is absent.

### Problem

Stream emit an effect until some end state offers no more values.

### Solution

repeatPresent is a factory method in the Stream class that creates a stream by repeatedly calling a function until it returns an absent value
1. Create a stream from a function that may return values multiple times
2. Automatically terminate the stream when the function returns an absent value
3. Control the chunk size of emitted values for performance optimization

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
@@ -569,18 +569,16 @@ object Stream:
       */
     def init[V, S](v: => Seq[V] < S, chunkSize: Int = DefaultChunkSize)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Stream[V, S] =
         Stream[V, S]:
-            v.map(seq => emitResized(seq, chunkSize))
-
-    private def emitResized[V](seq: Seq[V], chunkSize: Int)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Unit < Emit[Chunk[V]] =
-        val chunk: Chunk[V] = Chunk.from(seq)
-        val _chunkSize      = chunkSize max 1
-        Loop(chunk) { (c) =>
-            if _chunkSize >= c.length then
-                Emit.valueWith(c)(Loop.done)
-            else
-                Emit.valueWith(c.take(_chunkSize))(Loop.continue(c.dropLeft(_chunkSize)))
-        }
-    end emitResized
+            v.map { seq =>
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -569,14 +569,37 @@ object Stream:
...
-                val chunk: Chunk[V] = Chunk.from(seq)
-                val _chunkSize      = chunkSize max 1
-                Loop(chunk) { (c) =>
-                    if _chunkSize >= c.length then
-                        Emit.valueWith(c)(Loop.done)
-                    else
-                        Emit.valueWith(c.take(_chunkSize))(Loop.continue(c.dropLeft(_chunkSize)))
+            v.map(seq => emitResized(seq, chunkSize))
+
+    private def emitResized[V](seq: Seq[V], chunkSize: Int)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Unit < Emit[Chunk[V]] =
```
hearnadam, line 574: I don't think this is quite semantically correct for the `repeat` case as it may produce chunks like this:

given: `repeatPresent(Chunk(1, 2, 3, 4), 3)`
outcome: `Stream(Chunk(1, 2, 3), Chunk(4), Chunk(1, 2, 3), Chunk(4)...)`

In my opinion, it's reasonable for `.rechunk` to be used after the `repeat` operator, given rechunking will happen across suspension.

kyo-prelude/shared/src/test/scala/kyo/StreamTest.scala
```diff
@@ -60,6 +60,35 @@ class StreamTest extends Test:
         }
     }
 
+    "repeatPresent" - {
+        var i = 0
```
hearnadam, line 64: Please avoid mutability outside the scope of a single test if possible 🙏 it makes parallelizing/refactoring tests difficult.

Perhaps you can use the `Var` effect?

hearnadam (commented):
Thanks for your contribution! This definitely could be useful.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
vladpo: Thanks for the reply. I've changed to calling rechunk.

hearnadam (approved):

fwbrasil:
thank you @vladpo!

PR #1141, opened 2025-04-13
fix and enable JS tests

### LINKED ISSUE #1138: Review disabled tests in JS
The disabling of JS tests wasn't very strict in the initial development of the cross compilation. We need to review if they're correct. The main limitation are scenarios where two threads are required, for example when testing interruptions or waiting for some condition triggered by a concurrent computation.

fwbrasil:
Fixes #1138

### Problem

We have tests disabled in JS because they require multiple threads.

### Solution

Fix tests to avoid hot loops hogging the JS thread.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -157,6 +157,30 @@ object Async:
       */
     def never[E, A](using Frame): A < Async = Fiber.never[Nothing, A].get
 
+    /** Yields control to the scheduler, allowing other fibers to execute.
+      *
+      * Primarily useful in JavaScript environments for CPU-intensive loops without async operations, as JS lacks automatic preemption. On
+      * JVM/Native platforms, this is less necessary since they use automatic time-slice preemption between threads.
+      *
+      * @return
+      *   Unit wrapped in an Async effect
+      */
+    def yieldNow(using Frame): Unit < Async =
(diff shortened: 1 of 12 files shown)


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -157,6 +157,30 @@ object Async:
...
+      * automatic preemption makes this less critical.
+      *
+      * @param f
+      *   The computation to execute after yielding
+      * @return
+      *   The result of the computation wrapped in an Async effect
+      */
+    def yieldWith[A, S](f: => A < S)(using Frame): A < (Async & S) =
+        Async.run(()).map(_.use(_ => f))
+
```
fwbrasil, line 183: new methods to yield to the scheduler

kyo-core/shared/src/main/scala/kyo/internal/BaseKyoCoreTest.scala
```diff
@@ -18,4 +18,15 @@ private[kyo] trait BaseKyoCoreTest extends BaseKyoKernelTest[Abort[Any] & Async
             IO.Unsafe.evalOrThrow
         )
     end run
+
+    def untilTrue[S](f: => Boolean < S): Unit < (Async & S) =
```
fwbrasil, line 22: moved from the kernel so it's possible to use `Retry` and avoid a hot loop

hearnadam (approved):

PR #1143, opened 2025-04-17
[core] fix promise variance

### LINKED ISSUE #801: Review variance of IOPromise
The implementation currently has a few `.asInstanceOf`s. See https://github.com/getkyo/kyo/pull/797#discussion_r1825009882 for context

fwbrasil:
Fixes #801 

### Problem

Promises are currently unsafe. It's possible to complete them with the wrong types due to variance.

### Solution

Make `IOPromise`, `Fiber.Promise`, and `Fiber.Unsafe.Promise` invariant to ensure they can only be completed with the exact types. I've kept `Fiber` and `Fiber.Unsafe` with variance since, besides interruption, they're read only.

### Notes

Please see comments for some of the issues I found during this change.


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -155,7 +155,7 @@ object Async:
       * @return
       *   A computation that never completes
       */
-    def never[E, A](using Frame): A < Async = Fiber.never[Nothing, A].get
+    def never[A](using Frame): A < Async = Fiber.never[Nothing, A].get
```
fwbrasil, line 158: unrelated cleanup

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -396,7 +394,7 @@ object Fiber extends FiberPlatformSpecific:
                                     quickSort(indices, results, size)
 
                                     // Limit final result to max successful results
-                                    completeDiscard(Result.succeed(Chunk.fromNoCopy(results).take(size)))
+                                    completeDiscard(Result.succeed(Chunk.fromNoCopy(results).take(size).asInstanceOf[Chunk[A]]))
```
fwbrasil, line 397: The underlying `results` is untyped so this produces `Chunk[AnyRef]` but it's ok to cast here

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -263,80 +264,88 @@ private[kyo] object IOPromise:
...
-                def run[E2 >: E, A2 >: A](v: Result[E2, A2]) =
+                def run(v: Result[E, A]) =
                     eval(discard(f(v.asInstanceOf[Result[E, A]])))
                     self
                 end run
 
         final def interrupts(p: IOPromise[?, ?]): Pending[E, A] =
             new Pending[E, A]:
-                def interrupt[E2 >: E](error: Error[E2]) =
-                    discard(p.interrupt(error.asInstanceOf[Error[E]]))
```
fwbrasil, line 294: This code was incorrect. If the current promise is interrupted with `Result.Failure[E]`, it'd be propagated to the `p` promise without a guarantee that it actually has `Failure[E]` in its error channel, which could cause a class cast exception. As an alternative, I'm keeping the panic if possible but using a placeholder in case it's an actual `Failure`.

hearnadam (approved):

PR #1145, opened 2025-04-17
[scheduler] time slice preemption in JS

fwbrasil:
### Problem

https://github.com/getkyo/kyo/pull/1141 introduced `Async.yield*` as a way to avoid loops taking over the JS thread. The need for manual yielding only in JS ins't ideal when codebases have to support both JVM and JS.

### Solution

Introduce preemption in the JS scheduler based on time slices and remove the `Async.yield*` methods.


kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -36,12 +36,12 @@ sealed private[kyo] class IOTask[Ctx, E, A] private (
...
     private inline def locally[A](inline f: A): A = f
 
-    final private def eval(startMillis: Long, clock: InternalClock)(using Safepoint) =
+    final private def eval(startMillis: Long, clock: InternalClock, deadline: Long)(using Safepoint) =
         try
             curr = Isolate.restoring(trace, this) {
                 ArrowEffect.handlePartial(erasedAbortTag, Tag[Async.Join], curr, context)(
                     stop =
-                        shouldPreempt(),
+                        shouldPreempt() || (deadline != Long.MaxValue && clock.currentMillis() > deadline),
```
fwbrasil, line 44: This shouldn't introduce overhead for JVM but it can be costly for JS. I think we can optimize it later if needed.

hearnadam (approved):

PR #1148, opened 2025-04-18
[kernel] first-class effect nesting support

fwbrasil:
### Problem

Kyo's support for nested computations is limited. With exception of `flatten`, they can only be transformed in generic contexts given the `WeakFlat` constraint, and effects can only be handled when there's a `Flat` evidence available. The current approach has important downsides:

1. Passing the `Flat` evidence isn't always possible when integrating with other libraries. For example, `MonadLawsTest` currently needs to bypass the flat check due to that. Luckily, the sttp integration doesn't face the same issue but it could be the case.
2. Passing the `Flat` evidence around in the codebase makes the code more verbose and harder to follow. If users need to abstract effect handling, they'll also face the same complexity.
3. The compiler error messages when there's computation nesting can be confusing. It's problematic because developers used to other effect systems tend to try to nest computations even though it's mostly unnecessary in Kyo.
4. The requirement of a `Flat` evidence sometimes constrains designs. A good example is the `flatTransform` member [in isolates](https://github.com/getkyo/kyo/blob/5922778d313def530852b71807a1d3ea19890de8/kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala#L186).

### Solution

I've been working on a new kernel implementation (I know, I promise it's the last 😇) and noticed that we can apply a technique similar to `Maybe` and `Result` to automatically lift nested computations if necessary. Kyo's regular operation remains the same favoring flattening with `map` being the same as `flatMap` but nesting can be manually achieved via `Kyo.lift`. Under the hood, `<.lift` reflects on the type being lifted and creates a `Nested` wrapper in case the value is a computation. The same process happens when a value is automatically lifted in a generic context by `<.lift`, ensuring proper nested representation. 

### Notes

I've kept `WeakFlat` because without it, effect mismatches end up triggering nesting. If you pass a `Int < Memo` to method that takes `Int < Async`, the compiler automatically lifts the computation to `Int < Memo < Any`, which is a quite confusing behavior.


kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -87,12 +85,23 @@ object `<`:
           *   A computation producing the final result
           */
         inline def flatMap[B, S2](inline f: Safepoint ?=> A => B < S2)(
-            using
-            inline frame: Frame,
-            inline flatA: WeakFlat[A],
-            inline flatB: WeakFlat[B]
+            using inline _frame: Frame
         ): B < (S & S2) =
-            map(v => f(v))
```
fwbrasil, line 95: I had to duplicate the body because the compiler was inferring incorrect types and leaking the internal `Kyo` type.

kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -309,18 +326,25 @@ object `<`:
...
             v match
-                case kyo: Kyo[?, ?] => Maybe.empty
-                case v              => Maybe(v.asInstanceOf[A])
+                case kyo: KyoSuspend[?, ?, ?, ?, ?, ?] => Maybe.empty
+                case v                                 => Maybe(v.unsafeGet)
 
     end extension
 
+    extension [A, S](v: A < S)
+        private[kyo] def unsafeGet: A =
```
fwbrasil, line 342: This is the other end of `<.lift`, unwrapping the `Nested` if necessary.

hearnadam: How about `unnest`?

kyo-kernel/shared/src/main/scala/kyo/kernel/internal/KyoInternal.scala
```diff
@@ -27,6 +27,11 @@ import kyo.kernel.ArrowEffect
   */
 sealed abstract private[kernel] class Kyo[+A, -S] extends Serializable
 
+/** Nested serves as a special wrapper type that extends the core Kyo class to handle cases where one effect computation is embedded within
+  * another. It prevents unintended unsound flattening of effects and maintains the proper composition structure when effects are nested.
+  */
+case class Nested[+A](v: A) extends Kyo[A, Any]
```
hearnadam, line 33: I think this can be private?

kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -370,8 +394,25 @@ object `<`:
...
+      *   2. When the input is a regular value, it lifts it directly into the effect context through type casting.
+      *
+      * The WeakFlat constraint avoids unexpected lifting when the pending effect set of computations don't match.
+      *
+      * @param v
+      *   The value to lift into the effect context
+      * @return
+      *   A computation in the effect context
+      */
+    implicit def lift[A: WeakFlat, S](v: A): A < S =
```
hearnadam, line 412: Does this function need updating as well?

https://github.com/getkyo/kyo/blob/5922778d313def530852b71807a1d3ea19890de8/kyo-kernel/shared/src/main/scala/kyo/Kyo.scala#L31

kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
fwbrasil: I'd prefer `unsafeGet`. It better represents the risk of the operation since the code should first check that it isn't a suspended computation. In the new arrow-based kernel I'm working on, I'm trying to have a single method that does the introspection so we don't have this unsafe operation spread across the kernel.

kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
fwbrasil: No, this is the main method to nest computations using Kyo's APIs. I've added a few test scenarios demonstrating it. The other way we can end up with nesting is when Kyo computations are used in a generic context, which bypasses the `WeakFlat` check but gets properly nested in `<.lift`. This method actually ends up using `<.lift` as well when it returns `A` as `A < S`.

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -713,36 +713,34 @@ object Async:
...
+                            }.handle(IO.ensure {
+                                IO.Unsafe {
+                                    if !promise.done() then
+                                        ref.set(Absent)
+                                }
+                            })
+                        else
+                            loop()
+                        end if
+            Kyo.lift(IO(loop()))
```
steinybot, line 743: This particular example seems to work because `Async` includes `IO`. How would you defer some other `A < S` to `A < S < IO`? Wouldn't `Kyo.lift(IO(x))` become `A (S & IO) < IO`?

fwbrasil:
> This is interesting. So really `A < S1 < S2` isn't all that different from `A < (S1 & S2)`, they are both nested in a way, but the former forces the order that you have to handle the effects and in the later you can handle them in any order. Is that the right way of thinking about it?

I don't see them as similar since you can compose nested computations in multiple ways and handling ordering isn't the only relevant property. For example, you can map the nested computation in isolation or even discard nested effects. But yeah, nesting can be a way to encode ordering, which is the case for `Async.memoize`. I'm not expecting it to be part of the regular use of the library, though. It's more an edge case for integrations and use of Kyo in generic scopes.

kyo-core/shared/src/main/scala/kyo/Async.scala
fwbrasil: `IO(loop())` produces `A < (Async & IO)`, which is equivalent to `A < Async`. `Kyo.lift(IO(loop()))` then lifts it to `A < Async < Any`, which then becomes to `A < Async < IO` via the `IO` suspension at the beginning of the method. The outer suspension could be of any effect, it's separate from the nested `Async` and `IO` suspensions and unrelated to the fact that `Async` includes `IO`.

steinybot: Yep, I got that. I was struggling to see how to nest something where the outer effects were not a subset of the inner effects (as is the case with `IO` and `Async`).

It looks like in the general case you have to lift first and then add the outer effects.

For example, this does not work:
```
    "lift after" in run {
        val inner: Int < Abort[String]      = 1
        val outer: Int < Abort[String] < IO = Kyo.lift(IO(inner))
        outer.map(identity).map(result => assert(result == 1))
    }
```

But this does:
```
    "lift before" in run {
        val inner: Int < Abort[String]      = 1
        val outer: Int < Abort[String] < IO = IO(Kyo.lift(inner))
        outer.map(identity).map(result => assert(result == 1))
    }
```

fwbrasil: Exactly. `Kyo.lift` "isolates" the computation from the outer effect, becoming a regular immutable monadic value handled by the outer monad instance. There's no direct interaction between the inner and outer monads/effects. In traditional effect systems, that's typically achieved by returning a computation in a `map` transformation. This design optimizes for the most common scenario where nesting isn't necessary (`map` == `flatMap`) and makes nesting the exception by requiring an explicit `Kyo.lift` call to produce a nested computation. There isn't anything special about `Kyo.lift` by the way, the same can be achieved in generic methods by referencing Kyo computations via a single type parameter, which triggers the boxing (`Nested`).

steinybot (approved):
Nice!. Much better not having to have `Flat` everywhere.

hearnadam (approved):

Issue #1020, opened 2025-01-15
`Kyo.foreach*` should support more than Seq


hearnadam:
Either introduce specialized functions for Set, Map, etc or use an interface like `IterableOnce`
https://github.com/getkyo/kyo/blob/main/kyo-kernel/shared/src/main/scala/kyo/Kyo.scala#L77-L97

johnhungerford:
What about the zio approach?

```scala
def foreach[A, B, Coll[_], S](collection: Coll[A])(fun: A => B < S)(using bf: BuildFrom[Coll[A], A, Coll[B]): Coll[B] < S = ???
```

hearnadam:
Yes, that could work. The lack of specialization may be an issue for performance and or purity... `BuildFrom` doesn't specify (im)mutability IIRC.

ahoy-jon:
> Already fixed by [#1149](https://github.com/getkyo/kyo/pull/1149)

Yes, however, it doesn't return `Coll[B] < S`, we would need something like that for #1212 (on `List[?]`, `Vector[?]` ... )

hearnadam:
Thanks @ahoy-jon that was the intention of the issue. We should at least go from `Map` -> `Map` and `Set` -> `Set`...

PR #1149, opened 2025-04-20
Update `Kyo`'s batch actions to use IterableOnce

HollandDM:
Most of `Kyo`'s batch methods are receiving `Seq[A]` only, this PR update them to use `IterableOnce[A]` instead

see https://github.com/getkyo/kyo/issues/1020

### DIFF
--- a/kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
@@ -3,6 +3,7 @@ package kyo
 import kernel.Loop
 import kyo.kernel.internal.Safepoint
 import scala.annotation.tailrec
+import scala.collection.immutable.LinearSeq
 
 /** Object containing utility functions for working with Kyo effects.
   *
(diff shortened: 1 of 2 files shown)


HollandDM:
For some reason, `ChoiceTest` is failing. But `KyoTest` is still good

fwbrasil:
We can't leverage mutability in these methods due to Kyo's support for multi-shot continuations. See https://github.com/getkyo/kyo/pull/1027 for more background

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -532,14 +561,42 @@ object Kyo:
...
         }
 
-    private def toIndexed[A](seq: Seq[A]): Seq[A] =
-        seq match
-            case seq: IndexedSeq[A] => seq
-            case seq                => Chunk.from(seq)
+    /** A minimal wrapper that can lazily and immutably extract elements from an `IterableOnce[A]`. The implementation is closely modelled
+      * after `LazyList`, stripped down to bare metals.
+      */
+    final private class IterableExtractor[A](private var maybeIterator: Maybe[Iterator[A]]):
```
HollandDM, line 572: This was designed to maximize the lazy nature of iterator, and it introduces a small allocation overhead.
This has a trade off compare to `List`, as `List` is expected to be faster (no lazy val), but this is expected to be less memory consumption.
We only use this for the sake of `head` and `tail`, so every other collection will be worse than this and `List`

HollandDM: This follows almost exactly how `LazyList[A]` build itself from an `Iterator[A]`, so while internally it advance the iterator (like every other collection's `from` methods), the result is a immutable wrapper that can be reused.
While thinking about this is fun, but I agree with you that this seem over complicated. I'll update it.

HollandDM:
Updated to use `toIndexed`

fwbrasil (approved):

PR #1150, opened 2025-04-20
[kernel] introduce handleLoop to replace handleState

fwbrasil:
### Problem

There are scenarios where effect handling can perform new effect suspensions. The main example of that is `Stream` where `Emit` suspensions can be transformed and produce new emissions during handling. The only method that supports that is `handleState` so we need to use it in scenarios even when the handling doesn't require state.

### Solution

Introduce `ArrowEffect.handleLoop` methods as a replacement for `handleSate` following the usage pattern of `Loop` itself and the new `Actor.receiveLoop` methods. Besides allowing new effects during handling, `handleLoop` also gives more control to the user to stop the handling via `Loop.done`.

### Notes

With this change and https://github.com/getkyo/kyo/pull/1148, I think the new kernel I've been working on will require changes only in `kyo-kernel` itself.


kyo-kernel/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -393,9 +403,67 @@ object ArrowEffect:
...
+        inline _frame: Frame,
+        inline flat: Flat[A],
+        safepoint: Safepoint
+    ): A < (S & S2) =
+        @nowarn("msg=anonymous")
+        def handleLoopLoop(v: Loop.Outcome[A < (E & S), A] < S2, context: Context)(using Safepoint): A < (S & S2) =
+            v match
+                case _: KyoSuspend[?, ?, ?, ?, ?, ?] =>
+                    v.map(handleLoopLoop(_, context))
+                case continue: Loop.Continue[A < (E & S)] @unchecked =>
```
fwbrasil, line 440: The new implementation also has an optimization. If the user returns `Loop.continue` without suspensions, it skips the overhead of the `map` call and recurses directly.

kyo-kernel/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -404,40 +472,85 @@ object ArrowEffect:
...
+      *   The function implementation that returns a Loop.Outcome for each iteration
       * @return
-      *   The computation result with the stateful implementation provided
+      *   The computation result with the function implementation provided
       */
-    inline def handleState[I[_], O[_], E <: ArrowEffect[I, O], State, A, B, S, S2, S3](
+    inline def handleLoop[I[_], O[_], E <: ArrowEffect[I, O], A, S, S2, State](
         inline effectTag: Tag[E],
-        inline state: State,
+        state: State,
```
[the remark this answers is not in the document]
fwbrasil: I've been removing `inline` where possible to reduce compilation times. In this case, doesn't really do much since the parameter just gets passed to the loop. It's different for `Tag` since it'll typically be a compile-time static string that can inlined directly into the generated code, avoiding capturing in closures.

kyo-kernel/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -404,40 +472,85 @@ object ArrowEffect:
...
+      * @param v
+      *   The computation requiring the function implementation
+      * @param handle
+      *   The function implementation for each iteration
+      * @param done
+      *   The function to transform the final state and result
+      * @return
+      *   The computation result with the function implementation provided
+      */
+    inline def handleLoop[I[_], O[_], E <: ArrowEffect[I, O], A, B, S, S2, State](
```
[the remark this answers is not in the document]
fwbrasil: I believe it was a leftover from previous changes

kyo-kernel/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -29,6 +29,12 @@ import scala.util.control.NonFatal
   * handler takes two parameters: an input value of type I[C] that contains the input of the operation, and a continuation function
   * representing the remainder of the computation from the point where the effect was suspended to the point where it's being handled.
   *
+  * ArrowEffect provides two main kinds of handling methods with distinct capabilities:
+  *   - handle: Basic handler that doesn't allow introducing new effects during handling
+  *   - handleLoop: Enhanced handler that explicitly allows introducing new effects (via S2 type parameter) during handling:
```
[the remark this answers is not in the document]
fwbrasil: Since the `handle` returns a flat computation, the challenge is separating new effects from the one that is being handled. A good example is `filter` in `Stream` since it handles and emits suspensions of the same effect. It wouldn't be possible to distinguish which suspensions should be handled and when the loop should stop.

That said, this description isn't very accurate after https://github.com/getkyo/kyo/pull/1148. It's possible to introduce new effects with `handle` by nesting the computation now. The usability wouldn't be great, though. You'd need to first nest the result of the computation being handled for example. I think we should still keep `handleLoop` as the recommended API for this use case given the ergonomics.

Regarding allocations, the way the code is structured typically allow for the JIT to elide `Loop.continue` allocations. It isn't a guarantee but I haven't seen a profiling session where it shows as a hot allocation so far.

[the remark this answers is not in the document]
johnhungerford: > Regarding allocations, the way the code is structured typically allow for the JIT to elide Loop.continue allocations. It isn't a guarantee but I haven't seen a profiling session where it shows as a hot allocation so far.

Very cool -- I definitely like the `Loop` API for this if performance isn't an issue

kyo-kernel/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -404,40 +472,85 @@ object ArrowEffect:
...
+                            new KyoContinue[IX, OX, EX, Any, B, S & S2](kyo):
+                                def frame = _frame
+                                def apply(v: OX[Any], context: Context)(using Safepoint) =
+                                    handleLoopLoop(Loop.continue(state, kyo(v, context)), context)
+                            end new
+                        case kyo =>
+                            done(state, kyo.asInstanceOf[A])
+                    end match
+                case _ =>
+                    v.asInstanceOf[B < (S & S2)] // Loop.done
```
hearnadam, line 550: Could the cast be avoided using a non-opaque type for the input here? Or would it be too annoying for inference?

kyo-kernel/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
```diff
@@ -404,40 +472,85 @@ object ArrowEffect:
...
+    )(
+        inline handle: [C] => (I[C], State, Safepoint ?=> O[C] => A < (E & S)) => Loop.Outcome2[State, A < (E & S), B] < S2,
+        inline done: (State, A) => B < (S & S2)
+    )(
+        using
+        inline _frame: Frame,
+        inline flat: Flat[A],
+        safepoint: Safepoint
+    ): B < (S & S2) =
         @nowarn("msg=anonymous")
```
hearnadam, line 530: Is this still necessary?

hearnadam (approved):

kyo-kernel/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
fwbrasil: Yep, there's still a `new KyoContinue` in the body

kyo-kernel/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
fwbrasil: We could have wrapper class for `Loop.done` but we'd need to check if it doesn't end up allocating too much. An alternative is a `Loop.fold` method instead to make the code safer but still avoiding the allocation. I played with that recently, I'll explore it further.

PR #1154, opened 2025-04-22
Pass single transformation to handle by-value

steinybot:
### Problem

During some more spelunking I realised that the first transformation passed to the single `handle` function was being passed by-name unnecessarily. The other `handle` functions already take it by-value.

### Solution

Pass the first transformation by-value.

### DIFF
--- a/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
@@ -175,9 +175,8 @@ object `<`:
           * @return
           *   The result of applying the transformation
           */
-        inline def handle[B](inline f: (=> A < S) => B): B =
-            def handle1 = v
-            f(handle1)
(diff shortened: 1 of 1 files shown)


fwbrasil:
> I was playing around with effect tracking, specifically to show which effect was suspended when a Timeout occurs.

I'm not sure I understand how this relates to handlers taking a by-name param, can you elaborate?

> Is there a reason that handlers need to take the effect by-name rather than by-value?

The main ones I remember are `Abort.run` so `Abort.run(throw ex)` properly catches the failure and `Async.run` so the thunk is evaluated in the forked fiber. I see your point that we could require users to suspend evaluation but, if they don't, which should be common for users not experienced with FP, the behavior would be very surprising. Ah, there's also `IO.ensure` that takes a thunk so the finalizer is always executed even if there's a failure.

fwbrasil:
should we close this PR?

PR #1155, opened 2025-04-22
Show suspended frame when there is a timeout

steinybot:
### Problem

I was experiencing a `Timeout` and had no idea which effect was not completing. It seemed like it ought to be possible to show more information.

I was also debugging and wondered if there was a way to display the graph of effects and/or fibers. Something analogous to a thread dump.

### Solution

This is by no means a solution but hopefully it shows the problem. The basic idea is when creating a suspended computation because the other dependent computation is suspended then record the frame from that dependency. I'm sure you will tell me why this will never work but it feels like there might be something here.

### Notes

I was using the `kyo.kernel.internal.KyoSuspend#frame` but I think it would probably need to be tracked separately.

Also what would we do if there were multiple dependencies such as in `Fiber.race` and `Fiber.gather` etc? I'm not sure if we can track everything without hurting performance.

Obviously we cannot get the frame directly from an effect that does not exist yet, i.e. `Async(1)`

by-name effects are an issue, not because they can suspend computation, that is reasonable and we already capture the frame at that point, but because we use the same functions to handle strict and lazy effects. For example:
```scala
def slowMethod: Int = ???
Async.run(slowMethod)
```
The suspended computation from `Async.run` depends on the frame context parameter that is derived when `Async.run` is called.

However in:
```scala
val slowEffect: Int < Async = ???
Async.run(slowEffect)
```
It should depend on the frame that was used to create `slowEffect`.

I can't think of a good way to do this. Some ideas that haven't led anywhere:
- Change `Async.run` to be by-value. This makes the usability pretty bad.
- Overload `Async.run` somehow so that there is a by-name version for lifted effects and a by-value one for existing effects.
- Do some compile time trick to do something different depending on what is passed in. I couldn't think of a way to do this without a macro which was a bit yuck. There is `scala.compiletime.byName` but I don't think that helps.


fwbrasil:
Thanks for the clarifications, I can see now what you're trying to achieve. I think we have a better path to provide this feature. `IOTask` maintains the [execution trace](https://github.com/getkyo/kyo/blob/c67d64bab66bf152e0b4bfbf6d22eb555925e63b/kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala#L13). We could expose a method `fiber.dumpTrace` (not sure about the naming) that returns a `Chunk[Frame]` and use it from `Async.timeout`. Since `Fiber` can be a regular `IOPromise` as well, we'd need to return an empty `Chunk` in case the underlying value isn't `IOTask`.

PR #1156, opened 2025-04-23
Parallel stream map

johnhungerford:
Address #679 

### Problem

We need to be able to transform stream elements and chunks asynchronously

### Solution

Added the following methods to `Stream`:
- `mapPar`
- `mapParUnordered`
- `mapChunkPar`
- `mapChunkParUnordered`

### DIFF
--- a/kyo-core/jvm/src/test/scala/kyo/PathTest.scala
+++ b/kyo-core/jvm/src/test/scala/kyo/PathTest.scala
@@ -90,7 +90,7 @@ class PathTest extends Test:
                 _ <- useFile(first, text)
                 _ <- useFile(second, text)
                 v <- root.readAll("txt")
-            yield assert(v == List(first -> text, second -> text))
+            yield assert(v.toSet == Set(first -> text, second -> text))
             end for
         }
 
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -248,7 +248,7 @@ object Channel:
                 Loop[Unit, Unit, Abort[Closed] & Async & Emit[Chunk[A]]](()): _ =>
                     Channel.take(self).map: a =>
                         drainEffect.map: chunk =>
-                            Emit.valueWith(Chunk(a).concat(chunk))(Loop.continue(()))
+                            Emit.valueWith(Chunk.single(a).concat(chunk))(Loop.continue(()))
 
         /** Stream elements from channel, optionally specifying a maximum chunk size. In the absence of [[maxChunkSize]], chunk sizes will
           * be limited only by channel capacity or the number of elements in the channel at a given time. (Chunks can still be larger than
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -334,7 +334,8 @@ object StreamCoreExtensions:
                             Loop(()): _ =>
                                 channelIn.take.map:
                                     case Absent => Loop.done
-                                    case Present(v) => f(v).map: v2 =>
+                                    case Present(v) =>
+                                        f(v).map: v2 =>
                                             channelOut.put(Present(v2)).andThen(Loop.continue)
                         }.andThen(channelOut.put(Absent))
 
@@ -487,7 +488,8 @@ object StreamCoreExtensions:
                             Loop(()): _ =>
                                 channelIn.take.map:
                                     case Absent => Loop.done
-                                    case Present(c) => f(c).map: c2 =>
+                                    case Present(c) =>
+                                        f(c).map: c2 =>
                                             channelOut.put(Present(c2)).andThen(Loop.continue)
                         }.andThen(channelOut.put(Absent))
 
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -9,7 +9,8 @@ object StreamCoreExtensions:
     private def emitMaybeChunksFromChannel[V](channel: Channel[Maybe[Chunk[V]]])(using Tag[V], Frame) =
         val emit = Loop(()): _ =>
             channel.take.map:
-                case Absent => Loop.done
+                case Absent =>
+                    Loop.done
                 case Present(c) =>
                     Emit.valueWith(c)(Loop.continue)
         Abort.run(emit).unit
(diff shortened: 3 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -1,32 +1,48 @@
...
                     Emit.valueWith(c)(Loop.continue)
         Abort.run(emit).unit
     end emitMaybeChunksFromChannel
 
+    private def emitMaybeElementsFromChannel[V](channel: Channel[Maybe[V]])(using Tag[V], Frame) =
+        val emit = Loop(()): _ =>
+            channel.take.map: v =>
+                channel.drain.map: chunk =>
+                    val fullChunk      = Chunk(v).concat(chunk)
+                    val publishedChunk = fullChunk.collect({ case Present(v) => v })
```
fwbrasil, line 24: The transformation of chunks here isn't very efficient. What do you think about using a `ChunkBuilder?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -189,6 +195,333 @@ object StreamCoreExtensions:
...
+                            )
+
+                            Abort.fold[E | Closed](
+                                _ => Abort.run(channel.put(Absent)).unit,
+                                {
+                                    case _: Closed       => bug("buffer closed unexpectedly")
+                                    case e: E @unchecked => Abort.run(channel.put(Absent)).andThen(Abort.fail(e))
+                                },
+                                e => Abort.run(channel.put(Absent)).andThen(Abort.panic(e))
+                            )(handledStream)
```
fwbrasil, line 282: readability improvement
```suggestion
                            Abort.fold[E | Closed](
                                onSuccess = _ => Abort.run(channel.put(Absent)).unit,
                                onFail = {
                                    case _: Closed       => bug("buffer closed unexpectedly")
                                    case e: E @unchecked => Abort.run(channel.put(Absent)).andThen(Abort.fail(e))
                                },
                                onPanic = e => Abort.run(channel.put(Absent)).andThen(Abort.panic(e))
                            )(handledStream)
```

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -189,6 +195,333 @@ object StreamCoreExtensions:
...
+                                                    val nextParSection       = remainingChunk.take(adjustedRemainingPar)
+
+                                                    val nextChunkEffect = Async.foreach(nextParSection)(f)
+
+                                                    val newRemainingPar   = adjustedRemainingPar - nextParSection.size
+                                                    val newRemainingChunk = remainingChunk.drop(adjustedRemainingPar)
+
+                                                    if newRemainingPar <= 0 && newRemainingChunk.size <= 0 then
+                                                        nextChunkEffect.map: nextChunk =>
+                                                            prevChunkFiber.get.map: prevChunk =>
```
fwbrasil, line 241:
```suggestion
                                                            prevChunkFiber.use: prevChunk =>
```
`use` is more lightweight. In performance-sensitive code, it's always best to use it, even if the value is discarded (`.use(_ => ...)`). There are other instances in the code.

kyo-core/shared/src/test/scala/kyo/StreamCoreExtensionsTest.scala
```diff
@@ -104,6 +104,154 @@ class StreamCoreExtensionsTest extends Test:
...
+
+        val randomSleep = Random.nextInt(10).map(i => Async.sleep(i.millis))
+
+        "mapPar" - {
+            "should map all elements preserving order" in run {
+                val stream = Stream.init(1 to 4).concat(Stream.init(5 to 8)).concat(Stream.init(9 to 12))
+                val test =
+                    for
+                        par <- Choice.get(Seq(1, 2, 4, Async.defaultConcurrency))
+                        buf <- Choice.get(Seq(1, 4, 5, 8, 12))
```
fwbrasil, line 116:
```suggestion
                        par <- Choice.get(Seq(1, 2, 4, Async.defaultConcurrency, Int.MaxValue))
                        buf <- Choice.get(Seq(1, 4, 5, 8, 12, par, Int.MaxValue))
```
increasing coverage

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -189,6 +195,333 @@ object StreamCoreExtensions:
...
+        def mapPar[V2, S2](parallel: Int, bufferSize: Int = defaultAsyncStreamBufferSize)(f: V => V2 < (Abort[E] & Async & S2))(
+            using
+            t1: Tag[Emit[Chunk[V]]],
+            t2: Tag[Emit[Chunk[V2]]],
+            t3: Tag[V2],
+            i1: Isolate.Contextual[S & S2, IO],
+            i2: Isolate.Stateful[S & S2, Abort[E] & Async],
+            ev: SafeClassTag[E | Closed],
+            frame: Frame
+        ): Stream[V2, Abort[E] & Async & S & S2] =
```
fwbrasil, line 217: I still sometimes get impressed to see Kyo's effects all playing well together like here. Cool work!

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -189,6 +195,333 @@ object StreamCoreExtensions:
...
+            Stream[V2, S & S2 & Abort[E] & Async]:
+                Channel.initWith[Maybe[V]](bufferSize): channelIn =>
+                    Channel.initWith[Maybe[V2]](bufferSize): channelOut =>
```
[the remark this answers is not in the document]
fwbrasil: Agreed, we need to review the `init` methods to include `Resource` but it'd be nice to still provide APIs that don't include `Resource`. I think we'd need it for `Async.run` as well here?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: Given that this will only ever prepend `v` to a chunk once and is not building chunks by repeated concatenation, I suspect this is more efficent than using a `ChunkBuilder`, which would require iterating over the entire chunk that `v` is being prepended to.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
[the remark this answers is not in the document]
johnhungerford: done

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -1,32 +1,48 @@
...
                 case Present(c) =>
                     Emit.valueWith(c)(Loop.continue)
         Abort.run(emit).unit
     end emitMaybeChunksFromChannel
 
+    private def emitMaybeElementsFromChannel[V](channel: Channel[Maybe[V]])(using Tag[V], Frame) =
+        val emit = Loop(()): _ =>
+            channel.take.map: v =>
+                channel.drain.map: chunk =>
+                    val fullChunk      = Chunk(v).concat(chunk)
```
hearnadam, line 23:
```suggestion
                    val fullChunk      = Chunk.single(v).concat(chunk)
```

kyo-core/jvm/src/test/scala/kyo/PathTest.scala
```diff
@@ -90,7 +90,7 @@ class PathTest extends Test:
                 _ <- useFile(first, text)
                 _ <- useFile(second, text)
                 v <- root.readAll("txt")
-            yield assert(v == List(first -> text, second -> text))
+            yield assert(v.toSet == Set(first -> text, second -> text))
```
johnhungerford, line 93: @fwbrasil this has been breaking in CI -- are you okay with the fix? I'm not sure what the behavior should be.

fwbrasil: hmm I don't see anything on our side that would shuffle the data coming from the Java file APIs. I think it's ok to mirror the behavior of the underlying implementation. I'll push the fix to main to avoid issues in PRs.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
fwbrasil: My concern is more the `collect` call since it'll use Scala's implementation, which can be expensive. I think it's fine to optimize it later, though

johnhungerford: @hearnadam is this a concern? At this point, doesn't `Chunk` use `ChunkBuilder` by default to transformations via `StrictOptimizedSeqOps`?

fwbrasil: hmm good point! makes sense

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -248,7 +248,7 @@ object Channel:
                 Loop[Unit, Unit, Abort[Closed] & Async & Emit[Chunk[A]]](()): _ =>
                     Channel.take(self).map: a =>
                         drainEffect.map: chunk =>
-                            Emit.valueWith(Chunk(a).concat(chunk))(Loop.continue)
+                            Emit.valueWith(Chunk.single(a).concat(chunk))(Loop.continue)
```
fwbrasil, line 251: I've opened https://github.com/getkyo/kyo/pull/1161 so we don't need to explicitly call `Chunk.single`

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: If not I can add it to `Chunk` in a separate PR

fwbrasil (approved):

Issue #1124, opened 2025-04-02
[feature]: Sink/Pipe


johnhungerford:
### Problem Statement

I've found ZIO's `ZSink` and `ZPipeline` useful abstractions for manipulating streams. It would be nice to have something like this in Kyo.

### Proposed Solution

Now that Kyo has `Poll` to complement `Emit`, making equivalent types in Kyo is trivial. Each, like `Stream`, can just wrap an `Emit` or `Poll` effect or both:

- `Stream[V, S]` wraps `Unit < (Emit[Chunk[V]] & S)` <-- emits chunks of `V` without producing final value
- `Sink[V, A, S]` wraps `A < (Poll[Chunk[V]] & S)` <-- produces value `A` from consuming chunks of `V`
- `Pipe[V1, V2, S]` wraps `Unit < (Poll[Chunk[V1]] & Emit[Chunk[V2]] & S)` <-- consumes chunks of `V1` and emits chunks of `V2` without producing final value

### Alternative Solutions

N/A -- I think this is pretty straightforward. More a question of how much we want it.

### Current Workaround

Anything a `Pipe` or `Sink` is needed for can be accomplished just by handling `Stream` using existing methods. (E.g., `Stream#map` for `Pipe`, and `Stream#foldLeft` or `Stream#foreach` for `Sink`.)

hearnadam:
I think a Sink is a useful abstraction, especially for libraries/sharing code.

HollandDM:
@hearnadam I remember you had some works about pipe stream this in the past, right?

johnhungerford:
I've being playing around with this. One question is whether we want to have some version of `ZSink`'s "leftover" `L` type. I'm inclined against it for two reasons:
1. I have used `ZStream`, `ZPipeline`, and `ZSink` pretty extensively and have never once used `L`.
2. Whether or not there even *could* be leftovers is a fairly complicated function of the combined behavior of the stream and the sink, and is therefore hard to really capture in a type signature.
3. It is easy enough to reproduce the functionality of a result value and a leftover value without a separate `L` channel. (See below)

[...]

PR #1157, opened 2025-04-24
Sink

johnhungerford:
Addresses #1124 

### Problem

It would be useful to have an abstraction for processing streams similar to ZIO's `ZSink`

### Solution

Added a `Sink[V, A, S]` type, which is a wrapper of `A < (Poll[Chunk[V]] & S)`, as discussed in #1124 

Combinators:
- `contramap`
- `contramapChunk`
- `map`
- `zip`

Constructors:
- `drain`
- `collect`
- `count`
- `fold`
- `foldKyo`
- `foreach`
- `foreachChunk`

### Notes

I have not included any facility for "leftover" streams as discussed in #1124 as this has turned out to be substantially more complicated than I thought. Basically, since `Stream` and `Sink` emit and poll chunks of values respectively, remaining chunks would have to be re-emitted after being handled by the `poll` function in order to support sinks that terminate early.

This would break the symmetry of `Stream` and `Poll`, however, and would make methods like `contramap` impossible. For the time being, I recommend treating `Sink` as total. If there is desire for partial sinks, we can revisit. I personally have never used sinks partially and so don't have a good sense of need for them.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Discriminator.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Discriminator.scala
@@ -1,0 +1,9 @@
+package kyo
+
+/** A dummy class used to use to help the compiler distinguish between overloaded methods that have subtle type differences.
+  */
+sealed abstract class Discriminator {}
+
+object Discriminator:
+    private val cached  = new Discriminator {}
+    given Discriminator = cached
--- a/kyo-prelude/shared/src/main/scala/kyo/Sink.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Sink.scala
@@ -231,7 +231,7 @@ object Sink:
                 Poll.andMap[Chunk[V]]:
                     case Absent => Loop.done(currentChunk)
                     case Present(c) =>
-                        Loop.continue(currentChunk ++ c)
+                        Loop.continue(currentChunk.concat(c))
     end collect
 
     /** Construct a sink that counts streaming elements.
--- a/kyo-prelude/shared/src/main/scala/kyo/Sink.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Sink.scala
@@ -57,8 +57,14 @@ sealed abstract class Sink[V, A, -S] extends Serializable:
                             )
                         ,
                         done = a =>
-                            pollB.map: b =>
-                                Loop.done((a, b))
+                            ArrowEffect.handleFirst(tag, pollB)(
+                                handle = [C] =>
+                                    (_, contB) =>
+                                        contB(polledValue).map: b =>
+                                            Loop.done((a, b)),
+                                done = b =>
+                                    Loop.done((a, b))
+                            )
                     )
     end zip
 
--- a/kyo-prelude/shared/src/main/scala/kyo/Sink.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Sink.scala
@@ -103,7 +103,7 @@ sealed abstract class Sink[V, A, -S] extends Serializable:
         t1: Tag[Poll[Chunk[V]]],
         t2: Tag[Poll[Chunk[V2]]],
         ev: NotGiven[V2 <:< (Any < Nothing)],
-        d: Sink.Dummy,
+        d: Discriminator,
         fr: Frame
     ): Sink[V2, A, S & S2] =
         Sink:
@@ -156,7 +156,7 @@ sealed abstract class Sink[V, A, -S] extends Serializable:
         t1: Tag[Poll[Chunk[V]]],
         t2: Tag[Poll[Chunk[V2]]],
         ev: NotGiven[V2 <:< (Any < Nothing)],
-        d: Sink.Dummy,
+        d: Discriminator,
         fr: Frame
     ): Sink[V2, A, S & S2] =
         Sink:
@@ -446,11 +446,5 @@ object Sink:
     ): Sink[V, (A, B, C, D, E, F, G, H, I, J), S] =
         a.zip(zip(b, c, d, e, f, g, h, i, j)).map:
             case (a, tail) => a *: tail
-
-    /** A dummy type that can be used as implicit evidence to help the compiler discriminate between overloaded methods.
-      */
-    sealed class Dummy extends Serializable
-    object Dummy:
-        given Dummy = new Dummy {}
 
 end Sink
--- a/kyo-prelude/shared/src/main/scala/kyo/Sink.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Sink.scala
@@ -5,7 +5,6 @@ import kyo.kernel.ArrowEffect
 import kyo.kernel.internal.WeakFlat
 import scala.annotation.nowarn
 import scala.annotation.targetName
-import scala.util.NotGiven
 
 /** Processes a stream of type `V`, producing a value of type `A`.
   *
@@ -80,7 +79,6 @@ sealed abstract class Sink[V, A, -S] extends Serializable:
     final def contramap[V2](f: V2 => V)(using
         t1: Tag[Poll[Chunk[V]]],
         t2: Tag[Poll[Chunk[V2]]],
-        ev: NotGiven[V2 <:< (Any < Nothing)],
         fr: Frame
     ): Sink[V2, A, S] =
         Sink:
@@ -102,7 +100,6 @@ sealed abstract class Sink[V, A, -S] extends Serializable:
     final def contramap[V2, S2](f: V2 => V < S2)(using
         t1: Tag[Poll[Chunk[V]]],
         t2: Tag[Poll[Chunk[V2]]],
-        ev: NotGiven[V2 <:< (Any < Nothing)],
         d: Discriminator,
         fr: Frame
     ): Sink[V2, A, S & S2] =
@@ -128,7 +125,6 @@ sealed abstract class Sink[V, A, -S] extends Serializable:
     final def contramapChunk[V2](f: Chunk[V2] => Chunk[V])(using
         t1: Tag[Poll[Chunk[V]]],
         t2: Tag[Poll[Chunk[V2]]],
-        ev: NotGiven[V2 <:< (Any < Nothing)],
         fr: Frame
     ): Sink[V2, A, S] =
         Sink:
@@ -151,7 +147,6 @@ sealed abstract class Sink[V, A, -S] extends Serializable:
     final def contramapChunk[V2, S2](f: Chunk[V2] => Chunk[V] < S2)(using
         t1: Tag[Poll[Chunk[V]]],
         t2: Tag[Poll[Chunk[V2]]],
-        ev: NotGiven[V2 <:< (Any < Nothing)],
         d: Discriminator,
         fr: Frame
     ): Sink[V2, A, S & S2] =
--- a/kyo-prelude/shared/src/main/scala/kyo/Sink.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Sink.scala
@@ -37,24 +37,29 @@ sealed abstract class Sink[V, A, -S] extends Serializable:
       * @return
       *   A new sink that produces a tuple of the outputs of the source sinks.
       */
-    def zip[B, S2](other: Sink[V, B, S2])(using Tag[V], Frame): Sink[V, (A, B), S & S2] =
+    def zip[B, S2](other: Sink[V, B, S2])(using tag: Tag[Poll[Chunk[V]]], f: Frame): Sink[V, (A, B), S & S2] =
         Sink:
             Loop((poll, other.poll)): (pollA, pollB) =>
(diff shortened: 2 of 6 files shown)


johnhungerford:
The only way I can think of to support partial sinks is if we introduce a leftover/remainder type à la `ZSink`. This is necessary to support `contramap`, since transforming the input element type does not transform the output element type, which means these must be different types.

Adding that would also complicate `zip`. My implementation of `Sink` treats `zip` differently from `ZSink`. Rather than running sinks in sequence, it runs them in tandem on the same streamed elements and tuples their respective results. This seems much more useful to me than the ZIO version, but would make leftovers somewhat tricky since Kyo streams doesn't support union types yet. (Ideally the leftover of zipped sinks would be the union of their leftover types, but it would have to be an `Either` of them to resolve the tag.)

There's also the problem that `L` would have to be `Nothing` for complete sinks (i.e. sinks that never leave any remaining stream), but Kyo's `Tag` is not supported for `Nothing`.

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,456 @@
...
+                    case Present(c) =>
+                        Loop.continue(currentChunk ++ c)
+    end collect
+
+    /** Construct a sink that counts streaming elements.
+      *
+      * @return
+      *   A sink that produces the count of all elements emitted by source stream
+      */
+    def count[V](using Tag[V], Frame): Sink[V, Int, Any] =
```
hearnadam, line 243: Could this be `Sink[Any, Int, Any]`?

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -623,6 +623,19 @@ sealed abstract class Stream[V, -S] extends Serializable:
...
+      *
+      * @see
+      *   [[kyo.Sink.consume]]
+      *
+      * @param sink
+      *   Sink to process stream with
+      * @return
+      *   An effect producing a value of sink's output type `A`
+      */
+    def into[A, S2](sink: Sink[V, A, S2])(using Tag[V], Frame): A < (S & S2) =
```
hearnadam, line 637: pre-existing: I think we need to mark all the methods `final` in this class.

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,456 @@
...
+    )(using
+        Tag[V],
+        Frame
+    ): Sink[V, (A, B, C, D, E, F, G, H, I, J), S] =
+        a.zip(zip(b, c, d, e, f, g, h, i, j)).map:
+            case (a, tail) => a *: tail
+
+    /** A dummy type that can be used as implicit evidence to help the compiler discriminate between overloaded methods.
+      */
+    sealed class Dummy extends Serializable
```
hearnadam, line 452: Since this is useful in multiple classes, what would you think about having a `Discriminator.scala` file?

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,456 @@
...
+  *   The type of values that this sink can process
+  * @tparam A
+  *   The type of the value generated when processing a stream of `V` s
+  * @tparam S
+  *   The type of effects used in processing the stream
+  *
+  * @see
+  *   [[kyo.Poll]] for the underlying pull-based consumption mechanism
+  */
+sealed abstract class Sink[V, A, -S] extends Serializable:
```
hearnadam, line 25: I think we should convert stream to be `Stream[A, S]` and this `Sink[A, B, S]`. It can be a followup since it will touch many locs.

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,456 @@
...
+      *
+      * @see
+      *   [[kyo.Stream.into]]
+      *
+      * @param stream
+      *   Stream to be processed
+      * @return
+      *   An effect generating an output value from elements consumed from the stream
+      */
+    def consume[S2](stream: Stream[V, S2])(using Tag[V], Frame): A < (S & S2) =
```
hearnadam, line 200: Can all the methods be final?

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
johnhungerford: Sure -- where do you think it should go? kyo-data?

johnhungerford: (added it to kyo-data -- let me know if you think it should go somewhere else)

kyo-data/shared/src/main/scala/kyo/Discriminator.scala
```diff
@@ -0,0 +1,9 @@
+package kyo
+
+/** A dummy class used to use to help the compiler distinguish between overloaded methods that have subtle type differences.
+  */
+sealed abstract class Discriminator {}
+
+object Discriminator:
+    private val cached  = new Discriminator {}
+    given Discriminator = cached
```
hearnadam, line 9: a `given` should be a `lazy val` by default, right?
```suggestion
    given Discriminator = new Discriminator {}
```

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,450 @@
...
+      * @return
+      *   An effect generating an output value from elements consumed from the stream
+      */
+    final def consume[S2](stream: Stream[V, S2])(using Tag[V], Frame): A < (S & S2) =
+        Poll.run(stream.emit)(poll).map(_._2)
+end Sink
+
+object Sink:
+    @nowarn("msg=anonymous")
+    inline def apply[V, A, S](inline v: => A < (Poll[Chunk[V]] & S)): Sink[V, A, S] =
```
hearnadam, line 206: Do you think this should be public? It seems to make somewhat less sense than in Stream.

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,450 @@
...
+            def poll: A < (Poll[Chunk[V]] & S) = v
+
+    /** Construct a sink that runs a stream of element type `V` without producing any value.
+      *
+      * @return
+      *   A sink that runs a stream without producing a value
+      */
+    def drain[V](using Tag[V], Frame): Sink[V, Unit, Any] =
+        Sink:
+            Loop(()): _ =>
```
hearnadam, line 217: `Loop.foreach`

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,450 @@
...
+    /** Construct a sink that applies an effectful function to each element of a stream, discarding all resulting values
+      *
+      * @param f
+      *   Effectful function with no result to apply to each streaming element
+      * @return
+      *   A sink that applies an effectful function to each element of the source stream
+      */
+    def foreach[V, S](f: V => Unit < S)(using Tag[V], Frame): Sink[V, Unit, S] =
+        Sink:
+            Loop(()): _ =>
```
hearnadam, line 261: `Loop.foreach`

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,450 @@
...
+    /** Construct a sink that applies an effectful function to each chunk of a stream, discarding all resulting values
+      *
+      * @param f
+      *   Effectful function with no result to apply to each streaming element
+      * @return
+      *   A sink that applies an effectful function to each chunk of the source stream
+      */
+    def foreachChunk[V, S](f: Chunk[V] => Unit < S)(using Tag[V], Frame): Sink[V, Unit, S] =
+        Sink:
+            Loop(()): _ =>
```
hearnadam, line 277: `Loop.foreach`

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,450 @@
...
+    inline def apply[V, A, S](inline v: => A < (Poll[Chunk[V]] & S)): Sink[V, A, S] =
+        new Sink[V, A, S]:
+            def poll: A < (Poll[Chunk[V]] & S) = v
+
+    /** Construct a sink that runs a stream of element type `V` without producing any value.
+      *
+      * @return
+      *   A sink that runs a stream without producing a value
+      */
+    def drain[V](using Tag[V], Frame): Sink[V, Unit, Any] =
```
hearnadam, line 215: How about `discard`?

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,450 @@
...
+      * @return
+      *   A sink that produces a chunk of all elements emitted by the source stream
+      */
+    def collect[V](using Tag[V], Frame): Sink[V, Chunk[V], Any] =
+        Sink:
+            Loop(Chunk.empty[V]): currentChunk =>
+                Poll.andMap[Chunk[V]]:
+                    case Absent => Loop.done(currentChunk)
+                    case Present(c) =>
+                        Loop.continue(currentChunk ++ c)
```
hearnadam, line 235: Can you use `concat`? I think we need to update `Chunk` override `++` to be `concat`

hearnadam (commented):
Nice! This will be a great first cut. We can figure out the remainder later imo.

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,450 @@
...
+      *
+      * @see
+      *   [[kyo.Stream.into]]
+      *
+      * @param stream
+      *   Stream to be processed
+      * @return
+      *   An effect generating an output value from elements consumed from the stream
+      */
+    final def consume[S2](stream: Stream[V, S2])(using Tag[V], Frame): A < (S & S2) =
```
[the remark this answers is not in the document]
johnhungerford: Yeah that's better. Do you think it's an issue that `drain` is also the name of the sink that runs a stream without returning a value? (So you might have `Sink.drain.drain(someStream)`

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
johnhungerford: I don't see why we shouldn't expose the `Poll` API to the user. This is equivalent to ZIO's exposing `ZChannel`: most users don't use it, but it's necessary for some complicated use cases. (I've used it.)

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
johnhungerford: ah didn't catch that -- yes

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,449 @@
...
+      *   A new sink that processes streams of the new element type
+      */
+    final def contramap[V2](f: V2 => V)(using
+        t1: Tag[Poll[Chunk[V]]],
+        t2: Tag[Poll[Chunk[V2]]],
+        ev: NotGiven[V2 <:< (Any < Nothing)],
+        fr: Frame
+    ): Sink[V2, A, S] =
+        Sink:
+            ArrowEffect.handleState(t1, (), poll)(
```
hearnadam, line 87: Should these be using `handleLoop` now?

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
```diff
@@ -0,0 +1,449 @@
...
+      *
+      * @param f
+      *   Effectful function mapping new stream element type to original stream element type
+      * @return
+      *   A new sink that processes streams of the new element type
+      */
+    final def contramap[V2, S2](f: V2 => V < S2)(using
+        t1: Tag[Poll[Chunk[V]]],
+        t2: Tag[Poll[Chunk[V2]]],
+        ev: NotGiven[V2 <:< (Any < Nothing)],
```
hearnadam, line 105: Is this necessary?

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
johnhungerford: Good point -- just updated

kyo-prelude/shared/src/main/scala/kyo/Sink.scala
johnhungerford: Yeah just updated

fwbrasil (approved):

PR #1161, opened 2025-04-26
[data] introduce optimized override for Chunk.apply

fwbrasil:
### Problem

We recently introduced the `Single` version of `Chunk`. It's under the `private[kyo]` internal object so it shouldn't leak to users. Also, we'd need to migrate all single-elements calls to `Chunk.apply` to `Chunk.single` to leverage the optimization, which isn't ideal.

### Solution

Don't expose `Single` in user-facing APIs and override `Chunk.apply` to create `Single` internally in case a single element is passed.


hearnadam:
This is actually already the default. I think we can just remove `Chunk.single`

https://github.com/scala/scala/blob/cdb1a1d240c1ce79aa92429e0346964c2a864ee1/src/library/scala/collection/Factory.scala#L103

hearnadam (approved):

PR #1164, opened 2025-04-27
[core] add Stream's `unfoldKyo`

HollandDM:
### Solution
This PR introduce some method for `Stream[V, S]`
- [x] `duplicate`: which duplicate into `n` copies which respect to the original stream mutability
- [x] `unfoldKyo`: repeatedly build a `Stream[V, S]` using an effect

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
@@ -637,27 +637,18 @@ sealed abstract class Stream[V, -S] extends Serializable:
       *   A tuple containing chunk of the first n elements and the rest of the stream
       */
     def splitAt(n: Int)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): (Chunk[V], Stream[V, S]) < S =
-        val emptyEmit = Maybe.empty[Unit < (Emit[Chunk[V]] & S)]
-        ArrowEffect.handleLoop(tag, (Chunk.empty[V], emptyEmit), emit)(
-            handle = [C] =>
-                (input, state, cont) =>
-                    val (chunk, _)    = state
-                    val appendedChunk = chunk.concat(input)
-                    if (appendedChunk.size) < n then
-                        Loop.continue(appendedChunk -> emptyEmit, cont(()))
+        Loop(emit, Chunk.empty[V]): (curEmit, curChunk) =>
+            Emit.runFirst(curEmit).map:
+                case (Present(items), nextEmitFn) =>
+                    val nextChunk = curChunk.concat(items)
+                    if nextChunk.length < n then
+                        Loop.continue(nextEmitFn(), nextChunk)
                     else
-                        val (taken, rest) = appendedChunk.splitAt(n)
-                        val restEmit      = Maybe.Present(Emit.valueWith(rest)(cont(())))
-                        Loop.continue(taken -> restEmit, Kyo.unit)
+                        val (taken, rest) = nextChunk.splitAt(n)
+                        val restEmitFn    = () => Emit.valueWith(rest)(nextEmitFn())
+                        Loop.done(taken -> Stream(restEmitFn()))
                     end if
-            ,
-            done = (state, _) =>
-                val (chunk, lastEmit) = state
-                lastEmit match
-                    case Maybe.Present(emit) => (chunk, Stream(emit))
-                    case Maybe.Absent        => (chunk, Stream.empty)
-                end match
-        )
+                case (Absent, _) => Loop.done(curChunk -> Stream(Emit.value(Chunk.empty[V])))
     end splitAt
 
     /** Process with a [[Sink]] of corresponding streaming element type.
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -190,6 +191,148 @@ object StreamCoreExtensions:
...
+    extension [V, S](stream: Stream[V, S])
+
+        /** Creates multiple copies of this stream that can be consumed independently
+          *
+          * @param copies
+          *   Number of copies to create
+          * @return
+          *   A chunk containing the specified number of independent stream copies
+          */
+        def duplicate(copies: Int)(using
```
hearnadam, line 204: Would this be similar to ZStream broadcast? Would it be possible to implement this in terms of Hub?

What's the primary usecase?

HollandDM: Now you talk about it, it's true that this is similar to ZStream broadcast, but without the maximumLag param.
The primary use case I think is to implement `partition` and `partitionMap` for `Stream`

johnhungerford: This seems like an unnecessarily complicated implementation for broadcast. Can't you just use `Hub`?

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -756,6 +748,37 @@ object Stream:
...
+      *   The type of effects in the stream
+      * @param acc
+      *   The initial accumulator value
+      * @param chunkSize
+      *   The target size for each chunk (must be positive)
+      * @param f
+      *   A function that takes the current accumulator and returns Maybe of next value and accumulator
+      * @return
+      *   A stream containing the unfolded values
+      */
```
johnhungerford, line 767: Docs should mention somewhere that `Absent` signals end-of-stream

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -756,6 +748,37 @@ object Stream:
                     }
                 }
 
+    /** Creates a stream by repeatedly applying a function to accumulate values.
```
johnhungerford, line 751: Nice!

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -637,28 +637,20 @@ sealed abstract class Stream[V, -S] extends Serializable:
       *   A tuple containing chunk of the first n elements and the rest of the stream
       */
     def splitAt(n: Int)(using tag: Tag[Emit[Chunk[V]]], frame: Frame): (Chunk[V], Stream[V, S]) < S =
-        val emptyEmit = Maybe.empty[Unit < (Emit[Chunk[V]] & S)]
```
johnhungerford, line 640: What was wrong with the original implementation?

HollandDM: Nothing wrong, I think it'll be clearer this way. I implemented the first version when not knowing about `Emit.runFirst`

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
HollandDM: Yeah, maybe I'll remove this. It seems this is not needed that much

HollandDM:
Updated to only add `unfoldKyo` for now

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -637,27 +637,18 @@ sealed abstract class Stream[V, -S] extends Serializable:
...
+                    val nextChunk = curChunk.concat(items)
+                    if nextChunk.length < n then
+                        Loop.continue(nextEmitFn(), nextChunk)
                     else
-                        val (taken, rest) = appendedChunk.splitAt(n)
-                        val restEmit      = Maybe.Present(Emit.valueWith(rest)(cont(())))
-                        Loop.continue(taken -> restEmit, Kyo.unit)
+                        val (taken, rest) = nextChunk.splitAt(n)
+                        val restEmitFn    = () => Emit.valueWith(rest)(nextEmitFn())
+                        Loop.done(taken -> Stream(restEmitFn()))
```
hearnadam, line 649:
```suggestion
                        val rest    = Stream(Emit.valueWith(rest)(nextEmitFn()))
                        Loop.done(taken -> rest)
```

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -769,4 +760,41 @@ object Stream:
...
+                if curChunk.length == _chunkSize then
+                    Emit.valueWith(curChunk)(Loop.continue(Chunk.empty[V], curAcc))
+                else
+                    f(curAcc).map:
+                        case Present(value -> nextAcc) => Loop.continue(curChunk.append(value), nextAcc)
+                        case Absent                    => Emit.valueWith(curChunk)(Loop.done(()))
+
+    /** A dummy type that can be used as implicit evidence to help the compiler discriminate between overloaded methods.
+      */
+    sealed class Dummy extends Serializable
```
hearnadam, line 796: https://github.com/getkyo/kyo/blob/main/kyo-data/shared/src/main/scala/kyo/Discriminator.scala

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: You can also skip the emission in cases where `rest` is empty:
```scala
val remaining = Stream(if rest.isEmpty then nextEmitFn() else Emit.valueWith(rest)(nextEmitFn()))
Loop.done(taken -> remaining)
```
(@hearnadam 's suggested change has a double definition)

kyo-prelude/shared/src/test/scala/kyo/StreamTest.scala
```diff
@@ -44,6 +44,15 @@ class StreamTest extends Test:
             assert(size(301, 100) == Chunk(100, 100, 100, 1))
             assert(size(5, 0) == Chunk(1, 1, 1, 1, 1))
         }
+
+        "unfoldKyo" in {
```
johnhungerford, line 48: Test should probably include effects, since this is an effectful variant

HollandDM:
I've updated the feedbacks

fwbrasil (approved):

PR #1166, opened 2025-04-28
Pipe

johnhungerford:
Completes #1124 

### Problem

It would be good to have a Kyo version of ZIO's `ZPipeline`

### Solution

As discussed in #1124 , added `Pipe[A, B, S]`, a wrapper around `Unit < (Poll[Chunk[A]] & Emit[Chunk[B]] & S)`.

Combinators:

- contramap
- contramapChunk
- map
- mapChunk
- join (pipe1.join(pipe2) -> pipe, pipe.join(sink) -> sink)
- transform

Constructors:

- empty
- identity
- map
- mapChunk
- take
- drop
- takeWhile
- dropWhile
- filter
- collect
- collectWhile
- changes
- rechunk
- tap
- tapChunk

You can now process a stream via a sequence of pipes terminating with a sink:

`val processed = stream.into(pipe1).into(pipe2).into(pipe3).into(sink)`

### Notes

- `Pipe.transform` and `Sink.drain` no longer us `Poll.run` since this method handled `Emit` prior to `Poll`. `Pipe` and `Sink` delegate control to their own `Poll` effects, allowing early termination.
- Tests have been copied from `Stream`, since current `Pipe`s are all based on `Stream` methods. Users should expect identical behavior.


kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -660,10 +661,23 @@ sealed abstract class Stream[V, -S] extends Serializable:
...
+      *
+      * @see
+      *   [[kyo.Pipe.transform]]
+      *
+      * @param sink
+      *   Pipe to transform stream with
+      * @return
+      *   A new stream of transformed element type `A`
+      */
+    def into[A, S2](pipe: Pipe[V, A, S2])(using Tag[V], Frame): Stream[A, S & S2] =
```
fwbrasil, line 674: how about `map` instead of `into`? `Pipe` is similar to a `map` function

kyo-prelude/shared/src/main/scala/kyo/Pipe.scala
```diff
@@ -0,0 +1,721 @@
...
+        using
+        t1: Tag[Emit[Chunk[B]]],
+        t2: Tag[B1],
+        fr: Frame
+    ): Pipe[A, B1, S] =
+        Pipe:
+            ArrowEffect.handleLoop(t1, pollEmit)(
+                [C] =>
+                    (chunk, cont) =>
+                        Emit.valueWith(chunk.map(f))(Loop.continue(cont(())))
```
fwbrasil, line 142: Is there a way we could avoid the code duplication with `Stream`? Looking at [Stream.map](https://github.com/getkyo/kyo/blob/0e91844fa55cb8b32289d3283965aa1f673fa8e8/kyo-prelude/shared/src/main/scala/kyo/Stream.scala#L67), it seems this code is missing avoiding emits in case the chunk is empty. We might end up with multiple mismatches between the implementations. Maybe internal APIs that are used by both `Stream` and `Pipe` to avoid the `Pipe` allocation? I wonder if we're missing an effect in `kyo-prelude`.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: It's similar but different in important ways. `map` makes it seem like each element will be transformed, which is not necessarily the case. The input and output streams can be entirely different lengths, for instance.

kyo-prelude/shared/src/main/scala/kyo/Pipe.scala
fwbrasil: Thank you for the elaboration! I'd be ok with some tech debt due to code duplication. What do you think is the best path? With the 1.0-RC cycle coming up, I'd say it's defining the APIs we want to have is more relevant than accumulating some tech debt in in internal code. Performance is similar, it's a major focus of the project but we can  typically optimize things later

kyo-prelude/shared/src/main/scala/kyo/Pipe.scala
```diff
@@ -0,0 +1,721 @@
...
+  *   Input stream element type
+  * @tparam B
+  *   Output stream element type
+  * @tparam S
+  *   The type of effects associated with the transformation
+  */
+sealed abstract class Pipe[A, B, -S] extends Serializable:
+
+    /** Returns the underlying effect that polls chunks of type A and emits chunks of type B */
+    def pollEmit: Unit < (Poll[Chunk[A]] & Emit[Chunk[B]] & S)
```
hearnadam, line 29: I am slightly worried about the complexity/overhead here, but I think it should be okay for a first cut. It would be great if we could have comparative benchmarks between using `Pipe` and using the direct stream operators.

kyo-prelude/shared/src/main/scala/kyo/Pipe.scala
```diff
@@ -0,0 +1,721 @@
...
+            )
+
+    /** Join to another pipe producing a new pipe that performs both pipes' transformations in sequence.
+      *
+      * @param pipe
+      *   Pipe to prepend pipe to
+      * @return
+      *   New pipe that performs both transformations in sequence
+      */
+    def join[C, S1](pipe: Pipe[B, C, S1])(using Tag[B], Frame): Pipe[A, C, S & S1] =
```
hearnadam, line 219: Would this make sense as `andThen`? I think we have some other APIs with that naming.

kyo-prelude/shared/src/main/scala/kyo/Pipe.scala
```diff
@@ -0,0 +1,721 @@
...
+      *
+      * @param f
+      *   Pure function mapping new stream element type to original stream element type
+      * @return
+      *   A pipe that accepts streams of the new element type
+      */
+    def contramap[A1](f: A1 => A)(
+        using
+        t1: Tag[Poll[Chunk[A]]],
+        t2: Tag[A1],
```
hearnadam, line 41: I think you will need to merge mainline, then fix the usage of dynamic tags.

kyo-prelude/shared/src/main/scala/kyo/Pipe.scala
johnhungerford: If it's okay with you and @hearnadam I'd like to continue with the duplicated code for this PR. I'll make a benchmark to compare performance between the original stream methods and the pipe methods. If there isn't a performance loss, I'll just abstract the pipe implementations and they can be shared by stream and pipe. If there is, I'll consider other approaches.

kyo-prelude/shared/src/main/scala/kyo/Pipe.scala
johnhungerford: Agreed

kyo-prelude/shared/src/main/scala/kyo/Pipe.scala
johnhungerford: I prefer not to re-use extension method names as this can create annoying conflicts (`Pipe[A, B, S]` can be lifted to `Pipe[A, B, S] < Any`, the `andThen` extension on `<` could be accessed).

johnhungerford:
Looks like the failed build is unrelated to this PR

ahoy-jon:
@johnhungerford it's a draft, or ready to review + merge? (it looks good to me)

kyo-prelude/shared/src/main/scala/kyo/Pipe.scala
```diff
@@ -0,0 +1,727 @@
...
+        Pipe:
+            ArrowEffect.handleLoop(t1, pollEmit)(
+                [C] =>
+                    (unit, cont) =>
+                        Poll.andMap[Chunk[A1]]:
+                            case Absent => Loop.continue(cont(Absent))
+                            case Present(chunk) =>
+                                Kyo.foreach(chunk)(f).map: chunk2 =>
+                                    Loop.continue(cont(Present(chunk2)))
+            )
```
ahoy-jon, line 75: I guess it's going to be `contramapKyo`

ahoy-jon (approved):

fwbrasil (approved):
sorry for the delay to review

PR #1169, opened 2025-04-30
[stm] retry Abort when transaction is inconsistent

### LINKED ISSUE #1158: [BUG]: STM reading inconsistent values
### Version

0.18.0

### Scala Version

3.6.4

### Expected Behavior

Since the 2 `TRef`s initially contain equal values, and I only ever write into them equal values, I expect never to observe non-equal values in a single transaction. That is, I expect the program to exit without an exception.

### Actual Behavior

The program exits with an exception like this:
```
Exception in thread "main" java.lang.AssertionError: a111111111111111111111111 != a1111111111111111111111111
```

### Steps to Reproduce

```scala
object StmTest extends KyoApp {
  run {
    for {
      // initially they contain equal values:
      r1 <-STM.run(TRef.init("a"))
      r2 <-STM.run(TRef.init("a"))
      txn1 = for {
        v1 <- r1.get
        v2 <- r2.get
        // we only write equal values:
        _ <- r1.set(v1 + "1")
        _ <- r2.set(v2 + "1")
      } yield ()
      txn2 = r1.get.map { v1 =>
        r2.get.map { v2 =>
          if (v1 == v2) {
            ()
          } else {
            // observed non-equal values:
            Abort.panic(new AssertionError(s"${v1} != ${v2}")) // <- fails here
          }
        }
      }
      once = Async.zip(STM.run(txn1), STM.run(txn2))
      _ <- Async.foreachDiscard(1 to 100000, concurrency = 8) { _ => once }
    } yield ()
  }
}
```

### Current Workaround

_No response_

fwbrasil:
Fixes #1158

### Problem

When a transaction is rolled back due to an `Abort`, the `STM` effect simply propagates the error. This can be a surprising behavior when the `Abort` happened due to an inconsistent state during the transaction execution, which is an expected scenario given that consistency is checked at commit time. 

### Solution

In case of errors, check if the ref log is still valid and, if not, wrap the error into `FailedTransaction` to trigger a retry in case the schedule permits.

### Notes

- I've added `Abort.recoverError` and `Abort.foldError` since it seems a relatively common pattern.

### DIFF
--- a/kyo-stm/shared/src/main/scala/kyo/STM.scala
+++ b/kyo-stm/shared/src/main/scala/kyo/STM.scala
@@ -251,12 +251,12 @@ end STM
 
 /** A FailedTransaction exception that is thrown when a transaction fails to commit. Contains the frame where the failure occurred.
   */
-class FailedTransaction(error: Maybe[Result.Error[?]] = Absent)(using Frame)
+final class FailedTransaction(error: Maybe[Result.Error[?]] = Absent)(using Frame)
     extends KyoException(
         s"STM transaction failed!",
         error.fold("") {
             _.failureOrPanic match
                 case ex: Throwable => ex
-                case _             => s"Error: ${error.show}"
+                case _             => error.show
         }
     )
--- a/kyo-stm/shared/src/main/scala/kyo/STM.scala
+++ b/kyo-stm/shared/src/main/scala/kyo/STM.scala
@@ -163,9 +163,9 @@ object STM:
                     var i = 0
                     // Pre-validate and dump the log to the flat array
                     logMap.foreachEntry { (ref, entry) =>
-                        // This code uses exception throwing because
+                        // This code uses `boundary`/`break` because
                         // foreachEntry is the only way to traverse the
-                        // map without allocating tuples, so throwing
+                        // map without allocating tuples, so throwing via `break`
                         // is the workaround to short circuit
                         if !ref.validate(entry) then boundary.break(false)
                         array(i) = ref
--- a/kyo-stm/shared/src/main/scala/kyo/STM.scala
+++ b/kyo-stm/shared/src/main/scala/kyo/STM.scala
@@ -2,10 +2,7 @@ package kyo
 
 import kyo.Result.Failure
 import scala.annotation.tailrec
-
-/** A FailedTransaction exception that is thrown when a transaction fails to commit. Contains the frame where the failure occurred.
-  */
-class FailedTransaction()(using Frame) extends KyoException
+import scala.util.boundary
 
 /** Software Transactional Memory (STM) provides concurrent access to shared state using optimistic locking. Rather than acquiring locks
   * upfront, transactions execute speculatively and automatically retry if conflicts are detected during commit. While this enables better
@@ -70,7 +67,7 @@ object STM:
       * @return
       *   The result of the computation if successful
       */
-    def run[E, A, S](
+    def run[E: SafeClassTag, A, S](
         using Isolate.Stateful[S, Async & Abort[E | FailedTransaction]]
     )(v: A < (STM & Abort[E] & Async & S))(using frame: Frame): A < (S & Async & Abort[E | FailedTransaction]) =
         run(defaultRetrySchedule)(v)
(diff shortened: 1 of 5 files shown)


kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -202,6 +137,81 @@ object STM:
 
     end run
 
+    private def commit[A, S](tid: Long, log: TRefLog, probe: Boolean = false)(using AllowUnsafe): Boolean =
```
fwbrasil, line 140: The overall logic here is still the same. I've refactored it into a new method that returns a `Boolean` indicating if the commit was successful, introduced `boundary`/`break` to replace the exception throwing, and add a `probe` parameter to indicate that the commit should only validate and not write values to refs

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -2,10 +2,7 @@ package kyo
 
 import kyo.Result.Failure
 import scala.annotation.tailrec
-
-/** A FailedTransaction exception that is thrown when a transaction fails to commit. Contains the frame where the failure occurred.
-  */
-class FailedTransaction()(using Frame) extends KyoException
```
fwbrasil, line 8: moved to the end of the file

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -93,100 +90,38 @@ object STM:
...
-                                        end try
-                                    }
-                            end match
+                TID.useNew { tid =>
+                    v.handle(
+                        Abort.recoverError[E] { error =>
+                            // Retry arbitrary E failures in case the transaction is inconsistent
+                            Var.use[TRefLog] { log =>
+                                IO.Unsafe {
+                                    if !commit(tid, log, probe = true) then
```
fwbrasil, line 104: note `probe = true` to just probe the commit without updating refs

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -202,6 +137,81 @@ object STM:
...
+                // Commit multiple refs
+                // Flattened representation of the log
+                val array = new Array[Any](size * 2)
+
+                boundary {
+
+                    var i = 0
+                    // Pre-validate and dump the log to the flat array
+                    logMap.foreachEntry { (ref, entry) =>
+                        // This code uses exception throwing because
```
hearnadam, line 166: minor: could use an update.

durban:
I've tried this branch, and I can confirm that it solves my original problem in #1158. (It doesn't seem to provide full opacity, but as far as I can tell, it doesn't try to, so that's fine.)

One question: out of curiosity, I've tried replacing `Abort.panic(new AssertionError(...))` in my original test in #1158 with `throw new AssertionError(...)`. (I was interested in what would happen with a "random exception" thrown due to observing an inconsistency.) It fails with a seemingly unrelated error:

```
[error] Exception in thread "main" java.lang.NullPointerException: Cannot invoke "String.indexOf(int, int)" because "self" is null
[error]         at kyo.Frame$package$Frame$.findNextSeparator(Frame.scala:44)
[error]         at kyo.Frame$package$Frame$.snippet(Frame.scala:63)
[error]         at kyo.Frame$package$Frame$.snippetShort(Frame.scala:69)
[error]         at kyo.kernel.internal.Trace$Owner.parse$1(Trace.scala:107)
[error]         at kyo.kernel.internal.Trace$Owner.enrich(Trace.scala:111)
[error]         at kyo.kernel.internal.Trace$Owner.kyo$kernel$internal$Trace$Owner$$inline$enrich(Trace.scala:94)
[error]         at kyo.scheduler.IOTask.eval(IOTask.scala:39)
[error]         at kyo.scheduler.IOTask.run(IOTask.scala:84)
[error]         at kyo.scheduler.Worker.runTask(Worker.scala:257)
[error]         at kyo.scheduler.Worker.run(Worker.scala:219)
[error]         at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
[error]         at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
[error]         at java.base/java.lang.Thread.run(Thread.java:1583)
```

I don't know what this could be, but maybe it's useful information... :man_shrugging: 

(In general, is `Abort.panic` equivalent to throwing an exception? Or is that not a thing?)

durban:
Sure, I've opened #1172.

fwbrasil:
Merging since we have a confirmation that the issue is fixed

PR #1170, opened 2025-04-30
Added groupedWithin

johnhungerford:
Addresses #679 

### Problem

One of the APIs requested for `Stream` in #679 is `groupedWithin`, which chunks stream emissions that fall within the same given duration with a max chunk size.

### Solution

Collects emissions asynchronously to a buffer, with a second background thread that checks the current buffer at increments of `maxTime` parameter (duration) and publishes buffer to an intermediate channel.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -1,6 +1,7 @@
 package kyo
 
 import kyo.Async.defaultConcurrency
+import kyo.Schedule.forever
 import kyo.kernel.ArrowEffect
 
 object StreamCoreExtensions:
@@ -413,12 +414,9 @@ object StreamCoreExtensions:
                                                 Loop(initialPar): currentPar =>
                                                     if currentPar > 0 then
                                                         parRef.updateAndGet(_ - 1).andThen:
-                                                            // java.lang.System.err.println(s"RUNNING ASYNCHRONOUSLY $input")
                                                             Async.run {
                                                                 f(input).map: chunk =>
-                                                                    // java.lang.System.err.println(s"TRANSFORMED $input to $chunk")
                                                                     prevFiber.get.andThen:
-                                                                        // java.lang.System.err.println(s"PUTTING $chunk")
                                                                         channel.put(Present(chunk)).andThen:
                                                                             parRef.updateAndGet(_ + 1).unit
                                                             }.map: newFiber =>
@@ -532,6 +530,68 @@ object StreamCoreExtensions:
             frame: Frame
         ): Stream[V2, Abort[E] & Async & S & S2] =
             mapChunkParUnordered(Async.defaultConcurrency, defaultAsyncStreamBufferSize)(f)(using t1, t2, t3, i1, i2, ev, frame)
+
+        def groupedWithin(maxSize: Int, maxTime: Duration)(using
+            t1: Tag[Emit[Chunk[V]]],
+            t2: Tag[V],
+            i1: Isolate.Contextual[S, IO],
+            i2: Isolate.Stateful[S, Abort[E] & Async],
+            fr: Frame
+        ): Stream[Chunk[V], S & Async] =
+            Stream:
+                AtomicRef.initWith(Chunk.empty[V]): bufferRef =>
+                    Channel.initWith[Maybe[Chunk[V]]](defaultAsyncStreamBufferSize): channel =>
+                        AtomicRef.initWith(Fiber.unit[Closed]): fiberRef =>
+                            val flushLoop: Unit < (Async & Abort[Closed]) =
+                                Loop.foreach:
+                                    Async.sleep(maxTime).andThen:
+                                        bufferRef.getAndSet(Chunk.empty).map: drainedChunk =>
+                                            if drainedChunk.isEmpty then Loop.done
+                                            else
+                                                channel
+                                                    .put(Present(drainedChunk))
+                                                    .andThen(Loop.continue)
+                            end flushLoop
+
+                            def publishChunk(chunk: Chunk[V]) =
+                                for
+                                    _        <- if chunk.nonEmpty then channel.put(Present(chunk)) else Kyo.unit
+                                    newFiber <- Async.run(flushLoop)
+                                    _        <- fiberRef.set(newFiber)
+                                yield ()
+                                end for
+                            end publishChunk
+
+                            val handledEmit = ArrowEffect.handleLoop(t1, stream.emit)(
+                                handle = [C] =>
+                                    (chunk, cont) =>
+                                        if chunk.isEmpty then Loop.continue(cont(()))
+                                        else
+                                            bufferRef.getAndSet(Chunk.empty).map: buffer =>
+                                                val nextChunk = chunk.take(maxSize - buffer.size)
+                                                val newBuffer = buffer.concat(nextChunk)
+                                                val handleChunk = fiberRef.get.map: fiber =>
+                                                    if newBuffer.size >= maxSize then
+                                                        fiber.interrupt
+                                                            .andThen(publishChunk(newBuffer))
+                                                            .andThen(bufferRef.set(chunk.drop(maxSize - buffer.size)))
+                                                    else
+                                                        fiber.done.map: isDone =>
+                                                            if isDone then
+                                                                publishChunk(chunk)
+                                                                    .andThen(bufferRef.set(Chunk.empty))
+                                                            else bufferRef.set(newBuffer)
+                                                handleChunk.andThen(Loop.continue(cont(())))
+                            ).andThen(bufferRef.get.map(buffer => publishChunk(buffer))).andThen(channel.put(Absent))
+
+                            Async.run(Abort.run[Closed](publishChunk(Chunk.empty).andThen(handledEmit))).andThen:
+                                val emit = Loop.foreach:
+                                    channel.take.map:
+                                        case Absent =>
+                                            Loop.done
+                                        case Present(c) =>
+                                            Emit.valueWith(Chunk(c))(Loop.continue)
+                                Abort.run(emit).unit
 
     end extension
 end StreamCoreExtensions
--- a/kyo-core/shared/src/test/scala/kyo/StreamCoreExtensionsTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/StreamCoreExtensionsTest.scala
@@ -252,6 +252,38 @@ class StreamCoreExtensionsTest extends Test:
                 Choice.run(test).andThen(succeed)
             }
         }
+
+        "groupedWithin" - {
+            "should group by size" in run {
+                val stream = Stream:
+                    Loop(0): i =>
+                        if i > 10 then Loop.done
+                        else Emit.valueWith(Chunk(i))(Loop.continue(i + 1))
+                stream.groupedWithin(3, Duration.Infinity).run.map: result =>
+                    assert(result == Chunk(Chunk(0, 1, 2), Chunk(3, 4, 5), Chunk(6, 7, 8), Chunk(9, 10)))
+            }
+
+            "should group by time" in run {
+                val stream = Stream {
+                    Loop(0): i =>
+                        Emit.valueWith(Chunk(i)):
+                            Async.sleep(20.millis).andThen(Loop.continue(i + 1))
+                }.take(4)
+                stream.groupedWithin(Int.MaxValue, 10.millis).run.map: result =>
+                    assert(result == Chunk(Chunk(0), Chunk(1), Chunk(2), Chunk(3)))
+            }
+
+            "should group by size and time" in run {
+                val stream = Stream {
+                    Loop(1): i =>
+                        Emit.valueWith(Chunk(i)):
+                            (if i % 5 == 0 then Async.sleep(40.millis) else Kyo.unit)
+                                .andThen(Loop.continue(i + 1))
+                }.take(10)
+                stream.groupedWithin(3, 30.millis).run.map: result =>
+                    assert(result == Chunk(Chunk(1, 2, 3), Chunk(4, 5), Chunk(6, 7, 8), Chunk(9, 10)))
+            }
+        }
     }
 
 end StreamCoreExtensionsTest


kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -413,12 +414,9 @@ object StreamCoreExtensions:
                                                 Loop(initialPar): currentPar =>
                                                     if currentPar > 0 then
                                                         parRef.updateAndGet(_ - 1).andThen:
-                                                            // java.lang.System.err.println(s"RUNNING ASYNCHRONOUSLY $input")
```
johnhungerford, line 416: My sophisticated debugging strategy 🤣

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -533,6 +531,82 @@ object StreamCoreExtensions:
...
+        def groupedWithin(maxSize: Int, maxTime: Duration)(using
+            t1: Tag[Emit[Chunk[V]]],
+            t2: Tag[V],
+            i1: Isolate.Contextual[S, IO],
+            i2: Isolate.Stateful[S, Abort[E] & Async],
+            fr: Frame
+        ): Stream[Chunk[V], S & Async] =
+            Stream:
+                AtomicRef.initWith(Chunk.empty[V]): bufferRef =>
+                    Channel.initWith[Maybe[Chunk[V]]](defaultAsyncStreamBufferSize): channel =>
```
fwbrasil, line 552: can you add a method parameter with the default value so the buffer size can be specified by the user?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -533,6 +531,82 @@ object StreamCoreExtensions:
...
+                                                        channel
+                                                            .put(Present(drainedChunk))
+                                                            .andThen(Loop.continue)
+                                    end flushLoop
+
+                                    def publishChunk(chunk: Chunk[V]) =
+                                        for
+                                            _        <- if chunk.nonEmpty then channel.put(Present(chunk)) else Kyo.unit
+                                            newFiber <- Async.run(flushLoop)
+                                            _        <- fiberRef.set(newFiber)
```
fwbrasil, line 571: Can't this end up creating multiple flush loop fibers?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -533,6 +531,82 @@ object StreamCoreExtensions:
...
+        ): Stream[Chunk[V], S & Async] =
+            Stream:
+                AtomicRef.initWith(Chunk.empty[V]): bufferRef =>
+                    Channel.initWith[Maybe[Chunk[V]]](defaultAsyncStreamBufferSize): channel =>
+                        IO.ensure(channel.close):
+                            AtomicRef.initWith(Fiber.unit[Closed]): fiberRef =>
+                                IO.ensure(fiberRef.get.map(_.interrupt)):
+                                    val flushLoop: Unit < (Async & Abort[Closed]) =
+                                        Loop.foreach:
+                                            Async.sleep(maxTime).andThen:
```
fwbrasil, line 558: It seems the flushing can drift depending on the timing of the arrival of new items and if there's already a flush loop running or not. Maybe `Clock.repeat*` could be more predictable?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -533,6 +531,82 @@ object StreamCoreExtensions:
...
+                                                                        publishChunk(chunk)
+                                                                            .andThen(bufferRef.set(Chunk.empty))
+                                                                    else bufferRef.set(newBuffer)
+                                                        handleChunk.andThen(Loop.continue(cont(())))
+                                    ).andThen:
+                                        bufferRef.get.map: buffer =>
+                                            (if buffer.nonEmpty then channel.put(Present(buffer)) else Kyo.unit).andThen:
+                                                channel.put(Absent)
+
+                                    Async.run(Abort.run[Closed](publishChunk(Chunk.empty).andThen(handledEmit))).andThen:
```
fwbrasil, line 601: why is the initial `publishChunk` necessary?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -533,6 +531,82 @@ object StreamCoreExtensions:
...
+        /** Collects values that are emitted within the same duration [[maxTime]] or reach the amount [[maxSize]] and emits them as a chunk.
+          *
+          * @param maxSize
+          *   Maximum number of elements to be collected within a single duration
+          * @param maxTime
+          *   Maximum amount of time to collect elements to emit in chunk
+          * @return
+          *   A new stream that emits collected chunks of elements
+          */
+        def groupedWithin(maxSize: Int, maxTime: Duration)(using
```
fwbrasil, line 543: I worry that the implementation could have some race conditions with the different fibers and use of atomic refs. How about an approach with a single fiber + timeout for the channel take? If the `take` takes longer than `maxTime`, then the fiber can publish the pending items.

`Async.timeout(channel.take, maxTime)`

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
fwbrasil: There's `Debug.values` as well :) it prints with the frame location directly to the console without suspending

fwbrasil:
I've triggered the build again but this test is still failing. I can't see a problem in the test itself, it might be a bug:

```scala
[info]   - should map all chunks preserving order *** FAILED *** (8 seconds, 8 milliseconds)
[info]     kyo.Timeout: [2m   │ ──────────────────────────────
[info]    │ // BaseKyoCoreTest.scala:15:35 kyo.internal.BaseKyoCoreTest $anonfun
[info]    │ ──────────────────────────────
[info] 14 │ },
[info] 15 │ Async.timeout(timeout)📍,
[info] 16 │ Async.run,
[info]    │ ──────────────────────────────
[info]    │ ⚠️ KyoException
[info]    │ 
[info]    │ Computation has timed out.
[info]    │ ──────────────────────────────
[info]     ...
```

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -533,6 +532,81 @@ object StreamCoreExtensions:
...
+          *   A new stream that emits collected chunks of elements
+          */
+        def groupedWithin(maxSize: Int, maxTime: Duration, bufferSize: Int = defaultAsyncStreamBufferSize)(using
+            t1: Tag[Emit[Chunk[V]]],
+            t2: Tag[Emit[Chunk[Chunk[V]]]],
+            i1: Isolate.Contextual[S, IO],
+            i2: Isolate.Stateful[S, Abort[E] & Async],
+            fr: Frame
+        ): Stream[Chunk[V], S & Async] =
+            Stream:
```
hearnadam, line 551: Since Stream is a suspension, I think you can directly use `Channel.Unsafe.init` which will greatly reduce the nesting here.

fwbrasil: Yeah, it's tempting to reuse another suspension but I think it's best to properly suspend any side effects with `IO` otherwise we could end up with referential transparency issues as the the code evolves

johnhungerford:
I was seeing some very occasional failures where the last element was dropped when rerunning the test suite, so I reimplemented using a more principled approach. Now I've got a stateful event loop that handles (1) emissions from the original stream, (2) scheduled events signifying max time, and (3) end of stream. This way guarantees no emission is dropped.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -622,6 +620,91 @@ object StreamCoreExtensions:
...
+            import Event.*
+            enum Event derives CanEqual:
+                case Data(chunk: Chunk[V])
+                case Tick, Flush
+            end Event
+
+            Stream[Chunk[V], S & Abort[E] & Async]:
+                IO.Unsafe {
+                    val safeMax = 1 max maxSize
+                    val channel = Channel.Unsafe.init[Event](1 max bufferSize).safe
```
johnhungerford, line 653: Good call changing this to a single channel. I had needed two channels for some of the parallel map functions and made a habit of it. I think I simplify at least a couple of those method this way too.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -622,6 +620,91 @@ object StreamCoreExtensions:
...
+
+                    // Handle loop collecting emitted values and flushing them until completion
+                    val push: Fiber[E | Closed, Unit] < (IO & Resource & S) =
+                        Async.run[E | Closed, Unit, S & Resource]:
+                            Resource.ensure(channel.close).andThen:
```
johnhungerford: Shouldn't we just use IO.ensure for this? No need for the whole finalizer queue.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -622,6 +620,91 @@ object StreamCoreExtensions:
...
+                    // Single fiber emitting a tick at constant interval
+                    val tick: Fiber[Closed, Unit] < (IO & Resource) =
+                        if maxTime == Duration.Infinity then Fiber.unit
+                        else
+                            Resource.acquireRelease(Clock.repeatWithDelay(maxTime)(channel.put(Tick)))(_.interrupt)
+
+                    // Single fiber collecting emitted values and emitting them as chunks
+                    // when the buffer exceeds the max size or a flush is requested
+                    // TODO: flush when push fiber completes with abort
+                    val pull: Fiber[E | Closed, Unit] => Unit < (Abort[E | Closed] & Async & Emit[Chunk[Chunk[V]]]) = _ =>
```
johnhungerford: Not sure why this is a method if you're not doing anything with the input fiber

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: Also, you could drop the `done` parameter and just add `IO.ensure(channel.put(Flush))`. That way the stream is guaranteed to complete

kyo-core/shared/src/test/scala/kyo/StreamCoreExtensionsTest.scala
```diff
@@ -459,6 +457,76 @@ class StreamCoreExtensionsTest extends Test:
...
+                }
+
+                Env.run(7) {
+                    envStream.groupedWithin(3, 15.millis).run.map: result =>
+                        assert(result.flatten == (1 to 7))
+                        assert(result.size >= 2) // Should have multiple groups due to timing
+                }
+            }
+
+            // "with Abort" in run {
```
johnhungerford: Good call -- I've left this out of the other concurrent stream method suites and I suspect there are some issues there. I'll add an issue.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
hearnadam: My plan was to merge the fibers, but it was non-trivial. Joining the push fiber is better - which works when you properly finalize the channel flush.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -622,6 +620,89 @@ object StreamCoreExtensions:
...
+            Stream[Chunk[V], S & Abort[E] & Async]:
+                IO.Unsafe {
+                    val safeMax = 1 max maxSize
+                    val channel = Channel.Unsafe.init[Event](1 max bufferSize).safe
+
+                    // Handle loop collecting emitted values and flushing them until completion
+                    val push: Fiber[E | Closed, Unit] < (IO & S) =
+                        Async.run[E | Closed, Unit, S]:
+                            IO.ensure(Async.run[Closed, Unit, Any](channel.put(Flush))):
+                                ArrowEffect.handleLoop(t1, (), stream.emit)(
```
johnhungerford: You can just get rid of the empty state `()` and the `done = (_, _) => ()` parameter

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -622,6 +620,89 @@ object StreamCoreExtensions:
...
+
+            Stream[Chunk[V], S & Abort[E] & Async]:
+                IO.Unsafe {
+                    val safeMax = 1 max maxSize
+                    val channel = Channel.Unsafe.init[Event](1 max bufferSize).safe
+
+                    // Handle loop collecting emitted values and flushing them until completion
+                    val push: Fiber[E | Closed, Unit] < (IO & S) =
+                        Async.run[E | Closed, Unit, S]:
+                            IO.ensure(Async.run[Closed, Unit, Any](channel.put(Flush))):
```
johnhungerford, line 658: Is there a reason the finalizer is run in a separate fiber?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -622,6 +620,89 @@ object StreamCoreExtensions:
...
+
+                    // Single fiber emitting a tick at constant interval
+                    val tick: Fiber[Closed, Unit] < (IO & Resource) =
+                        if maxTime == Duration.Infinity then Fiber.unit
+                        else
+                            Resource.acquireRelease(Clock.repeatWithDelay(maxTime)(channel.put(Tick)))(_.interrupt)
+
+                    // Loop collecting values from the channel and re-emitting them as chunks.
+                    // Chunks are emitted when the buffer exceeds the max size or a flush is requested.
+                    val pull: Unit < (Abort[E | Closed] & Async & Emit[Chunk[Chunk[V]]]) =
```
johnhungerford: Can `pull` fail with `E`? I don't see how so, since it doesn't run the original effect.

johnhungerford (commented):
A couple questions, but LGTM

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
hearnadam: `IO.ensure` doesn't accept Async functions.

Issue #552, opened 2024-07-19
Tag is not a sub-type of Intersection


steinybot:
This fails:
```
        "intersection subtype 3" in {
            trait A
            trait B
            class C extends A with B
            assert(Tag[C] <:< Intersection[A & B])
        }
```

PR #1171, opened 2025-05-03
[data] new Tag with sub-type check caching

### LINKED ISSUE #551: Tag type tests fail if the parent has a different type argument
This fails:
```
        "subtype and supertype with different type argument" - {
            class Super[A]
            class Sub[A] extends Super[String]
            test[Sub[Int], Super[String]]
        }
```

with:
```
false did not equal true Tag[T1] <:< Tag[T2] is false but ITag[T1] <:< ITag[T2] is true
```

fwbrasil:
Fixes #551, #552

### Problem

The current `Tag` implementation has a few major limitations, which is the main reason we haven't made user-defined effects a first-class feature yet.

- Sub-type checking based on the tag string is too expensive. It's designed to avoid allocations but the logic itself is too heavyweight. Because of that, `ArrowEffect` and `ContextEffect` use `=:=` when matching effect suspensions to their handlers. Effects that have type parameters with variance currently need to rely erased tags due to that (see `Env`, `Abort`, and the lack of variance in `Emit`).
- The separate APIs for intersections and union types is clunky and limits their use. 
- Tags can't be composed dynamically at run time, unlike `SafeClassTag`, which offers the `|` and `&` methods.
- The implementation currently ignores type parameters in super classes, which causes #551 and #552.
- Literal types are unsupported, which is problematic since we have `Abort.literal` now.

### Solution

We've attempted to rewrite `Tag` following the current approach with all information in the string a few times already without success. The latest attempt was by @hearnadam in #948, which is the origin of `TagTestMacro` in this PR. I think we could eventually get to an implementation with the properties we need but the high complexity of the code and the constraints regarding allocations would likely become a maintenance burden in the long run.

This PR adopts a new approach to encode tags by representing them as regular case classes in the new `Tag.Type` hierarchy. When inferring tags, the macro reflects on the type and generates a compile-time constant string as before but the payload is the serialized `Type` object. If the tag can't be fully materialized statically because it depends on other runtime-provided tags, the macro then falls back to allocating an actual `Type` instance.

In addition, a highly-optimized caching mechanism is in place to provide low-overhead sub-type checking. It has two layers: a cache for the check itself and, in case of misses in this initial layer, another cache is used to ensure tags are parsed only once. Please see the scaladocs for more information and reasoning since it isn't a very orthodox approach.

I've also added a restriction: tags in the `kyo` package must be static to avoid allocations. For example, if a method takes just a `Tag[A]` but it requires a `Tag[Emit[A]]`, the macro will fail and the final complete tag should be made available instead. The codebase was adapted for this new restriction.

### Notes

The new implementation addresses several issues but still has some important limitations. I'm planning to open follow up tickets for the following:

- **Opaque type support**: The macro isn't able to introspect opaque types. I've attempted a few times with different APIs but it seems not possible without relying on compiler internals or `typeSymbol.info`, which is marked as experimental. Izumi just uses the internal compiler types directly. It'd be nice to have a proper supported solution so we can still keep Kyo without any use of internal or experimental language features. Due to this limitation, opaque types like `Maybe` need to define the tag manually exposing its internal types, which requires `Tag.deriveDynamic` for tags using `Maybe`.
- **Fix Async.Join**: I had trouble to make the tag inference work with the output type as `Result[Nothing, *]` and decided to go with `Any` instead to unblock this change. 
- **Support tags of tags**: The new impl can't infer `Tag[Tag[A]]`, which is why I used the `TypeMapMarker` in `TypeMap`
- **Compact string representation**: The generated tag string is quite large now since java serialization isn't very efficient. I've added gzip compaction to alleviate it a bit but we should introduce a custom serialization and dictionary-based compaction like in the previous version.
- **Tag cache customization**: The parameters of the cache are hardcoded. I'll follow up reviewing all static configs and introducing a better way to read user-defined configurations since this is popping up in multiple places of the codebase.
- **Remove erased tags**: Migrate the codebase to use `<:<` where possible and remove erased tags from effects.


kyo-bench/src/main/scala/kyo/bench/MtlBench.scala
```diff
@@ -21,13 +21,13 @@ class MtlBench extends BaseBench:
...
-                )
-            )
-        )
+
+        testKyo.handle(
+            Var.run(State(2)),
+            Emit.run(_),
+            Env.run(EnvValue("config")),
+            Abort.run
+        ).eval
```
fwbrasil, line 30: I did some initial performance testing and noticed that Kyo was essentially a no-op in this bench because it was missing `Abort.run` and `eval` 🤦

kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -16,7 +17,9 @@ export Maybe.Present
 /** Companion object for Maybe type */
 object Maybe:
     inline given [A, B](using inline ce: CanEqual[A, B]): CanEqual[Maybe[A], Maybe[B]] = CanEqual.derived
-    given [A]: Conversion[Maybe[A], IterableOnce[A]]                                   = _.iterator
+    implicit def toIterableOnce[A](v: Maybe[A]): IterableOnce[A]                       = v.iterator
```
fwbrasil, line 20: I've changed to `implicit def` because `Conversion` requires the user to add `import scala.language.implicitConversions`,  which isn't ideal

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,541 @@
...
+                case self: Type[A] => self
+
+        /** Returns a human-readable representation of the type structure. This is useful for debugging and understanding complex type
+          * relationships.
+          *
+          * @return
+          *   A string representation of the type
+          */
+        def show: String =
+            self.tpe.toString()
```
fwbrasil, line 188: I've removed `showTpe` and made `show` print the type directly instead. The distinction wasn't very useful.

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,541 @@
...
+        end ClassType
+
+        /** Represents a recursive type reference. This is used to handle recursive types without infinite recursion.
+          *
+          * @param id
+          *   The identifier of the referenced type
+          * @tparam A
+          *   The Scala type represented by this recursive reference
+          */
+        case class Recursive[A](id: String) extends Type[A]
```
fwbrasil, line 369: This is a case we didn't have before because `Tag` was ignoring type parameters of super classes. An example case is `String` since it extends from `Comparable[String]`, which would recurse the type in the AST.

kyo-data/shared/src/test/scala/kyo/TagTest.scala
```diff
@@ -2,48 +2,14 @@ package kyo
...
-import kyo.Tag.Union
+import kyo.Tag.Type.*
+import kyo.internal.RegisterFunction
+import kyo.internal.TagTestMacro.test
 import scala.annotation.nowarn
+import scala.concurrent.Future
 
 class TagTest extends Test:
 
-    inline def test[T1, T2](using k1: Tag[T1], i1: ITag[T1], k2: Tag[T2], i2: ITag[T2]): Unit =
```
johnhungerford, line 11: Why not use this?

kyo-bench/src/main/scala/kyo/bench/MtlBench.scala
johnhungerford: You can't be the performance of no-op!

kyo-data/shared/src/main/scala/kyo/Maybe.scala
```diff
@@ -16,7 +17,9 @@ export Maybe.Present
...
+
+    inline given [A]: Tag[Maybe[A]] = Tag[Absent | Present[A]]
```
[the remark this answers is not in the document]
johnhungerford: Are there any downsides? Will this break/fail under any circumstances?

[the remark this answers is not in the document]
fwbrasil: Yes, it's an important limitation because it's not possible to derive static tags for opaque types. They require `Tag.deriveDynamic` in the `kyo` package. In user code, `Tag.derive` automatically falls back to dynamic, which has a performance cost due to allocations.

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,541 @@
...
+    end Type
+
+    import Type.*
+
+    private[kyo] object internal:
+
+        private val threadSlots  = Runtime.getRuntime().availableProcessors() * 8
+        private val cacheEntries = 64
+        private val cacheSlots   = new Array[Array[Long]](threadSlots)
+        (0 until threadSlots).foreach(idx => cacheSlots(idx) = new Array(cacheEntries))
```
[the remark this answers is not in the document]
fwbrasil: cool! I imagine you meant `Array.ofDim`, I didn't know this method. The cache size of each thread slot is a separate config, though. The slots are just a way to have multiple cache instances to avoid interference in case multiple threads land on the same slot.

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,541 @@
...
-                        else
-                            loop(nextSubParam, nextSuperParam)
-                        end if
+        end isSubtype
+
+        private def isSubtype(a: Type[?], b: Type[?], seen: Map[String, Type[?]]): Boolean =
+            a match
+                case NothingType => true
+                case AnyType     => b eq AnyType
+                case NullType    => true
```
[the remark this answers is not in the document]
fwbrasil: hmm I normally keep the "simpler" cases first in the pattern match. Do you mean that there might be a behavior difference?

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,541 @@
...
+                                                    case Variance.Covariant     => isSubtype(aParam, bParam, updatedSeen)
+                                                    case Variance.Contravariant => isSubtype(bParam, aParam, updatedSeen)
+                                            ok && loopParams(paramIdx + 1)
+                                        }
+                                    loopParams(0)
+                                }
+                            }
+        end isSubtype
+
+        // avoids regular exists since it allocates an iterator
```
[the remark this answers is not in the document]
fwbrasil: yeah, the compiler doesn't allow to override the method from the stdlib because of `inline`. Maybe we could have a separate `Chain` API with optimized methods and then use it to replace `Chunk.Indexed`?

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,541 @@
...
+          *
+          * Since the set of unique types in a program is determined at compile time and is typically much smaller than the set of possible
+          * subtype relationships, this cache can safely grow without bounds. Each unique Tag is only decoded once during the lifetime of
+          * the application, regardless of how many times it's used.
+          */
+        private val decodeCache = new ConcurrentHashMap[String, Any]()
+
+        def encode[A](obj: Type[A]): String =
+            Using.resource(new ByteArrayOutputStream()) { baos =>
+                Using.resource(new GZIPOutputStream(baos)) { gzos =>
```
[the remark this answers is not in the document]
fwbrasil: I just saw that. I worry the payloads would become too large. I'll try a new more compact representation without gzip and update the PR

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,541 @@
...
-            val superSize = IntEncoder.decode(superTag.charAt(superIdx))
-            @tailrec def loop(subIdx: Int): Boolean =
-                if subIdx >= subTag.length() || superIdx >= superTag.length() then false
+    /** Base class for all type representations in the Tag system. Type provides a structured representation of Scala types that can be
+      * manipulated at runtime for operations like subtype checking and type composition.
+      *
+      * @tparam A
+      *   The Scala type represented by this Type
+      */
+    sealed abstract class Type[A] extends Serializable with Product:
```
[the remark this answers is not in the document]
fwbrasil: I had called it `Decoded` initially but I thought the name wasn't great. I think `Type` works well since the AST is public and the user can inspect it.

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,541 @@
...
+            loop(0)
+        end forall
+
+        private def resolve(tpe: Type[?], seen: Map[String, Type[?]]): Type[?] =
+            tpe match
+                case Recursive(id) => resolve(id, seen)
+                case _             => tpe
+
+        private def resolve(id: String, seen: Map[String, Type[?]]): Type[?] =
+            require(seen.contains(id), s"Invalid recursive type id: $id")
```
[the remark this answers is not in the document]
fwbrasil: `kyo.debug` is was `kyo-kernel` but I'm moving it to `kyo-data` and updating the macro errors

kyo-data/shared/src/main/scala/kyo/Tag.scala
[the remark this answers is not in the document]
hearnadam: No, I just think it's unlikely that `Null` be frequently the subtype checked against. Though it should be a pretty simple match?

kyo-data/shared/src/main/scala/kyo/Tag.scala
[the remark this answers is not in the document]
hearnadam: I think we could offer some inline extension methods for `Chunk.Indexed`... Perhaps `.inlined.foreach`? or similar.

kyo-kernel/shared/src/main/scala/kyo/kernel.scala
```diff
@@ -28,19 +28,10 @@ type Id[A] = A
...
     ): Nothing =
-        bug(s"Unexpected pending effect while handling ${expected.show}: " + kyo)
-
-    def check(cond: Boolean): Unit =
-        if !cond then throw new KyoBugException("Required condition is false.")
-
-    def apply(msg: String): Nothing =
-        throw KyoBugException(msg + " Please open an issue 🥹  https://github.com/getkyo/kyo/issues")
-end bug
+        self(s"Unexpected pending effect while handling ${expected.show}: " + kyo)
```
hearnadam, line 36:
```suggestion
        self(s"Unexpected pending effect while handling ${expected.show}: " + kyo.show)
```

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -361,6 +361,27 @@ object Abort:
             case Failure(e) => onFail(e)
             case Panic(thr) => throw thr
 
+    /** Recovers from an Abort failure or panic by applying the provided function to the Error value.
```
hearnadam, line 364: I think your diff from another PR got mixed in.

fwbrasil:
I've pushed additional changes:

- Compact flat representation
- Support for opaque types
- Lambda type fixes
- Support for nested records now that `Tag` can represent intersections + tests for it
- Remove `IntEncoder` since it isn't used anymore
- Removed `Tag.deriveDynamic` now that all necessary types in Kyo's codebase are supported

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,653 @@
...
+                                }
+
+                    case UnionEntry(aSet) =>
+                        forall(aSet) { aElemId =>
+                            isSubtype(aOwner, bOwner, aElemId, bId)
+                        }
+
+                    case LiteralEntry(widenedId, value) =>
+                        bEntry match
+                            case LiteralEntry(_, bValue) => value.equals(bValue)
```
hearnadam, line 372: Would the ids be the same?

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,653 @@
...
+                    case UnionEntry(set)              => s"U:${set.mkString(":")}"
+
+                    case LambdaEntry(params, lower, upper, body) =>
+                        s"M:$body:${params.size}:${params.mkString(":")}:${lower.mkString(":")}:${upper.mkString(":")}"
+
+                    case OpaqueEntry(name, lower, upper, variances, params) =>
+                        s"O:$name:$lower:$upper:${params.size}:${variances.map(_.ordinal).mkString(":")}:${params.mkString(":")}"
+
+                    case ClassEntry(className, variances, params, parents) =>
+                        require(params.size == variances.size)
```
hearnadam, line 564: Should this be in the constructor for the class?

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,653 @@
...
+          * significantly improving performance for repeated operations on the same types. Unlike the subtype cache, this cache is fully
+          * thread-safe using ConcurrentHashMap and never evicts entries.
+          *
+          * Since the set of unique types in a program is determined at compile time and is typically much smaller than the set of possible
+          * subtype relationships, this cache can safely grow without bounds. Each unique Tag is only decoded once during the lifetime of
+          * the application, regardless of how many times it's used.
+          */
+        private val decodeCache = new ConcurrentHashMap[String, Type[?]]()
+
+        private val decodeFunction: java.util.function.Function[String, Type[?]] =
```
hearnadam, line 585: Any reason this isn't `String => Type[?]`? I assume it's because it needs to be passed to `computeIfAbsent`?

kyo-data/shared/src/test/scala/kyo/TagTest.scala
```diff
@@ -576,71 +584,147 @@ class TagTest extends Test:
...
+    }
+
+    "base types" - {
+        "Nothing" - {
+            test[Nothing, Any]
+            test[Nothing, AnyRef]
+            test[Nothing, AnyVal]
+            test[List[Nothing], List[Int]]
+        }
+        "Null" - {
```
hearnadam, line 663: Can you add one for `Null` and a primitive?

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -1,239 +1,653 @@
...
+        import Type.*
+        import Type.Entry.*
+
+        private val typeCache = new ConcurrentHashMap[Tag[Any], Type[?]]()
+
+        private val threadSlots  = Runtime.getRuntime().availableProcessors() * 8
+        private val cacheEntries = 64
+        private val cacheSlots   = Array.ofDim[Long](threadSlots, cacheEntries)
+
+        case class Dynamic(tag: String, map: Map[Entry.Id, Any]):
```
hearnadam, line 267: final?

hearnadam (approved):

kyo-data/shared/src/main/scala/kyo/Tag.scala
fwbrasil: hmm I imagine there could be case where the string representation is the same for different numerical types. I'll fix it to check the tag as well

kyo-data/shared/src/main/scala/kyo/Tag.scala
fwbrasil: Yes, I was using an anon function before and it was allocating on every cache lookup

PR #1180, opened 2025-05-10
[data] Introduce Chain + migrate Tag to use it

fwbrasil:


kyo-bench/src/main/scala/kyo/bench/TypeMapBench.scala
```diff
@@ -164,14 +164,13 @@ class TypeMapBench extends BaseBench:
     def unionKyo() =
         kyoEnv.union(kyoEnvSmall)
 
-    // TODO: prune to an intersection once Kyo Tags support Intersections as String
```
fwbrasil, line 167: cleanup since tag supports intersections now

kyo-data/shared/src/main/scala/kyo/Tag.scala
```diff
@@ -291,8 +293,6 @@ object Tag:
         import Type.*
         import Type.Entry.*
 
-        private val typeCache = new ConcurrentHashMap[Tag[Any], Type[?]]()
```
fwbrasil, line 294: unused, leftover of previous iterations in the PR

kyo-data/shared/src/main/scala/kyo/KArray.scala
```diff
@@ -0,0 +1,350 @@
...
+  * memory-efficient when working with primitive values like Int, Long, Double, etc.
+  *
+  * Important: Unlike Chunk, KArray is NOT part of Scala's collection hierarchy. It doesn't extend Seq or any other collection trait, which
+  * means it won't work with methods expecting standard Scala collections. Consider using Chunk instead when you need compatibility with
+  * Scala's collection library or when you require efficient structural operations like slicing without copying data.
+  *
+  * @tparam A
+  *   the type of elements in this KArray
+  */
+opaque type KArray[A] = Array[A]
```
sideeffffect, line 27:
```suggestion
opaque type KArray[+A] = Array[A]
```
Could this be covariant since the array an be stored to?

fwbrasil: Good point. I wonder if we could have trouble with the fact that the arrays are specialized but it seems worth exploring. Do you mind creating an issue for it?

sideeffffect: Done
 * https://github.com/getkyo/kyo/issues/1185

PR #1190, opened 2025-05-14
[core] stable fiber identity

fwbrasil:
### Problem

When a fiber encounters a child fiber/promise to wait on during execution, `IOTask` currently resumes the parent in a new `IOTask` instance after the child completes. Besides the performance cost, this makes the identity of a fiber not stable since it might be executing in a linked `IOTask` instance if it got suspended by a child fiber. It's a blocker to implement fiber trace dumps for example since the actual execution trace might in a linked `IOTask` instance.

### Solution

Adapt the logic to simply resume the `IOTask` after a child completes and the execution of the parent needs to resume.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
+++ b/kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
@@ -34,55 +34,60 @@ sealed private[kyo] class IOTask[Ctx, E, A] private (
 
     private inline def locally[A](inline f: A): A = f
 
-    final private def eval(startMillis: Long, clock: InternalClock, deadline: Long)(using Safepoint) =
+    final private def eval(startMillis: Long, clock: InternalClock, deadline: Long)(using Safepoint): A < (Ctx & Async & Abort[E]) =
         try
-            curr = Isolate.restoring(trace, this) {
-                ArrowEffect.handlePartial(erasedAbortTag, Tag[Async.Join], curr, context)(
-                    stop =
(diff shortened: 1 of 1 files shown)


steinybot (approved):
This change looks right to me based on what I know although I don't have a full grasp on how the scheduler works.

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -34,71 +34,79 @@ sealed private[kyo] class IOTask[Ctx, E, A] private (
...
+        else
+            Task.Done
         end if
     end run
 
     private inline def nullResult = null.asInstanceOf[A < Ctx & Async & Abort[E]]
 
     override def toString =
-        s"IOTask(state = ${stateString()}, preempt = ${{ shouldPreempt() }}, finalizers = ${finalizers.size()}, curr = ${curr})"
+        s"IOTask(id = ${hashCode()}, state = ${stateString()}, preempt = ${{ shouldPreempt() }}, finalizers = ${finalizers.size()}, curr = ${curr})"
```
hearnadam, line 109: Do you think we should add an `id` method?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
fwbrasil: I considered exposing it but I haven't seen a good use case yet to justify the API commitment

PR #1191, opened 2025-05-14
[core] Channel/Queue closeAndWaitEmpty

### LINKED ISSUE #721: Add ability to close Channel after remaining elements have been taken
Currently when the writer calls `close` on a channel, if there are remaining values then those will be returned back to the writer and the channel will be closed immediately. Readers will not be able to read these remaining elements.

Another use case is that the writer should be able to signal that they are done writing and only when the reader reads the last remaining value should the channel then be closed.

It is quite tricky to do this with the current API. The writer has to only close if the channel is empty, otherwise set some other `AtomicBoolean` signal, and the reader has to handle both the channel being closed and this other signal.

fwbrasil:
Fixes #721

### Problem

`Queue` and `Channel` don't offer a way to close them only after all pending items are consumed. This is an important limitation in producer/consumer scenarios where the producer has has finished processing but it needs to wait for consumers to drain the queue or channel before closing it.

### Solution

Introduce `Channel.closeAndWaitEmpty` and `Queue.closeAndWaitEmpty`

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Queue.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Queue.scala
@@ -121,6 +121,18 @@ object Queue:
           */
         def close(using Frame): Maybe[Seq[A]] < IO = IO.Unsafe(self.close())
 
+        /** Closes the queue and asynchronously waits until it's empty.
+          * 
+          * This method closes the queue to new elements and returns a fiber that completes
+          * when all elements have been consumed. Unlike the regular `close` method, this
+          * allows consumers to process all remaining elements before considering the queue
+          * fully closed.
+          *
+          * @return
+          *   true if the queue was successfully closed and emptied, false if it was already closed
+          */
+        def closeAwaitEmpty(using Frame): Boolean < Async = IO.Unsafe(self.closeAwaitEmpty().safe.get)
+
         /** Checks if the queue is closed.
           *
           * @return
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/Queue.scala
```diff
@@ -339,42 +353,84 @@ object Queue:
...
 
             protected def _drain(max: Maybe[Int] = Maybe.Absent): Chunk[A]
+            protected def _isEmpty(): Boolean
+
+            private def opClosed: Maybe[Result.Error[Closed]] =
+                state.get() match
+                    case State.FullyClosed(r) => Present(r)
+                    case _                    => Absent
 
             protected inline def op[A](inline f: => A): Result[Closed, A] =
```
fwbrasil, line 405: The half closed state allows all operations (`op` method`) and fails in case of offers or adds (`offerOp`)

kyo-core/shared/src/main/scala/kyo/Queue.scala
```diff
@@ -339,42 +352,84 @@ object Queue:
...
+                    r
+                }
+
+            private def offerClosed: Maybe[Result.Error[Closed]] =
+                state.get() match
+                    case State.Open           => Absent
+                    case State.HalfOpen(_, r) => Present(r)
+                    case State.FullyClosed(r) => Present(r)
 
             protected inline def offerOp[A](inline f: => Boolean, inline raceRepair: => Boolean): Result[Closed, Boolean] =
```
steinybot, line 417: Unrelated to this change but it looks like `A` is unused here.

kyo-core/shared/src/main/scala/kyo/Queue.scala
```diff
@@ -121,6 +121,17 @@ object Queue:
...
         def close(using Frame): Maybe[Seq[A]] < IO = IO.Unsafe(self.close())
 
+        /** Closes the queue and asynchronously waits until it's empty.
+          *
+          * This method closes the queue to new elements and returns a computation that completes when all elements have been consumed.
+          * Unlike the regular `close` method, this allows consumers to process all remaining elements before considering the queue fully
+          * closed.
+          *
+          * @return
+          *   true if the queue was successfully closed and emptied, false if it was already closed
```
steinybot, line 131: Is it worth elaborating that it could be `false` if half open?

If two producers attempt to call `closeAwaitEmpty` on a non-empty queue, one will block, the other will return `false`. If the other then calls `closed` it would be `false` which might be confusing.

kyo-core/shared/src/main/scala/kyo/Queue.scala
```diff
@@ -339,42 +352,84 @@ object Queue:
...
+            private def opClosed: Maybe[Result.Error[Closed]] =
+                state.get() match
+                    case State.FullyClosed(r) => Present(r)
+                    case _                    => Absent
 
             protected inline def op[A](inline f: => A): Result[Closed, A] =
-                _closed.get().getOrElse(Result(f))
+                opClosed.getOrElse {
+                    val r = Result(f)
+                    handleHalfOpen()
```
steinybot, line 407: Nit picking but is it worth skipping this for operations that don't remove elements?

kyo-core/shared/src/main/scala/kyo/Queue.scala
```diff
@@ -455,6 +512,7 @@ object Queue:
                     loop(0)
                     b.result()
                 end _drain
+                protected def _isEmpty() = q.isEmpty()
```
steinybot, line 515: Nit: Should the others be protected too? Same thing with the `_drain`s.

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -924,4 +924,63 @@ class ChannelTest extends Test:
...
+            yield assert(result)
+        }
+
+        "returns true when channel becomes empty after closing" in run {
+            for
+                c      <- Channel.init[Int](10)
+                _      <- c.put(1)
+                _      <- c.put(2)
+                fiber  <- Async.run(c.closeAwaitEmpty)
+                _      <- c.take
```
steinybot, line 941: Should we also check that `closed` is `false`?

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
steinybot, line 1: How about checking:
- 2 producers calling `closeAwaitEmpty`
- 2 producers, one calling `closeAwaitEmpty` and another calling `close`

kyo-core/shared/src/test/scala/kyo/QueueTest.scala
steinybot, line 1: Same here too if you agree.

steinybot (approved):
Very neat. Thanks for doing this.

Approving since it looks good, only nits.

kyo-core/shared/src/main/scala/kyo/Queue.scala
fwbrasil: good catch!

kyo-core/shared/src/main/scala/kyo/Queue.scala
fwbrasil: ah, it's actually unnecessary since anonymous classes are final. I've removed `protected`

PR #1193, opened 2025-05-16
Fix channel drain close race

steinybot:
### Problem

Given:
- A channel.
- A producer that puts items into the channel.
- The produce closes the channel either:
  1) when it is empty via `close` or,
  2) using `closeAwaitEmpty`.
- A consumer that reads from the channel optimistically using `drain` and,
- if empty, the consumer uses `take` to block until there are new items or the channel is closed.

Then:
- There is a race condition where the consumer is in the middle of draining,
- the channel is not empty,
- the producer closes the channel,
- the consumer returns `Closed` rather than the items drained so far,
- and so items are effectively lost from the end of the channel.

### Solution

If draining encounters a `Failure[Closed]` and items have been drained, then returns those items.

### Notes

This is a little tricky to reproduce as it relies on timing. For whatever reason it was fairly reliable to reproduce when running all the gRPC tests.

I had asked about this and @fwbrasil had said:
> I'd say the behavior for the draining operations are expected since they just obtain the values but don't guarantee that new items won't arrive. The trouble is the closing since the pending backlog in this case is returned to the producer given it's the one to call close. The returned backlog to the producer should contain the missing message.

#1191 was an attempt to resolve this although unfortunately it does not (although it does make other parts of the above pattern much easier to implement).

With this change, it still doesn't guarantee that new items won't arrive. The guarantee is that if it wasn't closed when drain was first called (really when the underlying queue drain is first called) then it will always return a success.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -397,33 +397,54 @@ object Channel:
 
             private val closedResult = Result.fail(Closed("Channel", initFrame, "zero-capacity"))
 
-            private def asResult[A](value: A): Result[Closed, A] =
+            /** Succeeds with the value if the channel is still open, otherwise fails with a [[Closed]] error.
+             *
+             * @param value
+             *   The value to succeed with
+             * @return
+             *   The successful value or a [[Closed]] error
+             */
+            private def succeedIfOpen[B](value: B): Result[Closed, B] =
                 if isClosed.get() then closedResult else Result.succeed(value)
 
-            private def asOptimisticResult[A](value: Chunk[A]): Result[Closed, Chunk[A]] =
-                if value.isEmpty then asResult(value) else Result.succeed(value)
+            /** Succeeds with the value if it is non-empty or the channel is still open, otherwise fails with a
+             * [[Closed]] error.
+             *
+             * This is used in cases where the channel may be closed, but we still have a value to return. This
+             * typically occurs when the producer calls [[closeAwaitEmpty]] and the consumer calls [[drain]] or
+             * [[drainUpTo]] where the drain will close the channel once it has drained the last item, but we want the
+             * effect of the close occurring after the drain. This may also occur in a race condition where the producer
+             * checks if the channel is empty and closes it while the consumer is draining.
+             *
+             * @param value
+             *   The value to succeed with
+             * @return
+             *   The successful value or a [[Closed]] error
+             */
+            private def succeedIfNonEmptyOrOpen[B](value: Chunk[B]): Result[Closed, Chunk[B]] =
+                if value.nonEmpty then Result.succeed(value) else succeedIfOpen(value)
 
             def capacity = 0
 
-            def size()(using AllowUnsafe) = asResult(0)
+            def size()(using AllowUnsafe) = succeedIfOpen(0)
 
             def offer(value: A)(using AllowUnsafe) =
                 Maybe(takes.poll()) match
                     case Absent =>
-                        asResult(false)
+                        succeedIfOpen(false)
                     case Present(takePromise) =>
-                        if takePromise.complete(Result.succeed(value)) then asResult(true)
+                        if takePromise.complete(Result.succeed(value)) then succeedIfOpen(true)
                         else offer(value)
 
             def offerAll(values: Seq[A])(using AllowUnsafe): Result[Closed, Chunk[A]] =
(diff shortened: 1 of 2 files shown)


fwbrasil:
@steinybot not sure I understood, are the tests still flaky?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -395,8 +395,13 @@ object Channel:
...
             val isClosed = AtomicBoolean.Unsafe.init(false)
 
-            private val closedResult                     = Result.fail(Closed("Channel", initFrame, "zero-capacity"))
-            def asResult[A](value: A): Result[Closed, A] = if isClosed.get() then closedResult else Result.succeed(value)
+            private val closedResult = Result.fail(Closed("Channel", initFrame, "zero-capacity"))
+
+            private def asResult[A](value: A): Result[Closed, A] =
+                if isClosed.get() then closedResult else Result.succeed(value)
+
+            private def asOptimisticResult[A](value: Chunk[A]): Result[Closed, Chunk[A]] =
```
fwbrasil, line 403: The name doesn't seem very informative, how about `asDrainResult`? Could you also add a comment explaining the edge case the method handles?

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -633,19 +639,20 @@ object Channel:
 
             def drain()(using AllowUnsafe) =
                 @tailrec
-                def loop(current: Chunk[A]): Result[Closed, Chunk[A]] =
+                def loop(current: Chunk[A], i: Int): Result[Closed, Chunk[A]] =
```
fwbrasil, line 642: Is `i` unused?

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -1102,4 +1120,52 @@ class ChannelTest extends Test:
         }
     }
 
+    private def verifyRaceDrainWithClose(
```
fwbrasil, line 1123: Good catch! I imagine it was a though one to debug  🙏

kyo-core/shared/src/main/scala/kyo/Channel.scala
steinybot: Woops, this was left in for verifying the tests. Removed.

kyo-core/shared/src/main/scala/kyo/Channel.scala
steinybot: Sure thing. I went for a slightly different naming. Let me know what you think.

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
steinybot: Ha yeah, the closer I got to minimising the problem it started disappearing. Good ol' Heisenbug. It seems obvious in hindsight but I have been looking at that code for so long I failed to notice.

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -1102,4 +1120,52 @@ class ChannelTest extends Test:
...
+                Async.fill(100_000, concurrency = 1) {
+                    for
+                        c             <- ref.get
+                        closedOrEmpty <- Abort.recover[Closed](_ => true)(c.empty)
+                        _             <- if closedOrEmpty then Kyo.unit else drain(c)
+                    yield ()
+                }
+            }
+            _ <- producer.interrupt
+        yield assert(result.isSuccess)
```
steinybot, line 1167: This could be a bit better. It could actually check that the right number of elements were drained. I'd have to swap things around a bit. I should be able to simplify this a bit so long as it doesn't mess up the timings.

steinybot: I have a different version of these tests which are easier to read and test the actual counts but to get similar failure rates it runs about 5 times slower (~500ms on my M1): https://github.com/steinybot/kyo/pull/2/files

Not sure if it is worth it.

fwbrasil: I think having better coverage of an edge case is worth it! If we see it's impacting build times too much, we can follow up later

fwbrasil: I'm merging the PR to stabilize the build but feel free to post follow ups

PR #1194, opened 2025-05-17
[kernel][core] Resource improvements and fix

fwbrasil:
### Problem

When there's a failure, `Resource` is currently not properly waiting for the finalizers to execute before proceeding. Additionally, it doesn't provide a way for finalizers to conditionally execute depending on the outcome of the computation.

### Solution

Fix the handling in case of failure to ensure the computation waits for finalizers and adapt `Safepoint`, `IO`, and `Resource` to pipe through a `Maybe[Throwable]` to indicate if the computation failed or not so finalizers can implement custom logic.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/IO.scala
+++ b/kyo-core/shared/src/main/scala/kyo/IO.scala
@@ -1,5 +1,6 @@
 package kyo
 
+import kyo.Result.Error
 import kyo.Tag
 import kyo.kernel.*
 import kyo.kernel.internal.Safepoint
@@ -56,13 +57,31 @@ object IO:
       *   Implicit frame for the computation.
       * @tparam A
       *   The result type of the main computation.
+      */
+    inline def ensure[A, S](inline f: => Any < IO)(v: A < S)(using inline frame: Frame): A < (IO & S) =
+        ensure(_ => f)(v)
+
+    /** Ensures that a finalizer is run after the computation, regardless of success or failure.
+      *
+      * This version provides the finalizer with information about whether the computation completed successfully or failed with an
+      * exception. The finalizer receives a `Maybe[Error[Any]]` which will be `Absent` if the computation succeeded, or `Present` if it
+      * failed.
+      *
+      * @param f
+      *   The finalizer function that receives information about potential errors and performs cleanup actions.
+      * @param v
+      *   The computation.
+      * @param frame
+      *   Implicit frame for the computation.
+      * @tparam A
+      *   The result type of the computation.
       * @tparam S
       *   Additional effects in the computation.
       * @return
-      *   The result of the main computation, with the finalizer guaranteed to run.
+      *   The result of the computation, with the finalizer guaranteed to run.
       */
-    def ensure[A, S](f: => Any < IO)(v: A < S)(using frame: Frame): A < (IO & S) =
-        Unsafe(Safepoint.ensure(IO.Unsafe.evalOrThrow(f.unit))(v))
+    inline def ensure[A, S](inline f: Maybe[Error[Any]] => Any < IO)(v: A < S)(using inline frame: Frame): A < (IO & S) =
+        Unsafe(Safepoint.ensure(ex => IO.Unsafe.evalOrThrow(f(ex)))(v))
 
     /** Retrieves a local value and applies a function that can perform side effects.
       *
--- a/kyo-core/shared/src/main/scala/kyo/Resource.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Resource.scala
@@ -134,17 +134,10 @@ object Resource:
                     IO.ensure(finalizer.close),
                     Abort.run[Any]
                 ).map { result =>
-                    val io =
-                        result match
-                            case result @ Result.Error(e) =>
-                                val ex =
-                                    e match
-                                        case e: Throwable => e
-                                        case _            => new KyoException(s"Failed $e")
-                                finalizer.close(Present(Panic(ex)))
-                            case _ =>
-                                finalizer.close(Absent)
-                    io.andThen(finalizer.await).andThen(Abort.get(result.asInstanceOf[Result[Nothing, A]]))
+                    finalizer
+                        .close(result.error)
+                        .andThen(finalizer.await)
+                        .andThen(Abort.get(result.asInstanceOf[Result[Nothing, A]]))
                 }
         }
 
--- a/kyo-core/shared/src/main/scala/kyo/Resource.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Resource.scala
@@ -1,5 +1,7 @@
 package kyo
 
+import kyo.Result.Error
+import kyo.Result.Panic
 import kyo.Tag
 import kyo.kernel.ContextEffect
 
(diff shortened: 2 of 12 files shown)


kyo-core/shared/src/main/scala/kyo/IO.scala
```diff
@@ -56,13 +56,31 @@ object IO:
...
+      *   The result type of the computation.
       * @tparam S
       *   Additional effects in the computation.
       * @return
-      *   The result of the main computation, with the finalizer guaranteed to run.
+      *   The result of the computation, with the finalizer guaranteed to run.
       */
-    def ensure[A, S](f: => Any < IO)(v: A < S)(using frame: Frame): A < (IO & S) =
-        Unsafe(Safepoint.ensure(IO.Unsafe.evalOrThrow(f.unit))(v))
+    inline def ensure[A, S](inline f: Maybe[Throwable] => Any < IO)(v: A < S)(using inline frame: Frame): A < (IO & S) =
```
fwbrasil, line 82: I'm not sure `ensure` wasn't `inline` since that's typically the case for effect suspensions. Note that both `IO.ensure` and `Resource.ensure` now have an overriden version that takes a `Maybe[Throwable]`.

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -35,56 +35,6 @@ import kyo.kernel.ContextEffect
 sealed trait Resource extends ContextEffect[Resource.Finalizer]
 
 object Resource:
-    /** Represents a finalizer for a resource. */
```
fwbrasil, line 38: There are no changes in this section of the code. I've just moved it to the end of the file so the public APIs appear first in the file to improve readability.

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -124,7 +90,7 @@ object Resource:
       *   The acquired Closeable resource wrapped in Resource, IO, and S effects.
       */
     def acquire[A <: java.io.Closeable, S](resource: A < S)(using Frame): A < (Resource & IO & S) =
-        acquireRelease(resource)(r => IO(r.close()))
+        acquireRelease(resource)(_.close())
```
fwbrasil, line 93: The `IO` suspension here wasn't necessary since it's a by-name param that is already deferred.

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -160,10 +126,75 @@ object Resource:
...
+                    val io =
+                        result match
+                            case result @ Result.Error(e) =>
+                                val ex =
+                                    e match
+                                        case e: Throwable => e
+                                        case _            => new KyoException(s"Failed $e")
+                                finalizer.close(Present(ex))
+                            case _ => Kyo.unit
+                    io.andThen(finalizer.await).andThen(Abort.get(result.asInstanceOf[Result[Nothing, A]]))
```
fwbrasil, line 142: This code has some unsafe operations for two reasons:

1. It's necessary to handle `Abort` with `Any` to ensure all failures are caught. Without `Any`, it'd handle only panics in `Abort[Nothing]`. Given the guarantees of effect handling, it's ok to cast here since, if it's a `Failure[E]`, then there is a handler up in the stack that will handle the specific failure. I think this might be better encoded via an `Isolate` for `Abort`. I'll explore it.

3. Although `IO.ensure` now pipes the `Maybe[Throwable]`, in case of an interruption, the `IO.ensure` call will run only later when the fiber finalizes so I added special handling here. I had to make `finalizer.close` idempotent due to that.

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -124,7 +90,7 @@ object Resource:
       *   The acquired Closeable resource wrapped in Resource, IO, and S effects.
       */
     def acquire[A <: java.io.Closeable, S](resource: A < S)(using Frame): A < (Resource & IO & S) =
```
hearnadam, line 92:
```suggestion
    def acquire[A <: java.io.Closeable, S](resource: => A < S)(using Frame): A < (Resource & IO & S) =
```

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -160,10 +126,75 @@ object Resource:
...
+                    IO.ensure(finalizer.close),
+                    Abort.run[Any]
+                ).map { result =>
+                    val io =
+                        result match
+                            case result @ Result.Error(e) =>
+                                val ex =
+                                    e match
+                                        case e: Throwable => e
+                                        case _            => new KyoException(s"Failed $e")
```
hearnadam, line 139: I would be inclined to think this should be `Absent`? I suppose I understand the need to signal if it was a failure...

In ZIO, the finalizers are in the form: `Exit[Any, Any] => UIO[Any]` - perhaps we can pass forward `Maybe[Error[Any]]`?

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -160,10 +126,75 @@ object Resource:
...
+                                finalizer.close(Present(ex))
+                            case _ => Kyo.unit
```
hearnadam, line 141: Don't we need to run them here too?

kyo-core/shared/src/main/scala/kyo/scheduler/Finalizers.scala
```diff
@@ -6,67 +6,69 @@ import kyo.discard
...
 
     extension (e: Finalizers)
 
+        def isEmpty: Boolean = e eq Absent
+
         /** Adds a finalizer function. */
-        def add(f: () => Unit): Finalizers =
+        def add(f: Maybe[Throwable] => Unit): Finalizers =
             (e: @unchecked) match
                 case e if e.equals(Absent) || e.equals(f) => f
```
hearnadam, line 28:
```suggestion
                case e if e.isEmpty || e.equals(f) => f
```

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/Resource.scala
fwbrasil: the resource has to be evaluated first anyway so the by-name param doesn't seem necessary

kyo-core/shared/src/main/scala/kyo/Resource.scala
fwbrasil: good point, it's better to preserve the information. I'll update

kyo-core/shared/src/main/scala/kyo/Resource.scala
fwbrasil: It isn't strictly necessary since, if there wasn't a failure, then `IO.ensure` is always executed. I've updated it since the closing a finalizer is idempotent

kyo-core/shared/src/main/scala/kyo/Resource.scala
hearnadam: By name is important for trapping exceptions in a suspension, no?

fwbrasil: Good point. I'm updating `acquireRelease` to introduce an `IO` suspension as well

PR #1195, opened 2025-05-17
[core] move fromCompletionStage from Fiber to Async

fwbrasil:
### Problem

`Fiber` is a low-level effect with complex semantics for newcomers. Its functionality is essential for integrations and more low-level code but `Async` provides better usability and covers a majority of use cases. For this reason, `Fiber`'s API has become more limited over time and we don't even provide common combinators like `race` for it since the same functionality can be achieved via `Async` combinators and `Async.run` to obtain a `Fiber`. I hadn't noticed that `fromCompletionStage` is still in `Fiber` and using an old method naming pattern with a `Fiber` suffix.

### Solution

Move the method to `Async` and remove `fromCompletionStageFiber`.

### DIFF
--- a/kyo-core/jvm/src/main/scala/kyo/internal/AsyncPlatformSpecific.scala
+++ b/kyo-core/jvm/src/main/scala/kyo/internal/AsyncPlatformSpecific.scala
@@ -4,6 +4,10 @@ import java.util.concurrent.CompletionStage
 import kyo.*
 
 trait AsyncPlatformSpecific:
+
+    def fromFuture[A](cs: CompletionStage[A])(using Frame): A < Async =
+        fromCompletionStage(cs)
+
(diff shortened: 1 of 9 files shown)


kyo-core/jvm/src/main/scala/kyo/internal/AsyncPlatformSpecific.scala
```diff
@@ -3,17 +3,14 @@ package kyo.internal
 import java.util.concurrent.CompletionStage
 import kyo.*
 
-trait FiberPlatformSpecific:
+trait AsyncPlatformSpecific:
     def fromCompletionStage[A](cs: CompletionStage[A])(using Frame): A < Async =
```
hearnadam, line 7: I think we can alias this to `fromFuture` as well. It should typecheck as an overload (to Scala's future).

hearnadam (approved):

PR #1196, opened 2025-05-17
[core] remove putFiber/takeFiber from Channel and Hub

fwbrasil:
### Problem

Following the rationale of https://github.com/getkyo/kyo/pull/1195,  `Channel` and `Hub` should avoid exposing `Fiber` in its APIs.

### Solution

Remove the `takeFiber` and `putFiber` methods from both `Channel` and `Hub`. There's a performance cost in case a `Fiber` instance is needed but I think usability should be a priority.


hearnadam:
Looks like README needs updating

PR #1197, opened 2025-05-19
WIP: allow AsyncShift in kyo-direct

ahoy-jon:
### Problem

Kyo doesn't allow AsyncShift 
```scala
defer:
  xs.map(_.now) //Kyo.traverse(xs)(_.now)
```
and could be a nice to have, similar to direct-style
```scala
context:
  xs.map(_.?)
```

### Solution
Allow some Async Shifts during validation, like `Seq.map`, ... 

### Resources

* dotty-cps : https://github.com/dotty-cps-async/dotty-cps-async/blob/master/docs/HighOrderFunctions.rst#async-shift-substitution


ahoy-jon:
So, it works. I'm not a big fan of the current approach, which involves starting from the root tree, walking up to find the parent, and manually checking the type and method name.

I reviewed the code in `dotty-async-cps` and did not find any built-in support for this.

The plan is to go with a brute-force strategy:
- List the main combinations of types and method names (like `Seq.map`, `Option.map`, etc.)
- Build a custom `Trees.traverse` to handle everything in a single pass and avoid repeated tree walks.

kyo-direct/shared/src/main/scala/kyo/Validate.scala
```diff
@@ -126,7 +147,7 @@ private[kyo] object Validate:
                     |}""".stripMargin)}""".stripMargin.stripMargin
                 )
 
-            case tree @ DefDef(_, _, _, Some(body)) if !pure(body) =>
+            case tree @ DefDef(_, _, _, Some(body)) if !pure(body) && !validShift(body) =>
```
fwbrasil, line 150: If dotty-cps-async provides reasonable error messages for this scenario, another option is just removing the restriction. It'd be good if we can avoid the tree pattern matching

ahoy-jon: Ha, I see. The updated implementation if doing something else, just jump to part of the tree to check.

https://github.com/getkyo/kyo/pull/1197/files#diff-4f91081931ccb84da7b96c5f6764fae9e5c84903e1e521379f51dad387dd7e73R97-R99

We could remove the restriction, however there are constructions that are valid for `dotty-cps-async`, and not encouraged in the context of Kyo (using var, ...).

kyo-direct/shared/src/main/scala/kyo/Validate.scala
```diff
@@ -1,23 +1,114 @@
...
 
-        Trees.traverse(expr.asTerm) {
             case Apply(TypeApply(Ident("now" | "later"), _), List(qual)) =>
+                @tailrec
+                def dive(qual: Tree): Unit =
+                    qual match
+                        case Block(quals, last) =>
+                            quals.foreach(Trees.Step.goto)
+                            dive(last)
+                        case _ =>
```
ahoy-jon, line 108: improve current checks to restart traversing the tree after a match that doesn't necessarily lead to a failure.

ahoy-jon:
@fwbrasil The goal of this implementation is to let each tree match either:
- fail immediately,
- fail conditionally and continue into subtrees, 
- or descend selectively into subtrees. 

This enables restarting validation at various points in the tree.

The previous version would either: 
- fail immediately
- or fail conditionally (but stop the tree walk on that part).

---
I don't think lifting the restriction would work, that could allow constructs that are incompatible with Kyo, such as:
```scala
defer:
  Option(1).map { i => 
     def f(x: Int) = IO(x).now 
     f(i) 
  }
```
...

kyo-direct/shared/src/main/scala/kyo/Trees.scala
```diff
@@ -3,6 +3,33 @@ package kyo
...
 
 object Trees:
+
+    class Step(private var _gotoTrees: Seq[Any]):
+        def this() = this(Seq.empty)
+
+        protected[Trees] def addTree(using q: Quotes)(tree: q.reflect.Tree): Unit =
+            _gotoTrees = _gotoTrees :+ tree
+
+        protected[Trees] def gotoTrees(using q: Quotes): Seq[q.reflect.Tree] = _gotoTrees.asInstanceOf
```
ahoy-jon, line 13: it's painful to type it correctly, if someone has an idea on the best way to do that, without compromising on usability, that could help!

hearnadam (commented):
Very exciting

kyo-direct/shared/src/main/scala/kyo/Validate.scala
```diff
@@ -13,11 +16,80 @@ private[kyo] object Validate:
...
+                    qualifier.tpe <:< TypeRepr.of[ArrayOps[?]] |
+                    qualifier.tpe <:< TypeRepr.of[Either[?, ?]] |
+                    qualifier.tpe <:< TypeRepr.of[Either.LeftProjection[?, ?]]
+
+                // ?? WithFilter
+                // ?? PartialFunction[?, ?]
+                // ?? Function1[?, ?] (andThen / compose)
+                // ?? Using
+
+            inline def validName: Boolean = methodName match
```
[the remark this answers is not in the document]
ahoy-jon: That was the case a couple of commits before. 
It supposed to be faster using a match + we get some semantic grouping.

What do you think? (you choose)

kyo-direct/shared/src/main/scala/kyo/Validate.scala
```diff
@@ -13,11 +16,80 @@ private[kyo] object Validate:
...
             !Trees.exists(tree) {
                 case Apply(TypeApply(Ident("now"), _), _) => true
-                case _                                    => false
             }
+        end pure
+
+        def validAsyncShift(select: Select): Boolean =
+            val Select(qualifier, methodName) = select
+            inline def validType =
+                qualifier.tpe <:< TypeRepr.of[Iterable[?]] |
```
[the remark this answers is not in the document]
ahoy-jon: indeed, https://github.com/dotty-cps-async/dotty-cps-async/blob/b2233ff122bd1b06e83b0839f79dc73cbe5826fd/shared/src/main/scala/cps/runtime/IterableAsyncShift.scala#L211

adding it with a test

hearnadam (approved):

fwbrasil (approved):
🚀

PR #1199, opened 2025-05-21
Add `Async.yieldNow`

HollandDM:
### Problem
Other effect systems has a special effect that yield the current task, stopping it from its current execution and send it back to the scheduler.

### Solution
Add a `yieldNow` feature for `Async`, which will yield the current task

### Notes
This is a result of my learning & understanding of the current scheduler, so it can be wrong, or unnecessary. Feel free to close in that case.

### DIFF
--- a/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
@@ -1416,4 +1416,62 @@ class AsyncTest extends Test:
         }
     }
 
+    "yield" - {
+        "should yield" in run {
+            def computation(
+                name: String,
+                cur: Int < IO,
+                nextpred: Int => Boolean,
+                endPred: Int => Boolean,
+                next: Unit < IO
+            ) =
+                def loop: Unit < Async =
+                    for
+                        current <- cur
+                        _ <- if endPred(current) then
+                            Kyo.unit
+                        else if nextpred(current) then
+                            next.andThen(Async.yieldNow).andThen(loop)
+                        else
+                            Async.yieldNow.andThen(loop)
+                    yield ()
+                    end for
+                end loop
+                loop
+            end computation
+
+            for
+                counter <- AtomicInt.init(0)
+                list    <- AtomicRef.init[List[Int]](Nil)
+                _ <- Async.race(
+                    computation(
+                        "even",
+                        counter.get,
+                        _ % 2 == 0,
+                        _ >= 10,
+                        for
+                            cur <- counter.get
+                            _   <- list.getAndUpdate(cur :: _)
+                            _   <- counter.getAndIncrement
+                        yield ()
+                    ) ::
+                        computation(
+                            "odd",
+                            counter.get,
+                            _ % 2 == 1,
+                            _ >= 10,
+                            for
+                                cur <- counter.get
+                                _   <- list.getAndUpdate(cur :: _)
+                                _   <- counter.getAndIncrement
+                            yield ()
+                        ) ::
+                        Nil
+                )
+                result <- list.get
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
```diff
@@ -1416,4 +1416,62 @@ class AsyncTest extends Test:
         }
     }
 
+    "yield" - {
```
HollandDM, line 1419: Not sure how to test `yieldNow`, as this test will work without it anyway

fwbrasil:
I had [briefly added](https://github.com/getkyo/kyo/pull/1145) it recently. I think there can be scenarios where it might be useful but, in general, relying on the scheduler's preemption via time slices should be ok. Do you see some good application?

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
fwbrasil: `Safepoint` is able to preempt execution on any transformation, even when there isn't an effect suspension like in `(1: Int < Any).map(_ = 1)`. I had added the method in JS because the scheduler didn't have a preemption mechanism but the only case where a computation can take over a thread in the other platforms should be if it's not executing Kyo computations

fwbrasil:
> Oh, I think I missed the those updates.
yieldNows in other systems (Cats/ZIO) are mostly used when dealing with function calls that are known to be time consuming, because most of them yield automatically based on number of steps taken. Kyo use both steps taken && time slice, so I think it should be better in these cases.

It was a brief update :) Kyo actually doesn't do preemption based on the number of steps. Transformations always call `Safepoint.enter` before proceeding. It's also the way traces are collected. It might make sense to introduce step-based preemption as a performance optimization to reduce the overhead of `Safepoint.enter`, though.

> One other cases was like when I tried to https://github.com/zio/zio/pull/9662, one of my impl was to use yieldNow as a fiber version of onSpinwait. Its not on my final impl because yieldNow in ZIO does not work the way I want. Using yieldNow like this is rare, only when doing low level stuff.

If you're executing Kyo computations, even a hot loop should be fine. Interruptions use the same mechanism as preemption to stop the fiber. We have a few scenarios in the codebase showing that it's able to stop a hot loop. Example: https://github.com/getkyo/kyo/blob/f43acc7dad4cd760be64db2799daeb20ec84cbea/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala#L110

I think we don't need `yieldNow`. I imagine we could end up finding cases where it's useful but suspending with the existing effects might be enough.

HollandDM:
Ok, I'll close this then

HollandDM:
> It was a brief update :) Kyo actually doesn't do preemption based on the number of steps. Transformations always call Safepoint.enter before proceeding. It's also the way traces are collected. It might make sense to introduce step-based preemption as a performance optimization to reduce the overhead of Safepoint.enter, though.

I though `maxStackDepth` can be treated as a hard stop for steps, look likes it not the case

fwbrasil:
Ah, that only introduces a Defer suspension when the stack depth reaches the limit in recursive computations without effect suspensions. Any effect suspension automatically collapses the thread stack up to the point of the handler of the effect so these cases are quite rare. IO and Defer collapse the whole stack since they’re handled last. 

Kyo didn’t even provide stack safety for this case for a long time and I’m yet to see a realistic scenario where that would happen other than in benchmarking or testing. The automatic suspensions by thread stack depth were more a decision to improve usability given that people can make mistakes and to have one less way that Kyo is different 😅

HollandDM:
Ok. It seems like I only understand `kyo` enough to patch up some surface features. The internal runtime are still kind of alien to me 😅

Issue #903, opened 2024-12-05
defer blocks discarding unit effects


johnhungerford:
### Minimized example

```scala
import kyo.*

def runInner: Unit < Async = defer:
  await(Console.println("test"))

@main def main(): Unit =
  import AllowUnsafe.embrace.danger
  val _ =  KyoApp.Unsafe.runAndBlock(Duration.Infinity)(runInner)
```

scalac options in build.sbt:
```scala
scalacOptions ++= Seq(
  "-Wvalue-discard", 
  "-Wnonunit-statement", 
  "-Wconf:msg=(discarded.*value|pure.*statement):error",
),
```

### Expected result
Should cause compiler error: `discarded non-Unit value of type Unit < (kyo.IO & kyo.Abort[java.io.IOException])` (If the `defer` block in `runInner` is replaced with a direct call to `Console.println`, this is what happens)

### Actual result
No compiler error or warnings, and the application runs without displaying anything message

johnhungerford:
Workaround: annotating `runInner` as `Any < ...` will cause the compiler to catch the error. It seems to only be a problem for unit values. Anyone working with kyo-direct might consider using `Any` in place of `Unit` until some solution is found.

fwbrasil:
I took a look at this and couldn't find a way to fix it. It seems the compiler just bypasses the check. cc/ @rssh

ahoy-jon:
I may have found something similar 


```scala
//> using scala 3.7.0-RC3

//> using option -Wvalue-discard
//> using option -Wnonunit-statement
//> using option -Wconf:msg=(unused.*value|discarded.*value|pure.*statement):error
//> using option -language:strictEquality

//> using dep io.getkyo::kyo-prelude::0.18.0
//> using dep io.getkyo::kyo-core::0.18.0
//> using dep io.getkyo::kyo-direct::0.18.0
//> using dep io.getkyo::kyo-combinators::0.18.0

import kyo.*

object KyoAppTest extends KyoApp {
  val print: Unit < IO = Console.printLine("YO")
  val fail: Unit < Abort[Exception] = Abort.fail(new Exception("fail"))

  val prg = print *> fail

  val prgWithDefer = defer {
    print.now
    fail.now
  }

  val oups: Unit < (IO & Abort[Exception]) = defer {
    Console.printLine("YO").now
    Abort.fail(new Exception("fail")).now
  }

  val fixOups: Unit < (IO & Abort[Exception]) = defer {
    Console.printLine("YO").now
[...]

PR #1202, opened 2025-05-23
force compile error of discarded defer block

ahoy-jon:
#903 

### Problem
Due to https://github.com/scala/scala3/issues/23018, some defer block are silently transforming from 

```scala
val oups:Int < Abort[String] = Abort.fail("oups")

val x : Unit < Any = {
  defer:
    oups.now + 1
}
```

to 
```scala
val x: Unit < Any = {
  //discard
  kernel.`<`.lift(())
}
```
without any warning.

### Solution
make the compilation fail if we try to lift `Unit < A` into `Unit < B` without `B <:< A`.

### DIFF
--- a/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
@@ -427,8 +427,6 @@ object `<`:
                 |Please remove the type constraint on Left Hand Side.
                 |More info : https://github.com/getkyo/kyo/issues/903""".stripMargin
         )
-
-        '{ ??? }
     end liftUnitImpl
 
     /** Converts a pure single-argument function to an effectful computation. */
--- a/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
@@ -422,7 +422,7 @@ object `<`:
     def liftUnitImpl[S1: Type, S2: Type](v: Expr[Unit < S1])(using quotes: Quotes): Expr[Unit < S2] =
         import quotes.reflect.*
         val source = TypeRepr.of[S1].show
-        report.error(
+        report.errorAndAbort(
             s"""Cannot lift `Unit < $source` to the required type (`Unit < ?`).
                 |Please remove the type constraint on Left Hand Side.
                 |More info : https://github.com/getkyo/kyo/issues/903""".stripMargin
--- a/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
@@ -5,6 +5,9 @@ import kyo.kernel.internal.*
 import scala.annotation.nowarn
 import scala.annotation.tailrec
 import scala.language.implicitConversions
+import scala.quoted.Expr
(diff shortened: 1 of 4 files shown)


kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -414,6 +417,20 @@ object `<`:
             case kyo: Kyo[?, ?] => Nested(kyo)
             case _              => v.asInstanceOf[A < S]
 
+    implicit inline def liftUnit[S1, S2](inline v: Unit < S1): Unit < S2 = ${ liftUnitImpl[S1, S2]('v) }
+
+    def liftUnitImpl[S1: Type, S2: Type](v: Expr[Unit < S1])(using quotes: Quotes): Expr[Unit < S2] =
+        import quotes.reflect.*
+        val source = TypeRepr.of[S1].show
+        report.error(
```
hearnadam, line 425:
```suggestion
        report.errorAndAbort(
```

kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -414,6 +417,18 @@ object `<`:
...
             case _              => v.asInstanceOf[A < S]
 
+    implicit inline def liftUnit[S1, S2](inline v: Unit < S1): Unit < S2 = ${ liftUnitImpl[S1, S2]('v) }
+
+    private def liftUnitImpl[S1: Type, S2: Type](v: Expr[Unit < S1])(using quotes: Quotes): Expr[Unit < S2] =
+        import quotes.reflect.*
+        val source = TypeRepr.of[S1].show
+        report.errorAndAbort(
+            s"""Cannot lift `Unit < $source` to the required type (`Unit < ?`).
+                |Please remove the type constraint on Left Hand Side.
```
hearnadam, line 427: I think this may be too specific...

Perhaps better to ask the user if there is an effect mismatch

ahoy-jon: thanks. Changed it to:
```
Cannot lift `Unit < kyo.Abort[scala.Predef.String]` to the expected type (`Unit < ?`).
[info]       │ This may be due to an effect type mismatch.
[info]       │ Consider removing or adjusting the type constraint on the left-hand side.
[info]       │ More info : https://github.com/getkyo/kyo/issues/903
```

ahoy-jon:
it's good, it finds errors: 🥳
```
[error] -- Error: /Users/jon/Documents/GitHub/kyo/kyo-examples/jvm/src/main/scala/examples/ledger/db/Log.scala:42:71 
[error] 42 |        private[Log] def flushLoop(interval: Duration): Unit < Async = defer {
[error]    |                                                                       ^
[error]    |Cannot lift `Unit < kyo.Async$package.Async & kyo.Abort[kyo.Closed]` to the expected type (`Unit < ?`).
[error]    |This may be due to an effect type mismatch.
[error]    |Consider removing or adjusting the type constraint on the left-hand side.
[error]    |More info : https://github.com/getkyo/kyo/issues/903
[error] 43 |            Async.sleep(interval).now
[error] 44 |            val entries = q.drain.now
[error] 45 |            append(entries).now
[error] 46 |            flushLoop(interval).now
[error] 47 |        }
```

hearnadam (approved):

Issue #993, opened 2025-01-09
Queue and Channel can't be used with direct syntax


fwbrasil:
There's a compiler crash when opaque types are used with `kyo-direct`. It can be worked around by providing a type bound to the opaque type, which is why composite effects like `Async` work ok. Both `Queue` and `Channel` are opaque types and don't have a type bound. I'm not sure what we could put as the bound since we don't want to expose the APIs as their `Unsafe` counterparts. We might need a new wrapping class/allocation for them. More context: https://github.com/getkyo/kyo/pull/985/files#r1907969353

fwbrasil:
@rssh in case you have some time to have a look. I imagine it could be due to Kyo's wrapper on top of dotty-cps-async but we can't use opaque types without bounds with the direct syntax because of compiler crashes. To reproduce, you only need the `kyo-core` and `kyo-direct` modules and then try to use `Queue` or `Channel` with the direct syntax.

ahoy-jon:
still an issue: 
```scala
  val maybe1: Maybe[Int] < IO = Maybe(1)

  //works
  defer {
    val x = maybe1.now
    x.contains(1)
  }
  //does not compile
  defer {
    maybe1.now.contains(1)
  }
```
with 
```scala
opaque type Maybe[+A] >: (Absent | Present[A]) = Absent | Present[A]
```
___
```
Error: java.lang.AssertionError: assertion failed
        at scala.runtime.Scala3RunTime$.assertFailed(Scala3RunTime.scala:11)
        at dotty.tools.dotc.typer.Typer.retractDefDef(Typer.scala:2901)
        at dotty.tools.dotc.typer.Typer.typedDefDef(Typer.scala:2905)
        at dotty.tools.dotc.typer.Typer.typedNamed$1(Typer.scala:3542)
        at dotty.tools.dotc.typer.Typer.typedUnadapted(Typer.scala:3647)
```

fwbrasil:
@ahoy-jon does https://github.com/getkyo/kyo/pull/1204 fix this as well?

fwbrasil:
ah, github didn't make the transition to closed automatically. I'm not sure what syntaxes it picks up but I normally use `Fixes #999`

PR #1204, opened 2025-05-24
Fix/993

ahoy-jon:
On #993 

### Problem

cannot use constructs like:
```scala
val maybe: Maybe[Int] < IO = 1
defer: 
  maybe.now.contains(1)

```
without a compiler panic.

### Solution
more macros to turn this block into
```scala
defer: 
  val TEMP = maybe.now
  TEMP.contains(1)
```

### DIFF
--- a/kyo-direct/shared/src/main/scala/kyo/Direct.scala
+++ b/kyo-direct/shared/src/main/scala/kyo/Direct.scala
@@ -151,11 +151,11 @@ private def impl[A: Type](body: Expr[A])(using quotes: Quotes): Expr[Any] =
                     args0
                 ),
                 args1
-            ) if t.tpe <:< TypeRepr.of[Maybe[Any]] =>
+            ) if t.tpe.typeSymbol.flags.is(Flags.Opaque) =>
 
             val newVal: Symbol = Symbol.newVal(
                 Symbol.spliceOwner,
-                freshName("genKyoOpaque"),
+                freshName("opaqueAlias"),
                 t.tpe,
                 Flags.Private,
                 Symbol.noSymbol
--- a/kyo-direct/shared/src/main/scala/kyo/Direct.scala
+++ b/kyo-direct/shared/src/main/scala/kyo/Direct.scala
@@ -125,7 +125,7 @@ private def impl[A: Type](body: Expr[A])(using quotes: Quotes): Expr[Any] =
         loop(l, Nil)
     end flatten
 
-    val pending =
+    val pending: TypeRepr =
         flatten(effects) match
             case Nil =>
                 TypeRepr.of[Any]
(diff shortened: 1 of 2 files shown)


kyo-direct/shared/src/main/scala/kyo/Direct.scala
```diff
@@ -125,18 +125,54 @@ private def impl[A: Type](body: Expr[A])(using quotes: Quotes): Expr[Any] =
...
+                        types0
+                    ),
+                    args0
+                ),
+                args1
+            ) if t.tpe <:< TypeRepr.of[Maybe[Any]] =>
+
+            val newVal: Symbol = Symbol.newVal(
+                Symbol.spliceOwner,
+                freshName("genKyoOpaque"),
```
hearnadam, line 158: Is it possible to use the name of the symbol itself? Otherwise I would prefer something simple like `opaqueAlias_N1` or something.

ahoy-jon:
it was problematic for: 
- `contains`
- `fold`
- `when`

kyo-direct/shared/src/test/scala/kyo/HygieneTest.scala
```diff
@@ -236,4 +236,15 @@ class HygieneTest extends Test:
                """.stripMargin
         )("Cannot lift `Unit < kyo.Abort[scala.Predef.String]` to the expected type (`Unit < ?`).")
     }
+
+    "issue 993" in {
```
hearnadam, line 240:
```suggestion
    "opaque types issue #993" in {
```

kyo-direct/shared/src/main/scala/kyo/Direct.scala
```diff
@@ -125,18 +125,54 @@ private def impl[A: Type](body: Expr[A])(using quotes: Quotes): Expr[Any] =
...
+                        Apply(
+                            ta: TypeApply,
+                            List(a @ Apply(TypeApply(Ident("now"), List(t, s)), List(_)))
+                        ),
+                        types0
+                    ),
+                    args0
+                ),
+                args1
+            ) if t.tpe <:< TypeRepr.of[Maybe[Any]] =>
```
hearnadam, line 154: Does this just work with `Maybe`? Perhaps we can use a union of all the internal opaque types? Not too scalable...

ahoy-jon: I am doing the generalized version right now

kyo-direct/shared/src/main/scala/kyo/Direct.scala
ahoy-jon: noSymbols, it's doing "genKyoOpaque_N1", fixing it

hearnadam (approved):

fwbrasil:
You're on a roll!

ahoy-jon:
Thanks!

A couple of more issues, and I try to make a {cap}!

image

PR #1218, opened 2025-05-29
[feature] Update Kyo's foreach

### LINKED ISSUE #1020: `Kyo.foreach*` should support more than Seq

HollandDM:
### Problem
`foreach` only handle generic input, but not output the corresponding type
resolve #1020

### Solution
update to output the same collection type as input

#### Limitiations:
I do think there are some new allocations here, especially around empty cases (`newBuilder` && `result` call can have some allocations). Maybe we can create our own `BuildFrom` that can cache the empty cases, or just create specialized function for each collection types


HollandDM:
> There is one point, I think it's safer to use `source.iterableFactory.from(ch)` instead of source.newBuilder...

Yeah, this should be better for performance, as it's expose `empty` and `from` directly (which normally will be optimized on concrete classes), I'll update it later

HollandDM:
Updated

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -153,24 +154,31 @@ object Kyo:
...
+        Safepoint
+    ): CC[B] < S =
+        val factory = source.iterableFactory
         source.knownSize match
-            case 0 => Chunk.empty
+            case 0 => factory.empty[B]
             case 1 =>
                 val head = source.iterator.next()
-                f(head).map(Chunk(_))
+                f(head).map(factory(_))
```
HollandDM, line 166: we have to type `factory(_)` here, `factory` alone doesn't work

HollandDM:
Updated!

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -153,24 +154,31 @@ object Kyo:
...
-                            if seq.isEmpty then Loop.done(acc)
-                            else f(seq.head).map(u => Loop.continue(seq.tail, acc.append(u)))
-                    case other =>
-                        val indexedSeq = toIndexed(other)
-                        val size       = indexedSeq.length
-                        Loop.indexed(Chunk.empty[B]): (idx, acc) =>
-                            if idx == size then Loop.done(acc)
-                            else f(indexedSeq(idx)).map(u => Loop.continue(acc.append(u)))
+                for
+                    chunk <- source match
```
hearnadam, line 169: `for` nesting can get a bit awkward with significant indentation. Perhaps

`(...).map(factory.from)`?

kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -157,7 +157,7 @@ object Batch:
                     case (Expanded[A @unchecked, S @unchecked](_, source, _)) => source
                     case _ => () // Used as a placeholder for items that aren't source calls
                 }
-            Kyo.foreach(pending.toSeq) { tuple =>
+            Kyo.foreach(Chunk.from(pending)) { tuple =>
```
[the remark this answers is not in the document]
hearnadam: Do we need this? Shouldn't it be converted in `Kyo.foreach`?

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
HollandDM: For me these both are awkward, but I will update it if you want

ahoy-jon (commented):
I didn't see anything weird! Can wait for is to be merged to update what is done in direct.

I would recommend to abstract what is done with `source.iterableFactory`with a private extension. (cf. https://github.com/getkyo/kyo/blob/a9ca9d71220f1bc8337ba9c69a8b47d8dc126d3a/kyo-direct/shared/src/main/scala/kyo/internal/AsyncShift.scala#L28-L37 )

On master, there are a lot more `Kyo.*` methods, that woud be tideous to repeat:
```scala
val factory = source.iterableFactory
... .map(factoryFrom)
```
compare to 
```scala
given Builder = Builder.from(source)
...   .toCC
```

kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -603,20 +603,22 @@ object Chunk extends StrictOptimizedSeqFactory[Chunk]:
...
-                            val array = source.iterator.toArray(using erasedTag[A])
-                            array.length match
-                                case 0 => empty[A]
-                                case 1 => single(array(0))
-                                case _ => Compact(array)
-                            end match
+                            other match
+                                case seq: IndexedSeq[A] => FromSeq(seq)
+                                case _ =>
+                                    val array = source.iterator.toArray(using erasedTag[A])
```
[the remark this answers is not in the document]
fwbrasil: Yes, the lack of specialization in Scala 3 is a major factor for this design decision. If we don’t box the array, it’ll still box “on demand” in generic contexts. For example, given Chunk(1, 2, 3).map(f).map(g), every item will first get boxed to call f, unboxed to build the new unboxed Chunk, then boxed again to call g. By keeping chunks boxed, we avoid at least one allocation between the execution of f and g

HollandDM:
@fwbrasil I took the `ForeachBench` and ran only first benchmark. Here is the result:

```
[info] Benchmark                 (size)   Mode  Cnt       Score      Error  Units
(current)
[info] ForeachBench.kyoForeach     1024  thrpt    3  168414.079 ± 3714.576  ops/s
[info] ForeachBench.kyoForeach    65536  thrpt    3    2566.369 ±   88.075  ops/s
[info] ForeachBench.kyoForeach   262144  thrpt    3     405.964 ±   11.555  ops/s
[info] ForeachBench.kyoForeach  1048576  thrpt    3     137.332 ±  259.822  ops/s
(this PR)
[info] ForeachBench.kyoForeach     1024  thrpt    3  111718.799 ± 3588.155  ops/s
[info] ForeachBench.kyoForeach    65536  thrpt    3    1441.030 ±   12.437  ops/s
[info] ForeachBench.kyoForeach   262144  thrpt    3     336.556 ±    8.316  ops/s
[info] ForeachBench.kyoForeach  1048576  thrpt    3      18.033 ±    1.110  ops/s
```
It seems that the overhead are quite big here.
Should we try specialized methods instead of generic like this?

kyo-kernel/shared/src/main/scala/kyo/internal/KyoForeachSpecialized.scala
```diff
@@ -0,0 +1,1804 @@
...
+    final def foreachConcat[A, B, S](source: Chunk[A])(f: Safepoint ?=> A => IterableOnce[B] < S)(using
+        Frame,
+        Safepoint
+    ): Chunk[B] < S =
+        val len = source.length
+        len match
+            case 0 => Chunk.empty
+            case _ =>
+                Loop.indexed(Chunk.empty[Chunk[B]]): (idx, acc) =>
+                    if idx == len then Loop.done(acc.flattenChunk)
```
HollandDM: old code use `Chunk.concat`, which is super slow when benchmarking. So I adopted the strategy of `List` and `Vector` here. This yield a lot better performance (local benchmark show 100x improvement)

kyo-prelude/shared/src/main/scala/kyo/Batch.scala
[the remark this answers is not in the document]
HollandDM: pending is not `Chunk`, so the whole result will not be `Chunk`

kyo-prelude/shared/src/main/scala/kyo/Batch.scala
```diff
@@ -145,7 +145,7 @@ object Batch:
         def expand(items: Chunk[Item]): Chunk[Chunk[Item]] < S =
             Kyo.foreach(items) {
                 case ToExpand[A @unchecked, S @unchecked](seq: Seq[Any], cont) =>
-                    Kyo.foreach(seq)(v => capture(cont(v)))
+                    Kyo.foreach(Chunk.from(seq))(v => capture(cont(v)))
```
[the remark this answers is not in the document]
HollandDM: typing doesn't work because seq is `Seq`, and we need `Chunk`

HollandDM:
BTW, it seems that `build-pr` is flaky here, I run build and test in local successfully

hearnadam:
@HollandDM I appreciate the work you put into exploring this and would love if it didn't go to waste.

What do you think about introducing `Kyo.foreach` alternatives for common iterables for now? This could defer the need to have a truly generic implementation.

One note, I don't think we should use the `KyoForeachSpecialized` trait, as it can make it harder for users to find methods.

If we need some indirectly, perhaps we can use inline methods in an object.

HollandDM:
the trait was mainly used to separate the specialized code from the `Kyo.class` to make the file small.
I will try to come up with some implementation that doesn't blow out the code size. As of right now, it's quite big.

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -752,41 +432,2002 @@ object Kyo:
...
+    /** Applies an effect-producing function to each element of a `Set`.
+      *
+      * @param source
+      *   The input `Set`
+      * @param f
+      *   The effect-producing function to apply to each element
+      * @return
+      *   A new effect that produces a Set of results
+      */
+    def foreach[A, B, S](source: Set[A])(f: Safepoint ?=> A => B < S)(using Frame, Safepoint): Set[B] < S =
```
ahoy-jon, line 1788: it doesn't need a `CanEqual` on `B`? (not urgent)

kyo-kernel/shared/src/main/scala/kyo/internal/KyoForeachSpecialized.scala
```diff
@@ -0,0 +1,1804 @@
+package kyo.internal
+
+import kyo.*
+import kyo.kernel.internal.Safepoint
+import scala.annotation.tailrec
+import scala.collection.IterableOps
+import scala.collection.immutable.LinearSeq
+import scala.reflect.ClassTag
+
+private[kyo] trait KyoForeachSpecialized:
```
ahoy-jon: it's not deleted?

kyo-kernel/shared/src/test/scala/kyo/KyoTest.scala
```diff
@@ -614,18 +617,18 @@ class KyoTest extends Test:
...
+        //     val used = new java.util.concurrent.atomic.AtomicBoolean(false)
+        //     new Iterable[A]:
+        //         def iterator: Iterator[A] =
+        //             if !used.compareAndSet(false, true) then
+        //                 throw new IllegalStateException("Already consumed!")
+        //             else Iterator(a*)
+        //     end new
+        // end oneShot
+
+        // collectionTests[Iterable]("IterableOnce", [X] => seq => oneShot(seq*))
```
ahoy-jon: this test come from the work on `kyo-direct`, Just remove it.  (it should have passed)

kyo-prelude/shared/src/main/scala/kyo/Batch.scala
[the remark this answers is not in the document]
ahoy-jon: That's why I think, `Seq` is missing (and eventually all the super type of `Chunk` (`Iterable` / ...)).
There is no need to convert a `Chunk` to a `Seq` back.

(for another PR)

ahoy-jon (commented):
@HollandDM overall it's good, a couple of comments left. Thanks a lot for the tedious work.

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
HollandDM: should we, if we add that then we should add CanEqual for Map's K as well?

kyo-kernel/shared/src/main/scala/kyo/internal/KyoForeachSpecialized.scala
HollandDM: deleted

kyo-kernel/shared/src/test/scala/kyo/KyoTest.scala
HollandDM: will remove for now

ahoy-jon:
> updated!, please recheck

Sorry, I was off, taking a look at it as soon as possible!

ahoy-jon (commented):
Overall it's good, there are things I would have liked to change (probably I have to do it):
 - factorise the test (using inline/code generation/ ...)
 - performance check for Lists, currently a lot of methode are doing a List reverse build, maybe using a Chunk, then turning it to a List is better (and also factorise the implementation)
 
However, I think the best course of action is to merge, and rework it after.

HollandDM:
ok, can you re-run the CI, it seems I got a flakky test, because my local run did not fail.
For `List`, you can try some optimizations if you find the current impl is slow

ahoy-jon:
@HollandDM indeed, the CI fail on flaky.

What is your opinon on the implementation of `Kyo.foreach(List(1, 2, 3, ...))(_ + 1)` ? From what you have tested, is it better like that, or could it be better using `Kyo.foreach(Chunk.from ... )(_ + 1).map(_.toList)` ?

HollandDM:
Append without the builder like this will result in a deep ‘Append’ node. ‘List.from’ then consume this  by convert chunk into indexed form, then do the ListBuffer builder and then accumulate into a list.
All of this is way more overhead compared to a single reverse, which has optimized code in List (one traversal only)

ahoy-jon:
@HollandDM Indeed, ChunkBuilder is designed for concurrency and multi-shot usage, there's probably something worth exploring in that area.

Thanks again for the meticulous work 💪!

Note: This stuff is wild. I worked a bit on `Chunk`/`NonEmptyChunk` in ZIO back in 2020. There are multiple ways to implement a Chunk, plus Scala's boxing issues... honestly, this topic can drive anyone digging into it a little crazy 🤪

HollandDM:
`Chunk` is way more complicated as they try to handle all the primitives themself (it will fail when dealing with types like `Int | Boolean`), and also aimed for the collection to be replacement for every other `Seq` in an ZIO app.
Kyo's `Chunk` is kind of like a `Array[Object]` with some special ops made optimized. It seems to trade simplicity for general usability. For example, in this PR I have to switch `foreachConcat` from old implementation to the new one with a (kind of) unconventional implementation (instead of concat right way, I keep those as nested Chunk, then flattenChunk in the end). But without it, the `foreachConcat` will be suffer a lot (my local benchmark shows that the old implementation is slower than both `List` and `Vector`)

PR #1219, opened 2025-05-30
[improvement] var args to Choice.eval (fixes #1216)

### LINKED ISSUE #1216: [feature]: Choice.eval Varargs Ergonomics
### Problem Statement

```scala
Choice.eval(Seq(true, false))
```
Doesn't look as nice as
```scala
Choice.eval(true, false)
```

### Proposed Solution

```scala
    inline def eval[A](a:A, as:A*)(using inline frame: Frame): A < Choice = eval[A](a +: as)
```

ahoy-jon:
`Choice.eval` can be used as a var args. (cf #1216)

---

- [x] check if we need to create another method for seq.
- [x] update documentation

### DIFF
--- a/kyo-combinators/shared/src/main/scala/kyo/AbortCombinators.scala
+++ b/kyo-combinators/shared/src/main/scala/kyo/AbortCombinators.scala
@@ -462,7 +462,9 @@ class ForAbortOps[A, S, E, E1 <: E](effect: A < (Abort[E] & S)) extends AnyVal:
         reduce: Reducible[Abort[ER]],
         frame: Frame
     ): A < (S & reduce.SReduced & Choice) =
-        Abort.run[E1](effect.asInstanceOf[A < (Abort[E1 | ER] & S)]).map(e => Choice.eval(e.foldError(List(_), _ => Nil)*))
+        Abort.run[E1](effect.asInstanceOf[A < (Abort[E1 | ER] & S)]).map(result =>
+            result.foldError(value => Choice.eval(value), _ => Choice.drop)
+        )
 
     /** Translates the partial Abort[E1] effect to an Abort[Absent] effect in case of failure.
       *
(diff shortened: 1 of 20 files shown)


kyo-combinators/shared/src/main/scala/kyo/AbortCombinators.scala
```diff
@@ -71,7 +71,7 @@ extension [A, S, E](effect: A < (Abort[E] & S))
         ct: SafeClassTag[E],
         fr: Frame
     ): A < (S & Choice) =
-        effect.result.map(e => Choice.eval(e.foldError(List(_), _ => Nil)))
+        effect.result.map(e => Choice.eval(e.foldError(List(_), _ => Nil)*))
```
hearnadam, line 74: Does this work?
```suggestion
        effect.result.map(e => Choice.eval(e.toMaybe*))
```

hearnadam (approved):

ahoy-jon:
@hearnadam fix after merge + a bit of cleaning

kyo-prelude/shared/src/main/scala/kyo/Choice.scala
```diff
@@ -32,15 +32,29 @@ sealed trait Choice extends ArrowEffect[Seq, Id]
...
+      *   A sequence of candidate values to choose from.
       * @return
-      *   A computation that represents the selection of values from the sequence
+      *   A computation that represent multiple paths, one for each input value.
+      * @example
+      *   {{{Choice.evalFromSeq(Seq("a", "b", "c"))}}}
       */
-    inline def eval[A](seq: Seq[A])(using inline frame: Frame): A < Choice =
-        ArrowEffect.suspend[A](Tag[Choice], seq)
+    inline def evalFromSeq[A](seq: Seq[A])(using inline frame: Frame): A < Choice =
```
fwbrasil, line 56: I think From does it fit well here. How about evalSeq?

fwbrasil (approved):

hearnadam (approved):

PR #1222, opened 2025-05-30
[feature] Stream.fromIterator

ahoy-jon:
Implement `Stream.fromIterator`

---
- [x] add buffer
- [ ] ~add fromIteratorCatching~ (another PR)

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -73,20 +73,27 @@ object StreamCoreExtensions:
             tag: Tag[Emit[Chunk[V]]],
             frame: Frame
         ): Stream[V, IO] =
-            val builder = ChunkBuilder.init[V]
+            Stream:
+                IO:
+                    val it      = v
+                    val size    = bufferSize max 1
+                    val builder = ChunkBuilder.init[V]
+                    val nextElements: Chunk[V] < IO = IO:
+                        var count = 0
+                        while count < size && it.hasNext do
+                            builder.addOne(it.next())
+                            count += 1
 
-            def nextElement(it: Iterator[V])(using Frame): Maybe[Chunk[V]] < IO =
-                val size = bufferSize max 1
-                IO.ensure(builder.clear()):
-                    var count = 0
-                    while count < size && it.hasNext do
-                        builder.addOne(it.next())
-                        count += 1
+                        val res = builder.result()
+                        builder.clear()
+                        res
 
-                    Maybe.when(count > 0)(builder.result())
-            end nextElement
+                    Loop.foreach[Unit, IO & Emit[Chunk[V]]]:
+                        nextElements.map(chunk =>
+                            if chunk.isEmpty then Loop.done
+                            else Emit.valueWith(chunk)(Loop.continue)
+                        )
 
-            Stream.repeatPresent(nextElement(v))
         end fromIterator
 
         /** Merges multiple streams asynchronously. Stream stops as soon as any of the source streams complete.
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -61,6 +61,17 @@ object StreamCoreExtensions:
                             _ <- emitMaybeChunksFromChannel(channel)
                         yield ()
 
+        def fromIterator[V, S](v: => Iterator[V] < S)(using
+            tag: Tag[Emit[Chunk[V]]],
+            frame: Frame
+        ): Stream[V, (S & IO)] =
+            inline def nextElement(it: Iterator[V]): Maybe[Seq[V]] < IO =
+                IO(if it.hasNext then Maybe.Present(Seq(it.next())) else Maybe.Absent)
+
+            Stream.repeatPresent(v.map(nextElement))
+
+        end fromIterator
+
         /** Merges multiple streams asynchronously. Stream stops as soon as any of the source streams complete.
           *
           * @note
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -78,21 +78,17 @@ object StreamCoreExtensions:
                     val it      = v
                     val size    = bufferSize max 1
                     val builder = ChunkBuilder.init[V]
-                    val nextElements: Chunk[V] < IO = IO:
-                        var count = 0
-                        while count < size && it.hasNext do
-                            builder.addOne(it.next())
-                            count += 1
 
-                        val res = builder.result()
-                        builder.clear()
-                        res
+                    Stream.repeatPresent(
+                        IO.ensure(builder.clear()):
+                            var count = 0
+                            while count < size && it.hasNext do
+                                builder.addOne(it.next())
+                                count += 1
 
-                    Loop.foreach[Unit, IO & Emit[Chunk[V]]]:
-                        nextElements.map(chunk =>
-                            if chunk.isEmpty then Loop.done
-                            else Emit.valueWith(chunk)(Loop.continue)
-                        )
+                            val chunk = builder.result()
+                            Maybe.when(chunk.nonEmpty)(chunk)
+                    ).emit
 
         end fromIterator
 
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -1,6 +1,7 @@
 package kyo
 
 import kyo.Async.defaultConcurrency
+import kyo.ChunkBuilder
 import kyo.kernel.ArrowEffect
 
 object StreamCoreExtensions:
@@ -61,19 +62,6 @@ object StreamCoreExtensions:
                             _ <- emitMaybeChunksFromChannel(channel)
                         yield ()
 
-        private inline def nextElement[V](it: Iterator[V], bufferSize: Int)(using Frame): Maybe[Chunk[V]] < IO =
-            val size = bufferSize max 1
-            IO:
-                val builder = Chunk.newBuilder[V]
-                var count   = 0
-                builder.clear()
-                while count < size && it.hasNext do
-                    builder.addOne(it.next())
-                    count += 1
-
-                Maybe.when(count > 0)(builder.result())
-        end nextElement
-
         /** Creates a stream from an iterator.
           *
           * @param v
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/test/scala/kyo/StreamCoreExtensionsTest.scala
```diff
@@ -216,6 +216,61 @@ class StreamCoreExtensionsTest extends Test:
...
+
+            "reuse same stream" in run {
+                val it     = Iterator(1, 2, 3)
+                val stream = Stream.fromIterator(it)
+                for
+                    first  <- stream.run
+                    second <- stream.run
+                yield
+                    assert(first == Chunk(1, 2, 3))
+                    assert(second.isEmpty) // Iterator is exhausted after first use
```
ahoy-jon, line 240: Should it behave like that?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -61,6 +61,17 @@ object StreamCoreExtensions:
...
                         yield ()
 
+        def fromIterator[V, S](v: => Iterator[V] < S)(using
+            tag: Tag[Emit[Chunk[V]]],
+            frame: Frame
+        ): Stream[V, (S & IO)] =
+            inline def nextElement(it: Iterator[V]): Maybe[Seq[V]] < IO =
+                IO(if it.hasNext then Maybe.Present(Seq(it.next())) else Maybe.Absent)
+
+            Stream.repeatPresent(v.map(nextElement))
```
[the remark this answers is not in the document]
ahoy-jon: Loop inside IO can pull `n` directly

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -61,6 +61,33 @@ object StreamCoreExtensions:
...
                         yield ()
 
+        /** Creates a stream from an iterator.
+          *
+          * @param v
+          *   Iterator to create a stream from
+          * @param bufferSize
+          *   Size of the buffer that the iterator will write to and the stream will read from
+          */
+        def fromIterator[V, S](v: => Iterator[V] < S, bufferSize: Int = defaultAsyncStreamBufferSize)(using
```
fwbrasil, line 71: We typically avoid allowing effects in value parameters like the iterator here. It reduces complexity, overhead, and makes reasoning about where effects happen easier. Could you take the `Iterator` directly?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -61,6 +61,33 @@ object StreamCoreExtensions:
...
+            inline def nextElement(it: Iterator[V]): Maybe[Chunk[V]] < IO =
+                val size    = bufferSize max 1
+                val builder = Chunk.newBuilder[V]
+                IO:
+                    var count = 0
+                    builder.clear()
+                    while count < size && it.hasNext do
+                        builder.addOne(it.next())
+                        count += 1
+                    if count > 0 then Maybe.Present(builder.result()) else Maybe.Absent
```
fwbrasil, line 84: `Maybe.when(count > 0)(builder.result())`

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -61,6 +61,33 @@ object StreamCoreExtensions:
...
+          * @param bufferSize
+          *   Size of the buffer that the iterator will write to and the stream will read from
+          */
+        def fromIterator[V, S](v: => Iterator[V] < S, bufferSize: Int = defaultAsyncStreamBufferSize)(using
+            tag: Tag[Emit[Chunk[V]]],
+            frame: Frame
+        ): Stream[V, (S & IO)] =
+            inline def nextElement(it: Iterator[V]): Maybe[Chunk[V]] < IO =
+                val size    = bufferSize max 1
+                val builder = Chunk.newBuilder[V]
```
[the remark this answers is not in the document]
ahoy-jon: Btw, out of the scope of this PR, could we have: `Chunk.newBuilder[V](sizeHint = size)`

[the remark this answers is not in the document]
ahoy-jon: Would it be better to allocate a new Array with the right size, to directly build a `Chunk.Compact`?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -61,6 +62,33 @@ object StreamCoreExtensions:
...
+                    var count = 0
+                    while count < size && it.hasNext do
+                        builder.addOne(it.next())
+                        count += 1
+
+                    Maybe.when(count > 0)(builder.result())
+            end nextElement
+
+            Stream.repeatPresent(nextElement(v))
+        end fromIterator
```
ahoy-jon, line 90: @hearnadam following https://github.com/getkyo/kyo/pull/1222#discussion_r2118238010
It is done like that.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -61,6 +62,33 @@ object StreamCoreExtensions:
...
+          * @param v
+          *   Iterator to create a stream from
+          * @param bufferSize
+          *   Size of the buffer that the iterator will write to and the stream will read from
+          */
+        def fromIterator[V, S](v: => Iterator[V], bufferSize: Int = Stream.DefaultChunkSize)(using
+            tag: Tag[Emit[Chunk[V]]],
+            frame: Frame
+        ): Stream[V, IO] =
+            val builder = ChunkBuilder.init[V]
```
hearnadam: Sadly this needs to be suspended, so you should write:
```
Stream {
  IO {
    val builder = ...

  }
}
```

ahoy-jon: Sure. I just don't understand the form: 
```
Stream
 
     ... repeat
```

This will go back to a 
```
Stream
    Loop
```

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -61,6 +62,40 @@ object StreamCoreExtensions:
...
+                        builder.clear()
+                        res
+
+                    Loop.foreach[Unit, IO & Emit[Chunk[V]]]:
+                        nextElements.map(chunk =>
+                            if chunk.isEmpty then Loop.done
+                            else Emit.valueWith(chunk)(Loop.continue)
+                        )
+
+        end fromIterator
```
ahoy-jon: @hearnadam this is interesting, the previous implementation (using `repeatPresent`) had no issues with choice, this once has (`bufferSize = 1`):
```
[info]   fromIterator
[info]     bufferSize = 1
[info]     - call by name (1 millisecond)
[info]     - map with Choice *** FAILED *** (0 milliseconds)
[info]       Chunk.Indexed(Chunk.Indexed("A", "B", "C"), Chunk.Indexed("A", "B", "c"), Chunk.Indexed("A", "b"), Chunk.Indexed("a")) had size 4 instead of expected size 8 (StreamCoreExtensionsTest.scala:274)
[info]     bufferSize = 4
[info]     - call by name (4 milliseconds)
[info]     - map with Choice (4 milliseconds)
```

(however `repeatPresent` would fail the callByName test)

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -61,6 +62,40 @@ object StreamCoreExtensions:
...
+                    val size    = bufferSize max 1
+                    val builder = ChunkBuilder.init[V]
+                    val nextElements: Chunk[V] < IO = IO:
+                        var count = 0
+                        while count < size && it.hasNext do
+                            builder.addOne(it.next())
+                            count += 1
+
+                        val res = builder.result()
+                        builder.clear()
```
ahoy-jon: created an issue, cannot use IO.ensure: https://github.com/getkyo/kyo/issues/1228

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -61,6 +62,36 @@ object StreamCoreExtensions:
...
+            frame: Frame
+        ): Stream[V, IO] =
+            Stream:
+                IO:
+                    val it      = v
+                    val size    = bufferSize max 1
+                    val builder = ChunkBuilder.init[V]
+
+                    Stream.repeatPresent(
+                        IO.ensure(builder.clear()):
```
fwbrasil: There's no need to call `clear`. The underlying buffer is released and cleared automatically by `result`.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -61,6 +62,36 @@ object StreamCoreExtensions:
...
                         yield ()
 
+        /** Creates a stream from an iterator.
+          *
+          * @param v
+          *   Iterator to create a stream from
+          * @param bufferSize
+          *   Size of the buffer that the iterator will write to and the stream will read from
+          */
+        def fromIterator[V, S](v: => Iterator[V], bufferSize: Int = Stream.DefaultChunkSize)(using
```
fwbrasil: shouldn't it be `chunkSize`?

fwbrasil (approved):
LGTM besides the comments

fwbrasil:
I hadn't noticed the pending `add fromIteratorCatching` task. Could it be a separate PR?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
hearnadam: +1 - the existing methods in this file use `bufferSize` because they allocate a channel.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -61,6 +62,36 @@ object StreamCoreExtensions:
...
+                IO:
+                    val it      = v
+                    val size    = bufferSize max 1
+                    val builder = ChunkBuilder.init[V]
+
+                    Stream.repeatPresent(
+                        IO.ensure(builder.clear()):
+                            var count = 0
+                            while count < size && it.hasNext do
+                                builder.addOne(it.next())
```
hearnadam, line 86: Should we suspend on each element?

fwbrasil: I don't think so since there aren't effect suspensions between `next` calls and there's already an `IO` suspension

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
ahoy-jon:
```suggestion
                        IO:
```

hearnadam (approved):

PR #1229, opened 2025-06-01
[core] implement `Async.raceFirst`

hearnadam:
### Problem
`Async.race` completes on the the first success or the last failure. In certain cases, this is undesirable. For example racing a potentially failing computation with a promise that may never complete. The resulting computation will hang indefinitely.

### Solution
Create a new specialized `Race[E, A]` class extending `IOPromise` in `Fiber`. This will be used to implement customization of Race state without additional overhead of a separate class. This PR adds support for `Success` (existing) and `First` (new).

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -213,6 +213,9 @@ object Async extends AsyncPlatformSpecific:
       * WARNING: Executes all computations concurrently without bounds. Use with caution on large sequences to avoid resource exhaustion. On
       * platforms supporting parallel execution (like JVM), computations may run in parallel.
       *
+      * Note: Unlike `raceFirst`, this method will only complete when a successful computation completes. If all computations fail, it will
+      * wait for the last failure. If some fail while others never complete, it will wait indefinitely for a success.
+      *
       * @param seq
       *   The sequence of computations to race
       * @return
@@ -230,6 +233,9 @@ object Async extends AsyncPlatformSpecific:
     end race
 
     /** Races two or more computations and returns the result of the first successful computation to complete.
+      *
+      * Note: Unlike `race`, this method will only complete when a successful computation completes. If all computations fail, it will wait
+      * for the last failure. If some fail while others never complete, it will wait indefinitely for a success.
       *
       * @param first
       *   The first computation
@@ -254,6 +260,10 @@ object Async extends AsyncPlatformSpecific:
       * WARNING: Executes all computations concurrently without bounds. Use with caution on large sequences to avoid resource exhaustion. On
       * platforms supporting parallel execution (like JVM), computations may run in parallel.
       *
+      * Note: Unlike `race`, this method will complete as soon as any computation completes, regardless of whether it succeeded or failed.
+      * For example, if one computation fails while another never completes, this method will return the failure, interrupting the other
+      * computation(s).
+      *
       * @param seq
       *   The sequence of computations to race
       * @return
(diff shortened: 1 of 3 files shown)


ahoy-jon (approved):

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -298,27 +298,64 @@ object Fiber:
...
+        Race.success[E, A, S](using isolate)(iterable)(using frame)
+
+    private[kyo] def raceFirst[E, A, S](
+        using isolate: Isolate.Contextual[S, IO]
+    )(iterable: Iterable[A < (Abort[E] & Async & S)])(using frame: Frame): Fiber[E, A] < (IO & S) =
+        Race.first[E, A, S](using isolate)(iterable)(using frame)
+
+    /** Race state for a race between multiple Fibers.
+      */
+    sealed abstract private class Race[E, A](frame: Frame) extends IOPromise[E, A] with (Result[E, A] => Unit):
```
ahoy-jon, line 310: out of curiosity, why not using `using`
```scala
    sealed abstract private class Race[E, A](using Frame) extends IOPromise[E, A] with (Result[E, A] => Unit):
```

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -248,6 +248,47 @@ object Async extends AsyncPlatformSpecific:
     ): A < (Abort[E] & Async & S) =
         race[E, A, S](first +: rest)
 
+    /** Races multiple computations and returns the result of the first to complete. When one computation completes, all other computations
```
fwbrasil, line 257: Can you add some explanation of the difference from the regular `race`? It'd be nice to add a note to `race` as well

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -248,6 +248,47 @@ object Async extends AsyncPlatformSpecific:
...
+      *
+      * WARNING: Executes all computations concurrently without bounds. Use with caution on large sequences to avoid resource exhaustion. On
+      * platforms supporting parallel execution (like JVM), computations may run in parallel.
+      *
+      * @param seq
+      *   The sequence of computations to race
+      * @return
+      *   The result of the first computation to complete
+      */
+    def raceFirst[E, A, S](
```
fwbrasil, line 272: What do you think of `race` => `raceSuccess` and `raceFirst` => `race`?

kyo-core/shared/src/main/scala/kyo/Fiber.scala
fwbrasil: the explicit param makes it more difficult to end up capturing the wrong frame

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -298,27 +298,64 @@ object Fiber:
...
+                        fiber.onComplete(state)
+                    }
+                    state
                 }
-                state
             }
-        }
+        end apply
+
+        inline def success[E, A, S](using
```
fwbrasil, line 333: nice abstraction! I wonder if we could expose something similar in the user-facing API and have a single `race` method.

fwbrasil (approved):

kyo-core/shared/src/main/scala/kyo/Fiber.scala
hearnadam: I was thinking this as well, like `using RaceStrategy` - however, you don't want to accidentally share a strategy if the class is the maintaining the state.

Perhaps we can map `Strategy => RaceState[E, A]`.

kyo-core/shared/src/main/scala/kyo/Async.scala
ahoy-jon: Note: https://zio.dev/overview/basic-concurrency/#racing
I am more used to `race` is the first success. Another way to solve it, use `raceSuccess`/`raceFirst`.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
fwbrasil: Maybe something like:

```scala
Async.race(..) // the most common usage pattern
Async.race(_.firstSuccess)(..)
Async.race(_.first)(..)
```

but I think two race methods in Async would be fine too at least for now since we have only two variants. We might be able to even get rid of `gather` if we're able to express more advanced patterns in a more generic API

Issue #1232, opened 2025-06-03
[BUG]: compile error on IO.ensure


ahoy-jon:
### Version

0.19.0+40-6e6cea21+20250603-1734-SNAPSHOT

### Scala Version

3.7.0

### Expected Behavior

on:
```scala
IO.ensure(builder.clear()):
  var count = 0
 while count < bufferSize && it.hasNext do
   builder.addOne(it.next())
   count += 1

  builder.result()
```
from: https://gist.github.com/ahoy-jon/b2955cca7f9995c44acda20b08c0d0d9#file-lazychunk-scala-L73-L79

should compile.

### Actual Behavior

```
[error] class Ensure cannot be accessed as a member of (Safepoint$_this :
[error]   ($proxy81 : kyo.kernel.internal.Safepoint.type{type State = Long})) from anonymous class Safepoint$_this.Ensure {...}.
[error]   private[kyo] class Ensure can only be accessed from package kyo.
```

PR #1235, opened 2025-06-03
[core] fix for #1232

ahoy-jon:
Fix for #1232, changing the visibility of `Ensure`

---

```
[info] Issue1232:
[info] - IO.ensure *** FAILED *** (321 milliseconds)
[info]   Failure("class Ensure cannot be accessed as a member of (Safepoint$_this :
[info]     ($proxy10 : kyo.kernel.internal.Safepoint.type{type State = Long})) from anonymous class Safepoint$_this.Ensure {...}.
[info]     private[kyo] class Ensure can only be accessed from package kyo.") did not equal <(), the Unit value> (Issue1232.scala:11)
```
---
~I have no idea how to solve that, this doesn't work:~
Solved thanks to @hearnadam 
```diff
       * @tparam A
       *   The result type of the main computation.
       */
-    inline def ensure[A, S](inline f: => Any < IO)(v: A < S)(using inline frame: Frame): A < (IO & S) =
+    inline def ensure[A, S](inline f: Safepoint ?=> Any < IO)(v: A < S)(using inline frame: Frame): A < (IO & S) =
         ensure(_ => f)(v)

+    @targetName("ensureWithError")
     inline def ensure[A, S](inline f: Maybe[Error[Any]] => Any < IO)(v: A < S)(using inline frame: Frame): A < (IO & S) =
```

### DIFF
--- a/kyo-core/shared/src/test/scala/Issue1232.scala
+++ b/kyo-core/shared/src/test/scala/Issue1232.scala
@@ -1,0 +1,13 @@
+package example
+
+import kyo.*
+
+class Issue1232 extends Test:
+    "IO.ensure" in run {
+        val result = typeCheck("""
+          def hello: Unit = ()
+          val io = IO.ensure(hello)(1)""")
+
+        assert(result == Result.succeed(()))
+    }
+end Issue1232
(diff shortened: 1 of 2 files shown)


hearnadam:
I think we just need to make the class private, with a package private constructor:
https://github.com/getkyo/kyo/blob/main/kyo-kernel/shared/src/main/scala/kyo/kernel/internal/Safepoint.scala#L136

hearnadam (approved):

PR #1238, opened 2025-06-05
Stream handle method

johnhungerford:
Addresses [discussion of resource handling on discord](https://discord.com/channels/1087005439859904574/1087005440623259681/1379902422419509338)

### Problem

Handling stream effects is currently a little awkward. You have to apply a handler to the stream's `emit` effect and then construct a new Stream from the transformed effect. To handle resources within the scope of a stream, for instance, you would do:
```
val original: Stream[Int, Resource] = ???
val handled: Stream[Int, Any] = Stream(Resource.run(original.emit))
```

### Solution

This PR adds a `handle` method to `Stream` similar to the `handle` method on `Pending`: it accepts one or more functions transforming the underlying emit effect, and constructs a new stream using the transformed effect. You can handle multiple effects as follows:

```
val original: Stream[Int, Resource & Var[Int] & Abort[String] = ???
val handled: Stream[Int, Any] = original.handle(
  Resource.run(_),
  Var.run(23)(_),
  Abort.run[String](_),
)
```

The functions are expected to evaluate to a `Emit[Chunked[V1]]` effect where `V1` need not be the same as the original `V` type. It does not have to evaluate to `Unit < ...`, as the implementation calls `.unit` to ensure it will be a valid streaming effect. This means you can pass `Abort.run[String](_)` without having to worry about the fact that the transformed effect will evaluate to `Result[Unit, String]`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1366,6 +1366,22 @@ val result: Chunk[String] < (Env[Config] & Async) =
     c.run
 ```
 
+To handle one or more effects used by a stream, you can pass effect transformations to the `handle` method. These will be applied to the underlying streaming effect.
+
+```scala
+import kyo.*
+
+case class Config(someConfig: String)
+
+val original: Stream[Int, Resource & Env[Config] & Abort[String] & Async] = Stream(1)
+
+val handled: Stream[Int, Async] = original.handle(
+    Resource.run(_),
+    Env.run(Config("some config"))(_),
+    Abort.run[String](_)
+)
+```
+
 The `Stream` effect is useful for processing large amounts of data in a memory-efficient manner, as it allows for lazy evaluation and only keeps a small portion of the data in memory at any given time. It's also composable, allowing you to build complex data processing pipelines by chaining stream operations.
 
 Note that a number of `Stream` methods (e.g., `map`, `filter`, `mapChunk`) are overloaded to provide different implementations for pure vs effectful transformations. This can make a big difference for performance, so take care that the functions you pass to these methods are typed to return pure values if they do not include effects. Unnecessarily lifting them to return `A < Any` will result in performance loss.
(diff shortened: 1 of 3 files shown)


ahoy-jon (commented):
it's great.
The definition of handler in https://github.com/getkyo/kyo/blob/d90f29a06c06c1606e3bc5ac6f6b791c199de62b/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala#L181-L184
use a call-by-name `f:(=> <[A, S]) => B` 

It's not possible to do the same?

ahoy-jon:
I don't have the logic behind this call-by-name, but there might be something to do with inlining and allocations.

a potential signature could be:
```scala
 def handle[V1 >: V, S1](
        f:  (=> Unit < S) => (Unit < (Emit[Chunk[V1]] & S1))
    )(using Frame, Tag[Emit[Chunk[V1]]], Tag[Emit[Chunk[V]]]): Stream[V1, S1]
 ```
 
 I have a case for a `Stream[Int, IO & Abort]` (for `Stream.fromIteratorCatching`), that could be done like that:
 ```scala
stream.handle(
  _.fold(
    onSuccess = () => Chunk.empty, 
    onFail = _ => Chunk(-1), 
    onPanic = _ => Chunk(-2)
  ), 
   _.map(chunk => Emit.value(chunk))
 )
```

fwbrasil:
> @ahoy-jon good point, although do you know why the parameter is call-by-name? I don't really see the point. I notice that in Pending.scala handle is an inline method. Should Stream#handle also be inline? @fwbrasil do you have some guidance here?

The by-name param was introduced to allow `Abort` to handle the case when the computation panics without ever suspending. For example, `((throw ex): Int < Any).handle(Abort.run)` produces `Result.Panic(ex)` instead of just throwing. Ideally, users should ensure code that may fail is suspended but we try to provide some built-in safety in APIs where possible. I think we should do the same for `stream.handle`.

Regarding `inline`, I don't think we need it. Streams are typically built once and reused multiple times so the potential improvement is limited.

> I have a case for a Stream[Int, IO & Abort] (for Stream.fromIteratorCatching), that could be done like that:

We recommend `handle` only for effect handling, not regular transformations. I think it's important to avoid confusion and reduce the ways of doing the same thing.

johnhungerford:
I updated `Stream#handle` to accept functions with call-by-name inputs

PR #1239, opened 2025-06-05
Stream broadcast

johnhungerford:
Addresses #679 

### Problem

There should be an easy way to broadcast streams.

### Solution

Implemented a variety of `broadcast...` methods. The underlying approach was to make a `StreamHub`, simple wrapper of a `Hub` that communicates elements, failures, and end-of-stream to listeners. `StreamHub` is exposed in the `broadcastDynamic` and `broadcastDynamicWith` methods. Other broadcast methods use it under the hood.

`broadcast2` through `broadcast5` construct tuples of broadcasted streams from the original stream. These guarantee that all streams will consume all elements of the original.

`broadcastN` allows you to specify a number of streams and will construct a chunk of that size (this is based on ZIO's approach)

`broadcasted` generates a stream that subscribes to the hub each time its run so can be reused in parallel. Does not guarantee all elements will be streamed for each run.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -1,6 +1,7 @@
 package kyo
 
 import kyo.Async.defaultConcurrency
+import kyo.Result.Success
 import kyo.kernel.ArrowEffect
 
 object StreamCoreExtensions:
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -30,6 +30,87 @@ object StreamCoreExtensions:
         Abort.run(emit).unit
     end emitMaybeElementsFromChannel
 
+    sealed trait StreamHub[A, E]:
```
fwbrasil, line 34: I think this new behavior seems more appropriate in `Hub` for better modularity. The current design with `Listener` was based on earlier versions of Kyo without more advanced streaming functionality. Maybe we could clean it up to be based on streams for initialization as well? It seems we could hide `Listener` via `Resource` and remove the write APIs like `offer` as well

johnhungerford: I'm hesitant to make `Hub` take on the properties of `StreamHub` because the current version of `Hub` is still useful in other cases. The ability to broadcast streams is a special case that I've actually only used a couple of times. More commonly, I've used a `Hub` as a central input and then subscribed in different parts of my application as needed.

hearnadam: Do you think adding ZIO's Enqueue/Dequeue traits would help solve this issue?

More generally we should be able to convert a Listener or a Dequeue to a Stream.

johnhungerford: I do think those traits are useful, but for this use case I would still like to have a simple wrapper that offers only the `subscribe: Stream[...` method. For this I would need to hide more than just the `Enqueue`.

fwbrasil: makes sense 👍

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -30,6 +30,87 @@ object StreamCoreExtensions:
...
+        def consume[S](stream: Stream[A, Abort[E] & S & Async])(
+            using
+            i1: Isolate.Contextual[S, IO],
+            i2: Isolate.Stateful[S, Abort[E] & Async],
+            t1: SafeClassTag[E],
+            t2: Tag[Emit[Chunk[A]]],
+            fr: Frame
+        ): Unit < (Async & S & Resource) =
+            Resource.acquireRelease(Async.run {
+                Abort.run[E](
```
fwbrasil, line 89: It seems this could be `Abort.run[E | Closed]`?

fwbrasil (approved):

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: I only want to propagate the `E` failures to the output stream, so it's convenient to just handle them separately

PR #1241, opened 2025-06-06
Reimplement parallel map functions without state

### LINKED ISSUE #1192: [BUG]: Missed signals in Stream.mapChunkPar
### Version

0.19.0

### Scala Version

3.7.0

### Expected Behavior

_No response_

### Actual Behavior

`SignalRef` is a primitive that doesn't guarantee delivery of changes. As documented in the scaladoc, It's designed for UI-like use cases where missed signals are ok. `Stream.mapChunkPar`'s tests are failing I believe due to this characteristic. We need to reimplement the method without `SignalRef`.

### Steps to Reproduce

Revert https://github.com/getkyo/kyo/commit/9ed7aa269c102c72e58eb57c126b1f8bc27d30b2 and run `StreamCoreExtensionsTest`. It's flaky so it might need a few runs.

### Current Workaround

_No response_

johnhungerford:
Fixes #1192 

### Problem

`mapChunkPar` tests were flaky due to the use of `SignalRef`

### Solution

Use a channel of size `parallel` instead to limit concurrency.

Also reimplemented `mapPar`, `mapParUnordered`, and `mapChunkParUnordered` similarly without state (they were using `AtomicRef`). I think they are a bit easier to understand this way, and are more consistent with each other now.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -222,58 +222,58 @@ object StreamCoreExtensions:
                         def throttled[A](task: A < (Async & Abort[Closed | E] & S2)) =
                             parChannel.put(()).andThen(task.map(a => parChannel.take.andThen(a)))
 
-                        val background = Async.run:
-                            val handledStream = ArrowEffect.handleLoop(t1, initialState, stream.emit)(
-                                handle = [C] =>
-                                    (input, state, cont) =>
-                                        val (prevEmitFiber, initialRemainingPar) = state
-                                        Loop(Fiber.success[E | Closed, Chunk[V2]](Chunk.empty[V2]), input, initialRemainingPar):
-                                            (prevChunkFiber, remainingChunk, remainingPar) =>
-                                                val nextParSection = remainingChunk.take(remainingPar)
+                        val handledStream = ArrowEffect.handleLoop(t1, initialState, stream.emit)(
+                            handle = [C] =>
+                                (input, state, cont) =>
+                                    val (prevEmitFiber, initialRemainingPar) = state
+                                    Loop(Fiber.success[E | Closed, Chunk[V2]](Chunk.empty[V2]), input, initialRemainingPar):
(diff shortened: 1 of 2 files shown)


johnhungerford:
@ahoy-jon I added some comments [as promised on discord](https://discord.com/channels/1087005439859904574/1245754417840394313/1381275739869151242).  I also made a further simplification eliminating the need to use a stateful stream handler.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -304,80 +304,59 @@ object StreamCoreExtensions:
...
-                                    onPanic = e => Abort.run(channel.put(Absent)).andThen(Abort.panic(e))
-                                )(handledStream)
-
-                            background.map: backgroundFiber =>
-                                emitMaybeChunksFromChannel(channel).andThen:
-                                    backgroundFiber.get.unit
-                }
+                Channel.initWith[Maybe[Chunk[V2]]](bufferSize): channelOut =>
+                    // Staging channel acts as concurrency limiter: contains fibers that either contain a
+                    // chunk to be published, or a signal to continue or end
```
fwbrasil, line 312: thanks for the comments! it makes the review much easier

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -304,80 +304,59 @@ object StreamCoreExtensions:
...
+                    Channel.initWith[Fiber[E | Closed, Boolean | Chunk[V2]]](parallel - 1): stagingChannel =>
```
fwbrasil, line 313: no need to change it but wouldn't a `Meter` work well for this?

fwbrasil (approved):

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: The problem with `Meter` is ensuring that the end-of-stream signal is not published until all pending transformations have published.

PR #1244, opened 2025-06-07
[core] add (Stream < S).unwrap

ahoy-jon:
add `(Stream[V, S1] < S2).unwrap`

--- 

- [x] test
- [x] doc?


kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -565,6 +565,14 @@ object StreamCoreExtensions:
...
 
     end extension
+
+    extension [V, S, S2](stream: Stream[V, S] < S2)
+        /** Takes a Stream[V, S] in the context of S2 (i.e. Stream[V, S] < S2) and returns a Stream that fuses together both effect contexts
+          * S and S2 into a single Stream[V, S & S2].
+          */
+        inline def unwrap: Stream[V, S & S2] = Stream(stream.map(_.emit))
+    end extension
+
```
johnhungerford: This should probably be in kyo-prelude, since it is not restricted `IO` or `Async`. Also, can I recommend instead of making this an extension method, make it a method on the `Stream` companion object?:

```scala
object Stream:
   ...
   def unwrap[V, S, S2](stream: Stream[V, S] < S2): Stream[V, S & S2] = ???
```

This is more consistent with Kyo's core API. We can then add the extension method version in kyo-combinators.

README.md
```diff
@@ -1382,6 +1382,22 @@ val handled: Stream[Int, Async] = original.handle(
...
+To fuse one or more effects used to create and run a stream, you can use `unwrap` method.
+
+```scala
+import kyo.*
+
+case class Config(size:Int)
+
+val original: Stream[Int, Any] < Env[Config] = Env.get[Config].map(config => Stream.from(-9).take(config.size))
+
+val unwrapped: Stream[Int, Env[Config]] = original.unwrap
```
johnhungerford: You should probably either use the `Stream.unwrap` constructor or make some not that this method is in kyo-combinators.

johnhungerford (approved):
LGTM other than clarifying README

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -916,4 +916,13 @@ object Stream:
...
 
+    /** Takes a Stream[V, S] in the context of S2 (i.e. Stream[V, S] < S2) and returns a Stream that fuses together both effect contexts S
+      * and S2 into a single Stream[V, S & S2].
+      *
+      * @param stream
+      *   The stream to unwrap
+      * @return
+      *   A new stream that fuses together both effect contexts S and S2 into a single Stream[V, S & S2]
+      */
+    inline def unwrap[V, S, S2](stream: Stream[V, S] < S2): Stream[V, S & S2] = Stream(stream.map(_.emit))
```
fwbrasil, line 927: how about an extension method? I don't see the need to put it in the combinators module

README.md
ahoy-jon: back to `kyo-prelude`, following: https://github.com/getkyo/kyo/pull/1244#discussion_r2135631160

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
johnhungerford: It occurs to me any extension method should probably be `unwrapStream` in case there are other types we want to unwrap. (E.g., Pipe/Sink)

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -916,4 +916,22 @@ object Stream:
...
+
 end Stream
+
+extension [V, S, S2](stream: Stream[V, S] < S2)
+    /** Takes a Stream[V, S] in the context of S2 (i.e. Stream[V, S] < S2) and returns a Stream that fuses together both effect contexts S
+      * and S2 into a single Stream[V, S & S2].
+      */
+    @targetName("streamUnwrap")
```
ahoy-jon: @johnhungerford is the targetName enough?

README.md
```diff
@@ -1382,6 +1382,22 @@ val handled: Stream[Int, Async] = original.handle(
...
 ```
 
+To fuse one or more effects used to create and run a stream, you can use `unwrap` method.
+
+```scala
+import kyo.*
+
+case class Config(size:Int)
+
+val original: Stream[Int, Any] < Env[Config] = Env.get[Config].map(config => Stream.from(-9).take(config.size))
```
hearnadam:
```suggestion
val original: Stream[Int, Any] < Env[Config] = Env.get[Config].map(config => Stream.init(-9).take(config.size))
```

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
hearnadam: Perhaps we should just skip the object method? Will users ever prefer it?

ahoy-jon: > It occurs to me any extension method should probably be unwrapStream in case there are other types we want to unwrap. (E.g., Pipe/Sink)

we have:

- `Stream.unwrap`
- `(Stream[V, S] < S2).unwrap` (with the targetName `streamUnwrap`)

Not sure to follow, if we need to use `(Stream[V, S] < S2).unwrapStream`

ahoy-jon: > Perhaps we should just skip the object method? Will users ever prefer it?

I agree, however sometimes, it's just there for documentation when users are looking around.

johnhungerford: My two cents is that the kyo core API strongly prefers object methods to do transformations on effects. All of the handler methods, for instance, are on the object and not extension methods on `<[A, S]`. This is why I suggested putting the extension method in kyo-combinators, which is where we have historically been putting them. Adding `unwrap` as an extension method in kyo-prelude seems to break the pattern.

In either case I think it is worth keeping `unwrap` as a method on the `Stream` object simply because it isn't uncommon to run into issues with extension methods and type inference.

As for whether to name the extension method `unwrap` or `unwrapStream`, the issue is that compiler error about ambiguous extension method resolution often result when the same name is used for extension methods on different types. I don't think `targetName` addresses this problem. I have generally found it safer to make sure extension method names are unique, even if when the types *should* make it clear to the compiler.

fwbrasil: My bad, sorry. I’ve been trying to keep up with the PRs in limited time and didn’t realize that it’d be an extension on the pending type. We don’t have any outside of the combinators module. I agree we should not have an exception here

ahoy-jon: That's ok! done!

johnhungerford (approved):

PR #1245, opened 2025-06-07
[build] Scala Native 0.5.8 + enable tests

fwbrasil:
The new version should fix the CI issues 🤞


fwbrasil:
There's a different failure now: https://github.com/getkyo/kyo/actions/runs/15508180158/job/43665607127?pr=1245#step:8:12121

PR #1250, opened 2025-06-08
Stream variance

johnhungerford:
### Problem

The lack of variance in `Stream` is a problem for DX.
- It is difficult to merge streams of different types
- When defining streams using `Emit`, you have to explicitly set the type to the desired supertype

### Solution

Made `Emit[V]` contravariant with `V`, made `Poll[V]` covariant with `V`, made `Stream[V, S]` covariant with `V`.

Added partial handling to `Emit`. You can construct a `Emit[Int | String]` effect and then handle the `Int` or `String` emissions separately. (Haven't added this to `Poll` yet.)

### Notes

As a result of the covariance in `Stream`, I have had to lift a lot of the `V` parameters to `VV >: V`. This has apparently made it too difficult for the compiler to distinguish between our pure/effectful method overloads. Accordingly, I have added explicit `---Kyo` variants of each method. This solves the compiler issue, but I think it is also a better approach in the long run. Since we now support nested effects, there's no reason a user shouldn't be able to emit an effect (which was impossible, or very difficult, before), and now the user can be more assured that pure values aren't being lifted without their knowing.

### DIFF
--- a/kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
@@ -780,7 +780,7 @@ class ChannelTest extends Test:
             for
                 c <- Channel.init[Int](4)
                 _ <- Kyo.foreach(1 to 4)(c.put)
-                stream = c.stream(1).mapChunk(Chunk(_)).take(4)
+                stream = c.stream(1).mapChunk[Int, Chunk[Int]](Chunk(_)).take(4)
                 r <- stream.run
                 s <- c.size
             yield assert(r == Chunk(Chunk(1), Chunk(2), Chunk(3), Chunk(4)) && s == 0)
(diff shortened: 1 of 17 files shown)


fwbrasil:
If it doesn't complicate development too much, separating the changes on the different effects would make reviews easier. Don't worry if it isn't easy to make the changes in isolation

fwbrasil:
sorry, clicked the wrong button

kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
```diff
@@ -790,7 +790,7 @@ class ChannelTest extends Test:
             for
                 c <- Channel.init[Int](4)
                 _ <- Kyo.foreach(1 to 4)(c.put)
-                stream = c.stream().mapChunk(ch => Chunk(ch)).take(1)
+                stream = c.stream().mapChunk[Int, Chunk[Int]](ch => Chunk(ch)).take(1)
```
ahoy-jon, line 793: probably not needed anymore.

kyo-prelude/shared/src/test/scala/kyo/ChoiceTest.scala
```diff
@@ -232,7 +232,7 @@ class ChoiceTest extends Test:
             val stream      = Choice.runStream(computation)
 
             val result = stream
-                .filter(_ % 2 == 1)
+                .filter[Int](_ % 2 == 1)
```
ahoy-jon, line 235: same, it was before `filterKyo`?

johnhungerford: Not needed. I had put the explicit type parameters there prior to splitting the method names just to see if it would compile

PR #1251, opened 2025-06-08
[core] Stream.fromIteratorCatching

ahoy-jon:
follow-up of https://github.com/getkyo/kyo/pull/1222

* add `Stream.fromIteratorCatching[E]`
* rewrite `Stream.fromIterator` to use `Stream.fromIteratorCatching`

Note: previous implementation on `fromIterator` was passing the test on `Choice`, because `Stream.repeatPresent` is rechunking. `Choice` over map of impure/effectful construction is still on open subject (not for long).

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -1,9 +1,9 @@
 package kyo
 
-import kyo.Async.defaultConcurrency
 import kyo.ChunkBuilder
 import kyo.kernel.ArrowEffect
-import scala.compiletime.*
+import scala.annotation.implicitNotFound
+import scala.util.NotGiven
 
 object StreamCoreExtensions:
     val defaultAsyncStreamBufferSize = 1024
@@ -30,10 +30,6 @@ object StreamCoreExtensions:
                         else Loop.done
         Abort.run(emit).unit
     end emitMaybeElementsFromChannel
-
-    type DefaultIfNothing[E, Default] = E match
-        case Nothing => Default
-        case _       => E
 
     extension (streamObj: Stream.type)
         /** Merges multiple streams asynchronously. Stream stops when all sources streams have completed.
@@ -77,7 +73,29 @@ object StreamCoreExtensions:
         def fromIterator[V](v: => Iterator[V], chunkSize: Int = Stream.DefaultChunkSize)(using
             Tag[Emit[Chunk[V]]],
             Frame
-        ): Stream[V, IO] = _fromIteratorCatching[Nothing](v, chunkSize)
+        ): Stream[V, IO] =
+            val stream: Stream[V, IO] < IO = IO:
+                val it      = v
+                val size    = chunkSize max 1
+                val builder = ChunkBuilder.init[V]
+
+                val pull: Chunk[V] < IO =
+                    IO:
+                        var count = 0
+                        while count < size && it.hasNext do
+                            builder.addOne(it.next())
+                            count += 1
+
+                        builder.result()
+
+                Stream:
+                    Loop.indexed: _ =>
+                        pull.map:
+                            case chunk if chunk.isEmpty => Loop.done
+                            case chunk                  => Emit.valueWith(chunk)(Loop.continue)
+
+            Stream(stream.map(_.emit)) // unwrap
+        end fromIterator
 
         /** Creates a stream from an iterator.
           *
@@ -92,25 +110,13 @@ object StreamCoreExtensions:
             frame: Frame
         )[V](v: => Iterator[V], chunkSize: Int = Stream.DefaultChunkSize)(using
             tag: Tag[Emit[Chunk[V]]],
-            ct: SafeClassTag[E]
+            ct: SafeClassTag[E],
+            @implicitNotFound(
+                "Missing type param on `fromIteratorCatching[E = SomeException](iterator, chunkSize)`." +
+                    "\n - Cannot catch Exceptions using `E = Nothing`." +
+                    "\n - `fromIteratorCatching[E = Throwable]` catches all Exceptions."
+            ) notNothing: NotGiven[E =:= Nothing]
         ): Stream[V, IO & Abort[E]] =
-            inline erasedValue[E] match
-                case _: Nothing =>
-                    error(
-                        "Missing type param on `fromIteratorCatching[E = SomeException](iterator, chunkSize)`." +
-                            "\n - Cannot catch Exceptions using `E = Nothing`." +
-                            "\n - `fromIteratorCatching[E = Throwable]` catches all Exceptions."
-                    )
-                case _ => _fromIteratorCatching[E](v, chunkSize)
-        end fromIteratorCatching
-
-        private def _fromIteratorCatching[E](using
-            Frame
-        )[V](v: => Iterator[V], chunkSize: Int)(using
-            tag: Tag[Emit[Chunk[V]]],
-            ct: SafeClassTag[E]
-        ): Stream[V, IO & Abort[E]] =
-
             val stream: Stream[V, (IO & Abort[E])] < IO = IO:
                 val it      = v
                 val size    = chunkSize max 1
@@ -137,7 +143,7 @@ object StreamCoreExtensions:
                                     Emit.valueWith(lastElements)(Abort.error(error))
 
             Stream(stream.map(_.emit)) // unwrap
-        end _fromIteratorCatching
+        end fromIteratorCatching
 
         /** Merges multiple streams asynchronously. Stream stops as soon as any of the source streams complete.
           *
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -89,7 +89,7 @@ object StreamCoreExtensions:
                         builder.result()
 
                 Stream:
-                    Loop.indexed: _ =>
+                    Loop.foreach:
                         pull.map:
                             case chunk if chunk.isEmpty => Loop.done
                             case chunk                  => Emit.valueWith(chunk)(Loop.continue)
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -30,6 +33,9 @@ object StreamCoreExtensions:
         Abort.run(emit).unit
     end emitMaybeElementsFromChannel
 
+    @implicitNotFound("You must specify a concrete type when using `fromIteratorCatching[SomeException]`")
+    type CheckIteratorCatching[E] = NotGiven[E =:= Nothing]
```
hearnadam: Perhaps we should have a `ErrorIsNot` typeclass to be used outside of this class? Up to you

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -69,28 +75,60 @@ object StreamCoreExtensions:
...
-                    val it      = v
-                    val size    = chunkSize max 1
-                    val builder = ChunkBuilder.init[V]
 
-                    Stream.repeatPresent(
-                        IO:
+            def toPanic[S, E <: Throwable, A](x: => A < (S & Abort[E])) =
+                Abort.recoverOrThrow(Abort.panic)(x)
+
+            fromIteratorCatching[Throwable](v, chunkSize).handle(toPanic)
```
hearnadam: Isn't this the correct approach?
```suggestion
            fromIteratorCatching[Nothing](v, chunkSize)
```

kyo-core/shared/src/test/scala/kyo/StreamCoreExtensionsTest.scala
```diff
@@ -216,29 +216,29 @@ class StreamCoreExtensionsTest extends Test:
             }
         }
 
-        def fromIteratorTests(bufferSize: Int): Unit =
-            s"bufferSize = $bufferSize" - {
+        def fromIteratorTests(chunkSize: Int): Unit =
+            s"bufferSize = $chunkSize" - {
```
hearnadam:
```suggestion
            s"chunkSize = $chunkSize" - {
```

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -69,28 +75,60 @@ object StreamCoreExtensions:
...
 
-                            val chunk = builder.result()
-                            Maybe.when(chunk.nonEmpty)(chunk)
-                    ).emit
+                            builder.result()
 
-        end fromIterator
+                Stream:
+                    Loop.indexed: _ =>
+                        Abort.run(pull).map:
```
hearnadam, line 142: I think it's also fine to drop the elements outside the chunk - that's what FS2 does I believe. But this approach is slightly nicer as a user.

ahoy-jon: and it's less dependent on `chunkSize`

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
ahoy-jon: I'm trying something "better"

ahoy-jon:
@hearnadam thank you for the review, found a better way to handle `E`.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
ahoy-jon: Indeed, but with a private method.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -69,28 +74,70 @@ object StreamCoreExtensions:
...
+          * @param chunkSize
+          *   Size of the chunks that the iterator will produce and the stream will read from
+          */
+        inline def fromIteratorCatching[E <: Throwable](using
+            frame: Frame
+        )[V](v: => Iterator[V], chunkSize: Int = Stream.DefaultChunkSize)(using
+            tag: Tag[Emit[Chunk[V]]],
+            ct: SafeClassTag[E]
+        ): Stream[V, IO & Abort[E]] =
+            inline erasedValue[E] match
```
fwbrasil: can we use `NotGiven` + `@implicitNotFound` instead? I think it'd be more clear

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -69,28 +74,70 @@ object StreamCoreExtensions:
...
+        def fromIterator[V](v: => Iterator[V], chunkSize: Int = Stream.DefaultChunkSize)(using
             Tag[Emit[Chunk[V]]],
             Frame
-        ): Stream[V, IO] =
-            Stream:
-                IO:
-                    val it      = v
-                    val size    = chunkSize max 1
-                    val builder = ChunkBuilder.init[V]
+        ): Stream[V, IO] = _fromIteratorCatching[Nothing](v, chunkSize)
```
fwbrasil: can we have a specialized version without error handling? It'd reduce overhead

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -69,28 +74,70 @@ object StreamCoreExtensions:
...
+                            builder.result()
+
+                Stream:
+                    Loop.indexed: _ =>
+                        Abort.run(pull).map:
+                            case Result.Success(chunk) if chunk.isEmpty => Loop.done
+                            case Result.Success(chunk)                  => Emit.valueWith(chunk)(Loop.continue)
+                            case error: Result.Error[E] @unchecked =>
+                                IO:
+                                    val lastElements: Chunk[V] = builder.result()
```
fwbrasil, line 147: Nice behavior!

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
ahoy-jon: I didn't know before `@implicitNotFound` could be used directly on the method param.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -69,28 +70,80 @@ object StreamCoreExtensions:
...
+          *   Size of the chunks that the iterator will produce and the stream will read from
+          */
+        inline def fromIteratorCatching[E <: Throwable](using
+            frame: Frame
+        )[V](v: => Iterator[V], chunkSize: Int = Stream.DefaultChunkSize)(using
+            tag: Tag[Emit[Chunk[V]]],
+            ct: SafeClassTag[E],
+            @implicitNotFound(
+                "Missing type param on `fromIteratorCatching[E = SomeException](iterator, chunkSize)`." +
+                    "\n - Cannot catch Exceptions using `E = Nothing`." +
```
hearnadam: Perhaps pedantic, but this is not correct. `E =:= Nothing` will just let exceptions become Panics. This should be the same behavior of `fromIterator`, but this method just ensures that the start of the chunk is preserved.

It would be nice if we could have anti-inference typeclass (to avoid E always inferring as Nothing), but which would be ignored if explicitly set to Nothing...

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -69,28 +70,80 @@ object StreamCoreExtensions:
...
+                    IO:
+                        var count = 0
+                        while count < size && it.hasNext do
+                            builder.addOne(it.next())
+                            count += 1
+
+                        builder.result()
+
+                Stream:
+                    Loop.indexed: _ =>
```
hearnadam:
```suggestion
                    Loop.foreach:
```

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -69,28 +70,80 @@ object StreamCoreExtensions:
...
                                 count += 1
 
-                            val chunk = builder.result()
-                            Maybe.when(chunk.nonEmpty)(chunk)
-                    ).emit
+                            builder.result()
 
-        end fromIterator
+                Stream:
+                    Loop.indexed: _ =>
```
hearnadam:
```suggestion
                    Loop.foreach:
```

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
ahoy-jon: I don't mind being pedantic! The current message is:

```
Missing type param on `fromIteratorCatching[E = SomeException](iterator, chunkSize)`.
 - Cannot catch Exceptions using `E = Nothing`
 - `fromIteratorCatching[E = Throwable]` catches all Exceptions.
 ```
 
 Are you okay with:
 ```
 Missing *explicit* type param on `fromIteratorCatching[E = SomeException](iterator, chunkSize)`.
 - Cannot catch Exceptions as `Failure[E]` using `E = Nothing`, they are turned into `Panic`
          If you need this behaviour, use `fromIterator(iterator, chunkSize)` directly.
 - `fromIteratorCatching[E = Throwable]` catches all Exceptions.
``` 

> but this method just ensures that the start of the chunk is preserved.

Yes, even for `E =:= Nothing` , should I go back to a single implementation? (and test it)

---

I don't know how to make an anti-inference typeclass, that would be nice.

ahoy-jon: Tried something!

kyo-core/shared/src/test/scala/kyo/StreamCoreExtensionsTest.scala
```diff
@@ -268,12 +268,157 @@ class StreamCoreExtensionsTest extends Test:
...
+
+                    val panicTo42 = stream.handle(Abort.recoverError({
+                        case _: Result.Panic      => Emit.value(Chunk(42))
+                        case _: Result.Failure[?] => fail("should not happen")
+                    }))
+
+                    panicTo42.run.map: chunk =>
+                        assert(chunk == Chunk(0, 1, 2, 42))
+
+                }
```
ahoy-jon, line 288: @hearnadam so it's tested, we can use `fromIterator` instead of `fromIteratorCatching[Nothing]`.

PR #1252, opened 2025-06-09
update Contributing to add some design guidelines

ahoy-jon:
Could be called "things I wish I would have known before contributing to `Kyo`".

I will go back to my last PRs, and try to sum it up here.

@johnhungerford @fwbrasil @hearnadam don't hesitate to use directly PR suggestions on this one.

---
- [ ] add part about `Frame`, `Safepoint`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,3 +126,49 @@ We encourage contributors to leverage Large Language Models (LLMs) responsibly:
   ```sh
   sbt "scalafmtCheckAll"
   ```
+
+## Design Guidelines
+
+### Adding a New Method
+
+If you want to contribute a new method like `S.newMethod` or `s.newMethod`, feel free to:
+
+* Open an issue
+    
+* Discuss on Discord: [https://discord.gg/afch62vKaW](https://discord.gg/afch62vKaW)
+    
+* Share design examples:
+    
+    * Use cases
+        
+    * Equivalent in `ZIO` or `Cats Effect`
+        
+    * Other motivating patterns
+        
+
+### Where to add Your `newMethod`
+
+| Subproject | Use For |
+| --- | --- |
+| `kyo-data` | Data structures (`Chunk`, `Maybe`, `Result`, etc.) |
+| `kyo-prelude` | Effect types without `IO` (`Abort`, `Env`, `Var`, etc.) |
+| `kyo-core` | Methods requiring `IO`, `Async`, or `Resource` |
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -126,3 +126,49 @@ We encourage contributors to leverage Large Language Models (LLMs) responsibly:
...
+A new `Stream.fromSomething` method:
+
+* If it uses `IO`: place it in `kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala`
+    
+* If it doesn't: place it in `kyo-prelude/shared/src/main/scala/kyo/Stream.scala`
+    
+
+### Tips
+
+* Prefer [`call-by-name`](https://docs.scala-lang.org/tour/by-name-parameters.html) (`body: => A < S`) for deferred evaluation, especially when lifting to `IO`.
```
fwbrasil: This works well for `IO` but not other effect suspensions since we can't leverage any suspension to defer side effects. `IO` is special in this regard since it's the last effect to be handled. Let's make the tip more specific.

CONTRIBUTING.md
```diff
@@ -126,3 +126,49 @@ We encourage contributors to leverage Large Language Models (LLMs) responsibly:
...
+    
+* Use [`inline`](https://docs.scala-lang.org/scala3/reference/metaprogramming/inline.html) when beneficial.
```
fwbrasil: We were more liberal with the use of `inline` in the past but we need to keep it limited to reduce compilation times. Can you qualify that it should be used in APIs that are performance-sensitive?

CONTRIBUTING.md
```diff
@@ -126,3 +126,49 @@ We encourage contributors to leverage Large Language Models (LLMs) responsibly:
...
+    
+* Keep `S` open: use `A < (S & SomeEffect)` instead of `A < SomeEffect` to support composition and emulate effect row polymorphism.
```
fwbrasil: I don't think `emulate effect row polymorphism` is accurate. Kyo's tracking of effects isn't an emulation of row polymorphism, it's a distinct and proper tracking mechanism based on Scala's type system. It provides more advanced behaviors like parametrized effects with variance, which mixed with other type system features like union types, extension methods, typeclasses, I'd argue provides more flexibility and better usability than row polymorphism.

ahoy-jon: "Feigning ignorance to uncover the truth", thanks a lot, I will update!

Issue #894, opened 2024-12-04
Emit: convert to covariance


hearnadam:
To enable covariant Stream: `Stream[+V, -S]`, we probably want to introduce Emit chunking+covariance... The alternative involves type bounds and some coercion 🙂 
```
[error] 361 |    def run(using tag: Tag[Emit[Chunk[V]]], frame: Frame): Chunk[V] < S =
[error]     |    ^
[error]     |covariant type V occurs in invariant position in type (using tag: kyo.Tag[kyo.Emit[kyo.Chunk[V]]], frame: kyo.Frame): kyo.Chunk[V] < S of method run
[error] 362 |        ArrowEffect.handleState(tag, Chunk.empty[Chunk[V]], emit)(
[error] 363 |            handle = [C] =>
[error] 364 |                (input, state, cont) =>
[error] 365 |                    (state.append(input), cont(Continue())),
[error] 366 |            done = (state, _) => state.flattenChunk
```

PR #1254, opened 2025-06-10
Covariant Emit

johnhungerford:
Addresss #894 

See also #679 

### Problem

`Stream[V, S]` can't be covariant with `V` until `Emit[V]` is covariant with V. This is a pretty big problem for usability.

### Solution

Made `Emit[V]` covariant with `V` and updated `run` functions to support partial handling.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Emit.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Emit.scala
@@ -1,248 +1,238 @@
 package kyo
 
-import kyo.Tag
-import kyo.kernel.*
+import kyo.kernel.ArrowEffect
 
-/** The Emit effect allows producing multiple values alongside the main result of a computation.
+/** Represents polling values from a data source with backpressure control.
   *
-  * Emit implements a push-based model where values of type V are actively emitted from a producer without waiting for consumer readiness.
-  * This makes it useful for event emission, logging, or any scenario where values need to be produced during computation regardless of
-  * downstream consumption patterns.
+  * Poll implements a pull-based streaming model where consumers actively request data when they're ready to process it. This approach
+  * provides natural backpressure, as values are only produced in response to explicit requests, preventing overwhelmed consumers.
   *
-  * As a low-level primitive in Kyo's streaming architecture, Emit provides fundamental capabilities but lacks many conveniences. For most
-  * streaming use cases, prefer the higher-level Stream abstraction which offers richer transformation capabilities, automatic chunking, and
-  * better composition. Direct use of Emit should generally be reserved for specialized scenarios.
+  * Key behaviors:
+  *   - Poll returns Maybe[V], where:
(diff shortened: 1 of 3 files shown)


kyo-prelude/shared/src/main/scala/kyo/Emit.scala
```diff
@@ -171,17 +191,23 @@ object Emit:
...
-                // Effect not found, return empty input and a placeholder continuation
-                // that returns the result of the computation
-                (Maybe.empty[V], () => r: A < (Emit[V] & S))
-        )
+    def runFirst[V](using
+        Frame
+    )[A, VR, S](v: A < (Emit[V] & Emit[VR] & S))(using
+        tag: Tag[Emit[V]],
+        reduce: Reducible[Emit[VR]]
+    ): (Maybe[V], () => A < (Emit[V | VR] & S)) < (reduce.SReduced & S) =
```
hearnadam, line 199: I wonder if we can use `A < (...) < Any` now...

kyo-prelude/shared/src/main/scala/kyo/Emit.scala
```diff
@@ -86,11 +87,15 @@ object Emit:
...
       */
-    def run[V](using Frame)[A, S](v: A < (Emit[V] & S))(using tag: Tag[Emit[V]]): (Chunk[V], A) < S =
-        ArrowEffect.handleLoop(tag, Chunk.empty[V], v)(
-            handle = [C] => (input, state, cont) => Loop.continue(state.append(input), cont(())),
-            done = (state, res) => (state, res)
-        )
+    def run[V](using
+        tag: Tag[Emit[V]],
+        fr: Frame
+    )[A, VR, S](v: A < (Emit[V] & Emit[VR] & S))(using reduce: Reducible[Emit[VR]]): (Chunk[V], A) < (reduce.SReduced & S) =
```
hearnadam, line 93: It's a bit sad this introduces more `Reducible` signatures - they are a bit hard to read.

hearnadam (approved):

kyo-prelude/shared/src/main/scala/kyo/Emit.scala
johnhungerford: Interesting -- make the continuation a Kyo suspension rather than a lambda? I suspect that signature would be even more confusing than this one.

kyo-prelude/shared/src/main/scala/kyo/Emit.scala
johnhungerford: Unfortunately this is the only way to support partial handling

kyo-prelude/shared/src/main/scala/kyo/Emit.scala
hearnadam: True

PR #1255, opened 2025-06-11
[core] clean unwrap

ahoy-jon:


kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -98,7 +98,7 @@ object StreamCoreExtensions:
                                     val lastElements: Chunk[V] = builder.result()
                                     Emit.valueWith(lastElements)(Abort.panic(throwable))
 
-            Stream(stream.map(_.emit)) // unwrap
```
hearnadam, line 101: It looks like `unwrap` needs a frame implicit.

Issue #1221, opened 2025-05-30
[feature] [direct]:add support for Stream in direct


ahoy-jon:
### Problem Statement

Currently, this does not compile:
```scala
    "Stream" in {
        val stream = Stream.init(Seq(1, 2, 3))
        def f(i: Int): Int < Any =
            i + 1

        val d1 = defer:
            stream.map(f)

        val d2 = defer:
            stream.map(x => f(x).now)

    }
```

### Proposed Solution

- update macros to allow operations on Stream
- add an `AsyncShift` for `Stream[?, ?]`

### Current Workaround

for map, it's definied twice:
```scala
def map[V2, S2](f: V => V2 < S2)
def map[V2](f: V => V2)
```
so this works
```scala
    "Stream" in {
        val stream = Stream.init(Seq(1, 2, 3))
        def f(i: Int): Int < Any =
            i + 1

        val d = defer:
            stream.map(x => f(x).later)
    }
```

PR #1256, opened 2025-06-11
[direct][stream] add stream support in direct

ahoy-jon:
https://github.com/getkyo/kyo/issues/1221 add support for streams

Forms:

```scala
direct:
  stream.map(x => f(x).now)
```
* could be valid if unwraped. -> compile error

---

```scala
direct:
    stream.map(f)
```
* should be supported without a compile error

---

```scala
direct:
    stream.map(x => f(x).later)
```
* should be supported without a compile error

---

```scala
direct:
   stream.map(x => direct(f(x).now + 1))
```
* should be supported without a compile error

### DIFF
--- a/kyo-direct/shared/src/main/scala/kyo/internal/AsyncShift.scala
+++ b/kyo-direct/shared/src/main/scala/kyo/internal/AsyncShift.scala
@@ -7,6 +7,7 @@ import kyo.*
 import kyo.kernel.internal.Safepoint
 import scala.annotation.targetName
 import scala.collection.IterableOps
+import scala.util.NotGiven
 
 trait asyncShiftLowPriorityImplicit1:
 
@@ -20,6 +21,7 @@ object asyncShift extends asyncShiftLowPriorityImplicit1:
     transparent inline given shiftedChunk[A]: ChunkAsyncShift[A] = new ChunkAsyncShift[A]
     transparent inline given shiftedMaybe: MaybeAsyncShift       = new MaybeAsyncShift
     transparent inline given shiftedResult: ResultAsyncShift     = new ResultAsyncShift
+    transparent inline given shitferStream[V, S]: StreamAsyncShift[V, S] = new StreamAsyncShift
 
 end asyncShift
 
@@ -173,6 +175,18 @@ class MaybeAsyncShift(using Frame) extends AsyncShift[Maybe.type]:
                     case Maybe.Absent     => ifEmpty()
 
 end MaybeAsyncShift
+
+class StreamAsyncShift[V, S] extends AsyncShift[Stream[V, S]]:
+    def map[F[_], V2](stream: Stream[V, S], monad: CpsMonad[F])(f: V => F[V2])(using
+        t1: Tag[Emit[Chunk[V]]],
+        t2: Tag[Emit[Chunk[V2]]],
+        ev: NotGiven[V2 <:< (Any < Nothing)],
+        fr: Frame
+    ): F[Stream[V2, S]] =
+        monad match
+            case _: KyoCpsMonad[?] => monad.pure(stream.map(a => f(a)).asInstanceOf[Stream[V2, S]])
+    end map
+end StreamAsyncShift
 
 class ChunkAsyncShift[A](using Frame) extends KyoSeqAsyncShift[A, Chunk, Chunk[A]]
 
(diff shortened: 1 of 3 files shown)


kyo-direct/shared/src/main/scala/kyo/internal/AsyncShift.scala
```diff
@@ -16,10 +18,32 @@ end asyncShiftLowPriorityImplicit1
...
+            case _: KyoCpsMonad[?] => monad.pure(stream.map(a => f(a)).asInstanceOf[Stream[V2, S]])
+    end map
+
+    def filter[F[_]](stream: Stream[V, S], monad: CpsMonad[F])(f: V => F[Boolean])(using
+        tag: Tag[Emit[Chunk[V]]],
+        discr: Discriminator,
+        frame: Frame
+    ): F[Stream[V, S]] =
+        monad match
+            case _: KyoCpsMonad[?] => monad.pure(stream.filter(f).asInstanceOf[Stream[V, S]])
```
ahoy-jon: It's incorrect. It works for simple use cases, when you unwrap after, for other cases, that's incorrect.

Funny thing, it would run in the same issues as in https://github.com/rcardin/yaes/discussions/33.

I will write a test to show how unsafe it is, then fix to something better.

ahoy-jon: It behaves like context functions
```scala
        "right shift" in run {
            val stream = Stream.init(Seq(1, 2, 3))

            def f(i: Int): Int < Abort[Throwable] = i match
                case 3 => Abort.fail(new Exception("oups"))
                case _ => i

            val newStream: Stream[Int, Any] < Abort[Throwable] = direct:
                stream.rechunk(1).map(x => f(x).now)

            val unwraped = Stream.unwrap(newStream).handle(Abort.recover(_ => Kyo.unit))

            unwraped.run.map: chunk =>
                assert(chunk == Seq(1, 2))

            Abort.run(newStream).map:
                case _: Result.Error[?] => fail("error at stream creation")
                case Result.Success(stream) =>
                    stream.run.map: chunk =>
                        assert(chunk == Seq(1, 2)) //fail

        }
```

ahoy-jon:
```scala
def f(i: Int): String < Var[Int] =
  Var.updateWith[Int](_ + i)(i => i.toString)

val prg =
  direct:
    //it compiles
    val chunk: Chunk[String] = Stream.init(Seq(1, 2, 3)).map(f).run.now
    //chunk

    //it doesn't
    val stream = Stream.init(Seq(1, 2, 3)).map(f)
    ///////                  ^^^^^^^^^^^^      ^
    stream.run.now
```

ahoy-jon:
@fwbrasil @hearnadam there is a bit of cleaning work left, but a first review would help

kyo-direct/shared/src/main/scala/kyo/internal/AsyncShift.scala
```diff
@@ -7,6 +7,7 @@ import kyo.*
 import kyo.kernel.internal.Safepoint
 import scala.annotation.targetName
 import scala.collection.IterableOps
+import scala.util.NotGiven
```
hearnadam:
```suggestion
```

kyo-direct/shared/src/main/scala/kyo/internal/Validate.scala
```diff
@@ -83,6 +84,76 @@ private[kyo] object Validate:
...
+                        if hasNow then
+                            fail(
+                                body,
+                                """
+                                  |Calling `.now` inside a lazy structure breaks effect handling, and allow for escaping behavior.
+                                  |You have two options:
+                                  | - calling .now before building the structure :
+                                  |     def f(x: Int): Int < S
+                                  |     direct:
+                                  |        y = f(1).now
```
hearnadam: Should this be val/def?

kyo-direct/shared/src/main/scala/kyo/internal/Validate.scala
```diff
@@ -83,6 +84,76 @@ private[kyo] object Validate:
...
+                tree match
+                    case Block(List(DefDef(_, _, _, Some(body))), _) =>
+
+                        val hasNow: Boolean = Trees.exists(body)({
+                            case t @ (Apply(TypeApply(Ident("now"), _), List(_)) | Select(_, "now")) => true
+                            case DirectBlock()                                                       => false
+                        })
+
+                        if hasNow then
+                            fail(
```
hearnadam: Should `fail` use `reportErrorAndAbort`?

hearnadam (approved):

kyo-direct/shared/src/main/scala/kyo/internal/Validate.scala
ahoy-jon: No, but i find it confusing to have multiple error in a direct block, so yes!

fwbrasil (approved):

PR #1259, opened 2025-06-12
Covariant Poll

johnhungerford:
### Problem

While `Poll` does not need variance for `Stream`, now that `Emit[V]` is contravariant with `V`, it makes sense to make `Poll` covariant with `V`. This will allow us to have variance in `Pipe` and `Sink`.

### Solution

Made `Poll[V]` contravariant with `V` and made all handling methods support partial handling like in `Emit`.

### Notes

Renamed the `run` overload that runs `Poll` effects against `Emit` effects to `runEmit` since the variance led to type inference issues.


ahoy-jon (approved):
LGTM!

PR #1260, opened 2025-06-13
KyoApp: install signal handler for interruption

### LINKED ISSUE #491: Shutdown gracefully on SIGINT

hearnadam:
### Problem
Currently KyoApp doesn't terminate the computation on user signal (Control+C).

### Solution
Use Async.raceFirst to interrupt an executing computation via a race with a Promise. The promise is completed with an Interrupt on signal handle.

### Notes

- [X] Don't install Signal on windows
- [ ] Use Atomic to prevent multiple signal installs?
- [X] Testing?

/fixes #491


kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -12,8 +13,24 @@ import Result.Error
 abstract class KyoApp extends KyoAppPlatformSpecific:
     override def timeout: Duration = Duration.Infinity
 
+    private val awaitInterrupt =
```
ahoy-jon, line 16: (not important) a type here could make it easier to review.

ahoy-jon (approved):
LGTM

PR #1261, opened 2025-06-14
Add tests for allowed ops following closeAwaitEmpty

steinybot:

### DIFF
--- a/kyo-core/shared/src/test/scala/kyo/QueueTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/QueueTest.scala
@@ -409,6 +409,59 @@ class QueueTest extends Test:
     }
 
     "closeAwaitEmpty" - {
+        "allowed following ops when empty" in run {
+            for
+                q  <- Queue.init[Int](2)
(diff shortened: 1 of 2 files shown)


kyo-core/shared/src/test/scala/kyo/QueueTest.scala
```diff
@@ -409,6 +409,59 @@ class QueueTest extends Test:
...
+        "allowed following ops when not empty" in run {
+            for
+                q  <- Queue.init[Int](2)
+                _  <- q.offer(1)
+                _  <- q.offer(1)
+                f1 <- Async.run(q.closeAwaitEmpty)
+                v1 <- Abort.run(q.size)
+                v2 <- Abort.run(q.empty)
+                v3 <- Abort.run(q.full)
+                v4 <- Abort.run(q.offer(2))
```
steinybot, line 446: This doesn't fail because the `f1` fiber is not the same fiber returned from `closeAwaitEmpty` and so no guarantee that the state has been modified before this point.

Issue #1209, opened 2025-05-25
[feature]: add support for defer on Maybe operations


ahoy-jon:
### Problem Statement

`Maybe[?]` operations are not support in `defer`, compared to `Option[?]`.

```
    "Maybe" - {
        val x: Maybe[Int < Any] = Maybe(1)

        "map" in {
            val d = defer:
                x.map(_.now)

            assert(d.eval == Maybe(1))
        }
    }
```

---
```
[error] -- Error: kyo-direct/shared/src/test/scala/kyo/ShiftTest.scala:306:22 
[error] 306 |                x.map(_.now)
[error]     |                      ^^^^^
[error]     |Method definitions containing .now are not supported inside `defer` blocks.
```

### Proposed Solution

Create the asyncShift handler for `Maybe[?]` and update macros.

ahoy-jon:
I will take of it!

fwbrasil:
Cool! Do we need another for Result?

ahoy-jon:
Thanks to @rssh, https://github.com/dotty-cps-async/dotty-cps-async/issues/99 could be fix on the next release. I will create a draft PR in the meantime to fix the Validation for `Maybe[?]` and `Result[?]` to prepare the work.

hearnadam:
@ahoy-jon is this done? Should we close it?

PR #1263, opened 2025-06-14
direct on maybe/result

ahoy-jon:
thanks to https://github.com/dotty-cps-async/dotty-cps-async/issues/99#issuecomment-2972563568
it's possible to move forward with #1209

----

Maybe:

- - [x] map 
- - [x] flatMap
- - [x] filter
- - [x] exists
- - [x] forall
- - [x] foreach
- - [x] collect
- - [x] getOrElse

Result:
- - [x] map 
- - [x] flatMap
- - [x] filter
- - [x] collect
- - [x] getOrElse

----

For another PR:

Result:
- - [ ] exists
- - [ ] forall
- - [ ] foreach

### DIFF
--- a/kyo-direct/shared/src/main/scala/kyo/internal/AsyncShift.scala
+++ b/kyo-direct/shared/src/main/scala/kyo/internal/AsyncShift.scala
@@ -28,8 +28,8 @@ class ResultAsyncShift(using Frame) extends AsyncShift[Result.type]:
     def map[F[_], E, A](result: Result.type, monad: CpsMonad[F])(ma: Result[E, A])[B](f: A => F[B]): F[Result[E, B]] =
         monad match
             case _: KyoCpsMonad[?] => ma match
-                    case Result.Success(a) => f(a).map(Result.Success(_))
-                    case Result.Failure(e) => monad.pure(Result.Failure(e))
+                    case Result.Success(a)             => f(a).map(Result.Success(_))
+                    case e: Result.Error[E] @unchecked => monad.pure(e)
 
     def flatMap[F[_], E, A](
         result: Result.type,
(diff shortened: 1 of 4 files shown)


ahoy-jon:
blocked by https://github.com/getkyo/kyo/issues/1267 /  https://github.com/dotty-cps-async/dotty-cps-async/issues/100

kyo-direct/shared/src/main/scala/kyo/internal/AsyncShift.scala
```diff
@@ -17,9 +18,162 @@ end asyncShiftLowPriorityImplicit1
...
 
 end asyncShift
 
+class ResultAsyncShift(using Frame) extends AsyncShift[Result.type]:
+
+    def map[F[_], E, A](result: Result.type, monad: CpsMonad[F])(ma: Result[E, A])[B](f: A => F[B]): F[Result[E, B]] =
+        monad match
+            case _: KyoCpsMonad[?] => ma match
+                    case Result.Success(a) => f(a).map(Result.Success(_))
+                    case Result.Failure(e) => monad.pure(Result.Failure(e))
```
fwbrasil: it seems to be missing `Result.Panic`?

ahoy-jon: Yes indeed. 

I think `Result.Partial` should be the default `Result`, but that's another subject. 
```scala
type Result[E, A] = Result.Failure[E] | Result.Successful[A]
type Cause[E]     = Result.Failure[E] | Panic | Interrupt
```

fwbrasil: It seems to still be missing the handling of `Panic`?

kyo-direct/shared/src/main/scala/kyo/internal/AsyncShift.scala
```diff
@@ -17,9 +18,162 @@ end asyncShiftLowPriorityImplicit1
...
+    transparent inline given shiftedResult: ResultAsyncShift     = new ResultAsyncShift
 
 end asyncShift
 
+class ResultAsyncShift(using Frame) extends AsyncShift[Result.type]:
+
+    def map[F[_], E, A](result: Result.type, monad: CpsMonad[F])(ma: Result[E, A])[B](f: A => F[B]): F[Result[E, B]] =
+        monad match
+            case _: KyoCpsMonad[?] => ma match
+                    case Result.Success(a) => f(a).map(Result.Success(_))
```
fwbrasil: It seems to still be missing the handling of Panic?

ahoy-jon: Indeed, sorry, will write some tests for panic.

PR #1264, opened 2025-06-15
Channel stream optimistic drain

steinybot:
### Problem

When the producer is faster than a stream consumer (at least in terms of frequency of calls to put and emit, not necessarily overall throughput), the consumer will call take and drain and perform an unnecessary `Chunk.concat`.

### Solution

Optimistically call `drainUpTo` and only if it is non-empty then call `take`.

### Notes

Whenever the steam consumer blocks it will always emit a single element before draining the rest. This eliminates all `Chunk.concat` but the chunk sizes will be less consistent. If the consumer cares about this then they should call `rechunk`.

Here are the results of running the benchmark with both the old and new implementations:
```
[info] Benchmark                                      (maxChunkSize)   Mode  Cnt       Score       Error   Units
[info] ChannelBench.streamChunks                                   1  thrpt    5    1567.323 ±    46.675   ops/s
[info] ChannelBench.streamChunks:gc.alloc.rate                     1  thrpt    5    1203.317 ±    35.884  MB/sec
[info] ChannelBench.streamChunks:gc.alloc.rate.norm                1  thrpt    5  805148.166 ±     1.308    B/op
[info] ChannelBench.streamChunks:gc.count                          1  thrpt    5      10.000              counts
[info] ChannelBench.streamChunks:gc.time                           1  thrpt    5       7.000                  ms
[info] ChannelBench.streamChunks                                   2  thrpt    5    1629.221 ±    16.583   ops/s
[info] ChannelBench.streamChunks:gc.alloc.rate                     2  thrpt    5    1066.309 ±    10.867  MB/sec
[info] ChannelBench.streamChunks:gc.alloc.rate.norm                2  thrpt    5  686372.088 ±     1.475    B/op
[info] ChannelBench.streamChunks:gc.count                          2  thrpt    5       9.000              counts
[info] ChannelBench.streamChunks:gc.time                           2  thrpt    5       6.000                  ms
[info] ChannelBench.streamChunks                                   4  thrpt    5    2943.928 ±    38.511   ops/s
[info] ChannelBench.streamChunks:gc.alloc.rate                     4  thrpt    5    1116.141 ±    14.647  MB/sec
[info] ChannelBench.streamChunks:gc.alloc.rate.norm                4  thrpt    5  397602.393 ±     0.091    B/op
[info] ChannelBench.streamChunks:gc.count                          4  thrpt    5      10.000              counts
[info] ChannelBench.streamChunks:gc.time                           4  thrpt    5       6.000                  ms
[info] ChannelBench.streamChunks                                   8  thrpt    5    5059.374 ±   285.300   ops/s
[info] ChannelBench.streamChunks:gc.alloc.rate                     8  thrpt    5    1192.789 ±    63.451  MB/sec
[info] ChannelBench.streamChunks:gc.alloc.rate.norm                8  thrpt    5  247251.451 ±   942.549    B/op
[info] ChannelBench.streamChunks:gc.count                          8  thrpt    5      10.000              counts
[info] ChannelBench.streamChunks:gc.time                           8  thrpt    5       6.000                  ms
[info] ChannelBench.streamChunks2                                  1  thrpt    5    1582.921 ±    38.449   ops/s
[info] ChannelBench.streamChunks2:gc.alloc.rate                    1  thrpt    5    1216.563 ±    20.005  MB/sec
[info] ChannelBench.streamChunks2:gc.alloc.rate.norm               1  thrpt    5  806003.443 ±  7173.520    B/op
[info] ChannelBench.streamChunks2:gc.count                         1  thrpt    5      10.000              counts
[info] ChannelBench.streamChunks2:gc.time                          1  thrpt    5       6.000                  ms
[info] ChannelBench.streamChunks2                                  2  thrpt    5    2792.484 ±    39.788   ops/s
[info] ChannelBench.streamChunks2:gc.alloc.rate                    2  thrpt    5    1226.608 ±    64.519  MB/sec
[info] ChannelBench.streamChunks2:gc.alloc.rate.norm               2  thrpt    5  460646.177 ± 20508.503    B/op
[info] ChannelBench.streamChunks2:gc.count                         2  thrpt    5      10.000              counts
[info] ChannelBench.streamChunks2:gc.time                          2  thrpt    5       7.000                  ms
[info] ChannelBench.streamChunks2                                  4  thrpt    5    4841.032 ±    77.556   ops/s
[info] ChannelBench.streamChunks2:gc.alloc.rate                    4  thrpt    5    1280.094 ±    20.254  MB/sec
[info] ChannelBench.streamChunks2:gc.alloc.rate.norm               4  thrpt    5  277303.186 ±   744.321    B/op
[info] ChannelBench.streamChunks2:gc.count                         4  thrpt    5      11.000              counts
[info] ChannelBench.streamChunks2:gc.time                          4  thrpt    5       7.000                  ms
[info] ChannelBench.streamChunks2                                  8  thrpt    5    7295.941 ±   416.986   ops/s
[info] ChannelBench.streamChunks2:gc.alloc.rate                    8  thrpt    5    1322.592 ±    74.115  MB/sec
[info] ChannelBench.streamChunks2:gc.alloc.rate.norm               8  thrpt    5  190108.108 ±   278.695    B/op
[info] ChannelBench.streamChunks2:gc.count                         8  thrpt    5      11.000              counts
[info] ChannelBench.streamChunks2:gc.time                          8  thrpt    5       8.000                  ms
[info] Benchmark result is saved to jmh-result.json
```
`ChannelBench.streamChunks` was the old implementation. `ChannelBench.streamChunks2` is the new one.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -255,20 +255,21 @@ object Channel:
         ): Unit < (Emit[Chunk[A]] & Abort[Closed] & Async) =
             if maxChunkSize <= 0 then ()
             else if maxChunkSize == 1 then
-                Loop.foreach:
+                Loop.forever:
                     Channel.take(self).map: v =>
-                        Emit.valueWith(Chunk(v))(Loop.continue)
+                        Emit.value(Chunk(v))
             else
                 val drainEffect =
                     if maxChunkSize == Int.MaxValue then Channel.drain(self)
                     else Channel.drainUpTo(self)(maxChunkSize)
-                Loop[Unit, Unit, Abort[Closed] & Async & Emit[Chunk[A]]](()): _ =>
-                    for
-                        optimisticChunk <- drainEffect
-                        chunk: Chunk[A] <-
-                            if optimisticChunk.nonEmpty then Kyo.lift(optimisticChunk)
-                            else Channel.take(self).map(Chunk(_))
-                    yield Emit.valueWith(chunk)(Loop.continue)
+                Loop.forever:
+                    drainEffect.map:
+                        case chunk if chunk.nonEmpty => Emit.value(chunk)
+                        case _ =>
+                            for
+                                a  <- Channel.take(self)
+                                ch <- Channel.drainUpTo(self)(maxChunkSize - 1)
+                            yield Emit.value(Chunk(a).concat(ch))
 
         /** Stream elements from channel, optionally specifying a maximum chunk size. In the absence of [[maxChunkSize]], chunk sizes will
           * be limited only by channel capacity or the number of elements in the channel at a given time. (Chunks can still be larger than
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -261,11 +261,14 @@ object Channel:
             else
                 val drainEffect =
                     if maxChunkSize == Int.MaxValue then Channel.drain(self)
-                    else Channel.drainUpTo(self)(maxChunkSize - 1)
+                    else Channel.drainUpTo(self)(maxChunkSize)
                 Loop[Unit, Unit, Abort[Closed] & Async & Emit[Chunk[A]]](()): _ =>
-                    Channel.take(self).map: a =>
-                        drainEffect.map: chunk =>
-                            Emit.valueWith(Chunk(a).concat(chunk))(Loop.continue)
+                    for
+                        optimisticChunk <- drainEffect
+                        chunk: Chunk[A] <-
+                            if optimisticChunk.nonEmpty then Kyo.lift(optimisticChunk)
+                            else Channel.take(self).map(Chunk(_))
+                    yield Emit.valueWith(chunk)(Loop.continue)
 
         /** Stream elements from channel, optionally specifying a maximum chunk size. In the absence of [[maxChunkSize]], chunk sizes will
           * be limited only by channel capacity or the number of elements in the channel at a given time. (Chunks can still be larger than
--- a/kyo-bench/src/main/scala/kyo/bench/ChannelBench.scala
+++ b/kyo-bench/src/main/scala/kyo/bench/ChannelBench.scala
@@ -1,0 +1,33 @@
+package kyo.bench
+
+import kyo.*
+import org.openjdk.jmh.annotations.*
+import org.openjdk.jmh.infra.Blackhole
+import scala.compiletime.uninitialized
+import scala.reflect.ClassTag
+import scala.util.Random
+
+class ChannelBench extends BaseBench:
+
+    import AllowUnsafe.embrace.danger
+
+    var size: Int = 1024
+
+    @Param(Array("1", "2", "4", "8"))
+    var maxChunkSize: Int = uninitialized
+
+    def intFillFn: Int = Random.nextInt()
+
+    @Benchmark def streamChunks() =
+        val x =
+            for
+                channel <- Channel.init[Int](size)
+                _       <- channel.putBatch(Seq.fill(size)(intFillFn))
+                _       <- Async.run(channel.closeAwaitEmpty)
+                count   <- channel.streamUntilClosed(maxChunkSize).into(Sink.count)
+            yield count
+        val count = IO.Unsafe.evalOrThrow(Async.run(x).flatMap(_.block(Duration.Infinity))).getOrThrow
+        assert(count == size, s"Expected $size, got $count")
+    end streamChunks
+
+end ChannelBench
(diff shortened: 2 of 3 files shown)


kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -261,11 +261,14 @@ object Channel:
...
                 Loop[Unit, Unit, Abort[Closed] & Async & Emit[Chunk[A]]](()): _ =>
-                    Channel.take(self).map: a =>
-                        drainEffect.map: chunk =>
-                            Emit.valueWith(Chunk(a).concat(chunk))(Loop.continue)
+                    for
+                        optimisticChunk <- drainEffect
+                        chunk: Chunk[A] <-
+                            if optimisticChunk.nonEmpty then Kyo.lift(optimisticChunk)
+                            else Channel.take(self).map(Chunk(_))
+                    yield Emit.valueWith(chunk)(Loop.continue)
```
ahoy-jon: maybe something like that?
```scala
Loop.foreach[Unit, Abort[Closed] & Async & Emit[Chunk[A]]]:
    drainEffect.map:
        case chunk if chunk.nonEmpty =>
            Emit.valueWith(chunk)(Loop.continue)
        case _ => Channel.take(self).map: a =>
            Emit.valueWith(Chunk(a))(Loop.continue)
```

steinybot: That looks much better 👍

steinybot: Or this?
```scala
drainEffect.map:
    case chunk if chunk.nonEmpty => chunk
    case _                       => Channel.take(self).map(Chunk(_))
.map(Emit.valueWith(_)(Loop.continue))
```

ahoy-jon: I would not do that, I have this weird _personal_ coding rule "duplication start at 3". When you have just two branches, the gain of factoring is limited.

Aside from that, it's maybe a use case for `Loop.forever`. The whole function looks like that to me:
```scala
        private def emitChunks(maxChunkSize: Int = Int.MaxValue)(
            using
            Tag[Emit[Chunk[A]]],
            Frame
        ): Unit < (Emit[Chunk[A]] & Abort[Closed] & Async) =
            require(maxChunkSize > 0)
            if maxChunkSize == 1 then
                Loop.forever:
                    Channel.take(self).map: v =>
                        Emit.value(Chunk(v))
            else
                val drainEffect: Chunk[A] < (IO & Abort[Closed]) =
                    if maxChunkSize == Int.MaxValue then Channel.drain(self)
                    else Channel.drainUpTo(self)(maxChunkSize)

                Loop.forever:
                    drainEffect.map:
                        case chunk if chunk.nonEmpty => Emit.value(chunk)
                        case _ => Channel.take(self).map(a => Emit.value(Chunk(a)))


```

--- 

Note, that's personal taste to this point. Even for the `require(maxChunkSize > 0)`

johnhungerford: This is going to fail to batch emissions properly if the producer is slower than the consumer but produces in chunks. The consumer will have a suspended `take`, the producer will publish a batch to the channel, and the consumer will emit just one of them before emitting a chunk. One of the elements will always end up being split off from a batch in its own `Chunk.Single`

johnhungerford: Could we make the second branch something like:
```scala
                    drainEffect.map:
                        case chunk if chunk.nonEmpty => Emit.value(chunk)
                        case _ => Channel.take(self).map: a => 
                            Channel.drainUpTo(maxChunkSize - 1).map: ch => 
                                Emit.value(Chunk(a).concat(ch))
```
This will drain without suspending when possible, but will try to grab a batch when it does suspend

steinybot: I've updated but I think we should avoid take, drain, concat if possible since it is a pointless split and rejoin. We ought to be able to emit each put as it is without unnecessary ops. A single value should emitted as a single value and a batch as a batch (assuming it is within the maxChunkSize).

johnhungerford: I agree the split and rejoin is not ideal, but unfortunately it is necessary to "emit each put as is." Puts can include both batches and single elements, whereas only single takes can be suspended. I proposed supporting batched takes in a [previous PR](https://github.com/getkyo/kyo/pull/913#discussion_r1876087828) but @fwbrasil thought it introduced more overhead than was gained by eliminating the need to concat chunks. Maybe we should revisit that decision? (I don't have a good sense of the tradeoffs here.)

ahoy-jon: A couple of ideas:

* Maybe there is a solution by just being rechunk friendly? If you pull `n`, you can drain `chunkSize - 1` ?

* maybe there is a need for `Channel.pool`, 
```scala
drainEffect.map:
  case chunk if chunk.nonEmpty => Emit.value(chunk)
  case _ => Channel.poll
```
Just to continue with drain after.

steinybot: That's a cool idea although drain is going to concat anyway. Also all the Channels from Java Queues are going to use a builder anyway so this one extra concat seems fine.

PR #1266, opened 2025-06-15
remove scala 2.12

ahoy-jon:


ahoy-jon:
I don't understand why we have still references to scala 2.12

PR #1268, opened 2025-06-16
Stream variance

johnhungerford:
Addresses #679 

### Problem

To this point `Stream[V, S]` was invariant with `V`. This has led to some usability issues, especially when merging streams. For instance, in an application you may have an event type `DomainEvent` with multiple subtypes, e.g., `UserEvent` and `SystemEvent`. If you had separate streams of these that eventually need to be merged, you could not simply call `userEventStream.merge(systemEventStream)`, as these have different types and could not be lifted by the compiler to `DomainEvent` (or `UserEvent | SystemEvent` in the case where there is no common supertype).

### Solution

- Now that tags support subtyping and unions, and `Emit[V]` is contravariant with `V`, we can simply make `Stream[V, S]` covariant with `V`.
- Since pretty much every method takes `Tag[Emit[Chunkl[V]]]` as an implicit parameter, however, we've needed to lift `V` to `VV >: V` for all these methods to satisfy the compiler's variance constraints.
- This change to the Stream methods' type signatures has made it impossible for the compiler to distinguish between the pure and effectful overloads of functions like `map`, `filter`, `takeWhile`, etc (any method that takes a lambda from `V` to some value). Accordingly, these methods have been split into `map`/`mapKyo`, `filter`/`filterKyo`, `takeWhile`/`takeWhileKyo`, etc. While it's unfortunate that the compiler can't distinguish these automatically, I believe it's for the best since the implementations are substantially different and it's probably a good thing that the user is forced to be explicit about it.

### Notes

Also added a `dropWhile` method alongside `dropWhileKyo`. I believe it was an oversight that we didn't have overloads for this one.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -46,7 +46,7 @@ object StreamCoreExtensions:
         private def emit(listener: Hub.Listener[Result.Partial[E, Maybe[Chunk[A]]]])(using Frame) =
             listener
                 .stream(1)
-                .collectWhile[Chunk[A], Abort[E]] {
+                .collectWhile {
                     case Result.Success(maybeChunk) => maybeChunk
                     case Result.Failure(e)          => Abort.fail(e)
                 }


ahoy-jon:
Not 100% sure about the `mapKyo`, however it looks good, and that will help me with https://github.com/getkyo/kyo/pull/1256 (better errors, easier to support some constructs)

ahoy-jon (approved):

README.md
```diff
@@ -1339,12 +1339,12 @@ case class Config(someConfig: String)
 // Stream with IO effect
 val a: Stream[String, IO] =
     Stream.init(Seq("file1.txt", "file2.txt"))
-        .map(fileName => IO(scala.io.Source.fromFile(fileName).mkString))
+        .mapKyo(fileName => IO(scala.io.Source.fromFile(fileName).mkString))
```
ahoy-jon:
```suggestion
        .map(fileName => IO(scala.io.Source.fromFile(fileName).mkString))
```

README.md
```diff
@@ -1339,12 +1339,12 @@ case class Config(someConfig: String)
...
 
 // Stream with Abort effect
 val b: Stream[Int, Abort[NumberFormatException]] =
     Stream.init(Seq("1", "2", "abc", "3"))
-        .map(s => Abort.catching[NumberFormatException](s.toInt))
+        .mapKyo(s => Abort.catching[NumberFormatException](s.toInt))
```
ahoy-jon:
```suggestion
        .map(s => Abort.catching[NumberFormatException](s.toInt))
```

README.md
```diff
@@ -1400,7 +1400,7 @@ assert(handled.run.eval == Chunk(-9, -8, -7))
...
 
-Note that a number of `Stream` methods (e.g., `map`, `filter`, `mapChunk`) are overloaded to provide different implementations for pure vs effectful transformations. This can make a big difference for performance, so take care that the functions you pass to these methods are typed to return pure values if they do not include effects. Unnecessarily lifting them to return `A < Any` will result in performance loss.
+Note that a number of `Stream` methods (e.g., `map`, `filter`, `mapChunk`) are separated into pure and effectful versions, which have a `-Kyo` suffix. For instance, `map` transforms streamed elements using a function in the form `V => V1`, whereas `mapKyo` transforms elements with a function `V => V1 < S1`, allowing you to use arbitrary effects in the transformation. Any of the `-Kyo` methods will also accept pure functions, since any value can be lifted to a pending type, but the performance will be worse. For optimal performance, be sure to use the pure versions when you can.
```
ahoy-jon:
```suggestion
Note that a number of `Stream` methods (e.g., `map`, `filter`, `mapChunk`) are separated into pure and effectful versions. For instance, `mapPure` transforms streamed elements using a pure function of the form `V => V1`, whereas `map` uses a function `V => V1 < S1`, allowing you to introduce effects in the transformation. While `map` can also accept pure functions (as any value can be lifted), this comes with a performance cost. For optimal performance, prefer `mapPure` when no effects are needed.
```

README.md
```diff
@@ -2034,15 +2034,15 @@ val lines: Stream[String, Resource & IO] =
 
 // Process the stream
 val result: Unit < (Resource & Console & Async & Abort[IOException]) =
-    lines.map(line => Console.printLine(line)).discard
+    lines.mapKyo(line => Console.printLine(line)).discard
```
ahoy-jon:
```suggestion
    lines.map(line => Console.printLine(line)).discard
```

README.md
```diff
@@ -2034,15 +2034,15 @@ val lines: Stream[String, Resource & IO] =
...
 
 // Walk a directory tree
 val tree: Stream[Path, IO] =
     Path("tmp").walk
 
 // Process each file in the tree
 val processedTree: Unit < (Console & Async & Abort[IOException]) =
-    tree.map(file => file.read.map(content => Console.printLine(s"File: ${file}, Content: $content"))).discard
+    tree.mapKyo(file => file.read.map(content => Console.printLine(s"File: ${file}, Content: $content"))).discard
```
ahoy-jon:
```suggestion
    tree.map(file => file.read.map(content => Console.printLine(s"File: ${file}, Content: $content"))).discard
```

ahoy-jon (approved):

fwbrasil (approved):
Maybe let's separate the change to avoid the `*Pure` methods?

PR #1269, opened 2025-06-17
Introduce PendingFn type to simplify Stream.map

hearnadam:
### Problem
`Stream.map` currently uses NotGiven to discriminate between pure and effectful map. This is problematic for generic contexts and for introducing variance.

### Solution
Use a custom trait that preserves the original `A => B` function without the Pending `Any`.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Stream.scala
@@ -64,31 +64,23 @@ sealed abstract class Stream[V, -S] extends Serializable:
       * @return
       *   A new stream with transformed values
       */
-    def map[V2](f: V => V2)(using
+    def map[V2, S2](f: V => V2 < S2)(using
         t1: Tag[Emit[Chunk[V]]],
         t2: Tag[Emit[Chunk[V2]]],
-        ev: NotGiven[V2 <:< (Any < Nothing)],
         fr: Frame
-    ): Stream[V2, S] =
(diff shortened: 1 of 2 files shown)


kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -64,32 +64,24 @@ sealed abstract class Stream[V, -S] extends Serializable:
...
+            f match
+                case f: kernel.`<`.PendingFn1[V, V2] @unchecked =>
+                    ArrowEffect.handleLoop(t1, emit)(
+                        [C] =>
+                            (input, cont) =>
+                                val c = input.map(f.pure)
+                                if c.isEmpty then Loop.continue(cont(()))
+                                else Emit.valueWith(c)(Loop.continue(cont(())))
+                    )
+                case _ => mapChunk(c => Kyo.foreach(c)(f)).emit
```
johnhungerford, line 82: I tried this and found the `PnedingFn1` branch never executed. Were you able to test this successfully?

hearnadam: The issue is probably that the method itself is not being lifted, but the higher priority lift of the element returned in the function. There should be another technique we can use...

johnhungerford: Yes I believe this is the case. I think I can make a working approach with ad-hoc polymorphism, but I don't love the signature that it produces.

PR #1273, opened 2025-06-19
[core] add wall clock delay to Clock.TimeControl + fix flaky tests

fwbrasil:
### Problem

`Clock.TimeControl` only advances the time field and fires the pending async ops without waiting for them to execute. The behavior is actually correct since the time control properly handles the time passing and sleeping fibers but not having a mechanism to wait for async ops limits its usability in tests.

### Solution

Introduce a new parameter `wallClockDelay` that specifies for how lock the time control should wait before continuing execution.

### DIFF
--- a/kyo-core/shared/src/test/scala/kyo/FiberTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/FiberTest.scala
@@ -165,7 +165,7 @@ class FiberTest extends Test:
                         else
                             loop(i - 1, s)
                     }
-                Fiber.raceFirst(Seq(loop(100, "a"), loop(Int.MaxValue, "b"), loop(200, "c"))).map(_.getResult).map { r =>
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/test/scala/kyo/FiberTest.scala
```diff
@@ -165,7 +165,7 @@ class FiberTest extends Test:
                         else
                             loop(i - 1, s)
                     }
-                Fiber.raceFirst(Seq(loop(100, "a"), loop(Int.MaxValue, "b"), loop(200, "c"))).map(_.getResult).map { r =>
+                Fiber.raceFirst(Seq(loop(100, "a"), loop(Int.MaxValue, "b"), loop(2000, "c"))).map(_.getResult).map { r =>
```
fwbrasil, line 168: unrelated flaky test

PR #1276, opened 2025-06-21
add ArrowEffect.wireTap

ahoy-jon:
Small tool to help with simple transformations, without handling the effect.

### DIFF
--- a/kyo-kernel/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/kernel/ArrowEffect.scala
@@ -100,6 +100,103 @@ object ArrowEffect:
                     continue = f(v)
                 )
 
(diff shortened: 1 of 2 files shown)


ahoy-jon:
@fwbrasil i would say it's more a beginner tool on ArrowEffect:
- to show how it's done/used,
- to `tap` into an existing effect, without handling it, eg:
  - addicting checks, transformation to `Var[V]` (if the update is not correct, fix, abort, log)
  - log specific part of an effect handling,
  - ...

ahoy-jon:
@fwbrasil closing it, I think we need to create `kyo-contribs` for things we are not sure they are part of `kio-kernel` and al.

ahoy-jon:
@fwbrasil closing it, I think we need to create `kyo-contribs` for things we are not sure they are part of `kio-kernel` and al.

PR #1281, opened 2025-06-23
Pipe variance

johnhungerford:
### Problem

Composability of `Pipe`, `Sink`, and `Stream` is improved vastly with variance in their streaming element types. Now that `Sink` and `Stream` have been updated, `Pipe` is all that's left.

### Solution

Now that `Poll[V]` is covariant with `V` and `Emit[V]` is contravariant with `V`, `Pipe[A, B, S]` can be made contravariant with `A` and covariant with `B` simply by adding the variance and adding a type parameter narrowing and widening `A` and `B` respectively.

### Notes

Also added `empty` methods to `Sink` and `Pipe`, and added scaladoc to all `empty` methods

### DIFF
--- a/kyo-prelude/shared/src/test/scala/kyo/PipeTest.scala
+++ b/kyo-prelude/shared/src/test/scala/kyo/PipeTest.scala
@@ -166,6 +166,42 @@ class PipeTest extends Test:
             }
         }
 
+        "takeWhilePure" - {
+            "take none" in {
+                assert(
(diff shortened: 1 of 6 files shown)


kyo-prelude/shared/src/test/scala/kyo/PipeTest.scala
```diff
@@ -494,7 +746,7 @@ class PipeTest extends Test:
             }
             "empty stream" in {
                 val stream = Stream
-                    .empty[Int]
+                    .empty
```
ahoy-jon, line 749: nice!

ahoy-jon (approved):

PR #1283, opened 2025-06-23
Kyo test

johnhungerford:
This is a draft POC to float a possible approach to testing in Kyo going forward

### Problem

- Users should have a standard way to write and run Kyo tests without having to use zio-test (only current offering is `kyo-zio-test`)
- Writing a whole testing library is really hard and requires resolving a lot of solved problems

### Solution

This PR is a draft POC proposes defining a standard Kyo testing API that can be implemented for any Scala testing library:
- Model tests using an abstract `Assert` effect that be suspended and then run differently depending on how assertions are modeled by the underlying library
  - Abstract method `assertKyo(condition: Boolean)` to suspend assertion
  - Optional overload `assertKyo(assertion: Assertion)` where `Assertion` is a type belonging to the underlying library: support translating custom assertion types
- Provide methods to translate `Assert` effects into tests. Support at least the following two forms:
  - `Any < (Assert & Abort[Any] & Memo & IO)` (sync)
  - `Any < (Assert & Abort[Any] & Memo & Resource & Async)` (async)

There are some drawbacks to this:
- Some amount of Boilerplate, as users will have to call `runKyo` or some variant for each effect.
- We would miss out on a bunch of test features Kyo's effect system could provide: for instance, scoping resources or dependencies to different levels of nested suites like zio-test allows

Ultimately, I think the power and flexibility of this approach make it good option for the time being. The amount of work required to generate a proper testing library is better spent on other aspects of Kyo, especially given that this approach  works quite well for most use cases. Moreover, it provides an incremental approach to developing a full-fledged testing library, as we can work out a suitable testing API prior to solving all the other problems.

Other advantages:
- Easier transition to kyo: users can use their favorite testing libraries (including their favorite assertions DSLs)
- Easier documentation: we can show what to import for each testing library, but any example tests should be able to be written in the same way
- Easy integration maintenance: rather than writing a bespoke implementation of each library's base test suite class, we only have to implement a few methods of our own methods for each test suite. I found I was able to get a working implementation for zio and scalatest very quickly without having to understand too much about each library
  - We also don't have to write separate integrations for different "flavors" of a given library. For instance, a single implementation can support the flatspec, freespec, funspec, etc. styles in scalatest
- Using the `Assert` effect affords some nice flexibility. For instance, you can easily choose between a "fail-fast" approach (like scalatest's default) or an approach that runs everything before submitting assertions to the testing framework.


kyo-test-zio/shared/src/test/scala/kyo/test/KyoScalatestApiTest.scala
```diff
@@ -0,0 +1,45 @@
...
+                            i2 <- Var.get[Int]
+                            _  <- assertKyo(assertTrue(i2 == i + 1))
+                            _  <- assertKyo(i2 == i + 1)
+                        yield ()
+                    Choice.run:
+                        for
+                            i <- Choice.evalSeq(Range(0, 100))
+                            _ <- Var.run(i)(varEffect)
+                        yield ()
+                        end for
```
johnhungerford, line 24: Here we see another benefit to this approach: while it is nice that zio-test evaluates all assertions instead of only one, having to yield the assertion at the end of the effect can be pretty annoying. In this example, it makes more sense to define the main assertion in `varEffect` than in the `Choice.run` block. The way I've defined the `Assert` handler, however, just concatenates all assertions together (`Assert` is just `Emit` under the hood), the result is the same as if all assertions had been provided at the end.

kyo-test-scalatest/shared/src/main/scala/kyo/test/Assert.scala
```diff
@@ -0,0 +1,25 @@
+package kyo.test
+
+import kyo.*
+import org.scalatest.exceptions.TestCanceledException
```
hearnadam, line 4: Can we instead avoid any dependence on scalatest? I think we should ideally use our own assertions, then create modules for registering tests with different testing suites.

It shouldn't be too hard, as we can always propagate assertions as values instead of exceptions.

kyo-test-scalatest/shared/src/main/scala/kyo/test/KyoScalatestApi.scala
```diff
@@ -0,0 +1,55 @@
...
+import org.scalatest.Assertion
+import org.scalatest.Suite
+import scala.concurrent.Future
+
+trait KyoScalatestApi extends KyoTestApiSync[Assertion] with KyoTestApiAsync[Future[Assertion]] with KyoTestApiSpecialAssertion[Assertion]:
+    self: Suite =>
+
+    type Assert = KyoAssert.Assert
+
+    override inline def assertKyo(inline condition: Boolean)(using Frame): Unit < Assert =
```
hearnadam, line 13: `assertTrue` or `assert` IMO

kyo-test-scalatest/shared/src/main/scala/kyo/test/Assert.scala
johnhungerford: This opens up a can of worms in terms of implementation. Scalatest's `assert` method, for instance, is a macro (as is utest's) that needs to introspect the assertion at compiletime to work correctly. If we made our own assertion type, we wouldn't be able to leverage that functionality.

kyo-test-scalatest/shared/src/main/scala/kyo/test/KyoScalatestApi.scala
johnhungerford: Yes, but name clashes. Since we want to expose all of the libraries assertion mechanism to the user, we need to make sure they can easily use the original assertion DSL.

PR #1285, opened 2025-06-27
restrict lift to avoid lifting Abort.type and others

ahoy-jon:
When working with kyo-combinators, as a user we are proposed to lift anything which lead to the IDE proposing:

```scala
Abort.foldAbort

//with foldAbort from AbortCombinators:

def foldAbort[B, S1](
        onSuccess: A => B < S1,
        onFail: E => B < S1
    )(
        using
        ct: SafeClassTag[E],
        fr: Frame
    ): B < (Abort[Nothing] & S & S1) =
        Abort.fold[E](onSuccess, onFail)(effect)
```

This PR makes those construct invalid, and try to improve the developper experience.

---

There is a reduction of generated bytecode (not by a lot), but complety remove calls on trivial types.

### DIFF
--- a/kyo-kernel/jvm/src/test/scala/kyo/kernel/BytecodeTest.scala
+++ b/kyo-kernel/jvm/src/test/scala/kyo/kernel/BytecodeTest.scala
@@ -32,12 +32,12 @@ class BytecodeTest extends Test:
 
     "map" in {
         val map = methodBytecodeSize[TestMap]
-        assert(map == Map("test" -> 26, "anonfun" -> 11, "mapLoop" -> 170))
+        assert(map == Map("test" -> 26, "anonfun" -> 11, "mapLoop" -> 158))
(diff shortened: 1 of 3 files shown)


kyo-kernel/jvm/src/test/scala/kyo/kernel/BytecodeTest.scala
```diff
@@ -32,12 +32,12 @@ class BytecodeTest extends Test:
 
     "map" in {
         val map = methodBytecodeSize[TestMap]
-        assert(map == Map("test" -> 26, "anonfun" -> 11, "mapLoop" -> 170))
+        assert(map == Map("test" -> 26, "anonfun" -> 11, "mapLoop" -> 158))
```
ahoy-jon, line 35: -7%, that's nice!

PR #1287, opened 2025-06-27
Avoid marking synthetic files as internal

### LINKED ISSUE #1286: [BUG]: Kyo cannot be used at top level
### Version

0.19.0

### Scala Version

3.7.1

### Expected Behavior

The following samples should compile and run:

```scala
//> using scala "3.7.1"
//> using dep "io.getkyo::kyo-core::0.19.0"
//> using dep "io.getkyo::kyo-prelude::0.19.0"

import kyo.*

val a: Int < Any = 1
val b: Int = a.eval
```

```scala
//> using scala "3.7.1"
//> using dep "io.getkyo::kyo-core::0.19.0"
//> using dep "io.getkyo::kyo-prelude::0.19.0"

import kyo.*

def example1(
    a: Int < Async,
    b: Int < Abort[Exception]
): Int < (Async & Abort[Exception]) =
    a.flatMap(v => b.map(_ + v))
```

### Actual Behavior

Both fail at compile-time with "Frame cannot be derived within the kyo package." error (Scala CLI 1.8.3):

```scala
[error] Frame cannot be derived within the kyo package.
[error] 
[error] To resolve this issue:
[error] 1. Propagate the Frame from user code via a `using` parameter.
[error] 2. If absolutely necessary, you can use Frame.internal, but this is not recommended for general use.
[error] 
[error] Example of propagating Frame:
[error] def myMethod[A](a: A)(using Frame): Unit = ...
[error] val b: Int = a.eval
```

```scala
[error] Frame cannot be derived within the kyo package.
[error] 
[error] To resolve this issue:
[error] 1. Propagate the Frame from user code via a `using` parameter.
[error] 2. If absolutely necessary, you can use Frame.internal, but this is not recommended for general use.
[error] 
[error] Example of propagating Frame:
[error] def myMethod[A](a: A)(using Frame): Unit = ...
[error]     a.flatMap(v => b.map(_ + v))
```

### Steps to Reproduce

Run any of these samples with Scala CLI, either in a `.sc` or `.scala` file.

### Current Workaround

Wrapping the code in an object solves the problem:

```scala
//> using scala "3.7.1"
//> using dep "io.getkyo::kyo-core::0.19.0"
//> using dep "io.getkyo::kyo-prelude::0.19.0"

import kyo.*

object Foo:
  // An 'Int' is also an 'Int < Any'
  val a: Int < Any = 1

  // Since there are no pending effects,
  // the computation can produce a pure value
  val b: Int = a.eval

  // Kyo still supports both `map`
  // and `flatMap`.
  def example1(
      a: Int < Async,
      b: Int < Abort[Exception]
  ): Int < (Async & Abort[Exception]) =
      a.flatMap(v => b.map(_ + v))
```

hearnadam:
Fixes #1286

### Problem
Synthetic files (scala repl, worksheet, etc) are currently triggering `FindEnclosing.isInternal`

### Solution
Manually exclude empty files and files starting with `<` - like `<repl>` from internal.

### Notes
I tested with a publishLocal, but not sure how robust that is. I can't think of a great way to test.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/internal/FindEnclosing.scala
+++ b/kyo-data/shared/src/main/scala/kyo/internal/FindEnclosing.scala
@@ -11,9 +11,12 @@ private[kyo] object FindEnclosing:
     def isInternal(using Quotes): Boolean =
         val pos      = quotes.reflect.Position.ofMacroExpansion
         val fileName = pos.sourceFile.name
-        val path     = pos.sourceFile.path
-        val excluded = (path.contains("src/test/") && testFileSuffixes.exists(fileName.endsWith)) || fileName.endsWith("Bench.scala")
-        apply(sym => sym.fullName.startsWith("kyo") && !excluded).nonEmpty
+        if fileName.isEmpty || fileName.startsWith("<") then false // synthetic file, like scala-cli/repl
+        else
+            val path     = pos.sourceFile.path
+            val excluded = (path.contains("src/test/") && testFileSuffixes.exists(fileName.endsWith)) || fileName.endsWith("Bench.scala")
+            apply(sym => sym.fullName.startsWith("kyo") && !excluded).nonEmpty
+        end if
     end isInternal
 
     def apply(using Quotes)(predicate: quotes.reflect.Symbol => Boolean): Maybe[quotes.reflect.Symbol] =


ahoy-jon:
> I tested with a publishLocal, but not sure how robust that is. I can't think of a great way to test.

I think we should add some Scala-Cli fragments as tests at some point.

PR #1289, opened 2025-06-27
try enabling Native tests again + update dependencies + fix flaky tests

fwbrasil:
🤞

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
+++ b/kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
@@ -14,7 +14,10 @@ object StreamCoreExtensions:
                 case Absent =>
                     Loop.done
                 case Present(c) =>
-                    Emit.valueWith(c)(Loop.continue)
+                    if c.nonEmpty then
+                        Emit.valueWith(c)(Loop.continue)
+                    else
+                        Loop.continue
         Abort.run(emit).unit
     end emitMaybeChunksFromChannel
 
--- a/kyo-core/shared/src/test/scala/kyo/ClockTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/ClockTest.scala
@@ -425,7 +425,7 @@ class ClockTest extends Test:
                     _        <- task.interrupt
                     instants <- queue.drain
                 yield
-                    intervals(instants).foreach(v => assert(v == 1.millis))
+                    intervals(instants).foreach(v => assert(v <= 2.millis))
                     succeed
             }
         }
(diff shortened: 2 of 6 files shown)


kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -14,7 +14,10 @@ object StreamCoreExtensions:
                 case Absent =>
                     Loop.done
                 case Present(c) =>
-                    Emit.valueWith(c)(Loop.continue)
+                    if c.nonEmpty then
+                        Emit.valueWith(c)(Loop.continue)
+                    else
+                        Loop.continue
```
fwbrasil, line 20: @johnhungerford there was a flaky test due to an empty chunk sometimes being emitted. I've fixed it here, not sure if it's the best place. Failure: https://github.com/getkyo/kyo/actions/runs/15930299494/job/44937602854#step:6:7402

kyo-core/shared/src/test/scala/kyo/ClockTest.scala
```diff
@@ -425,7 +425,7 @@ class ClockTest extends Test:
                     _        <- task.interrupt
                     instants <- queue.drain
                 yield
-                    intervals(instants).foreach(v => assert(v == 1.millis))
+                    intervals(instants).foreach(v => assert(v <= 2.millis))
```
fwbrasil, line 428: SN has more drift in thread wake ups. That might be the cause of the issue with the scheduler regulator in SN. A different tuning might be enough cc/ @hearnadam

kyo-scheduler/jvm-native/src/test/scala/kyo/scheduler/SchedulerTest.scala
```diff
@@ -30,7 +31,7 @@ class SchedulerTest extends AnyFreeSpec with NonImplicitAssertions {
         }
 
         "handles task that throws exception" in withScheduler { scheduler =>
-            val task = TestTask(_run = () => throw new RuntimeException("Test exception."))
+            val task = TestTask(_run = () => throw new RuntimeException("Test exception.") with NoStackTrace)
```
fwbrasil, line 34: SN also seems to have some concurrency issue to report stack traces

kyo-scheduler/jvm-native/src/test/scala/kyo/scheduler/WorkerTest.scala
```diff
@@ -18,6 +18,8 @@ import org.scalatest.freespec.AnyFreeSpec
 
 class WorkerTest extends AnyFreeSpec with NonImplicitAssertions {
 
+    val executor = Executors.newCachedThreadPool(Threads("test-worker"))
```
fwbrasil, line 21: This was the hardest case. SN seems to have an issue with shutting down daemon thread pools with active threads (workers in this case). It's not ideal to leave the thread pool open but it's daemon so it's ok

project/plugins.sbt
```diff
@@ -14,6 +14,8 @@ addSbtPlugin("org.scoverage" % "sbt-scoverage" % "2.3.1")
 
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "1.1.4")
 
+addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.6.3")
```
fwbrasil, line 17: I've added this plugin while we don't have a solution for automatic updates

hearnadam (approved):

PR #1290, opened 2025-06-27
specialize >.lift for AnyVal and Unit

ahoy-jon:
Following https://github.com/getkyo/kyo/pull/1285 a lighter version.

- No visible compileTime increase
- Less bytecode on simple cases

### DIFF
--- a/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
@@ -417,9 +417,13 @@ object `<`:
             case kyo: Kyo[?, ?] => Nested(kyo)
             case _              => v.asInstanceOf[A < S]
 
-    implicit inline def liftUnit[S1, S2](inline v: Unit < S1): Unit < S2 = ${ liftUnitImpl[S1, S2]('v) }
+    implicit inline def liftAnyVal[A <: AnyVal, S](inline v: A): A < S = v.asInstanceOf[A < S]
 
-    private def liftUnitImpl[S1: Type, S2: Type](v: Expr[Unit < S1])(using quotes: Quotes): Expr[Unit < S2] =
+    implicit inline def liftUnit[S](inline v: Unit): Unit < S = v.asInstanceOf[Unit < S]
+
+    implicit inline def abortCastUnit[S1, S2](inline v: Unit < S1): Unit < S2 = ${ abortCastUnitImpl[S1, S2]('v) }
+
+    private def abortCastUnitImpl[S1: Type, S2: Type](v: Expr[Unit < S1])(using quotes: Quotes): Expr[Unit < S2] =
         import quotes.reflect.*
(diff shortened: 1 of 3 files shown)


kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -417,9 +417,13 @@ object `<`:
...
 
-    private def liftUnitImpl[S1: Type, S2: Type](v: Expr[Unit < S1])(using quotes: Quotes): Expr[Unit < S2] =
+    implicit inline def liftUnit[S](inline v: Unit): Unit < S = v.asInstanceOf[Unit < S]
+
+    implicit inline def abortCastUnit[S1, S2](inline v: Unit < S1): Unit < S2 = ${ abortCastUnitImpl[S1, S2]('v) }
+
+    private def abortCastUnitImpl[S1: Type, S2: Type](v: Expr[Unit < S1])(using quotes: Quotes): Expr[Unit < S2] =
         import quotes.reflect.*
         val source = TypeRepr.of[S1].show
         report.errorAndAbort(
```
hearnadam, line 429: pre-existing, but can you show `S2` instead of `?`

ahoy-jon: I will try, but from what I remember, it's not worth it!

ahoy-jon: indeed, not worth it, for example the compiler is feeding `Any`, when S2 is `Emit[Int]`
```
[info]       │   "expected" -> "Cannot lift `Unit < kyo.Abort[scala.Predef.String]` to the expected type (`Unit < ?`).",
[info]       │   "actual" -> """Cannot lift `Unit < kyo.Abort[scala.Predef.String]` to the expected type (`Unit < scala.Any`).
```

so `?` was the right choice !

fwbrasil:
is this to fix the same issue as https://github.com/getkyo/kyo/pull/1291?

ahoy-jon:
| **Branch**                  | **Run** | **User Time** | **System Time** | **CPU Usage** | **Total Time** |
|----------------------------|--------|--------------|---------------|--------------|--------------|
| `feat/avoid-lifting-modules` | 1      | 527.57s      | 20.68s        | 441%         | 2:04.24      |
| `feat/avoid-lifting-modules` | 2      | 513.18s      | 19.29s        | 453%         | 1:57.41      |
| `main`                      | 1      | 521.05s      | 19.72s        | 420%         | 2:08.75      |
| `main`                      | 2      | 521.46s      | 19.44s        | 456%         | 1:58.40      |
| **`feat/specialize-lift`**      | 1      | 531.72s      | 18.35s        | 430%         | 2:07.79      |
| **`feat/specialize-lift`**      | 2      | 535.36s      | 20.55s        | 451%         | 2:03.09      |

PR #1291, opened 2025-06-27
avoid lifting modules (like Abort.type/Var.type, ...)

ahoy-jon:
Plain solution about this problem. There is probably a better way.

@johnhungerford that's what I understood from: "Could we use implicit evidence to exclude certain objects from lifting?", you tell me!


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -42,7 +43,7 @@ import scala.util.control.NonFatal
   */
 opaque type Async <: (Sync & Async.Join) = Async.Join & Sync
 
-object Async extends AsyncPlatformSpecific:
+object Async extends AsyncPlatformSpecific with KyoModule:
```
johnhungerford: Yeah this is what I had in mind

kyo-kernel/shared/src/main/scala/kyo/kernel/internal/WeakFlat.scala
```diff
@@ -30,7 +32,8 @@ trait WeakFlatLowPriority:
 object WeakFlat extends WeakFlatLowPriority:
     inline given WeakFlat[Unit] = unsafe.bypass
 
-    inline given [A](using inline ng: NotGiven[A <:< (Any < Nothing)]): WeakFlat[A] = unsafe.bypass
+    inline given [A](using inline ng: NotGiven[A <:< (Any < Nothing)], inline ngModule: NotGiven[A <:< KyoModule]): WeakFlat[A] =
```
johnhungerford: Do we need to make this a part of `WeakFlat`? Why not make it a second evidence for the `lift` implicit? (This doesn't seem to have anything to do with flatness.)

ahoy-jon: +100, it's just it will be "NotKyoModule"!

kyo-bench/src/main/scala/kyo/bench/ChannelBench.scala
```diff
@@ -26,7 +26,7 @@ class ChannelBench extends BaseBench:
                 _       <- Async.run(channel.closeAwaitEmpty)
                 count   <- channel.streamUntilClosed(maxChunkSize).into(Sink.count)
             yield count
-        val count = IO.Unsafe.evalOrThrow(Async.run(x).flatMap(_.block(Duration.Infinity))).getOrThrow
+        val count = Sync.Unsafe.evalOrThrow(Async.run(x).flatMap(_.block(Duration.Infinity))).getOrThrow
```
fwbrasil: did we miss the rename in the bench module? it's odd since the build isn't failing

fwbrasil:
I don't like the overall approach since it pollutes the codebase and we can forget to add the marker to new code. How about renaming `WeakFlat` to `CanLift` and disallowing lifting of any companion object in the macro itself?

kyo-bench/src/main/scala/kyo/bench/ChannelBench.scala
ahoy-jon: indeed, I have since that in logs.

ahoy-jon:
@fwbrasil 

> I don't like the overall approach since it pollutes the codebase

True 💯 

> How about renaming WeakFlat to CanLift and disallowing lifting of any companion object in the macro itself?

Like in this PR : https://github.com/getkyo/kyo/pull/1285 with `lift` as a macro? or `CanLift` as a macro?

kyo-kernel/shared/src/main/scala/kyo/kernel/internal/CanLift.scala
```diff
@@ -0,0 +1,64 @@
...
+     (x: (Int < S1) < S2).flatten  // Result: Int < (S1 & S2)
+
+2. Split the computation into multiple statements:
+     val x: Int < S1 = computeFirst()
+     val y: Int < S2 = useResult(x)
+""")
+opaque type CanLift[A] = Null
+
+object CanLiftMacro:
+    inline given derived[A](using inline ng: NotGiven[A <:< (Any < Nothing)]): CanLift[A] = ${ liftImpl[A] }
```
ahoy-jon, line 42: circular references don't help with the macro:
- `NotGiven[A <:< (Any < Nothing)]` is better than doing it in the macro.
- the macro have to be isolated from `inline given CanLift[Nothing]`, hence the `CanLiftMacro`, then an export.

fwbrasil: why not inspect the `A` type for nesting in the macro code? There's also `Expr.summon` I think that can be used in the macro but avoiding another implicit search could be good for compilation times.

kyo-kernel/shared/src/main/scala/kyo/kernel/internal/CanLift.scala
```diff
@@ -0,0 +1,64 @@
...
+
+    private[internal] def liftImpl[A: Type](using Quotes): Expr[CanLift[A]] =
+        import quotes.reflect.*
+        val tpe = TypeRepr.of[A].dealias
+
+        val sym = tpe.typeSymbol
+        if sym.fullName.startsWith("kyo.") && sym.flags.is(Flags.Module) && !sym.flags.is(Flags.Case) then
+            report.error(s"Cannot lift '${sym.fullName}' to a '${sym.name} < S'", Position.ofMacroExpansion)
```
hearnadam:
```suggestion
            report.errorAndAbort(s"Cannot lift '${sym.fullName}' to a '${sym.name} < S'", Position.ofMacroExpansion)
```

ahoy-jon: ok, will try that, thanks!

kyo-kernel/shared/src/main/scala/kyo/kernel/internal/CanLift.scala
ahoy-jon: > why not inspect the `A` type 

It was not working a all, with painful compile errors due to cyclic definitions.

Now it could be working and make PendingTest.scala crash with 32 compile errors:
```
[error] -- Error: /Users/jon/Documents/GitHub/kyo/kyo-kernel/shared/src/test/scala/kyo/kernel/PendingTest.scala:456:48 
[error] 456 |                TestEffect1(1).map(_ => Kyo.lift(TestEffect2("hello")))
[error]     |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]     |Type '<' may contain a nested effect computation.
```

```scala
        if tpe <:< TypeRepr.of[Any < Nothing] then
            report.errorAndAbort(s"""Type '${sym.name}' may contain a nested effect computation.
                                   |
                                   |This usually means you have a value of type 'X < S' where you need a plain value.
                                   |To fix this, you can:
                                   |
                                   |1. Call .flatten to combine the nested effects:
                                   |     (x: (Int < S1) < S2).flatten  // Result: Int < (S1 & S2)
                                   |
                                   |2. Split the computation into multiple statements:
                                   |     val x: Int < S1 = computeFirst()
                                   |     val y: Int < S2 = useResult(x)""".stripMargin)
        end if
```

It's better to not change this part now, and have a discussion when possible to plan the work on lift.

ahoy-jon:
With the current implementation

| **Branch**                  | **Run** | **User Time** | **System Time** | **CPU Usage** | **Total Time** |
|----------------------------|--------|--------------|---------------|--------------|--------------|
| `feat/avoid-lifting-modules` | 1      | 527.57s      | 20.68s        | 441%         | 2:04.24      |
| `feat/avoid-lifting-modules` | 2      | 513.18s      | 19.29s        | 453%         | 1:57.41      |
| `main`                      | 1      | 521.05s      | 19.72s        | 420%         | 2:08.75      |
| `main`                      | 2      | 521.46s      | 19.44s        | 456%         | 1:58.40      |

ahoy-jon:
| **Branch**                  | **Run** | **User Time** | **System Time** | **CPU Usage** | **Total Time** |
|----------------------------|--------|--------------|---------------|--------------|--------------|
| `feat/avoid-lifting-modules` | 1      | 527.57s      | 20.68s        | 441%         | 2:04.24      |
| `feat/avoid-lifting-modules` | 2      | 513.18s      | 19.29s        | 453%         | 1:57.41      |
| `main`                      | 1      | 521.05s      | 19.72s        | 420%         | 2:08.75      |
| `main`                      | 2      | 521.46s      | 19.44s        | 456%         | 1:58.40      |
| **`feat/specialize-lift`**      | 1      | 531.72s      | 18.35s        | 430%         | 2:07.79      |
| **`feat/specialize-lift`**      | 2      | 535.36s      | 20.55s        | 451%         | 2:03.09      |

kyo-kernel/shared/src/main/scala/kyo/kernel/internal/CanLift.scala
```diff
@@ -0,0 +1,64 @@
...
+@implicitNotFound("""
+Type '${A}' may contain a nested effect computation.
+
+This usually means you have a value of type 'X < S' where you need a plain value.
+To fix this, you can:
+
+1. Call .flatten to combine the nested effects:
+     (x: (Int < S1) < S2).flatten  // Result: Int < (S1 & S2)
+
+2. Split the computation into multiple statements:
```
hearnadam, line 33: Why does this fix it? 😄 Maybe mention it's due to inference?

kyo-kernel/shared/src/main/scala/kyo/kernel/internal/CanLift.scala
```diff
@@ -0,0 +1,64 @@
...
+    private[internal] def liftImpl[A: Type](using Quotes): Expr[CanLift[A]] =
+        import quotes.reflect.*
+        val tpe = TypeRepr.of[A].dealias
+
+        val sym = tpe.typeSymbol
+        if sym.fullName.startsWith("kyo.") && sym.flags.is(Flags.Module) && !sym.flags.is(Flags.Case) then
+            report.errorAndAbort(s"Cannot lift '${sym.fullName}' to a '${sym.name} < S'", Position.ofMacroExpansion)
+            return '{ ??? }
+
+        // Passed checks
```
hearnadam:
```suggestion

```

kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -405,21 +405,25 @@ object `<`:
...
       *   A computation in the effect context
       */
-    implicit def lift[A: WeakFlat, S](v: A): A < S =
+    implicit def lift[A: CanLift, S](v: A): A < S =
         v match
             case kyo: Kyo[?, ?] => Nested(kyo)
             case _              => v.asInstanceOf[A < S]
 
-    implicit inline def liftUnit[S1, S2](inline v: Unit < S1): Unit < S2 = ${ liftUnitImpl[S1, S2]('v) }
+    implicit inline def liftAnyVal[A <: AnyVal, S](inline v: A): A < S = v.asInstanceOf[A < S]
```
hearnadam, line 420: Do we need `CanLift`? Technically you could put an AnyVal around a computation, right?

ahoy-jon: I am sorry, what do you mean?

kyo-kernel/shared/src/main/scala/kyo/kernel/internal/CanLift.scala
ahoy-jon: No idea, it's the original documentation of WeakFlat. At least with the current implementation, you either get this paragraphe, or the one on modules depending of the type of compile errors.

kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
ahoy-jon: Maybe we should do a call on the rest, their are a lot of interesting things to do with `lift`.

Eg. optimizing all the `Const[A <: AnyVal][?]` could be a huge improvement (perf / bytecode). as far as I see, they are not using `liftAnyVal`, and it's use a lot in `ArrowEffect`.

kyo-kernel/shared/src/main/scala/kyo/kernel/internal/CanLift.scala
```diff
@@ -0,0 +1,64 @@
...
+This usually means you have a value of type 'X < S' where you need a plain value.
+To fix this, you can:
+
+1. Call .flatten to combine the nested effects:
+     (x: (Int < S1) < S2).flatten  // Result: Int < (S1 & S2)
+
+2. Split the computation into multiple statements:
+     val x: Int < S1 = computeFirst()
+     val y: Int < S2 = useResult(x)
+""")
```
ahoy-jon: @hearnadam something like that?

```suggestion
Type '${A}' may contain a nested effect computation.
This usually means you have a value of type `X < S1 < S2` (i.e. `(X < S1) < S2`) where a plain value `X < S` is expected.

This often happens due to *type inference*: some effect computations are nested when chaining operations, and Scala infers a value with nested effects instead of merging them.

To fix this, you can:

1. Call `.flatten` to merge the nested effects:
    val x: (Int < S1) < S2 = ...
    val y: Int < (S1 & S2) = x.flatten


   This collapses the nested effect layers into a single computation with a combined effect set.

2. Split the computation into multiple statements:
   Breaking the code into smaller expressions helps Scala infer the correct types incrementally, avoiding nested effects.
    val x: Int < S1 = computeFirst()
    val y: Int < S2 = useResult(x)

""")
```

Issue #1278, opened 2025-06-22
[feature]: ZStream Support


hearnadam:
### Problem Statement

Adding ZStream support will unlock alternative Server/Client libraries, such as zio-http.

### Proposed Solution

tbh 🙂

ahoy-jon:
probably closed by #1298 , need to check

hearnadam:
The current support is not bi-directional

PR #1298, opened 2025-06-29
add ZStream interop to Stream

HollandDM:
This add conversions between `ZStream[Any, E, A]` and `Stream[A, Abort[E] & Async]`

### Problem
[Details here](https://github.com/getkyo/kyo/issues/1278)

### Solution
Use ZIO's toPull in to create a recursive loop pulling elements from the stream.

### Notes
Will add tests later


kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -45,6 +49,34 @@ object ZIOs:
     inline def get[R: zio.Tag, E, A](v: ZIO[R, E, A])(using Tag[Env[R]], Frame, zio.Trace): A < (Env[R] & Abort[E] & Async) =
         compiletime.error("ZIO environments are not supported yet. Please handle them before calling this method.")
```
hearnadam: Can you make a new ZStreams file? I also wonder if we should be using a separate module? perhaps too much overhead.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -45,6 +49,34 @@ object ZIOs:
...
+      * @return
+      *   A Kyo's Stream that, when run, will execute the zio.stream.ZStream
+      */
+    def get[E, A](stream: => ZStream[Any, E, A])(using
+        frame: Frame,
+        trace: zio.Trace,
+        tag: kyo.Tag[kyo.Emit[kyo.Chunk.Indexed[A]]],
+        classTag: ClassTag[A]
+    ): Stream[A, Abort[E] & Async] =
+        zio.Unsafe.unsafely { unsafe ?=>
```
hearnadam: Can you use `Stream` (and potentially a single `Sync` at the top level for a suspension? `Unsafe` doesn't suspend.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -45,6 +49,34 @@ object ZIOs:
...
+            val emit = Resource
+                .acquireRelease(Scope.unsafe.make(using unsafe))(_.close(Exit.unit))
```
hearnadam: In my next PR I will add a change to convert `Error[A]` to `Cause[A]`. We can correctly close the scope with the exit.

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
HollandDM: Resource already has an outer Sync, but I can add it to this too.
I’m not quite understand about using Stream, can you elaborate ?

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
hearnadam: See: https://github.com/getkyo/kyo/pull/1299

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
hearnadam: You are referencing a by-name inside a suspension, but it's not 100% clear that's the case - seems prone to breaking when someone refactors. Something like this can help prevent bugs, since the top level code is always suspended.
```scala
Stream {
  Sync.Unsafe {
    val scope = Scope.unsafe
    ...
```

Might be good to add a `Stream.Unsafe {` btw!

kyo-zio/shared/src/test/scala/kyo/ZStreamsTest.scala
```diff
@@ -0,0 +1,94 @@
...
+            val kyoStream = Stream.init(1 to 10, 2)
+                .concat(Stream.init(Abort.fail(Error)))
+                .concat(Stream.unfold(0)(prev => Maybe.Present(prev -> (prev + 1))))
+            val zioStream = ZStreams.run(kyoStream)
+            zioStream.runDrain.exit.map {
+                case zio.Exit.Failure(cause) if cause.contains(zio.Cause.fail(Error)) => assert(true)
+                case _                                                                => assert(false)
+            }
+        }
+        // "convert correctly with parallel kyo's stream" in runZIO {
```
HollandDM, line 81: There is a problem with this test cases, I'll try come up with a solution later. For now leave it as is. Here the detail:

`mapParUnordered` right now only work if consume the resulted stream inside Kyo boundary. But currently `ZStreams.run` consume the stream by "pulling" signal, which is emitted from ZChannel runtime and will run a single `emit` inside Kyo boundary. Because of that, the second time ZChannel call pull and continue the multi shot `emit`, its scope has already been cleaned, and will result in an error.

This happened because `mapParUnordered` is using `Channel` as its underlying mechanism, which will be closed when jump out of Kyo boundary. A fix for this will likely required `mapParUnordered`'s implementation to change a lot.

kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -426,11 +426,14 @@ object `<`:
     private def abortCastUnitImpl[S1: Type, S2: Type](v: Expr[Unit < S1])(using quotes: Quotes): Expr[Unit < S2] =
         import quotes.reflect.*
         val source = TypeRepr.of[S1].show
+        val target = TypeRepr.of[S2].show
         report.errorAndAbort(
-            s"""Cannot lift `Unit < ${source}` to the expected type (`Unit < ?`).
+            s"""Cannot lift `Unit < ${source}` to the expected type (`Unit < ${target}`).
```
hearnadam: I think @ahoy-jon mentioned this would result in `Target =:= Any` always. Is that true?

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -76,7 +80,7 @@ object ZIOs:
         }
     end run
 
-    extension [E, A](exit: Exit[E, A])
+    extension [E, A](exit: zio.Exit[E, A])
```
hearnadam: Can you revert the changes to this file?

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
```diff
@@ -0,0 +1,80 @@
...
+      *   A Kyo's Stream that, when run, will execute the zio.stream.ZStream
+      */
+    def get[E, A](stream: => ZStream[Any, E, A])(using
+        frame: Frame,
+        trace: zio.Trace,
+        tag: kyo.Tag[kyo.Emit[kyo.Chunk.Indexed[A]]],
+        classTag: ClassTag[A]
+    ): Stream[A, Abort[E] & Async] =
+        Stream(Sync {
+            zio.Unsafe.unsafely { unsafe ?=>
```
hearnadam: I'm not sure I see where this is used.

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
```diff
@@ -0,0 +1,80 @@
...
+                val scope = Unsafe.unsafely(Scope.unsafe.make)
+                Resource.run {
+                    Resource.ensure(ex => ZIOs.get(scope.close(ex.fold(Exit.unit)(_.toExit)))).andThen:
```
hearnadam: @fwbrasil this seems like a good candidate for `Scope.locally` or equivalent which would avoid introducing the `Resource` (soon to be Scope) effect.

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
```diff
@@ -0,0 +1,80 @@
...
+                        ZIOs.get(stream.channel.toPullIn(scope)).map: pullIn =>
+                            Loop.foreach:
+                                ZIOs.get(pullIn).map {
+                                    case Right(zioChunk) => Emit.valueWith(Chunk.from(zioChunk.toArray))(Loop.continue)
```
hearnadam: Since `zio.Chunk` extends `IndexedSeq`, you can actually just do `Chunk.from(zioChunk)` - though I am not sure which is more cost, copying the buffers or the indexing cost on the additional layer.

@fwbrasil might have an opinion.

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
```diff
@@ -0,0 +1,80 @@
...
+    /** Lifts a zio.stream.ZStream into a Kyo's Stream.
+      *
+      * @param stream
+      *   The zio.stream.ZStream to lift
+      * @return
+      *   A Kyo's Stream that, when run, will execute the zio.stream.ZStream
+      */
+    def get[E, A](stream: => ZStream[Any, E, A])(using
+        frame: Frame,
+        trace: zio.Trace,
```
hearnadam: Do the implicits need names?

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
```diff
@@ -0,0 +1,80 @@
...
+    def run[A, E](stream: => Stream[A, Abort[E] & Async])(using
+        frame: Frame,
+        trace: zio.Trace,
+        tag: kyo.Tag[kyo.Emit[kyo.Chunk[A]]],
+        classTag: ClassTag[A]
+    ): ZStream[Any, E, A] =
+
+        def loop(emit: Unit < (Emit[Chunk[A]] & Abort[E] & Async)): ZChannel[Any, Any, Any, Any, E, zio.Chunk[A], Any] =
+            ZChannel
+                .fromZIO(ZIOs.run(Emit.runFirst(emit)))
```
hearnadam: I think we may need to rethink this approach. A few possibilities:

- Use a `Channel` to buffer conversion
- Write to the channel in a loop, convert to ZIO, then flatten the stream? `ZIOs.get(Emit.runFirst(...).map(_.fold(ZChannel.unit)(c => ZChannel.write(...) *> suspend)`...

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
```diff
@@ -0,0 +1,80 @@
...
+                .flatMap: (chunkMaybe, cont) =>
+                    chunkMaybe match
+                        case Present(chunk) =>
+                            ZChannel.write(zio.Chunk.fromArray(chunk.toArray)) *>
```
hearnadam: We can use `toArrayInternal`, though i am not sure we should expose that method to outside the `Chunk` class.

hearnadam (commented):
Nice work

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
HollandDM: Yeah, this need a clean up

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
HollandDM: Yeah, I'm also planning on rewrite this, current implementation does not work well (the disable test is failing)

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
HollandDM: I think we can do `private[kyo]` for `toArrayInternal`

kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
ahoy-jon: yep, it doesn't provide a better message! I should maybe let a comment on this block of code. "Tried it, doesn't work"

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
fwbrasil: I think `Chunk.from` is best since we avoid the copy but it still allows efficient indexed access

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
```diff
@@ -0,0 +1,181 @@
...
+
+    /** Interprets a Kyo Stream to ZStream. Note that this method only accepts Abort[E] and Async pending effects. Plase handle any other
+      * effects before calling this method.
+      *
+      * @param v
+      *   The Kyo effect to run
+      * @return
+      *   A ZStream that, when run, will drain the Kyo Stream
+      */
+    def run[A, E](stream: => Stream[A, Abort[E] & Async])(using
```
HollandDM: New implementation for this conversion. Maybe I'll add some comments for this in later to detail the logic. But we need to go this way be cause simply call `ZIOs.run(Emit.runFirst)` will trigger `ensure` chain of `emit`, which in `mapPar*` cases can fail the whole stream

kyo-zio/shared/src/test/scala/kyo/ZStreamsTest.scala
```diff
@@ -0,0 +1,103 @@
...
+            val kyoStream = Stream.init(1 to 10, 2)
+                .concat(Stream.init(Abort.fail(Error)))
+                .concat(Stream.unfold(0)(prev => Maybe.Present(prev -> (prev + 1))))
+            val zioStream = ZStreams.run(kyoStream)
+            zioStream.runDrain.exit.map {
+                case zio.Exit.Failure(cause) if cause.contains(zio.Cause.fail(Error)) => assert(true)
+                case _                                                                => assert(false)
+            }
+        }
+        "convert correctly with parallel kyo's stream" in runZIO {
```
HollandDM: This is flakky locally, sometime it lose elements, sometime it close unexpectedly, will look into it further

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
```diff
@@ -0,0 +1,181 @@
...
+object ZStreams:
+
+    /** Lifts a zio.stream.ZStream into a Kyo's Stream.
+      *
+      * @param stream
+      *   The zio.stream.ZStream to lift
+      * @return
+      *   A Kyo's Stream that, when run, will execute the zio.stream.ZStream
+      */
+    def get[E, A](stream: => ZStream[Any, E, A])(using
```
HollandDM, line 25: Maybe I will need to update this logic to be similar with `ZStreams.run`, will think about this later

kyo-zio/shared/src/test/scala/kyo/ZStreamsTest.scala
hearnadam: Instead of using a `par` operator, you could use something like `buffer` or equivalent to have some async suspensions without a risk of order problems.

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -832,6 +832,8 @@ object Stream:
         new Stream[V, S]:
             def emit: Unit < (Emit[Chunk[V]] & S) = v
 
+    abstract private[kyo] class StreamLike[+V, -S] extends Stream[V, S]
```
hearnadam: I think we can use the same strategy as `Ensure`:
https://github.com/getkyo/kyo/blob/c650a37cd85805fac5c479d1b5823d4377f0a550/kyo-kernel/shared/src/main/scala/kyo/kernel/internal/Safepoint.scala#L137-L139

```
abstract class Stream @publicInBinary private[kyo] ()
```

That will avoid a distinct class hierarchy in the binary.

kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -426,6 +426,8 @@ object `<`:
     private def abortCastUnitImpl[S1: Type, S2: Type](v: Expr[Unit < S1])(using quotes: Quotes): Expr[Unit < S2] =
         import quotes.reflect.*
         val source = TypeRepr.of[S1].show
+        val target = TypeRepr.of[S2].show
+        // Tried printing target, doesn't work since it's always Any
```
hearnadam:
```suggestion
```

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -1,16 +1,20 @@
 package kyo
 
 import kyo.Result.*
+import scala.reflect.ClassTag
```
hearnadam:
```suggestion
```

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -1,16 +1,20 @@
...
 import zio.FiberId
 import zio.Runtime
+import zio.Scope
 import zio.StackTrace
 import zio.Trace
 import zio.Unsafe
 import zio.ZEnvironment
 import zio.ZIO
+import zio.stream.ZChannel
+import zio.stream.ZStream
```
hearnadam:
```suggestion
```

kyo-zio/shared/src/main/scala/kyo/ZIOs.scala
```diff
@@ -1,16 +1,20 @@
 package kyo
 
 import kyo.Result.*
+import scala.reflect.ClassTag
 import zio.Cause
 import zio.Chunk
 import zio.Exit
 import zio.FiberId
 import zio.Runtime
+import zio.Scope
```
hearnadam:
```suggestion
```

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
```diff
@@ -0,0 +1,181 @@
...
+
+        zio.Unsafe.unsafely:
+            import AllowUnsafe.embrace.danger
+
+            val handoff = new java.util.concurrent.atomic.AtomicReference[Handoff](Empty)
+
+            // We only handle happy case here
+            val kyoSide =
+                Loop(stream.emit): emit =>
+                    Emit.runFirst(emit).map: (maybeChunk, nextFn) =>
```
hearnadam: I think it might be better to just use a `handleLoop` and `Channel` for now. We should be able to avoid all unsafe runs.

We can also followup if you would like, to get the `get` method in now, then cut a followup PR

kyo-zio/shared/src/main/scala/kyo/ZStreams.scala
hearnadam: I think this is fine

kyo-prelude/shared/src/main/scala/kyo/Stream.scala
```diff
@@ -44,7 +45,7 @@ import scala.util.NotGiven
   * @see
   *   [[kyo.Poll]] for pull-based consumption with backpressure
   */
-sealed abstract class Stream[+V, -S] extends Serializable:
+abstract class Stream[+V, -S] @publicInBinary private[kyo] () extends Serializable:
```
hearnadam, line 48: this change is not strictly necessary, but it's the pattern we should follow going forward. It will be necessary for the `ZStream.run` method I removed.

ahoy-jon (approved):

PR #1302, opened 2025-06-29
Specialized AsyncShift for List/Set, add codegeneration for TestVariants

ahoy-jon:
following https://github.com/getkyo/kyo/pull/1218 from @HollandDM 

Changing implementation in `direct` to call the corresponding methods in `kyo.Kyo`.

- [x] List
- [ ] ~Map?~
- [x] Set
- [x] Factorize tests
- [x] Clean AsyncShiftTests

--- 

refactor tests in `kyo-kernel` on `Kyo.foreach` using `TestVariant` and code generation.


hearnadam (approved):

project/TestVariant.scala
```diff
@@ -0,0 +1,150 @@
...
+            val zero: State = State(Vector.empty, Mode.Continue)
+        }
+    }
+
+    val generate = Def.task {
+
+        import internal.*
+
+        implicit val log: ManagedLogger = streams.value.log
+        val files: Seq[File]            = (Test / unmanagedSources).value
```
johnhungerford: Can `files` include non-Scala files, and will that break anything?

project/TestVariant.scala
```diff
@@ -0,0 +1,150 @@
...
+            file
+        }
+
+        files.flatMap(file => {
+            val content = IO.read(file, StandardCharsets.UTF_8)
+
+            if (content.contains("@TestVariant")) {
+                log.debug(s"Processing file $file for variants")
+
+                var expecedSize: Option[Int] = None
```
johnhungerford: "expectedSize"

project/TestVariant.scala
```diff
@@ -0,0 +1,150 @@
...
+                        case None => Mode.Continue
+                    }
+
+                    val line: Line = state.mode match {
+                        case Mode.Continue => Line.Raw(str)
+                        case Mode.Replace(testVariant) =>
+                            if (str.contains(testVariant.base))
+                                Line.Variants(testVariant.replacements.map(r => str.replace(testVariant.base, r)))
+                            else {
+                                log.error(s"in $file: \ncannot find '${testVariant.base}' in this line: \n|  $str")
```
johnhungerford: Nice-to-have: since we have the file, it should be possible to log `$path:$lineNumber` so that the user can navigate to the file and line in the IDE. I think this should just require `content.split(...).zipWithIndex.foldLeft(...`?

johnhungerford (approved):
LGTM!

project/TestVariant.scala
ahoy-jon: oups 💤

project/TestVariant.scala
ahoy-jon: Probably, let's check for '*.scala'!

PR #1304, opened 2025-06-30
Kyo.when

### LINKED ISSUE #1297: [feature]: Branching on effectful condition
### Problem Statement

There's currently no way to do `if (x) then y else z` in the case where `x` is a pending boolean (e.g., `Boolean < IO`).

### Proposed Solution

Add an extension method to the `Kyo` companion object: 
```
def when[S](condition: Boolean < S)[A, S1](ifTrue: A < S1, ifFalse: A < S1): A < (S & S1)
```

### Alternative Solutions

Other proposed names: ifKyo, ifEff, ifM

### Current Workaround

If you have an effect `ifTrue` and an effect `ifFalse` and an effect `condition`, you can use the `when` combinator on `A < S` to do the following:
```
ifTrue.when(condition).map:
  case Present(res) => res
  case Absent => ifFalse
```

### Additional Context

See [discussion on discord](https://discord.com/channels/1087005439859904574/1087005440623259681/1388578560670826506)

johnhungerford:
Fixes #1297 

### Problem

There's no easy way to branch effects based on an effectful condition

### Solution

Added `Kyo.when`, which accepts a `condition` effect along with two branch effects, and runs one of them based on the result of the condition.


kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -44,6 +44,20 @@ object Kyo:
...
+      * @param condition
+      *   Effectful boolean condition
+      * @param ifTrue
+      *   Effect to run if condition is true
+      * @param ifFalse
+      *   Effect to run if condition is false
+      * @return
+      *   An effect that runs ifTrue when the condition is evaluated to true, otherwise ifFalse
+      */
+    def when[S](condition: Boolean < S)[A, S1](ifTrue: A < S1, ifFalse: A < S1)(using Frame): A < (S & S1) =
```
hearnadam: Can we make `ifFalse = unit[S]` by default? These should also be by-name parameters.

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -44,6 +44,20 @@ object Kyo:
       */
```
hearnadam, line 44: Can you add `unless` as well?

Do you think these should be extension methods?

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
johnhungerford: That would screw up the types (the result would be `(A | Unit) < (S & S1)`). I'll make another version with one branch with an output type of `Maybe[A] < (S & S1)`.

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
johnhungerford: @fwbrasil suggested putting these on the `Kyo` companion object. I don't have a strong opinion either way.

ahoy-jon: The idea was both? `Kyo.___` and in kyo-combinators `(A < S).___`?

I vote for both!

johnhungerford: We've already got `when` and `unless` in combinators. This PR is just for the `Kyo.__` versions. The question is whether to add them as extensions to `Kyo.type` in kyo-combinators or put them on the actual object.

ahoy-jon: True, forgot about that. In the actual object for `when`, not sure about `unless`.

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -44,6 +44,46 @@ object Kyo:
...
+      * @param condition
+      *   Effectful boolean condition
+      * @param ifTrue
+      *   Effect to run if condition evaluates to true
+      * @param ifFalse
+      *   Effect to run if condition evaluates to false
+      * @return
+      *   An effect that runs ifTrue when the condition is evaluated to true, otherwise ifFalse
+      */
+    def when[S](condition: => Boolean < S)[A, S1](ifTrue: => A < S1, ifFalse: => A < S1)(using Frame): A < (S & S1) =
```
hearnadam: Should `condition` be by name?

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
```diff
@@ -44,6 +44,46 @@ object Kyo:
...
+      * run and [[Absent]] if not.
+      *
+      * @param condition
+      *   Effectful boolean condition
+      * @param ifFalse
+      *   Effect to run if condition evaluates to false
+      * @return
+      *   An effect that runs [[ifFalse]] if [[condition]] evaluates to false
+      */
+    def unless[S](condition: => Boolean < S)[A, S1](ifFalse: => A < S1)(using Frame): Maybe[A] < (S & S1) =
```
hearnadam: Is there a reason there isn't a `ifTrue` overload here?

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
johnhungerford: I guess it doesn't really end up helping at all does it

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
johnhungerford: Because then `unless` just ends up being `when` with the parameters reversed. Do you think it's still needed?

hearnadam: Sure, but it avoids a `map(!_)` call to use `unless`, right?

johnhungerford: No, if you want to invert `Kyo.when(condition)(ifTrue, ifFalse)` you could just do `Kyo.when(condition)(ifFalse, ifTrue)` -- i.e., switch the branches

kyo-kernel/shared/src/main/scala/kyo/Kyo.scala
fwbrasil: It's unnecessary since the thunk is evaluated immediately. I think it's best avoid since it can give the impression that the computation will be suspended.

fwbrasil (approved):
LGTM besides the by-name conditions

PR #1305, opened 2025-06-30
feat: Support generic input for Parse effect

Iltotore:
### Problem
Unlike other parsing libraries such as ZIO Parser, Kyo's `Parse` effect was only made for `Text`/`Char` input and does not support generic inputs. A common use-case are tokens to separate parsing and lexing.

### Solution
Changed `ParseState` to consume `Chunk[In]` where `In` is the type of the tokens the parser consumes. The `Parse` is now also parametized with the input type.

Here is a working example (Scala CLI, put the files in the same directory) with a JSON parser: https://gist.github.com/Iltotore/8d4315e1c668645907f6d678c7c202b3

### Notes
- This PR includes breaking changes since `Parse` and some combinators now require a type parameter.
- Had to do a "dirty trick" with `readAspect` to make it work with generic input. See its definition and `localReadAspect`'s.
- Despite some methods such as `char` being superseded by their generic counterpart (e.g `literal[In](expected: In)`), I currently kept them for backward compatibility and in case we divide `ParseState` into specialized `TextParseState` and another one for generic chunks.
- Should we make a specialized `ParseState` for `Text` or keep the current representation of this PR (stream of `Char`)?

**DRAFT: Some methods and comments might need to be changed before marking this PR as ready. I'd still like to receive feedback on its current state though**


kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -36,18 +36,24 @@ import kyo.kernel.*
...
     /** Aspect that modifies how text is read and parsed. This is the core parsing aspect that other parsing operations build upon. It takes
       * the current text input and a parsing function, allowing for preprocessing of input or postprocessing of results.
       *
       * @return
       *   An Aspect that transforms from Const[Text] to Maybe[(Text, C)]
       */
-    val readAspect: Aspect[Const[Text], [C] =>> Maybe[(Text, C)], Parse] =
-        Aspect.init(using Frame.internal)
+    def readAspect[A]: Aspect[Const[Chunk[A]], [C] =>> Maybe[(Chunk[A], C)], Parse[A]] =
+        localReadAspect.asInstanceOf
```
hearnadam: Please add the type 🙏

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -160,10 +166,10 @@ object Parse:
       * @return
       *   Result from first successful parser, drops the parse branch if none succeed
       */
-    def firstOf[A, S](
-        parser1: => A < (Parse & S),
-        parser2: => A < (Parse & S)
-    )(using Frame): A < (Parse & S) =
+    def firstOf[A, In, S](
```
hearnadam, line 172: Do you think clause interleaving would be helpful for the type annotations?

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -834,9 +875,19 @@ object Parse:
       * @return
       *   Unit if character matches
       */
-    def char(c: Char)(using Frame): Char < Parse =
+    def char(c: Char)(using Frame): Char < Parse[Char] =
+        literal(c) // Could be changed once we specialize Text state out of Chunk state
```
hearnadam: I am not sure I understand the comment. Do you expect changes here?

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -879,22 +951,22 @@ object Parse:
       * @return
       *   Matched text
       */
-    def regex(pattern: String)(using Frame): Text < Parse =
+    def regex(pattern: String)(using Frame): Text < Parse[Char] =
```
hearnadam: pre-existing: I think we should overload this to accept regex

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -879,22 +951,22 @@ object Parse:
...
         Parse.read { s =>
             val regex = pattern.r
-            Maybe.fromOption(regex.findPrefixOf(s.show).map(m => (s.drop(m.length), Text(m))))
+            Maybe.fromOption(regex.findPrefixOf(s.mkString).map(m => (s.drop(m.length), Text(m))))
```
hearnadam: in general, `mkString` is not performant. We should add a specialized operator to Chunk.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -688,6 +717,18 @@ object Parse:
...
+      *
+      * @param input
+      *   Text to parse
+      * @param v
+      *   Parser to run
+      * @return
+      *   Parsed result if successful
+      */
+    def run[A, S](input: Text)(v: A < (Parse[Char] & S))(using Frame): A < (S & Abort[ParseFailed]) =
+        run(Chunk.from(input.toString))(v)
```
hearnadam: Perhaps `Text` can have an optimized conversion to Chunk.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -36,18 +36,24 @@ import kyo.kernel.*
   * @see
   *   [[kyo.Var]], [[kyo.Choice]], [[kyo.Abort]] for the underlying effects that Parse composes
   */
-opaque type Parse <: (Var[Parse.State] & Choice & Abort[ParseFailed]) = Var[Parse.State] & Choice & Abort[ParseFailed]
+opaque type Parse[A] <: (Var[Parse.State[A]] & Choice & Abort[ParseFailed]) =
+    Var[Parse.State[A]] & Choice & Abort[ParseFailed]
```
ahoy-jon, line 40: Thanks a lot for this PR. Maybe there is a trick with two type param, `Parse[-A1, +A2]`, but I am not sure. 

@johnhungerford What do you think?

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
ahoy-jon: it's not related to https://github.com/getkyo/kyo/pull/1305#discussion_r2174999491 ?

That's what I understood.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
ahoy-jon: do you mean?
```scala
def firstOf[In](using Frame)[A, S](...
```

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
Iltotore: I forgot to remove this one but it is related to this bullet:

> Should we make a specialized ParseState for Text or keep the current representation of this PR (chunk of Char)?

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
Iltotore: I don't understand what you mean by "pre-existing". Also do you mean Scala stdlib's [Regex](https://www.scala-lang.org/api/3.x/scala/util/matching/Regex.html), Java's [Pattern](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/regex/Pattern.html) or something else?

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
hearnadam: Yes, though I think `A` would be best?

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
hearnadam: I just meant that this was there before your change. Scala's regex seems reasonable.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
hearnadam: No. `Text` can add `toChunk` potentially, avoiding the intermediate String buffer

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
Iltotore: Are you sure it was there before? I don't see it in the diff :thinking:

Anyway I gonna add it

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
Iltotore: When playing with the changed `Parse`, I found that the type that had most trouble to be infered correctly is the input type `In` rather than `A`.

kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -507,6 +507,19 @@ sealed abstract class Chunk[+A]
             case c: Compact[B] @unchecked => c.array
             case c                        => c.toArray
 
+    /** Converts this [[Chunk]] to a [[String]]. This operation is a more performant version of [[mkString]].
+      *
+      * @return
+      *   a [[String]] of all elements concatenated without any separator
+      */
+    def show: String =
```
hearnadam: Ah, let me followup here. Feel free to keep using `mkString` for now. Sorry for the confusion.

kyo-data/shared/src/main/scala/kyo/Chunk.scala
```diff
@@ -507,6 +507,19 @@ sealed abstract class Chunk[+A]
...
+        val it = iterator
+        val sb = StringBuilder()
+        while it.hasNext do
+            sb.append(it.next())
+        sb.toString
+    end show
+
```
ahoy-jon: @hearnadam is that as the right place?

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -138,10 +145,10 @@ object Parse:
       * @return
       *   Result from the single successful parser, fails if zero or multiple parsers succeed
       */
-    def anyOf[A, S](seq: (A < (Parse & S))*)(using Frame): A < (Parse & S) =
+    def anyOf[A, In, S](seq: (A < (Parse[In] & S))*)(using Frame): A < (Parse[In] & S) =
         Choice.evalWith(seq)(identity)
```
ahoy-jon: same here, clause interleaving?

kyo-data/shared/src/test/scala/kyo/TextTest.scala
```diff
@@ -667,4 +667,9 @@ class TextTest extends Test:
         }
     }
 
+    "toChunk" in {
+        assert(Text("").toChunk == Chunk.empty)
+        assert(Text("123").toChunk == Chunk('1', '2', '3'))
```
hearnadam: Can you add a case for `drop` or another version of Text?

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
hearnadam: I think it's fine if the state is parameterized for now.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
hearnadam: I think we should add it in all the functions (with A, In, S) if we plan to use it.

kyo-data/shared/src/main/scala/kyo/Chunk.scala
Iltotore: So should I remove this change?

Iltotore:
For some reason I cannot directly reply to the following comment so I'll do it here:

> I think it's fine if the state is parameterized for now.

Do you mean that I can remove the comment or also remove the `char` function since it's kind of superseded by `literal`?

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
Iltotore: So something like

```scala
def anyOf[A](seq: (A < (Parse[In] & S))*)[In, S](using Frame): A < (Parse[In] & S)
```

?

ahoy-jon:
```scala
def anyOf[A](using Frame)[In, S](seq: (A < (Parse[In] & S))*): A < (Parse[In] & S)
```

kyo-data/shared/src/main/scala/kyo/Chunk.scala
ahoy-jon: Yes!

ahoy-jon:
a couple of changes, and it's good to go! thanks a lot @Iltotore!

Iltotore:
> both? just keep `literal`?

That's the question actually :sweat_smile:. Should we keep both or just `literal`?

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
Iltotore: I guess I should do the same for every other combinators parametized with `A, In, S` such as `firstOf`, `inOrder`, `skipUntil`...?

If so, how should we handle implicit parameters depending on `In` like:

```scala
def firstOf[A, In, S](
        parser1: => A < (Parse[In] & S),
        parser2: => A < (Parse[In] & S)
    )(using Frame, StateTag[In]): A < (Parse[In] & S)
```

ahoy-jon: like that?

```scala
def firstOf[A](using Frame)[In, S](
        parser1: => A < (Parse[In] & S),
        parser2: => A < (Parse[In] & S)
    )(using StateTag[In]): A < (Parse[In] & S)
```

Iltotore: I interleaved the type parameters for other methods but `firstOf` and `inOrder` seem to have problem with clause interleaving:

```scala
def inOrder[A, B, C](using
        Frame
)[In, S](
    parser1: => A < (Parse[In] & S),
    parser2: => B < (Parse[In] & S),
    parser3: => C < (Parse[In] & S)
)(using
    StateTag[In]
): (A, B, C) < (Parse[In] & S)

//Same for other variants...
```

From ParseTest.scala line 655:
```scala
"consumes input sequentially" in run {
    val parser = Parse.inOrder(
        Parse.literal("a").andThen(1),
        Parse.literal("b").andThen(2),
        Parse.literal("c").andThen(3)
    )
    Parse.run("abc")(parser).map { result =>
        assert(result == (1, 2, 3))
    }
}
```

<details>

<summary>Error message</summary>

```scala
[error] -- [E134] Type Error: /home/fromentin/IdeaProjects/kyo/kyo-prelude/shared/src/test/scala/kyo/ParseTest.scala:655:35 
[error] 655 |                val parser = Parse.inOrder(
[error]     |                             ^^^^^^^^^^^^^
[error]     |None of the overloaded alternatives of method inOrder in object Parse with types
[error]     | [A, B, C, D, E, F, G, H]
[error]     |  (using x$1: kyo.Frame)
[error]     |    [In, S]
[error]     |      (parser1: => A < (kyo.Parse[In] & S),
[error]     |        parser2: => B < (kyo.Parse[In] & S),
[error]     |        parser3: => C < (kyo.Parse[In] & S),
[error]     |        parser4: => D < (kyo.Parse[In] & S),
[error]     |        parser5: => E < (kyo.Parse[In] & S),
[error]     |        parser6: => F < (kyo.Parse[In] & S),
[error]     |        parser7: => G < (kyo.Parse[In] & S), parser8: => H < (kyo.Parse[In] & S)
[error]     |        )
[error]     |        (using x$10: kyo.Parse.StateTag[In]): (A, B, C, D, E, F, G, H) < (
[error]     |          kyo.Parse[In] & S)
[error]     | [A, B, C, D, E, F, G]
[error]     |  (using x$1: kyo.Frame)
[error]     |    [In, S]
[error]     |      (parser1: => A < (kyo.Parse[In] & S),
[error]     |        parser2: => B < (kyo.Parse[In] & S),
[error]     |        parser3: => C < (kyo.Parse[In] & S),
[error]     |        parser4: => D < (kyo.Parse[In] & S),
[error]     |        parser5: => E < (kyo.Parse[In] & S),
[error]     |        parser6: => F < (kyo.Parse[In] & S), parser7: => G < (kyo.Parse[In] & S)
[error]     |        )
[error]     |        (using x$9: kyo.Parse.StateTag[In]): (A, B, C, D, E, F, G) < (
[error]     |          kyo.Parse[In] & S)
[error]     | [A, B, C, D, E, F]
[error]     |  (using x$1: kyo.Frame)
[error]     |    [In, S]
[error]     |      (parser1: => A < (kyo.Parse[In] & S),
[error]     |        parser2: => B < (kyo.Parse[In] & S),
[error]     |        parser3: => C < (kyo.Parse[In] & S),
[error]     |        parser4: => D < (kyo.Parse[In] & S),
[error]     |        parser5: => E < (kyo.Parse[In] & S), parser6: => F < (kyo.Parse[In] & S)
[error]     |        )
[error]     |        (using x$8: kyo.Parse.StateTag[In]): (A, B, C, D, E, F) < (kyo.Parse[In]
[error]     |           & S)
[error]     | [A, B, C, D, E]
[error]     |  (using x$1: kyo.Frame)
[error]     |    [In, S]
[error]     |      (parser1: => A < (kyo.Parse[In] & S),
[error]     |        parser2: => B < (kyo.Parse[In] & S),
[error]     |        parser3: => C < (kyo.Parse[In] & S),
[error]     |        parser4: => D < (kyo.Parse[In] & S), parser5: => E < (kyo.Parse[In] & S)
[error]     |        )
[error]     |        (using x$7: kyo.Parse.StateTag[In]): (A, B, C, D, E) < (kyo.Parse[In] &
[error]     |          S)
[error]     | [A, B, C, D]
[error]     |  (using x$1: kyo.Frame)
[error]     |    [In, S]
[error]     |      (parser1: => A < (kyo.Parse[In] & S),
[error]     |        parser2: => B < (kyo.Parse[In] & S),
[error]     |        parser3: => C < (kyo.Parse[In] & S), parser4: => D < (kyo.Parse[In] & S)
[error]     |        )(using x$6: kyo.Parse.StateTag[In]): (A, B, C, D) < (kyo.Parse[In] & S)
[error]     | [A, B, C]
[error]     |  (using x$1: kyo.Frame)
[error]     |    [In, S]
[error]     |      (parser1: => A < (kyo.Parse[In] & S),
[error]     |        parser2: => B < (kyo.Parse[In] & S), parser3: => C < (kyo.Parse[In] & S)
[error]     |        )(using x$5: kyo.Parse.StateTag[In]): (A, B, C) < (kyo.Parse[In] & S)
[error]     | [A, B]
[error]     |  (using x$1: kyo.Frame)
[error]     |    [In, S]
[error]     |      (parser1: => A < (kyo.Parse[In] & S), parser2: => B < (kyo.Parse[In] & S))
[error]     |        (using x$4: kyo.Parse.StateTag[In]): (A, B) < (kyo.Parse[In] & S)
[error]     |match arguments ((Int < kyo.Parse[Char], Int < kyo.Parse[Char], Int < kyo.Parse[Char]))
[error]     |
[error]     |where:    A is a type variable
[error]     |          B is a type variable
[error]     |          C is a type variable
[error]     |          D is a type variable
[error]     |          E is a type variable
[error]     |          F is a type variable
[error]     |          G is a type variable
[error]     |          H is a type variable
```

</details>

Passing the output types (A, B, C...) explicitly seems to solve the problem but still a worse type inferrence :shrug:

For instance this compiles:

```scala
val parser = Parse.inOrder[Int, Int, Int](
    Parse.literal("a").andThen(1),
    Parse.literal("b").andThen(2),
    Parse.literal("c").andThen(3)
)
```

Note: I did not push the change for `firstOf` and `inOrder` due to this issue. Should we still get along with that and just pass explicitly the type parameters in the test cases?

ahoy-jon:
> > both? just keep `literal`?
> 
> That's the question actually 😅. Should we keep both or just `literal`?

just `literal`

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -58,8 +65,8 @@ object Parse:
       * @return
       *   Parsed value if successful, drops the current parse branch if unsuccessful
       */
-    def read[A](f: Text => Maybe[(Text, A)])(using Frame): A < Parse =
-        Var.use[State] { state =>
+    def read[A, In](f: Chunk[In] => Maybe[(Chunk[In], A)])(using Frame, StateTag[In]): A < Parse[In] =
+        Var.use[State[In]] { state =>
```
ahoy-jon:
```suggestion
    def read[In](using Frame)[A](f: Chunk[In] => Maybe[(Chunk[In], A)])(using StateTag[In]): A < Parse[In] =
        Var.use[State[In]] { state =>
```

but I am not sure about it ...

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -429,17 +455,19 @@ object Parse:
...
-    def inOrder[A, B, C, D, E, F, G, S](
-        parser1: => A < (Parse & S),
-        parser2: => B < (Parse & S),
-        parser3: => C < (Parse & S),
-        parser4: => D < (Parse & S),
-        parser5: => E < (Parse & S),
-        parser6: => F < (Parse & S),
-        parser7: => G < (Parse & S)
+    def inOrder[A, B, C, D, E, F, G, In, S](
+        parser1: => A < (Parse[In] & S),
```
ahoy-jon, line 461: @fwbrasil do you know why `parser1: => A < (Parse[In] & S)` is a by-name? (it's probably just me not getting something about laziness in the last mile)

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -834,9 +880,19 @@ object Parse:
       * @return
       *   Unit if character matches
       */
-    def char(c: Char)(using Frame): Char < Parse =
+    def char(c: Char)(using Frame): Char < Parse[Char] =
+        literal(c) // Could be changed once we specialize Text state out of Chunk state
```
ahoy-jon:
```suggestion
    def char(c: Char)(using Frame): Char < Parse[Char] =
        literal(c)
```
keeping it!

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
johnhungerford: Yeah that's to defer computation in the case of where the user passes an unsuspended effect

Iltotore: An example of what happens if we remove the by-name parameters:

```scala
import kyo.*

enum Expr:
  case Lit(value: Int)
  case Add(left: Expr, right: Expr)

val parseLit: Expr < Parse[Char] = Parse.int.map(Expr.Lit.apply)

def parenthesized: Expr < Parse[Char] = Parse.between(Parse.char('('), parseExpr, Parse.char(')'))

def parseInvokable: Expr < Parse[Char] = Parse.firstOf(parseLit, parenthesized)

def parseAdd: Expr < Parse[Char] = Parse
  .separatedBy(parseInvokable, Parse.char('+'))
  .map(_.reduceLeft(Expr.Add.apply))

def parseExpr: Expr < Parse[Char] = parseAdd
```

```scala
Exception in thread "main" java.lang.StackOverflowError
	at java.base/java.lang.invoke.DirectMethodHandle.allocateInstance(DirectMethodHandle.java:520)
	at kyo.Parse$package$Parse$.literal(Parse.scala:878)
	at kyo.Parse$package$Parse$.char(Parse.scala:58)
	at ExprTest$package$.parenthesized(ExprTest.scala:28)
	at ExprTest$package$.parseInvokable(ExprTest.scala:11)
	at ExprTest$package$.parseAdd(ExprTest.scala:14)
	at ExprTest$package$.parseExpr(ExprTest.scala:17)
	at ExprTest$package$.parenthesized(ExprTest.scala:9)
```

fwbrasil: yes, I added the by-name params to avoid stack overflows in recursive parsers

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -36,18 +37,24 @@ import kyo.kernel.*
...
     /** Aspect that modifies how text is read and parsed. This is the core parsing aspect that other parsing operations build upon. It takes
       * the current text input and a parsing function, allowing for preprocessing of input or postprocessing of results.
       *
       * @return
       *   An Aspect that transforms from Const[Text] to Maybe[(Text, C)]
       */
-    val readAspect: Aspect[Const[Text], [C] =>> Maybe[(Text, C)], Parse] =
-        Aspect.init(using Frame.internal)
+    def readAspect[A]: Aspect[Const[Chunk[A]], [C] =>> Maybe[(Chunk[A], C)], Parse[A]] =
+        localReadAspect.asInstanceOf[Aspect[Const[Chunk[A]], [C] =>> Maybe[(Chunk[A], C)], Parse[A]]]
```
fwbrasil: I think this isn't safe. If there are multiple parsers in a computation handling different input types, aspects will be applied to all regardless of their type, which can produce a class cast exception. We could make `Aspect` be based on tags to distinguish aspects for different input types. A workaround might be putting the aspect in the parse `State` itself but then we need to handle their removal at the end of the scope and composition of nested aspects. Can you open a separate to explore a way to encode this in `Aspect`? I'm afraid we'd need to fix this to merge.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
fwbrasil: yeah, I think it'd be ideal to always have `In` as a single type param group first

fwbrasil:
congrats on your first contribution @Iltotore! this is a good improvement and the json parser is a nice example 🎉

I think the only blocker is the `Aspect` issue. The source file is getting quite large. What do you think about extracting a new generic `Read[A]` effect and then keep `Parse` specific for `Text`? I wonder if we should move them to a separate module as well since we'll likely need to add more features or different kinds of parsers.

fwbrasil:
Yeah, I think we should move it to a new `kyo-parse` module. I suggested separating the generic version as a way to separate the "default" parsing methods specific to different input types so we don't have to deal with a single large file. We could have methods for binary inputs as well for example. Another alternative might be input-type-specific files with extension methods of the `Parse` companion but there's the downside of code completion showing methods for input types that aren't being used.

> Having discussed with @ahoy-jon , at the end of the day it would be good to have a better implementation of Parse. The current Var+Abort combination is similar to other parser combinators libraries but it is not suited to real use cases where you need to correctly handle error cases (emit all parsing errors and not just one, still recover a partial AST...) rather than just the happy path.

Agreed! The way I structured the effect was also meant to validate the pattern using an opaque type that composes other effects. If I remember correctly, it was the first effect using this pattern. Ideally, we should have a proper `ArrowEffect` with richer operations like error accumulation and reporting. It should also provide better performance.

fwbrasil:
I've just posted https://github.com/getkyo/kyo/pull/1327 with a solution for parametrized generic aspects.

> Should we rename already the current Parse?

Do you mean move to a new module? That'd be good but it isn't a requirement to merge.

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
fwbrasil:
```suggestion
    def readAspect[A](using Tag[(Const[Chunk[A]], [C] =>> Maybe[(Chunk[A], C)])]): Aspect[Const[Chunk[A]], [C] =>> Maybe[(Chunk[A], C)], Parse[A]] =
        Aspect.init(using Frame.internal)
```

fwbrasil (approved):
LGTM after the `Aspect` fix. A nice to have would be testing multiple parsers for different input types in the same computation

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -36,17 +37,23 @@ import kyo.kernel.*
...
   */
-opaque type Parse <: (Var[Parse.State] & Choice & Abort[ParseFailed]) = Var[Parse.State] & Choice & Abort[ParseFailed]
+opaque type Parse[A] <: (Var[Parse.State[A]] & Choice & Abort[ParseFailed]) =
+    Var[Parse.State[A]] & Choice & Abort[ParseFailed]
 
 object Parse:
 
+    type StateTag[A] = Tag[Var[State[A]]]
+
+    private val localReadAspect: Aspect[Const[Chunk[Any]], [C] =>> Maybe[(Chunk[?], C)], Parse[Any]] =
```
fwbrasil: remove?

Iltotore: Are you sure about the `StateTag`?

kyo-prelude/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -915,17 +1006,17 @@ object Parse:
...
+    def charNotIn(chars: String)(using Frame): Char < Parse[Char] =
         charIf(c => !chars.contains(c))
 
     /** Succeeds only at end of input
       *
       * @return
       *   Unit if at end of input
       */
-    def end(using Frame): Unit < Parse =
+    def end[In](using Frame, Tag[In], StateTag[In], AspectTag[In]): Unit < Parse[In] =
```
ahoy-jon: @hearnadam do you have why we need `Tag`, `StateTag` and `AspectTag` here?

```scala
    type StateTag[A]  = Tag[Var[State[A]]]
    type AspectTag[A] = Tag[(Chunk[A], Maybe[(Chunk[A], Any)])]
```

fwbrasil:
Sorry for the back and forth here folks! I think this is a good case to bypass the dynamic tag check since we shouldn't be inlining methods that could generate larger trees. I've just posted https://github.com/getkyo/kyo/pull/1389, I think it enables removing `inline` here

Iltotore:
> @fwbrasil updated with master, the same 8 tests are still failing, probably due to Tags on Aspect.
> 
> Will try to dig into it next week.

Isn't it because `readAspect` is a `def`? Looks like the same problem that forced me to do the `asInstanceOf` trick. IIRC, there is a new way to work around this problem (there was a PR I think?) ?

fwbrasil:
I've tried pushing a fix but it didn't work. The main issue is that `Aspect` is key is based on the `Frame` as well so and we need to keep it constant instead of varying based on what method calls `read`. I think we should follow up removing the `Frame` from the key. Here's the fix patch for this PR: https://gist.github.com/fwbrasil/2028dd1d0e237698b75610f6be108bcf

PR #1307, opened 2025-06-30
fixes #1293, add zippable to combinators

### LINKED ISSUE #1293: [feature]: flattening tuple on <&> (<*>)
### Problem Statement

following conversation on discord with @lbialy, 
`a <&> b <&> c`  should probably be `(A, B, C) < S`, instead of `((A, B), C) < S`

ZIO and Besom do that, using things like Zippable: https://github.com/zio/zio/blob/fc30e80b0a14fae778b3342224a9a541e05f2a51/core/shared/src/main/scala/zio/Zippable.scala#L5-L12

### Current Workaround

Use `Async.zip`

ahoy-jon:


ahoy-jon:
Failling FiberTest: https://github.com/getkyo/kyo/actions/runs/15983796477/job/45088274507?pr=1307#step:6:7755

kyo-combinators/shared/src/test/scala/kyo/internal/ZippableTest.scala
```diff
@@ -0,0 +1,33 @@
...
+        assert(Zippable.zip((), 2) == 2)
+        assert(Zippable.zip((1, 2), 3) == (1, 2, 3))
+        assert(Zippable.zip((1, 2), ()) == (1, 2))
+        assert(Zippable.zip((), (3, 4)) == (3, 4))
+        assert(Zippable.zip((1, 2), (3, 4)) == (1, 2, 3, 4))
+    }
+
+    "prepend" in {
+        assert(Zippable.zip(2, (3, 4)) == (2, (3, 4)))
+    }
```
ahoy-jon: note, we can make
```scala
Zippable.zip(2, (3, 4)) == (2, 3, 4)
```

It's a small change, but not implemented usually(as least in ZIO / Besom).

kyo-combinators/shared/src/main/scala/kyo/internal/Zippable.scala
```diff
@@ -0,0 +1,23 @@
+package kyo.internal
+
+trait Zippable[-A, -B]:
```
hearnadam: I think we can use abstract classes for all the traits here.
```suggestion
abstract class Zippable[-A, -B]:
```

kyo-combinators/shared/src/test/scala/kyo/internal/ZippableTest.scala
hearnadam: Why is not usually implemented?

ahoy-jon: it was maybe complicated to implement in Scala2? `ZIO` implement `Zippable[A, Unit]`, `Zippable[Unit, B]`, but not `Besom`.

I will try it quicky.

kyo-combinators/shared/src/test/scala/kyo/internal/ZippableTest.scala
```diff
@@ -0,0 +1,33 @@
...
+    "zip" in {
+        assert(Zippable.zip(1, 2) == (1, 2))
+        assert(Zippable.zip(1, ()) == 1)
+        assert(Zippable.zip((), 2) == 2)
+        assert(Zippable.zip((1, 2), 3) == (1, 2, 3))
+        assert(Zippable.zip((1, 2), ()) == (1, 2))
+        assert(Zippable.zip((), (3, 4)) == (3, 4))
+        assert(Zippable.zip((1, 2), (3, 4)) == (1, 2, 3, 4))
+        assert(Zippable.zip(2, (3, 4)) == (2, 3, 4))
```
ahoy-jon, line 31: @hearnadam done.

```scala
trait ZippableLowPrio0 extends ZippableLowPrio1:
    given prepend[A, B <: Tuple]: (Zippable[A, B] { type Out = A *: B }) = (left: A, right: B) => left *: right
    given append[A <: Tuple, B]: (Zippable[A, B] { type Out = A :* B })  = (left: A, right: B) => left :* right
end ZippableLowPrio0
```

kyo-combinators/shared/src/main/scala/kyo/internal/Zippable.scala
```diff
@@ -0,0 +1,27 @@
...
+    given left[A]: (Zippable[A, Unit] { type Out = A })  = (left: A, _) => left
+    given right[B]: (Zippable[Unit, B] { type Out = B }) = (_, right: B) => right
+
+    given concat[A <: Tuple, B <: Tuple]: (Zippable[A, B] { type Out = A ++ B }) = (left: A, right: B) => left ++ right
+
+    def apply[A, B](using z: Zippable[A, B]): Zippable[A, B]  = z
+    def zip[A, B](a: A, b: B)(using z: Zippable[A, B]): z.Out = z.zip(a, b)
+end Zippable
+
+trait ZippableLowPrio0 extends ZippableLowPrio1:
```
hearnadam: minor, but let's use full names.

kyo-combinators/shared/src/main/scala/kyo/internal/Zippable.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import scala.Tuple.*
+
+abstract class Zippable[-A, -B]:
+    type Out
+    def zip(left: A, right: B): Out
+
+object Zippable extends ZippableLowPrio0:
+
+    given left[A]: (Zippable[A, Unit] { type Out = A })  = (left: A, _) => left
```
hearnadam, line 11: Never actually seen this syntax for this purpose, I assume it avoids the `Aux` type?

kyo-combinators/shared/src/main/scala/kyo/internal/Zippable.scala
```diff
@@ -0,0 +1,27 @@
+package kyo.internal
```
hearnadam, line 1: Should this be `internal`?

hearnadam (approved):

kyo-combinators/shared/src/main/scala/kyo/internal/Zippable.scala
ahoy-jon: and it uses SAM cf:  https://github.com/scala/scala3/pull/20092 it was in the source code of Besom in comments. The original exemple of this PR is ... `Zippable[-A, -B]`

kyo-combinators/shared/src/main/scala/kyo/internal/Zippable.scala
ahoy-jon: It's not a type user should/would use directly, so internal.

PR #1308, opened 2025-06-30
[core] improve IO suspension naming

fwbrasil:
### Problem

Since the renaming of `IO` to `Sync`, its `apply` method doesn't seem as intuitive anymore since it doesn't indicate suspension of side effects. Also, `Async.apply` is confusing because it's meant to suspend side effects but it can give the impression that it'll run the computation concurrently. 

### Solution

Rename the `apply` method of both `Sync` and `Async` to `io` since it's a common naming to indicate side effects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ Kyo also offers an `andThen` method for more fluent code, combining two computat
 import kyo.*
 
 val a: Unit < Sync =
-    Sync(println("hello"))
+    Sync.defer(println("hello"))
 
 val b: String < Sync =
     a.andThen("test")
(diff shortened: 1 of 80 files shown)


ahoy-jon (approved):
I would prefer `Sync.suspend`/`Sync.defer`.

hearnadam:
I don't really like `.io`. Perhaps `succeed`, `suspend`, or `defer`. But `apply` is ok too.

fwbrasil:
@hearnadam the issue with `apply` is that it seems natural to assume that `Async(someCode)` will fork execution. I like `Sync.defer`/`Async.defer`. I can make the change later today if there aren't concerns

fwbrasil:
I've renamed the `io` methods to `defer` according to the offline discussion

PR #1313, opened 2025-07-01
Channel resource safety

### LINKED ISSUE #1168: [feature]: Channel memory safety
### Problem Statement

Channels can easily introduce memory leaks into an application since they maintain queues of suspended fibers. This is especially a danger for frequently used methods that allocate a channel as a short-lived asynchronous utility (e.g., merging streams). If the channel is not closed, any fibers that are suspended at the time that the method returns its value will remain in memory indefinitely. 

### Proposed Solution

I propose that:
- `Channel.init`/`Channel.initWith` use `Resource` to manage channel closure
- Add a `Channel.initLocal`/`Channel.initLocalWith` to provide a channel that is managed more efficiently by `IO.ensure` for use within a single function/process
- Add a `Channel.initUnsafe`/`Channel.initUnsafeWith` to provide a channel that is managed manually (current behavior)

### Alternative Solutions

Alternatively we could keep the current defaults and just add `Channel.initResource`/`Channel.initResourceWith`

### Current Workaround

You can just use `IO.ensure` or `Resource.acquireRelease` to make it safe.

### Additional Context

_No response_

johnhungerford:
Fixes #1168 

### Problem

There is no resource safe `Channel` constructor.

### Solution

- Update `Channel.init`/`Channel.initWith` to use the `Resource` effect, adding a finalizer that calls `channel.close`
- Add `Channel.initLocalWith` to construct a channel, and use `IO.ensure` to guarantee it gets closed after it's used
- Add `Channel.initUnscoped`/`Channel.initUnscopedWith` to construct channels without resource safety

### Notes

A lot of benchmark implementations do not include cleanup. I have kept this implementation, updating the `init` calls to `initUnsafe`. We might consider changing this `init` and using `Resource` to cleanup, as the `ZIO` implementations we are comparing with do include cleanup.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -318,7 +318,7 @@ object Channel:
       * @warning
       *   The actual capacity may be larger than the specified capacity due to rounding.
       */
-    def init[A](capacity: Int, access: Access = Access.MultiProducerMultiConsumer)(using Frame): Channel[A] < Sync =
+    def init[A](capacity: Int, access: Access = Access.MultiProducerMultiConsumer)(using Frame): Channel[A] < (Sync & Resource) =
         initWith[A](capacity, access)(identity)
 
     /** Uses a new Channel with the provided configuration.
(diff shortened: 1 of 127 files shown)


kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -329,8 +329,60 @@ object Channel:
       */
     inline def initWith[A](capacity: Int, access: Access = Access.MultiProducerMultiConsumer)[B, S](
         inline f: Channel[A] => B < S
+    )(using inline frame: Frame): B < (S & Sync & Resource) =
+        Sync.Unsafe:
+            val channel = Unsafe.init[A](capacity, access)
+            Resource.ensure(Channel.close(channel)).andThen:
```
hearnadam, line 335: I assume this wasn't selecting because you forgot to write `.safe`? I suppose this is better anyways.

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -329,8 +329,60 @@ object Channel:
...
+                f(channel)
+
+    /** Uses a new Channel with the provided configuration, closing the channel after usage.
+      * @param f
+      *   The function to apply to the new Channel
+      * @return
+      *   The result of applying the function
+      */
+    inline def initLocalWith[A](capacity: Int, access: Access = Access.MultiProducerMultiConsumer)[B, S](
```
hearnadam: I'm hesitant to introduce this method because `B =:= Channel[A]` is unsafe... perhaps it's not a risk, but I could see myself letting it accidentally escape the scope.

fwbrasil: I think it's something good to document but resources escaping is something that can happen with `Resource` as well. Capture checking might enable a stronger guarantee

fwbrasil: I don't like `local` for the name since we have `Local` but I can't think of a better name. Maybe something like `initLocallyScoped`?

johnhungerford: I agree that `initLocalWith` is confusing. `initLocallyScoped` is better but too long. How about `use`? I was hoping to have all constructors have `init` prefix, but `use` seems like a good idiomatic way to describe using within a bounded scope. `initUsing` maybe?

@hearnadam @ahoy-jon want to weigh in here? We need a good standard set of names for:
1. Default constructor that introduces `Resource` (currently `init`/`initWith`
2. Constructor with no resource safety (currently `initUnscoped`/`initUnscopedWith`
3. Constructor for a bounded scope, i.e. cleans up via `Sync.ensure` after callback completes:
    - `initLocallyScoped`?
    - `initUsing`?
    - `use`?
    - ???

johnhungerford: - `initBounded`?
- `initScoped`? (not great if we're changing `Resource` to `Scope`...)

johnhungerford: Moved naming discussion to new thread.

I agree this is a little unsafe, but as @fwbrasil pointed out, you end up with the same issue with `Resource` as well. We'd need capture checking to solve this problem! I do think it's an important method, since there a fair number of use cases requiring a channel or other similar resourceful primitives within just the scope of a function. I think it makes sense to provide a public way to accomplish this without the overhead of `Resource` and with greater safety than if users were forced to resort to the unsafe API.

ahoy-jon:
`Channel.use` looks good to me

kyo-core/shared/src/main/scala/kyo/Channel.scala
johnhungerford: There was actually a different problem was preventing me from running `.close` even when prefixed with `.safe` and I forgot change back.

Issue #1224, opened 2025-05-31
[feature]: Strengthen guarantees of Resource/bracket


calvinlfer:
### Problem Statement

There is an edge case that can occur in some circumstances between acquiring a resource and registering a finalizer - an interruption occurs causing the resource to be acquired but the finalizer not being registered 

### Proposed Solution

@fwbrasil mentioned that the way finalizers are added to the safe-point may need to be reworked. Instead of just a finalizer thunk, something more structured representing the acquisition and finalizer

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/1988

PR #1315, opened 2025-07-02
introduce Scope + ensureMap

fwbrasil:
Fixes https://github.com/getkyo/kyo/issues/1224

--- 

This PR reworks resource-handling abstraction and improves safety of finalizer registration.

| **Change**                  | **Description**                                                                 |
|----------------------------|---------------------------------------------------------------------------------|
| `Resource` to `Scope`     |  (`opaque type Scope <: Sync`) replacing `Resource`; |
| Finalizer registration  | `ensureMap` avoids races by registering finalizers synchronously after acquisition. |
| Interrupt safety       | Prevents interruption between acquisition and finalizer registration — not previously guaranteed. |
| `awaitClose` flag       | Allows finalizers to run in background (no backpressure) or wait for completion. |
| `runLocally` method     | Isolated execution without propagating outer finalizers; renamed from `runIsolated`. |
| *est additions         | Adds hierarchical, interrupt, and isolation behavior tests for `Scope`. |

**Impact:**  
- Safer, more flexible resource lifecycle management.  
- Cleaner semantics for nested and interruptible scopes.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Scope.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Scope.scala
@@ -196,7 +196,7 @@ object Scope:
             case Absent             => bug("Missing finalizer from context")
         }
 
-    given Isolate.Contextual[Scope, Sync] = Isolate.Contextual.derive[Sync, Sync]
+    given isolate: Isolate[Scope, Sync, Any] = Isolate.derive[Sync, Sync, Any]
 
     /** Represents a finalizer for a scope. */
     opaque type Finalizer = Finalizer.Unsafe
@@ -229,7 +229,7 @@ object Scope:
             import AllowUnsafe.embrace.danger
 
             private val queue   = Queue.Unbounded.Unsafe.init[Callback](Access.MultiProducerSingleConsumer)
-            private val promise = Promise.Unsafe.init[Nothing, Unit]()
+            private val promise = Promise.Unsafe.init[Unit, Any]()
 
             private[Scope] def tryEnsure(callback: Callback): Unit =
                 discard(queue.offer(callback))
@@ -247,18 +247,18 @@ object Scope:
                     case Absent => ()
                     case Present(tasks) =>
                         if tasks.isEmpty then
-                            promise.completeDiscard(Result.unit)
+                            promise.completeDiscard(Result.succeed(()))
                         else
                             Sync.Unsafe.evalOrThrow(
                                 Async.foreachDiscard(tasks, parallelism) { task =>
                                     Abort.run[Throwable](task(ex))
                                         .map(_.foldError(_ => (), ex => Log.error("Scope finalizer failed", ex.exception)))
                                 }
-                                    .handle(Fiber.init[Nothing, Unit, Any])
+                                    .handle(Fiber.init[Nothing, Unit, Any, Any])
                                     .map(promise.safe.becomeDiscard)
                             )
 
-            def await()(using AllowUnsafe): Fiber.Unsafe[Nothing, Unit] =
+            def await()(using AllowUnsafe): Fiber.Unsafe[Unit, Any] =
                 promise
 
             def safe: Finalizer = this
--- a/kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
+++ b/kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
@@ -22,9 +22,9 @@ extension [A, E, S](effect: A < (Abort[E] & Async & S))
     ): Fiber[A, reduce.SReduced & S2] < (Sync & S) =
         Fiber.init(effect)
 
-    /** Forks this computation using the Async effect and returns its result as a `Fiber[A, Abort[E]]`, managed by the Scope effect. Unlike `fork`,
-      * which creates an unmanaged fiber, `forkScoped` ensures that the fiber is properly cleaned up when the enclosing scope is closed,
-      * preventing resource leaks.
+    /** Forks this computation using the Async effect and returns its result as a `Fiber[A, Abort[E]]`, managed by the Scope effect. Unlike
+      * `fork`, which creates an unmanaged fiber, `forkScoped` ensures that the fiber is properly cleaned up when the enclosing scope is
+      * closed, preventing resource leaks.
       *
       * @return
       *   A computation that produces the result of this computation with Async and Scope effects
--- a/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
@@ -752,7 +752,7 @@ class AsyncTest extends Test:
 
             Memo.run {
                 Kyo.zip(
-                    Async.collectAll(
+                    Async.collectAll(using Memo.isolate)(
                         Seq(
                             f(1),
                             f(1),
--- a/kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
+++ b/kyo-core/shared/src/test/scala/kyo/ChannelTest.scala
@@ -374,13 +374,13 @@ class ChannelTest extends Test:
             yield assert(result.sorted == Chunk(1, 2, 3) && finalSize == 0)
             end for
         }
-        "race with close" in run {
+        "race with close" in runNotJS {
             verifyRaceDrainWithClose(2, _.drain, _.close)
         }
-        "race with closeAwaitEmpty" in run {
+        "race with closeAwaitEmpty" in runNotJS {
             verifyRaceDrainWithClose(2, _.drain, _.closeAwaitEmpty)
         }
-        "race with close and zero capacity" in run {
+        "race with close and zero capacity" in runNotJS {
             verifyRaceDrainWithClose(2, _.drain, _.close)
         }
     }
@@ -450,13 +450,13 @@ class ChannelTest extends Test:
                 finalSize <- c.size
             yield assert(result.size == 3 && finalSize == 0)
         }
-        "race with close" in run {
+        "race with close" in runNotJS {
             verifyRaceDrainWithClose(2, _.drainUpTo(2), _.close)
         }
-        "race with closeAwaitEmpty" in run {
+        "race with closeAwaitEmpty" in runNotJS {
             verifyRaceDrainWithClose(2, _.drainUpTo(2), _.closeAwaitEmpty)
         }
-        "race with close and zero capacity" in run {
+        "race with close and zero capacity" in runNotJS {
             verifyRaceDrainWithClose(0, _.drainUpTo(Int.MaxValue), _.close)
         }
     }
--- a/kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
@@ -77,7 +77,7 @@ import scala.quoted.*
   * @tparam Restore
   *   Effects that become available after isolation completes
   */
-abstract class Isolate[Remove, -Keep, Restore]:
+abstract class Isolate[Remove, -Keep, -Restore]:
     self =>
 
     /** The type of state being managed */
--- a/kyo-stm/shared/src/test/scala/kyo/STMTest.scala
+++ b/kyo-stm/shared/src/test/scala/kyo/STMTest.scala
@@ -250,7 +250,7 @@ class STMTest extends Test:
             Memo.run {
                 for
                     ref <- TRef.init(1)
-                    result <- STM.run {
+                    result <- STM.run(using Memo.isolate) {
                         for
                             _      <- ref.set(2)
                             v1     <- f(2)


kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,126 @@
+package kyo
+
+import kyo.Result.Error
+import kyo.Result.Panic
+import kyo.Tag
+import kyo.kernel.ContextEffect
+
+opaque type Scope = Sync
```
fwbrasil: This is planned to be the replacement for `Resource`. Instead of using `ContextEffect`, which has more strict type-level tracking, I'm using an opaque type + `Local`. The effect tracking is meant to ensure the `Local` is set in a scope to allow `runFinalizer` to dynamically check if there's an outer `Scope` finalizer. It's similar to how `STM` handles transactions.

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,126 @@
...
+    def ensure(v: => Any < (Async & Abort[Throwable]))(using Frame): Unit < Scope =
+        ensure(_ => v)
+
+    def ensure(callback: Maybe[Error[Any]] => Any < (Async & Abort[Throwable]))(using Frame): Unit < Scope =
+        withFinalizerUnsafe(f => Abort.get(f.ensure(callback)))
+
+    def acquireRelease[A, S](acquire: => A < S)(release: A => Any < (Async & Abort[Throwable]))(using Frame): A < (Scope & S) =
+        withFinalizerUnsafe { finalizer =>
+            acquire.ensureMap { resource =>
+                val result = finalizer.ensure(_ => release(resource))
```
fwbrasil: `ensureMap` avoids potential interruptions after `acquire` finishes. Paired with an unsafe `finalizer.ensure`, there's no possibility of interruption between the acquisition finishing and the finalizer callback being registered.

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,126 @@
...
+        Sync.Unsafe(run(new Finalizer.Unsafe, closeParallelism, awaitClose)(v))
+
+    def runFinalizer[A, S](closeParallelism: Int, awaitClose: Boolean = true)(v: Finalizer => A < (Scope & S))(
+        using Frame
+    ): A < (Async & S) =
+        Sync.withLocal(local) { parent =>
+            import AllowUnsafe.embrace.danger
+            val child = new Finalizer.Unsafe
+            parent.foreach(p => discard(p.ensure(child.close(_, closeParallelism))))
+            run(child, closeParallelism, awaitClose)(v(child))
```
fwbrasil: @hearnadam does this cover the scenario you were exploring? The finalizer will be registered with the parent to ensure closing if the parent is closed

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+    private def run[A, S](finalizer: Finalizer, closeParallelism: Int, awaitClose: Boolean)(v: A < (Scope & S))(
```
fwbrasil: `awaitClose` is a new feature and indicates if the execution should wait for the finalizer to close. If `false`, the closing happens in the background without backpressure. We could have separate apis returning a pending `Sync` instead of `Async` when it's `false` but it doesn't seem worth it.

kyo-core/shared/src/test/scala/kyo/ScopeTest.scala
```diff
@@ -0,0 +1,449 @@
...
+            io.handle(
+                Scope.run,
+                Abort.run
+            ).map { _ =>
+                assert(recoveryAction == "IllegalState")
+            }
+        }
+    }
+
+    "can't be interrupted without registering the finalizer" in run {
```
fwbrasil: this isn't supported by the current `Resource`

kyo-core/shared/src/main/scala/kyo/Scope.scala
fwbrasil: Actually, it shouldn't set the local so the resources are isolated. Updating

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,133 @@
...
+            def ensure(callback: Callback)(using Frame): Unit < Sync =
+                Sync.Unsafe(Abort.get(self.ensure(callback)))
+
+            def unsafe: Unsafe = self
+
+        end extension
+
+        type Callback = Maybe[Error[Any]] => Any < (Async & Abort[Throwable])
+
+        final class Unsafe(using frame: Frame, allow: AllowUnsafe):
```
hearnadam: minor: I would make the AllowUnsafe an inline given in the body.

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,133 @@
...
+
+object Scope:
+
+    private val local = Local.init(Maybe.empty[Finalizer])
+
+    def ensure(v: => Any < (Async & Abort[Throwable]))(using Frame): Unit < Scope =
+        ensure(_ => v)
+
+    def ensure(callback: Maybe[Error[Any]] => Any < (Async & Abort[Throwable]))(using Frame): Unit < Scope =
+        withFinalizerUnsafe(f => Abort.get(f.ensure(callback)))
```
hearnadam: Is this not selecting the extension method on the safe variant?  Can you use `Finalizer.ensure(f)(callback)`?

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,133 @@
...
+
+            private val queue   = Queue.Unbounded.Unsafe.init[Callback](Access.MultiProducerSingleConsumer)
+            private val promise = Promise.Unsafe.init[Nothing, Unit]()
+
+            def ensure(callback: Callback)(using AllowUnsafe): Result[Nothing, Unit] =
+                if !queue.offer(callback).contains(true) then
+                    Result.panic(new Closed(
+                        "Finalizer",
+                        frame,
+                        "This finalizer is already closed. This may happen if a background fiber escapes the scope of a 'Resource.run' call."
```
hearnadam:
```suggestion
                        "This finalizer is already closed. This may happen if a background fiber escapes the scope of a 'Scope.run' call."
```

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,133 @@
...
+                queue.close() match
+                    case Absent => ()
+                    case Present(tasks) =>
+                        if tasks.isEmpty then
+                            promise.completeDiscard(Result.unit)
+                        else
+                            Sync.Unsafe.evalOrThrow(
+                                Async.foreachDiscard(tasks, parallelism) { task =>
+                                    Abort.run[Throwable](task(ex))
+                                        .map(_.foldError(_ => (), ex => Log.error("Resource finalizer failed", ex.exception)))
```
hearnadam:
```suggestion
                                        .map(_.foldError(_ => (), ex => Log.error("Scope finalizer failed", ex.exception)))
```

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,133 @@
...
+            val finalizer = new Finalizer.Unsafe
+            runAndClose(finalizer, closeParallelism, awaitClose) {
+                local.let(Present(finalizer))(v)
+            }
+        }
+
+    def runIsolated[A, S](f: Finalizer => A < (Scope & S))(using Frame): A < (Async & S) =
+        runIsolated(1)(f)
+
+    def runIsolated[A, S](closeParallelism: Int, awaitClose: Boolean = true)(f: Finalizer => A < (Scope & S))(
```
hearnadam: I don't think the name Isolated is helpful to me.

Perhaps `runLocally`?

kyo-core/shared/src/main/scala/kyo/Scope.scala
fwbrasil: I've added an import since it's simpler and I think it's equivalent to an inline given

kyo-core/shared/src/main/scala/kyo/Scope.scala
fwbrasil: good catch, I've added `f.unsafe.ensure` to select the unsafe version since there's no need for an additional suspension

fwbrasil: actually it isn't necessary. `Unsafe` is a class, not an `opaque type` so it's not ambiguous

kyo-core/shared/src/main/scala/kyo/Scope.scala
fwbrasil: updated

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,269 @@
...
+      * @param acquire
+      *   The effect to acquire the resource.
+      * @param release
+      *   The function to release the acquired resource.
+      * @param frame
+      *   The implicit Frame for context.
+      * @return
+      *   The acquired resource wrapped in Scope and S effects.
+      */
+    def acquireRelease[A, S](acquire: => A < S)(release: A => Any < (Async & Abort[Throwable]))(using Frame): A < (Scope & S) =
```
ahoy-jon, line 86: I would have used `A < (Scope & S & Sync)` instead of `A < (Scope & S)` 🤔

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,269 @@
...
+  *   [[kyo.Scope.acquireRelease]] For creating resources with custom acquire and release functions
+  * @see
+  *   [[kyo.Scope.acquire]] For creating resources from Java Closeables
+  * @see
+  *   [[kyo.Scope.ensure]] For registering cleanup actions
+  * @see
+  *   [[kyo.Scope.run]] For executing resource-managed computations
+  */
+opaque type Scope <: Sync = Sync
+
```
ahoy-jon, line 42:
```suggestion

@deprecated("use `Scope`", "1.0-RC")
type Resource = Scope

@deprecated("use `Scope`", "1.0-RC")
val Resource = Scope

```

I think we could remove them after the 1.0

kyo-core/shared/src/main/scala/kyo/Scope.scala
ahoy-jon: Ok, only for `Kyo. acquireRelease`!

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,269 @@
...
+  * @see
+  *   [[kyo.Scope.acquireRelease]] For creating resources with custom acquire and release functions
+  * @see
+  *   [[kyo.Scope.acquire]] For creating resources from Java Closeables
+  * @see
+  *   [[kyo.Scope.ensure]] For registering cleanup actions
+  * @see
+  *   [[kyo.Scope.run]] For executing resource-managed computations
+  */
+opaque type Scope <: Sync = Sync
```
hearnadam, line 41: Will this result in `Scope & Sync` becoming `Sync`? I am not sure that's desirable for users.

kyo-core/shared/src/main/scala/kyo/Scope.scala
hearnadam: That's fine, but can you note that in the message?

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,269 @@
...
+                finalizer.close(result.error, closeParallelism)
+                Kyo.when(awaitClose)(finalizer.await().safe.get)
+                    .andThen(Abort.get(result.asInstanceOf[Result[Nothing, A]]))
+            }
+        }
+
+    private def withFinalizerUnsafe[A, S](f: AllowUnsafe ?=> Finalizer => A < S)(using Frame): A < (Scope & S) =
+        local.use {
+            case Present(finalizer) => f(using AllowUnsafe.embrace.danger)(finalizer)
+            case Absent             => bug("Missing finalizer from context")
```
hearnadam, line 196: I do wonder if this is the correct approach, but I suppose type level tracking is all that's needed.

kyo-core/shared/src/test/scala/kyo/ScopeTest.scala
```diff
@@ -0,0 +1,651 @@
...
+            io.handle(
+                Scope.run,
+                Abort.run
+            ).map { _ =>
+                assert(recoveryAction == "IllegalState")
+            }
+        }
+    }
+
+    "hierarchical behavior" - {
```
fwbrasil, line 425: new tests from here

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,269 @@
...
+                            Sync.Unsafe.evalOrThrow(
+                                Async.foreachDiscard(tasks, parallelism) { task =>
+                                    Abort.run[Throwable](task(ex))
+                                        .map(_.foldError(_ => (), ex => Log.error("Scope finalizer failed", ex.exception)))
+                                }
+                                    .handle(Fiber.init[Nothing, Unit, Any])
+                                    .map(promise.safe.becomeDiscard)
+                            )
+
+            def await()(using AllowUnsafe): Fiber.Unsafe[Nothing, Unit] =
```
hearnadam: Maybe this should be `fiber`?

kyo-core/shared/src/test/scala/kyo/ScopeTest.scala
```diff
@@ -0,0 +1,651 @@
...
+        "resource leak detection on fiber escape" in run {
+            var leakedResourceUsed = false
+            val resource           = TestResource(1)
+
+            for
+                l <- Latch.init(1)
+                escapedFiber <- Scope.run {
+                    for
+                        r <- Scope.acquire(resource)
+                        fiber <- Fiber.init {
```
hearnadam, line 533: I think this is a bit awkward (and pre-existing). I am not sure we should have closed as a result here, but see if we can just fire the finalizer immediately. Or forking should by default create a child scope that's closed when the parent is.

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -0,0 +1,269 @@
...
+      * @param awaitClose
+      *   Whether to wait for all finalizers to complete before returning the result.
+      * @param f
+      *   The function that receives a finalizer and returns the effect to run.
+      * @param frame
+      *   The implicit Frame for context.
+      * @return
+      *   The result of the effect wrapped in Async and S effects.
+      */
+    def runLocally[A, S](closeParallelism: Int, awaitClose: Boolean = true)(f: Finalizer => A < (Scope & S))(
```
hearnadam, line 176: I don't think this should handle `Scope` in it's current state - it doesn't set the Finalizer to `Present`, right?

kyo-core/shared/src/test/scala/kyo/ScopeTest.scala
```diff
@@ -0,0 +1,651 @@
...
+                result <- escapedFiber.getResult
+            yield
+                assert(resource.closes == 1)
+                assert(!leakedResourceUsed)
+                assert(result.panic.exists(_.isInstanceOf[Closed]))
+            end for
+        }
+    }
+
+    "runLocally" - {
```
hearnadam, line 551: Can you add a test that stacks `runLocally` instances? I also think we need to use `finalizer.ensure` in some tests.

kyo-core/shared/src/main/scala/kyo/Scope.scala
fwbrasil: no, just that Scope can subsume Sync. Sync can't subsume Scope

kyo-core/shared/src/main/scala/kyo/Scope.scala
fwbrasil: this should never happen given the type-level tracking. STM does something similar

kyo-core/shared/src/main/scala/kyo/Scope.scala
fwbrasil: the unsafe apis normally return unsafe types?

kyo-core/shared/src/test/scala/kyo/ScopeTest.scala
fwbrasil: it's actually something I implemented (see `Scope.internal.tryEnsure`). If the outer scope is already closed, I don't see a reason to fail. For example some daemon fiber that forks others. Note that the scope always propagates

kyo-core/shared/src/main/scala/kyo/Scope.scala
fwbrasil: good catch

ahoy-jon:
I think we can close it, that will be done in the next kernel?

fwbrasil:
yes, it'll be addressed by the new kernel

PR #1318, opened 2025-07-02
fixes #1317: cast A => B to A => B < Any

### LINKED ISSUE #1317: [performance] `A => B` can be casted directly into `A => B < Any`
following https://github.com/getkyo/kyo/pull/1314:

This can be just a cast, without creating a closure.
https://github.com/getkyo/kyo/blob/eec7cf6c428d0eb76a4bc7ab7cdf955ebcf98aee/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala#L438-L471

ahoy-jon:


ahoy-jon:
@fwbrasil do you see a case where that would be incorrect?

ahoy-jon:
- [ ] test with: https://github.com/getkyo/kyo/pull/1300

kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -435,40 +435,32 @@ object `<`:
     end abortCastUnitImpl
 
     /** Converts a pure single-argument function to an effectful computation. */
-    implicit inline def liftPureFunction1[A1, B](inline f: A1 => B)(
-        using inline flat: CanLift[B]
-    ): A1 => B < Any =
-        a1 => f(a1)
+    implicit inline def liftPureFunction1[A1, B](inline f: A1 => B)(using inline canLift: CanLift[B]): A1 => B < Any = f.asInstanceOf
```
fwbrasil, line 438: This is incorrect. Even if `B` isn't statically known as a Kyo computation via `CanLift`, it may still be a computation in generic scopes and we need to ensure the `Nested` wrapper is instantiated by reflecting at runtime

ahoy-jon: @fwbrasil ok, can we make a test for that? there is a way to optimize it (same as lift).

ahoy-jon: This example doesn't run, even in `main`.

```scala
        "nested effect suspension as function" in {
            def f(str:String): Int < TestEffect2 = TestEffect2(str)
            
            def g[B:WeakFlat](f: String => B):B < TestEffect1 = { 
                val g: String => B < Any = f
                TestEffect1(1).map(g)
            } //(<.liftPureFunction1(f))
            
            inline def h[B](f: String => B): B < TestEffect1 = g(f)
            
            val nested: Int < TestEffect2 < TestEffect1 = h(f)

            val result = TestEffect1.run(nested.map(TestEffect2.run))
            assert(result.eval == 19)

            val result2 = TestEffect1.run(TestEffect2.run(nested.flatten))
            assert(result2.eval == 19)
        }

```

fwbrasil: do you mean it doesn't compile?  I think you'd need to remove the `WeakFlat` constraint from `g` and also not handle both effects before calling `eval`. That should trigger this branch: https://github.com/getkyo/kyo/blob/803e8420a8563500e39b3e7fdbf708af1b48fd06/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala#L389

ahoy-jon: `h` is removing the `WeakFlat` on `g` ... it does compile, it doesn't run.

```
BUG Unexpected pending effect while handling scala.Any: Kyo(kyo.kernel.PendingTest._$TestEffect2, Input(Effect1:1), PendingTest.scala:327:49, cont(s"Effect1:$input"))) Please open an issue 🥹 https://github.com/getkyo/kyo/issues
kyo.bug$KyoBugException: BUG Unexpected pending effect while handling scala.Any: Kyo(kyo.kernel.PendingTest._$TestEffect2, Input(Effect1:1), PendingTest.scala:327:49, cont(s"Effect1:$input"))) Please open an issue 🥹 https://github.com/getkyo/kyo/issues
```

Tried it on `main` before the work on `CanLift`/`lift`.

ahoy-jon: `h`was not necessary, so the test on `main` is:
```scala
"nested effect suspension as function" in {
            def f(str:String): Int < TestEffect2 = TestEffect2(str)
            
            def g[B](f: String => B):B < TestEffect1 = { 
                val liftedF: String => B < Any = f
                TestEffect1(1).map(liftedF)
            }
            
            val nested: Int < TestEffect2 < TestEffect1 = g(f)

            val result = TestEffect1.run(nested.map(TestEffect2.run))
            assert(result.eval == 19)

            val result2 = TestEffect1.run(TestEffect2.run(nested.flatten))
            assert(result2.eval == 19)
        }
```

fwbrasil: does it pass or fail on main?

ahoy-jon: it fails, sadly! If it passes, I would have made a PR to had it, and then restart some work to optimize function lift.

fwbrasil: that'd be a bug, can you open an issue? Feel free to add the test marked with `pendingUntilFixed` if you like as well.

PR #1319, opened 2025-07-02
Combinators house-keeping

johnhungerford:
### Problem

kyo-combinators is overdue to some cleanup. A few issues that I found:

- `Reducible` used unnecessarily to eliminate `Abort[Nothing]` in the context of `Sync`/`Async`. (I hadn't understood that `Abort[Nothing]` is included in `Sync` when I first implemented these.
- Some methods and tests were in the wrong file
- Some test names were out of date ("fibers" instead of "async")
- `Kyo.async` did not support errors
- Have `Kyo.emit` but not `Kyo.poll`
- No need for `Kyo.traverse` alias in place of `Kyo.collectAll` (in general we're moving to greater consistency between combinators and base API)
- `Kyo.suspend` should be `Kyo.defer` (in general we're moving to greater consistency between combinators and base API)
- `retry`/`retryForever` after `forAbort` probably should not retry on panic, since it is designed to target a specific failure

### Solution

Updates are straightforward

### DIFF
--- a/README.md
+++ b/README.md
@@ -1386,7 +1386,7 @@ import kyo.*
 
 case class Config(someConfig: String)
 
-val original: Stream[Int, Resource & Env[Config] & Abort[String] & Async] = Stream(1)
+val original: Stream[Int, Resource & Env[Config] & Abort[String] & Async] = Stream.init(Seq(1))
 
 val handled: Stream[Int, Async] = original.handle(
     Resource.run(_),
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3524,8 +3524,6 @@ Being more modular than ZIO, Kyo separates effect constructors in the companion
 1. Factory methods on the `Kyo` object, styled after those found on `ZIO`, for many of the core Kyo effect types.
 2. Extension methods on Kyo effects modeled similarly to ZIO's methods.
 
-Generally speaking, the names of `kyo-combinators` methods are the same as the corresponding methods in ZIO. When this is not possible or doesn't make sense, `kyo-combinators` tries to keep close to ZIO conventions.
-
```
johnhungerford, line 3528: I don't think this holds true anymore

ahoy-jon: or close to. Indeed, doesn't seem to be true anymore.

PR #1323, opened 2025-07-03
Kyo-rules

ahoy-jon:


.scalafmt.conf
```diff
@@ -17,6 +17,9 @@ indent.main = 4
 indent.callSite = 4
 newlines.source = keep
 fileOverride {
+  "glob:**/scalafix/rules/**" {
+    runner.dialect = scala212
```
hearnadam: ?
```suggestion
    runner.dialect = scala213
```

build.sbt
```diff
@@ -721,3 +721,50 @@ def mimaCheck(failOnProblem: Boolean) =
...
+        scalaVersion                           := scalaFixScalaVersion
+    )
+
+lazy val `kyo-scalafix-input` = (project in file("scalafix/input"))
+    .settings(
+        publish / skip                     := true,
+        scalaVersion                       := scala3Version,
+        semanticdbEnabled                  := true,
+        semanticdbVersion                  := scalafixSemanticdb.revision,
+        libraryDependencies += "io.getkyo" %% "kyo-direct" % "0.19.0"
```
hearnadam, line 748: Is there a way to support multiple versions in a single module? or do we just write over the previous versions?

scalafix/rules/src/main/scala/kyo/rules/KyoUpdateToV1_0.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import scala.meta._
+import scalafix.v1._
+
+class KyoUpdateToV1_0 extends SemanticRule("KyoUpdateToV1_0") {
+
+    override def fix(implicit doc: SemanticDocument): Patch = {
+        println("Tree.syntax: " + doc.tree.syntax)
+        println("Tree.structure: " + doc.tree.structure)
+        println("Tree.structureLabeled: " + doc.tree.structureLabeled)
```
hearnadam: leftover?

build.sbt
```diff
@@ -721,3 +721,50 @@ def mimaCheck(failOnProblem: Boolean) =
...
+
+lazy val V = _root_.scalafix.sbt.BuildInfo
+
+lazy val scalaFixScalaVersion = V.scala213
+
+lazy val `kyo-scalafix` = (project in file("scalafix"))
+    .aggregate(`kyo-rules`, `kyo-scalafix-input`, `kyo-scalafix-output`, `kyo-scalafix-test`)
+    .settings(publish / skip := true)
+
+lazy val `kyo-rules` = (project in file("scalafix/rules"))
```
hearnadam, line 735: For the publishing of this as a binary, do we just use the same version as the rest?

scalafix/rules/src/main/scala/kyo/rules/KyoUpdateToV1_0.scala
```diff
@@ -0,0 +1,33 @@
+package kyo.rules
+
+import scala.meta._
+import scalafix.v1._
+
+class KyoUpdateToV1_0 extends SemanticRule("KyoUpdateToV1_0") {
```
hearnadam, line 6: Happy to followup, but should we do this for all the renames?

.scalafmt.conf
ahoy-jon: 🤔 we me have to go back to 2.12 after (hope not), but yes.

scalafix/rules/src/main/scala/kyo/rules/KyoUpdateToV1_0.scala
ahoy-jon: We still have rules to write!

scalafix/rules/src/main/scala/kyo/rules/KyoUpdateToV1_0.scala
ahoy-jon: I will check the changelog, I guess the biggest ones (like `Resource`)

build.sbt
ahoy-jon: it's for tests, so we can write different test modules (one for 0.19 to 1, ...)
I guess we will only keep the last two versions in the future!

so `kyo-scalafix-input`  and `kyo-scalafix-previousrelease-input`

hearnadam (approved):

PR #1325, opened 2025-07-03
[kernel] lifted functions fixes

fwbrasil:
### Problem

@ahoy-jon found a couple of issues with lifted functions. Due to the definition of the implicit conversions being in the same scope as the definition of `opaque type <`, the compiler doesn't introduce a `lift` conversion, which creates issues with nested computations in generic scopes. Additionally, the compiler is having issues with inlining of pure functions in generic contexts that requires an explicit type annotation.

### Solution

Introduce the missing `lift` calls in the function conversions and add explicit type annotations in methods that take functions.

### DIFF
--- a/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
@@ -72,7 +72,7 @@ object `<`:
                         val value = v.unsafeGet
                         Safepoint.handle(value)(
                             suspend = mapLoop(value),
-                            continue = f(value)
+                            continue = f(value): B < S2
                         )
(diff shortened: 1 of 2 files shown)


kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala
```diff
@@ -72,7 +72,7 @@ object `<`:
                         val value = v.unsafeGet
                         Safepoint.handle(value)(
                             suspend = mapLoop(value),
-                            continue = f(value)
+                            continue = f(value): B < S2
```
ahoy-jon, line 75: @fwbrasil 👏 👏  that's an amazing catch

PR #1327, opened 2025-07-04
[prelude] parametrized generic aspects

fwbrasil:
### Problem

`Aspect`s can't be used in generic contexts when their inputs or outputs depend on a parametrized type. This issue was raised as part of the work to generalize `Parse` in https://github.com/getkyo/kyo/pull/1305.

### Solution

Instead of relying on the object identity of the `Aspect` instance, manage `Cut` activations based on a pair of the type signature of the `Aspect` via `Tag` and its allocation site via `Frame`. 

### Notes

- This approach doesn't keep a strict identity as before since the `Frame` can be shared in case it's available a scope with multiple `Aspect` instantiations but that's something that should happen only in Kyo's codebase since users shouldn't manage `Frame`s.
- I've removed the `default` cut to simplify the implementation since it's unused.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Aspect.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Aspect.scala
@@ -34,7 +34,7 @@ import Aspect.*
   */
 final class Aspect[Input[_], Output[_], S] private[kyo] (
     using
-    tag: Tag[(Input[Any], Output[Any], S)],
+    tag: Tag[(Input[Any], Output[Any])],
     frame: Frame
 ) extends Serializable:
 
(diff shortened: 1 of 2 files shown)


fwbrasil:
I had to remove `S` from the `Tag` due to a failure in the readme. `TagMacro` is failing to materialize a tag for this `Aspect`:

```scala
val numberAspect = Aspect.init[Const[Int], Const[Int], Abort[Throwable] & Sync]
```

There's a risk of a dynamic `S` param mismatch, which could lead to unhandled effects, but I don't think it's a blocker. I'll open an issue to follow up with a `Tag` fix.

Issue #1247, opened 2025-06-08
Remove usage of erased Tags


hearnadam:
Now that Tag supports low cost subtype checking, we should switch to using it in ContextEffect handling (if not already), then remove usage of `erasedTag`.

avasis-ai:
I was looking at how Tags are used across the codebase — is the intent to replace the erased Tags with the new Kyo Tags that carry type information at runtime, or is this about removing the `erased` keyword usage specifically? The Tag machinery in kyo-prelude seems to already have the non-erased variants, wondering if this is mostly a mechanical find-and-replace or if there are places where the type-level semantics change.

PR #1333, opened 2025-07-04
Abort: remove usage of erased Tags

hearnadam:
Related to https://github.com/getkyo/kyo/issues/1247

### Problem
Erased tags are leftover from earlier limitations with splicing and subtype checking of tags.

### Solution
Require tag evidences in all Abort methods.

### Notes

- I think we should consider making less arguments/methods inline in Abort. Perhaps just the most common cases need to be inline. This will help prevent large bytecode sizes.
- I moved `Frame` to the first implicit for consistency where possible, but we can followup for the rest.


kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
```diff
@@ -669,7 +669,7 @@ class AbortTest extends Test:
                 }
                 "subclass" in {
                     assert(
-                        Abort.run[RuntimeException](Abort.catching[RuntimeException](test(0))).eval ==
+                        Abort.run[RuntimeException](Abort.catching[Ex1](test(0))).eval ==
```
steinybot, line 672: Wow how did you spot that this wasn't catching a subclass?

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
```diff
@@ -1272,7 +1272,7 @@ class AbortTest extends Test:
         }
 
         "generic" in {
-            def test[A](a: A): Nothing < Abort[A] =
+            def test[A](a: A)(using tag: Tag[a.type]): Nothing < Abort[A] =
```
steinybot, line 1285: Is the singleton type necessary or could it be:
```suggestion
            def test[A](a: A)(using tag: Tag[A]): Nothing < Abort[A] =
```

kyo-combinators/shared/src/main/scala/kyo/AbortCombinators.scala
```diff
@@ -418,14 +442,16 @@ class ForAbortOps[A, S, E, E1 <: E](effect: A < (Abort[E] & S)) extends AnyVal:
...
     ): [A1 >: A, S1] => (E1 => A1 < S1) => A1 < (S & S1 & reduce.SReduced) =
         [A1 >: A, S1] =>
             (fn: E1 => A1 < S1) =>
                 reduce(Abort.run[E1](effect.asInstanceOf[A < (Abort[E1 | ER] & S)]).map {
-                    case Result.Failure(e1)   => fn(e1)
-                    case Result.Success(v)    => v
-                    case ab @ Result.Panic(_) => Abort.get(ab.asInstanceOf[Result[Nothing, Nothing]])
+                    case Result.Failure(e1)  => fn(e1)
+                    case Result.Success(v)   => v
+                    case panic: Result.Panic => Abort.error(panic)
```
steinybot, line 454: Nice.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -848,15 +878,16 @@ object StreamCoreExtensions:
...
             i2: Isolate.Stateful[S, Async],
             t1: Tag[V],
             t2: Tag[Emit[Chunk[V]]],
             t3: Tag[Emit[Chunk[Chunk[V]]]],
             t4: ConcreteTag[E],
-            fr: Frame
+            tag: Tag[Abort[E]]
         ): Stream[V, Abort[E] & Async & Resource] < (Resource & Async & S) =
-            broadcastDynamic(bufferSize).map: streamHub =>
+            broadcastDynamic(bufferSize)(using fr, i1, i2, t1, t2, t3, t4, tag).map: streamHub =>
```
steinybot: Do we have to pass these explicitly?

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
hearnadam: It's necessary since this is aborting with a literal type

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
hearnadam: No, I want to remove all the explicit usages if possible. It made this refactor harder 😆

kyo-prelude/shared/src/test/scala/kyo/AbortTest.scala
hearnadam: The test could be correct in either case. In the original, the catch was a subclass of `RuntimeException` - though I wanted to show that the `run` would correctly capture the RE.

ahoy-jon (approved):

PR #1334, opened 2025-07-05
Kyo grpc

### LINKED ISSUE #390: gRPC Support

steinybot:
Fixes #390

/claim #390

### Problem

Built-in support in Kyo for gRPC.

### Solution

The code generation uses [paiges](https://github.com/typelevel/paiges) which I don't love. I started using Scalameta but it [doesn't support comments](https://github.com/scalameta/scalameta/issues/3372). It is possible this has changed since then. Either way, perhaps switching back to Scalameta would be a better choice. This code got a bit funky.

This is similar to [zio-grpc](https://zio.dev/ecosystem/community/zio-grpc/) although their `ServerLayer` and `ServiceList` stuff seemed unnecessary.

I am using the high level API. In future I'll try implementing it using `ServerCallHandler` which would probably give much better flow control and backpressure. It seems to still be pretty performant even without it (I need to run the benchmarks again).

### Notes

TODO during PR:
- [ ] Rename ArenaBench2
- [ ] Rerun benchmarks
- [x] Check for IOs
- [ ] Update READMEs
- [ ] Ensure everything has ScalaDocs
- [ ] Ensure that the example compiles as part of the build
- [ ] Decide what to use for the StreamChannel capacity
- [ ] Make sure that the right stuff is private
- [ ] Figure out what to do with the kyo version in the example project
- [ ] Remove `closeAwaitEmptyFiber`

TODO after PR:
- [ ] Reimplement the code gen using scalameta instead of paiges (ideally with a version of scalameta that supports comments)
- [x] Implement ServerCallHandler like ZIO does
- [x] Support Metadata and Attributes
- [ ] Make sure observers backpressure and can be interrupted correctly.
- [ ] Potentially split kyo-grpc-core into client / server
- [ ] Potentially split kyo-grpc-core-gen into client / server / shared
- [ ] Add link to https://grpc.io/docs/languages/
- [ ] Check all of these https://grpc.io/docs/guides/
- [ ] Split CodeGenerator into client / server
- [ ] Add JS client tests


kyo-bench/src/main/scala/kyo/bench/arena/ArenaBench2.scala
```diff
@@ -0,0 +1,75 @@
+package kyo.bench.arena
+
+import kyo.bench.BaseBench
+import org.openjdk.jmh.annotations.*
+import scala.compiletime.uninitialized
+
+// TODO: What to call this?
+abstract class ArenaBench2[A](val expectedResult: A) extends BaseBench:
```
steinybot, line 8: This approach allows the runtimes and other state to be provided via `@State` objects. For example `GrpcClientUnaryBench` requests the state for each effect system from `GrpcClientBench`. `GrpcClientBench` defines the states such that they start the server and shut it down after the benchmark.

This could be used everywhere although that is a bit much for this PR. If people like this approach I can do a follow up to change other benchmarks to use this. Alternatively we can keep both approaches.

For now, need to come up with a better name for this.

kyo-grpc/core/jvm/src/main/scala/kyo/grpc/ServerHandler.scala
```diff
@@ -0,0 +1,114 @@
...
+      *   the handler function that processes the request
+      * @tparam Request
+      *   the type of the incoming request
+      * @tparam Response
+      *   the type of the outgoing response
+      * @return
+      *   a gRPC [[ServerCallHandler]] for unary calls
+      */
+    def unary[Request, Response](f: Request => Response < Grpc)(using Frame): ServerCallHandler[Request, Response] =
+        ServerCalls.asyncUnaryCall { (request, responseObserver) =>
```
steinybot: Using the `ServerCalls` stubs is not ideal. We are not able to do proper flow control. We also don't get the `Metadata` headers which would be pretty important in a real production usage.

I will need to follow up this PR with another one that implements `ServerCallHandler` manually. This is what `zio-grpc` does.

fwbrasil:
@copilot resolve the merge conflicts in this pull request

PR #1335, opened 2025-07-05
[build] use setup-action for jdk+sbt setup

hearnadam:
### Problem
The CI was using an old, specific version of JDK21.

### Solution
Use coursier setup to manage JDK versioning.

### Notes
I kept bench on JDK21 as I suspect the current async-profiler setup is not correct for JDK24.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -58,9 +58,10 @@ jobs:
         with:
           fetch-depth: 0
       - uses: coursier/cache-action@v6
-      - uses: olafurpg/setup-scala@v14
+      - uses: coursier/setup-action@v1
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/bench.yml
```diff
@@ -58,9 +58,10 @@ jobs:
         with:
           fetch-depth: 0
       - uses: coursier/cache-action@v6
-      - uses: olafurpg/setup-scala@v14
+      - uses: coursier/setup-action@v1
```
steinybot, line 61: As per https://docs.github.com/en/actions/how-tos/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions we should consider using the hash rather than a tag as the tag could be updated to a new hash like what happened with the changed-files action.

steinybot (approved):

.github/workflows/bench.yml
hearnadam: That's annoying - let's followup for all the actions if that's alright?

PR #1337, opened 2025-07-06
[combinators] clean up Async Combinators + Constructors

hearnadam:
### Problem
This PR refines some of the signatures in Async combinators and Constructors to more closely align with ZIO + what I expect as a user.

### Changes
- Add `Reducible` to `Fiber.join`
- Refactor `awaitCompletion` to `await` - let users discard the `Result`
- Fix `Kyo.addFinalizer` to support all effects supported by `Resource`.
- remove `deferAttempt
- Enable `Resource.acquire` to support `AutoClosable` (parent type of `Closable`

### DIFF
--- a/kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
+++ b/kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
@@ -14,9 +14,7 @@ extension [A, E, Ctx](effect: A < (Abort[E] & Async & Ctx))
       * @return
       *   A computation that produces the result of this computation with Async effect
       */
-    inline def fork(
-        using frame: Frame
-    ): Fiber[E, A] < (Sync & Ctx) =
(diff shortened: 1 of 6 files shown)


kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
```diff
@@ -104,14 +100,14 @@ extension [A, E, S](fiber: Fiber[E, A] < S)
       * @return
       *   A computation that produces the result of this computation with Async effect
       */
-    def join(using Frame): A < (S & Abort[E] & Async) =
+    def join(using frame: Frame, reduce: Reducible[Abort[E]]): A < (S & reduce.SReduced & Async) =
```
ahoy-jon: Is it still useful `reduce: Reducible[Abort[E]]`, with `Async <: Abort[Nothing]`?

hearnadam:
```suggestion
    def join(using Frame): A < (S & Abort[E] & Async) =
```

ahoy-jon (approved):

PR #1347, opened 2025-07-07
update scala 3.7.1

ahoy-jon:
ignoring a test in `kyo-data`

### DIFF
--- a/kyo-data/shared/src/test/scala/kyo/RecordTest.scala
+++ b/kyo-data/shared/src/test/scala/kyo/RecordTest.scala
@@ -318,7 +318,7 @@ class RecordTest extends Test:
             inline def stage[Name <: String, Value](field: Field[Name, Value]): Column[Value] =
                 Column[Value](field.name)(using summonInline[AsColumn[Value]])
 
-        "build record if all inlined" ignore {
+        "build record if all inlined" pending {
             assertCompiles("""
             type Person = "name" ~ String & "age" ~ Int
 


kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -318,7 +318,7 @@ class RecordTest extends Test:
             inline def stage[Name <: String, Value](field: Field[Name, Value]): Column[Value] =
                 Column[Value](field.name)(using summonInline[AsColumn[Value]])
 
-        "build record if all inlined" in {
+        "build record if all inlined" ignore {
             assertCompiles("""
```
hearnadam, line 322: We should change these methods to use `typeCheck`.

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -318,7 +318,7 @@ class RecordTest extends Test:
             inline def stage[Name <: String, Value](field: Field[Name, Value]): Column[Value] =
                 Column[Value](field.name)(using summonInline[AsColumn[Value]])
 
-        "build record if all inlined" in {
+        "build record if all inlined" ignore {
```
hearnadam: Can you use `pending` or whatever so the test begins failing if the change is fixed?

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
```diff
@@ -318,8 +318,8 @@ class RecordTest extends Test:
             inline def stage[Name <: String, Value](field: Field[Name, Value]): Column[Value] =
                 Column[Value](field.name)(using summonInline[AsColumn[Value]])
 
-        "build record if all inlined" in {
-            assertCompiles("""
+        "build record if all inlined" pending {
+            typeCheck("""
```
ahoy-jon: @hearnadam are you sure? 
```scala
transparent inline def typeCheck(inline code: String): Result[String, Unit]

transparent inline def assertCompiles(inline code: String): Assertion
```

hearnadam: You need to assert that it's a success - perhaps we should have `typeCheck` (success), `typeCheckWith` (private) and `typeCheckFailure` 🤔

ahoy-jon: ok, but new PR! going back to `assertCompiles`

hearnadam (approved):

kyo-data/shared/src/test/scala/kyo/RecordTest.scala
fwbrasil: @ahoy-jon please create an issue for this cc/ @road21

PR #1348, opened 2025-07-08
[core] respect Thread interruption on blocking boundaries

hearnadam:
Merge Interrupt and Interrupted exceptions

### Problem
Imperative code using Thread interruption will not propagate interruption to the Fiber/Promise being awaited.

### Solution
Check the Thread interrupt status + restore it.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
+++ b/kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
@@ -3,7 +3,6 @@ package kyo.scheduler
 import IOPromise.*
 import java.util.concurrent.locks.LockSupport
 import kyo.*
-import kyo.Fiber.Interrupted
 import kyo.Result.Error
 import kyo.kernel.internal.Safepoint
 import scala.annotation.tailrec
@@ -76,7 +75,7 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
                 case l: Linked[E, A] @unchecked =>
                     interruptsLoop(l.p)
                 case _ =>
-                    discard(other.interrupt(Result.Panic(Interrupt())))
+                    discard(other.interrupt(Result.Panic(Interrupted(frame))))
         interruptsLoop(this)
     end interrupts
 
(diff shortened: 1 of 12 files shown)


kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -255,7 +260,12 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
                     blockLoop(l.p)
                 case v =>
                     v.asInstanceOf[Result[E | Timeout, A]]
-        blockLoop(this)
+        try blockLoop(this)
+        catch
+            case interrupted: java.lang.InterruptedException =>
```
hearnadam: We probably can remove this? `LockSupport` just continues after interruption.

kyo-core/shared/src/test/scala/kyo/scheduler/SyncPromiseTest.scala
```diff
@@ -220,6 +220,26 @@ class SyncPromiseTest extends Test:
...
+            thread.start()
+
+            // wait for parking
+            while !threadStarted do Thread.sleep(10)
+            Thread.sleep(10)
+
+            thread.interrupt()
+            thread.join(1000)
+
+            assert(p.block(deadline()).isPanic)
```
hearnadam: For some reason this always times out? somehow the promise is not being completed??

From my debugging `run()` correctly reports the Result as Interrupted. @fwbrasil am I missing something obvious here?

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
fwbrasil: yeah, I just checked it doesn't throw

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -236,14 +235,20 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
...
-                                    LockSupport.parkNanos(this, timeLeft.toNanos)
+                                    val timeLeft = deadline.timeLeft()
+                                    if !timeLeft.isFinite then
+                                        LockSupport.park(this)
+                                    else
+                                        LockSupport.parkNanos(this, timeLeft.toNanos)
+                                    end if
+                                    if Thread.interrupted() then
+                                        Thread.currentThread().interrupt()
+                                        Result.Panic(Interrupted(frame, s"Thread ${Thread.currentThread().getName}"))
```
fwbrasil: I think we need to interrupt the fiber here + throw `InterruptedException`

fwbrasil:
it seems we still have references to `Fiber.Interrupted`: https://github.com/getkyo/kyo/actions/runs/16142273575/job/45553293634?pr=1348#step:6:4796

ahoy-jon (approved):

Issue #1320, opened 2025-07-02
[BUG]: Contextual isolation error when adding type signature


gcsolaroli:
### Version

0.19.0

### Scala Version

3.7.1

### Expected Behavior

Specifying the type signature for a valid symbol, defined in a for comprehension, should not raise a compilation error.

### Actual Behavior

Declaring a symbol in a for comprehension specifying its type cause a compilation error.

The following statement works fine:
```scala
    run:
        for
            _ <- Console.printLine(">>> Task Start")

            (job1: (Unit < Async), signal1: Stream[String, Async]) <- job(1, 10, 3.seconds)
            (job2: (Unit < Async), signal2: Stream[String, Async]) <- job(2, 10, 5.seconds)

            updates = Stream.collectAll(Seq(signal1, signal2))
            ui      = updates.into(logSink)
            jobs    = Async.collectAll(Seq(job1, job2))

            _ <- Async.race(jobs, ui)
            _ <- Console.printLine(s"<<< Task Done")
        yield ()
```

When adding the type signature to the definition of `updates` the compiler reports the following error:
```
[error] -- Error: /Main.scala:38:85
[error] 38 |            updates: Stream[String, Async] = Stream.collectAll(Seq(signal1, signal2))
[error]    |                                                                                     ^
[error]    |This operation requires Contextual isolation for effects:
[error]    |
[error]    |  kyo.Async$package.Async
[error]    |
[error]    |Common mistake: Using operations like Async.run with effects that need complex state management.
[error]    |
[error]    |You have a couple of options, from simplest to most advanced:
[error]    |
[error]    |1. Handle these effects before the operation:
[error]    |   Async.run(MyEffect.run(computation))
[error]    |
[error]    |2. Use an operation that supports complex state:
[error]    |   Instead of Async.run, use Async.parallel
[error] one error found
```

### Current Workaround

The only solution I found at the moment is to  just leave the type signature out.

ahoy-jon:
it minimizes to a bug with

```scala
package testSignal

import kyo.*

val job = Stream.init(Seq.empty[String])

val prg = Stream.collectAll[String, Nothing, Async](Seq(job))

@main def main(): Unit = ()

//given Contextual[Async, Any] = null
```
This operation requires Contextual isolation for effects:

  kyo.Async$package.Async

Common mistake: Using operations like Async.run with effects that need complex state management.

You have a couple of options, from simplest to most advanced:

1. Handle these effects before the operation:
[...]

PR #1349, opened 2025-07-08
[prelude][core] unify contextual and stateful isolates + use a pending set in Fiber

fwbrasil:
Fixes https://github.com/getkyo/kyo/issues/1320

---

This PR reworks core abstractions to improve composability, inference, and runtime semantics:

| **Change**                    | **Description**                                                                 |
|------------------------------|---------------------------------------------------------------------------------|
| **Isolate unification**      | Merges `Isolate.Contextual` and `Isolate.Stateful` into `Isolate[Remove, Keep, Restore]`, resolving composability and inference issues. |
| **Fiber pending set**        | Adds a `pending` set to track suspended computations, improving control over lifecycle and interruption. |
| **Decoupling from `Abort[E]`** | Removes `Abort[E]` from core `Fiber` logic, allowing failure-free fiber use and cleaner effect sets. |
| **Invariant extension workaround** | Duplicates extension methods for `Promise`/`Fiber` to bypass variance restrictions in `Abort[E] & S`. |
| **Effect combinator updates** | Adapts `AsyncCombinators`, `Stream`, `Resource`, etc. to the new isolate model and improved inference. |
| **Utility improvements**     | Refactors `Clock.repeat`; moves `Async.runAndBlock` to `KyoApp` for clearer user API. |
| **Test updates**             | Updates tests for `panic`, `await`, `getResult`, etc., matching the refined semantics. |


README.md
```diff
@@ -1858,26 +1858,26 @@ val a: Unit < Sync =
 
 // Recurring task with a delay between
 // executions
-val b: Fiber[Nothing, Unit] < Sync =
+val b: Fiber[Unit, Abort[Nothing]] < Sync =
     Clock.repeatWithDelay(
```
fwbrasil, line 1903: I'll review the `Clock` methods to use `Reducible` and `Isolate`

kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
```diff
@@ -41,8 +41,8 @@ extension [A, E, Ctx](effect: A < (Abort[E] & Async & Ctx))
     @targetName("zipRightPar")
     def &>[A1, E1, Ctx1](next: A1 < (Abort[E1] & Async & Ctx1))(
         using
-        i1: Isolate.Contextual[Ctx, Sync],
-        i2: Isolate.Contextual[Ctx1, Sync],
+        i1: Isolate[Ctx, Sync, Any],
+        i2: Isolate[Ctx1, Sync, Any],
```
fwbrasil: these methods could be more flexible with `Isolate[Ctx1, Sync, Any]`. I'm planning to follow up on it

kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -525,17 +525,17 @@ object Channel:
                         case Absent =>
                             Absent
                         case Present(Put.Value(value, promise)) =>
-                            discard(promise.complete(Result.unit))
+                            discard(promise.complete(Result.succeed(())))
```
fwbrasil: this is something not great about this new encoding of `Fiber`.  The completion now takes a `Result[E, A < S]` so `succeed` is necessary to lift `()` to `Unit < Any`

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -1,746 +1,527 @@
...
+    end extension
 
-        /** Gets the result of the Fiber.
-          *
-          * @return
-          *   The result of the Fiber
-          */
-        def get(using reduce: Reducible[Abort[E]], frame: Frame): A < (reduce.SReduced & Async) =
-            Async.get(self.asPromise)
+    extension [E, A, S](self: Fiber[A, Abort[E] & S])
```
fwbrasil, line 198: separate set of extensions for methods that require the `E` failure type

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -1,746 +1,527 @@
...
+            def poll(using Frame): Maybe[Result[Nothing, A < S]] < Sync =
+                Sync.Unsafe(Unsafe.poll(self)())
+
+            def waiters(using Frame): Int < Sync =
+                Sync.Unsafe(Unsafe.waiters(self)())
+
+            def unsafe: Unsafe[A, S] = self
+        end extension
+
+        extension [E, A, S](self: Promise[A, Abort[E] & S])
```
fwbrasil, line 486: I had to repeat all methods in both extensions due to `Promise` being invariant, which doesn't allow extracting `Abort[E]` if the promise doesn't have the pending effect

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -1,746 +1,527 @@
...
-                    if i < s then
-                        f(i, l(i))
-                        loop(i + 1)
-                loop(0)
-            case _ =>
-                val it = l.iterator
-                @tailrec def loop(i: Int): Unit =
-                    if it.hasNext then
-                        f(i, it.next())
+    private[kyo] object internal:
```
fwbrasil, line 650: I've moved these impls to facilitate reading of the source file since they're internal

kyo-core/shared/src/main/scala/kyo/Resource.scala
```diff
@@ -141,8 +141,6 @@ object Resource:
                 }
         }
 
-    given Isolate.Contextual[Resource, Any] = Isolate.Contextual.derive[Resource, Any]
```
fwbrasil, line 144: the derivation macro now automatically handles `ContextEffect`s

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -82,8 +82,7 @@ object StreamCoreExtensions:
 
         def consume[S](stream: Stream[A, Abort[E] & S & Async])(
             using
-            i1: Isolate.Contextual[S, Sync],
-            i2: Isolate.Stateful[S, Abort[E] & Async],
+            i: Isolate[S, Sync, S],
```
fwbrasil, line 85: This change was motivated by the composability issues of the two isolates here + inference issue reported in https://github.com/getkyo/kyo/issues/1320. It's much more flexible now!

kyo-core/shared/src/test/scala/kyo/AsyncTest.scala
```diff
@@ -371,7 +371,7 @@ class AsyncTest extends Test:
             val future: Future[Int] = Future.failed(exception)
             for
                 result <- Abort.run(Async.fromFuture(future))
-            yield assert(result.failure.contains(exception))
+            yield assert(result.panic.contains(exception))
```
fwbrasil, line 374: I've moved potential failures from `Future`s to panic

kyo-aeron/jvm/src/main/scala/kyo/Topic.scala
```diff
@@ -262,6 +262,6 @@ object Topic:
         }
     end stream
 
-    given isolate: Isolate.Contextual[Topic, Any] = Isolate.Contextual[Topic, Any]
+    given isolate: Isolate[Topic, Any, Any] = Isolate.derive[Env[Aeron], Any, Any]
```
ahoy-jon, line 265: so, there is only one `Isolate`, that's nice!

kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
```diff
@@ -1,359 +1,140 @@
...
-                            |   Instead of Fiber.init, use Async.parallel 
-                            |""".stripMargin
-                    )
-                end if
-
-                '{ cached.asInstanceOf[Isolate.Contextual[R, P]] }
-            }
-
-    end Contextual
+abstract class Isolate[Remove, -Keep, Restore]:
```
ahoy-jon, line 80: 🧠, and the `-Keep` 🤌 !

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
ahoy-jon: 💯, ~I think it's `Isolate[S, Async, S]` (not sure)~ tried it on my machine, I was wrong!

kyo-core/shared/src/main/scala/kyo/Channel.scala
ahoy-jon: With an extension:
```scala
promise.complete(Result.succeed(()))
promise.done
\\
promise.completeDiscard(Result.succeed(()))
promse.doneDiscard
```

kyo-bench/src/main/scala/kyo/bench/arena/ForkChainedBench.scala
```diff
@@ -22,12 +22,12 @@ class ForkChainedBench extends ArenaBench.ForkOnly(0):
     override def kyoBenchFiber() =
         import kyo.*
 
-        def iterate(p: Promise[Nothing, Unit], n: Int): Unit < Sync =
-            if n <= 0 then p.complete(Result.unit).unit
+        def iterate(p: Promise[Unit, Any], n: Int): Unit < Sync =
+            if n <= 0 then p.complete(Result.succeed(())).unit
```
ahoy-jon, line 26: with a extension 
```scala
p.doneDiscard
```

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -1,746 +1,527 @@
...
-        /** Uses the result of the Fiber to compute a new value.
-          *
-          * @param f
-          *   The function to apply to the Fiber's result
-          * @return
-          *   The result of applying the function to the Fiber's result
-          */
-        def use[B, S](f: A => B < S)(using reduce: Reducible[Abort[E]], frame: Frame): B < (reduce.SReduced & Async & S) =
-            Async.use(self.asPromise)(f)
+        def get(using reduce: Reducible[Abort[E]], frame: Frame): A < (reduce.SReduced & Async & S) =
```
ahoy-jon: is Reducible usefull here?

kyo-core/shared/src/main/scala/kyo/Channel.scala
ahoy-jon: `done` is not the right term, maybe `markDone`/`succeed`, I tried, might be interesting with a `completeSucceed`/`succeed`
```scala
        extension [S](self: Promise[Unit, S])
            def completeUnit(using Frame): Boolean < Sync = Sync.Unsafe(Unsafe.complete(self)(Result.succeed(Kyo.unit)))
```

ahoy-jon:
```scala
        extension [S](self: Promise[Unit, S])
            def succeed(using Frame): Boolean < Sync = succeedWith(Kyo.unit)

            def succeedWith(using Frame)(v: Unit < S): Boolean < Sync =
                Sync.Unsafe(Unsafe.complete(self)(Result.succeed(v)))
        end extension

        extension [A, S](self: Promise[A, S])(using NotGiven[S <:< Abort[Any]])

            def succeedWith(v: A < S)(using Frame): Boolean < Sync = self.complete(Result.succeed(v))

Comment (fwbrasil) [reply kyo-core/shared/src/main/scala/kyo/Channel.scala]: I've added `completeUnit`. We use the `With` suffix to indicate that the method takes a continuation function

File: kyo-core/shared/src/main/scala/kyo/Fiber.scala (hunk 13 above)
Comment (fwbrasil) [reply kyo-core/shared/src/main/scala/kyo/Fiber.scala]: good point! it isn't since `Async` includes `Abort[Nothing]`. Removed

File: README.md (hunk 1 above)
Comment (fwbrasil) [reply README.md:1903]: fixed

File: kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala [hunk 14]
```diff
@@ -93,21 +103,21 @@ extension [A, E, Ctx](effect: A < (Abort[E] & Async & Ctx))
...
+    def join(using Frame): A < (S & S2 & Abort[E] & Async) =
         fiber.map(_.get)
 
     /** Awaits the completion of the fiber and returns its result as a `Unit`.
       *
       * @return
       *   A computation that produces the result of this computation with Async effect
       */
-    def await(using Frame): Result[E, A] < (S & Async) =
+    def await(using Frame): Result[E, A < S2] < (S & Async) =
```
hearnadam: This is a bit odd, though I suppose it makes sense... it might be nice to have a method that converts to `Result[E, A]` without aborting (since join does).

I am inclined to believe most users will want the 'joining' effects on await...

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -195,13 +195,13 @@ object Async extends AsyncPlatformSpecific:
...
     def race[E, A, S](
-        using isolate: Isolate.Stateful[S, Abort[E] & Async]
+        using isolate: Isolate[S, Abort[E] & Async, S]
     )(iterable: Iterable[A < (Abort[E] & Async & S)])(
         using frame: Frame
     ): A < (Abort[E] & Async & S) =
         require(iterable.nonEmpty, "Can't race an empty collection.")
         isolate.capture { state =>
-            Fiber.race(iterable.map(isolate.isolate(state, _))).map(fiber => isolate.restore(fiber.get))
+            Fiber.internal.race(iterable.map(isolate.isolate(state, _))).map(fiber => isolate.restore(fiber.get))
```
hearnadam, line 204: I think it makes sense to keep the Fiber combinator implementations internal, though some can be useful, like `collect`

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -801,21 +801,11 @@ object Async extends AsyncPlatformSpecific:
...
 
-    private[kyo] def use[E, A, B, S](v: IOPromise[? <: E, ? <: A])(f: A => B < S)(
-        using
-        reduce: Reducible[Abort[E]],
-        frame: Frame
-    ): B < (S & reduce.SReduced & Async) =
-        val x = useResult(v)(_.fold(f, Abort.fail, Abort.panic))
-        reduce(x)
-    end use
+    private[kyo] def use[E, A, B, S](v: IOPromise[? <: E, ? <: A])(f: A => B < S)(using Frame): B < (Abort[E] & Async & S) =
```
hearnadam: Should this be inline?

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -28,35 +26,32 @@ import scala.util.control.NoStackTrace
...
   * @see
   *   [[Promise]] for creating and completing Fibers manually
   * @see
   *   [[Fiber.Unsafe]] for low-level operations requiring [[AllowUnsafe]]
   */
-opaque type Fiber[+E, +A] = IOPromiseBase[E, A]
+opaque type Fiber[+A, -S] = IOPromiseBase[Any, A < (Async & S)]
 
-object Fiber:
+export Fiber.Promise
```
hearnadam, line 43: 🥳

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -28,35 +26,32 @@ import scala.util.control.NoStackTrace
...
 
-    private val _unit  = IOPromise(Result.unit).mask()
-    private val _never = IOPromise[Nothing, Unit]().mask()
+object Fiber:
 
-    private[kyo] inline def fromTask[E, A](inline ioTask: IOTask[?, E, A]): Fiber[E, A] = ioTask
+    private val _unit = IOPromise(Result.succeed((): Unit < Any))
```
hearnadam, line 47: Does this need a mask?

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -28,35 +26,32 @@ import scala.util.control.NoStackTrace
...
       */
-    def unit[E]: Fiber[E, Unit] = _unit.asInstanceOf[Fiber[E, Unit]]
-
-    /** Creates a never-completing Fiber.
-      *
-      * @return
-      *   A Fiber that never completes
-      */
-    def never[E, A]: Fiber[E, A] = _never.asInstanceOf[Fiber[E, A]]
+    def unit: Fiber[Unit, Any] = _unit
```
hearnadam, line 54: ?
```suggestion
    def unit[S]: Fiber[Unit, S] = _unit
```

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -83,23 +78,32 @@ object Fiber:
...
+
+    /** Creates a never-completing Fiber.
+      *
+      * @return
+      *   A Fiber that never completes
+      */
+    def never(using Frame): Fiber[Nothing, Any] < Sync =
+        Sync.defer(IOPromise[Nothing, Nothing < Any]())
+
+    def fromResult[E, A, S](result: Result[E, A < S]): Fiber[A, Abort[E] & S] =
```
hearnadam: Does this need a reduction to avoid `Abort[Nothing]`?

kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
```diff
@@ -7,24 +7,34 @@ import scala.annotation.tailrec
...
       * @return
       *   A computation that produces the result of this computation with Async effect
       */
-    inline def fork(using Frame, Isolate.Contextual[Ctx, Sync]): Fiber[E, A] < (Sync & Ctx) =
+    def fork[S2](
+        using
+        isolate: Isolate[S, Sync, S2],
+        reduce: Reducible[Abort[E]],
+        frame: Frame
+    ): Fiber[A, reduce.SReduced & S2] < (Sync & S) =
```
hearnadam, line 22: Outside the scope of this PR, but I think we should explore distinct aliases for `Reducible`, given a signature with a reduced effect is not very clear. Perhaps `abort.Error` or something.

This would be useful if a signature has more than one reducible effect.

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -361,7 +361,7 @@ object Clock:
                                     Maybe.empty
                         } match
                             case Present(task) =>
-                                task.completeDiscard(Result.unit)
+                                task.completeDiscard(Result.succeed(()))
```
hearnadam, line 364: missed one 🙂

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -642,12 +718,12 @@ object Clock:
...
                     Promise.Unsafe.fromIOPromise {
-                        new IOPromise[Nothing, Unit] with Callable[Unit]:
+                        new IOPromise[Any, Unit < Any] with Callable[Unit]:
                             val task = executor.schedule(this, duration.toNanos, TimeUnit.NANOSECONDS)
-                            override def interrupt(error: Result.Error[Nothing]): Boolean =
+                            override def interrupt(error: Result.Error[Any]): Boolean =
                                 discard(task.cancel(true))
                                 super.interrupt(error)
-                            def call(): Unit = completeDiscard(Result.unit)
+                            def call(): Unit = completeDiscard(Result.succeed(()))
```
hearnadam, line 726: Another?

kyo-stm/shared/src/test/scala/kyo/STMTest.scala
```diff
@@ -802,10 +802,10 @@ class STMTest extends Test:
             r.get
         }
 
-        val task = Async.runAndBlock(Duration.Infinity)(Async.fromFuture(unsafeToFuture(STM.run(faultyTransaction))))
+        val task = KyoApp.runAndBlock(Duration.Infinity)(Async.fromFuture(unsafeToFuture(STM.run(faultyTransaction))))
 
         Abort.run(task).map { result =>
-            assert(result == Result.fail(ex))
+            assert(result == Result.panic(ex))
```
hearnadam, line 808: Is there a reason this changed?

kyo-reactive-streams/shared/src/test/scala/kyo/interop/flow/PublisherToSubscriberTest.scala
```diff
@@ -226,7 +226,7 @@ abstract private class PublisherToSubscriberTest extends Test:
                 }
                 _      <- Resource.run(subscriber.stream.map(_.take(10).discard))
                 result <- promise.getResult
-            yield assert(result == Success(()))
+            yield assert(result.map(_.eval) == Success(()))
```
hearnadam:
```suggestion
            yield assert(result.map(_.eval) ==  Result.unit)
```

kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
```diff
@@ -60,300 +70,232 @@ import scala.quoted.*
...
 
-                val statefulIsolates =
-                    isolates.flatMap {
-                        case (_, Some(Right(isolate))) => Some(isolate)
-                        case _                         => None
+        /** No-op isolate that performs no state management.
+          *
+          * Used as a base case for isolate composition and when no isolation is needed.
+          */
+        object noop extends Isolate[Any, Any, Any]:
```
hearnadam: I wonder if this should be `Identity`

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -83,23 +78,32 @@ object Fiber:
...
       *   The Future to convert into a Fiber
-      * @tparam E
-      *   The expected error type that the Future might fail with. Use Throwable if you don't need to catch specific exceptions.
       * @tparam A
       *   The type of the successful result
       * @return
       *   A Fiber that completes with the result of the Future
       */
-    def fromFuture[A](future: => Future[A])(using frame: Frame): Fiber[Throwable, A] < Sync =
+    def fromFuture[A](future: => Future[A])(using Frame): Fiber[A, Any] < Sync =
```
hearnadam, line 116: outside the scope of this PR, but I think we can use a ConcreteTag to determine which Throwables should be tracked for CF and Scala's Future, enabling `F[A] => F[E, A]` without `E =:= Throwable`

kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
```diff
@@ -7,24 +7,34 @@ import scala.annotation.tailrec
 import scala.annotation.targetName
 import scala.util.NotGiven
 
-extension [A, E, Ctx](effect: A < (Abort[E] & Async & Ctx))
+extension [A, E, S](effect: A < (Abort[E] & Async & S))
```
johnhungerford, line 10: We can probably get rid of the `Abort[E]` term here, since Fiber is no longer bound up with E. I can follow up separately if you don't want to go through that in this PR.

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -13,12 +13,12 @@ import kyo.internal.OsSignal
...
     private val awaitInterrupt =
         given AllowUnsafe = AllowUnsafe.embrace.danger
-        val promise       = Promise.Unsafe.init[Nothing, Nothing]()
+        val promise       = Promise.Unsafe.init[Nothing, Any]()
 
         val interrupt = (signal: String) =>
             () =>
                 promise
-                    .completeDiscard(Result.panic(Interrupted(Frame.internal, s"Interrupt Signal: $signal")))
+                    .completeDiscard(Result.panic(Interrupted(Frame.internal, s"Interrupt Signal Reached: $signal")))
```
hearnadam: The fiber message now says 'Fiber interrupted by $xyz' when a message is present, so this currently says: FIber Interrupted by Interrupt Signal INT - or whatever.
```suggestion
                    .completeDiscard(Result.panic(Interrupted(Frame.internal, s"Interrupt Signal: $signal")))
```

I think this is just a git rebase issue.

kyo-core/shared/src/main/scala/kyo/KyoApp.scala
```diff
@@ -37,6 +37,24 @@ object KyoApp:
...
+    /** Runs an asynchronous computation in a new fiber and blocks until completion or timeout.
+      *
+      * @param timeout
+      *   The maximum duration to wait
+      * @param v
+      *   The computation to run
+      * @return
+      *   The result of the computation, or a Timeout error
+      */
+    private[kyo] def runAndBlock[E, A, S](
```
hearnadam: I think this needs to be public

kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
fwbrasil: thanks!

kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
fwbrasil: makes sense. I've also updated `fiber.getResult` and `fiber.block`. The only case that remains is `fiber.onComplete` since there isn't a way to flatten the effects

kyo-core/shared/src/main/scala/kyo/Async.scala
fwbrasil: I think we might need to expose them at some point but let's see if the `Async` combinators are enough since it's more high-level

kyo-core/shared/src/main/scala/kyo/Fiber.scala
fwbrasil: no, given that it's fulfilled mask calls are ignored

kyo-core/shared/src/main/scala/kyo/Fiber.scala
fwbrasil: there's no need since `S` is contravariant

kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
fwbrasil: agreed

kyo-stm/shared/src/test/scala/kyo/STMTest.scala
fwbrasil: It's unrelated. I don't see much value in tracking `Throwable`

kyo-core/shared/src/main/scala/kyo/Clock.scala
fwbrasil: actually these two are calling `IOPromise`, which doesn't have the new methods. I don't think it's necessary

kyo-combinators/shared/src/main/scala/kyo/AsyncCombinators.scala
ahoy-jon: good idea!

hearnadam (approved):

PR #1355, opened 2025-07-09
add rule Resource => Scope + clean @deprecated 1.0-RC

ahoy-jon:


ahoy-jon:
@hearnadam with tests

kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -36,6 +36,12 @@ import kyo.kernel.ContextEffect
   */
 sealed trait Scope extends ContextEffect[Scope.Finalizer]
 
+@deprecated("use `Scope`", "1.0-RC")
+type Resource = Scope
+
+@deprecated("use `Scope`", "1.0-RC")
```
hearnadam: Can you note that these will be removed for 1.0?

hearnadam: Can you note that these will be removed for 1.0?

PR #1356, opened 2025-07-09
Rename: Resource -> Scope

hearnadam:
### Problem
- We plan to add support for heirarchical Resource scopes (see: #1131), but the naming of `Resource` makes the behavior less clear. It seems much more intuitive to 'nest' scopes.

### Solution
- Rename Resource to Scope. In #1315 Flavio will add a followup for local scoping.


ahoy-jon (approved):
?? 
```
@deprecated("use `Scope`", "1.0-RC")
type Resource = Scope

@deprecated("use `Scope`", "1.0-RC")
val Resource = Scope
```

PR #1369, opened 2025-07-10
Allow nested `.now`/`.later` in `direct`

### LINKED ISSUE #1366: [improvement] Improve `.now`/`.later` validation in `kyo-direct`
Allow `.now`/`.later` inside `.now`, only restrict inside `.later`.

The current validation in `kyo-direct` forbids nested `.now` or `.later` calls regardless of context. This is overly strict. The real constraint is that only `.now` inside `.later` should be forbidden.

```scala
def f(i:Int): Int < Any = i + 1

direct:
  Console.printLine(f(1).now).now  // ❌ Currently rejected, but safe
```

ahoy-jon:
fixes #1366

* allow (`.now`|`.later`) in `.now`
* allow `.later` in `.later`
* restrict `.now` in `.later`


kyo-direct/shared/src/test/scala/kyo/HygieneTest.scala
```diff
@@ -276,4 +276,26 @@ class HygieneTest extends Test:
...
+        def f(i: Int): Int < Var[Int] = Var.set(i)
+        val stream                    = Stream.init(Seq(1, 2, 3))
+
+        val res: Stream[Int, Var[Int]] < Abort[String] = direct:
+            val g = f(x.now).later
+            stream.map(_ => g)
+
+        Abort.run(res).eval match
+            case Result.Success(_) => assertionFailure("oups")
+            case Result.Error(_)   => assertionSuccess
```
ahoy-jon: @fwbrasil ~not sure if we should allow `.now` in `.later`, however, it's doing the correct behavior (compared to what you would get in direct style).~

ahoy-jon: "fixed" it, it is not allowed.

kyo-direct/shared/src/main/scala/kyo/Direct.scala
```diff
@@ -169,20 +169,24 @@ private def impl[A: Type](body: Expr[A])(using quotes: Quotes): Expr[Any] =
...
-                }
+            val transformer: TreeMap = new TreeMap:
+                override def transformTerm(term: Term)(owner: Symbol): Term =
+                    term match
+                        case Apply(TypeApply(Ident("now"), List(t, s2)), List(qual)) =>
+                            (t.tpe.asType, s2.tpe.asType) match
+                                case ('[t], '[s2]) => '{
+                                        given KyoCpsMonad[s2] = KyoCpsMonad[s2]
+                                        cps.await[[A] =>> A < s2, t, [A] =>> A < s2](${
+                                            transformTerm(qual)(owner).asExprOf[t < s2]
```
ahoy-jon, line 180: had to use the original TreeMap to help with recursion.

kyo-direct/shared/src/main/scala/kyo/internal/Validate.scala
```diff
@@ -192,52 +204,35 @@ private[kyo] object Validate:
...
-                            tree,
-                            s"""${".now".cyan} and ${".later".cyan} can only be used directly inside a ${"`direct`".yellow} block.
-                            |
-                            |Common mistake: You may have forgotten to wrap an effectful computation in ${"`direct`".yellow}:
-                            |${highlight("""
-                            |// Missing direct when handling effects:
-                            |val result = Emit.run {      // NOT OK - missing direct
-                            |    Emit.value(1).now
-                            |    Emit.value(2).now
-                            |}
```
steinybot, line 218: Is there no error message for when direct is missing now?

ahoy-jon: there is, in the macro, for `.now` and `.later`.

https://github.com/getkyo/kyo/blob/bfec5e5cf0d0358fa8a2b433c67957e8acc7a80a/kyo-direct/shared/src/main/scala/kyo/Direct.scala#L57-L74

ahoy-jon: It's not well tested, i will had a test, thanks.

kyo-direct/shared/src/test/scala/kyo/HygieneTest.scala
```diff
@@ -276,4 +276,51 @@ class HygieneTest extends Test:
...
+
+    ".now outside of direct" in {
+        val x: Int < Any = 1
+        typeCheckFailure("x.now")(".now must be used within a `direct` block.")
+    }
+
+    ".later outside of direct" in {
+        val x: Int < Any = 1
+        typeCheckFailure("x.later")(".later must be used within a `direct` block.")
+    }
```
ahoy-jon, line 288: @steinybot here!

Issue #1368, opened 2025-07-10
[BUG]: Tag expansion fails with opaque type with bounds + variance in Stream


hearnadam:
### Version

0.19.0

### Scala Version

3.7.1

### Expected Behavior

```scala
import kyo.*

opaque type B <: Vector[Any] = Vector[Any]

def x: Stream[B, Any] = ???

x.foreach(println)
```

### Actual Behavior

```
Exception occurred while executing macro expansion.
java.lang.IllegalArgumentException: requirement failed
	at scala.Predef$.require(Predef.scala:324)
	at kyo.internal.TagMacro$.loop$1(TagMacro.scala:136)
	at kyo.internal.TagMacro$.visit$1(TagMacro.scala:150)
	at kyo.internal.TagMacro$.$anonfun$6(TagMacro.scala:105)
	at scala.collection.immutable.List.map(List.scala:247)
	at kyo.internal.TagMacro$.loop$1(TagMacro.scala:105)
	at kyo.internal.TagMacro$.visit$1(TagMacro.scala:150)
	at kyo.internal.TagMacro$.$anonfun$6(TagMacro.scala:105)
	at scala.collection.immutable.List.map(List.scala:247)
	at kyo.internal.TagMacro$.loop$1(TagMacro.scala:105)
	at kyo.internal.TagMacro$.visit$1(TagMacro.scala:150)
	at kyo.internal.TagMacro$.deriveDB(TagMacro.scala:157)
	at kyo.internal.TagMacro$.deriveImpl(TagMacro.scala:19)
```

### Steps to Reproduce

https://scastie.scala-lang.org/LB328Vv5QsK3lS6bKicq4A

PR #1370, opened 2025-07-11
[data] reproduce TagMacro issue with opaque types + variance

hearnadam:
### Problem
Bug #1368

### Solution
For now, just reproduce the issue.

### DIFF
--- a/kyo-data/shared/src/test/scala/kyo/TagTest.scala
+++ b/kyo-data/shared/src/test/scala/kyo/TagTest.scala
@@ -1057,6 +1057,18 @@ class TagTest extends Test:
         test[C, A & B]
     }
 
+    opaque type V <: Vector[Any] = Vector[Any]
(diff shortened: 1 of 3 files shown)


kyo-data/shared/src/test/scala/kyo/TagTest.scala
```diff
@@ -1057,6 +1057,17 @@ class TagTest extends Test:
         test[C, A & B]
     }
 
+    "opaque type bounds with variance (bug #1368)" in {
+        opaque type V <: Vector[Any] = Vector[Any]
+        abstract class Variant[+A]:
+          def method[AA >: A](using Tag[AA])
```
hearnadam:
```suggestion
          def method[AA >: A](using Tag[AA]): Unit
```

ahoy-jon (approved):

Issue #1377, opened 2025-07-14
error propagation in Async.gather


johnhungerford:
### Version

1.0-RC1+7-326ab911+20250714-1255-SNAPSHOT

### Scala Version

3.7.1

### Expected Behavior

The following test should pass
```
            "Async.gather propagates failures" in run {
                AtomicRef.init[Set[Int]](Set.empty).map: ref =>
                    Abort.run[String](
                        Async.gather((0 to 10).map(i => if i == 5 then Abort.fail("failure") else ref.getAndUpdate(_ + i).unit))
                    ).map: result =>
                        ref.get.map: set =>
                            assert(result == Result.Failure("failure") && set.size < 49)
            }
```

### Actual Behavior

The result is a chunk of 49 `()` (unit) values. This means the error is not propagated, and all successful values are included in the result.

This may be intended, but it means `Async.gather` has different behavior from rest of the `Async` collections API without any documentation.

If it is intended, there should be a variant that allows propagating errors like `Async.foreach`.

### Steps to Reproduce

Past the above test in `AsyncTest` in the kyo-core test suite and run `sbt kyo-core/testOnly **AsyncTest -- -z "Async.gather propagates failures"`

### Current Workaround

`Async.foreach(Seq(effect1, effect2, ...))(identity)`

fwbrasil:
The behavior is actually intentional and briefly documented in the scaladoc but not clearly. The method was designed to return successful results and fail only when all computations fail. I'm using it for scenarios where an LLM evaluates multiple possible executions where failures are typically expected.

That said, I feel we're missing something more high-level for how to define these patterns of async operations.

fwbrasil:
The original PR had better scaladocs, it seems we lost them at some point: https://github.com/getkyo/kyo/pull/860/files#diff-9346f8b93b46e185b02d2937bf06e79a461a46c609f1257a81310c1a99949390R241

Issue #1258, opened 2025-06-12
Add Abort tests to concurrent stream methods


johnhungerford:
Pretty much all the concurrent streaming methods in `StreamCoreExtensions` require propagating errors from a background fiber to the emitting fiber. There should be at least one test per method ensuring that this works correctly.

PR #1378, opened 2025-07-14
Stream: parallel map improvements

johnhungerford:
Addresses #1258 

### Problem

Regarding `mapPar`/`mapChunkPar` and their `-Unordered` variants:
1. The existing implementations have a memory leak: they do not bound the lag between long running transformations and subsequent transformations, allowing an arbitrary number of fibers to accumulate
2. There are no error handling tests
3. Now that there is `Channel#closeAwaitEmpty`, we no longer need to signal end-of-stream

### Solution

1. Use `Meter` to limit concurrency and use a bounded queue of fibers to track running transformations. For `mapPar` and `mapChunkPar`, we can get away with a single channel. For the `-Unordered` variants, use two channels, one simply to track fibers, and the other to publish their results
2. Added simple error handling tests. Turns out `Async.gather` was not propagating errors as I thought it was. See #1377 
3. No longer use `Maybe.Absent` to signal end of stream: call `channel.closeAwaitEmpty` when producer is finished and handle `Abort[Close]` on consumers.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Text.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Text.scala
@@ -549,7 +549,7 @@ object Text:
           */
         def dropUntilNext(p: Predicate): Text =
             val idx = self.indexWhere(!p(_))
-            if idx == -1 then Text.empty else self.drop(idx + 1)
+            if idx == -1 then Text.empty else self.drop(idx)
 
     end extension
 
--- a/kyo-data/shared/src/test/scala/kyo/TextTest.scala
+++ b/kyo-data/shared/src/test/scala/kyo/TextTest.scala
@@ -661,7 +661,7 @@ class TextTest extends Test:
         }
 
         "dropUntilNext" in {
-            assert(Text("hello123world").dropUntilNext(_.isLetter).toString == "23world")
+            assert(Text("hello123world").dropUntilNext(_.isLetter).toString == "123world")
             assert(Text("hello").dropUntilNext(_.isLetter).isEmpty)
             assert(Text("").dropUntilNext(_.isLetter).isEmpty)
         }


kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -380,54 +376,51 @@ object StreamCoreExtensions:
         ): Stream[V2, Abort[E] & Async & S & S2] =
             given CanEqual[Boolean | Chunk[V2], Boolean | Chunk[V2]] = CanEqual.derived
             Stream[V2, S & S2 & Abort[E] & Async]:
```
hearnadam, line 378: Would you be able to avoid the `use` methods and directly handle instantiation + cleanup?

```scala
Stream[A, S]:
  Sync.Unsafe:
    val channel = 
    val meter = 
```

We could optionally add an `Unsafe` method to `StreamCoreExtensions`, essentially Stream.apply+Sync.Unsafe

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -380,54 +376,51 @@ object StreamCoreExtensions:
...
-                            stagingChannel.take.map: fiber =>
-                                fiber.get.map:
-                                    case true             => Loop.continue
-                                    case false            => channelOut.put(Absent).andThen(Loop.done)
-                                    case chunk: Chunk[V2] => channelOut.put(Present(chunk)).andThen(Loop.continue)
-
-                        // Run stream and staging handlers in background, handling failures (end stream)
+                                    // Fork async generation of chunks and publish fiber to output channel
+                                    Fiber.init(Async.foreach(input)(v => semaphore.run(f(v)))).map: chunkFiber =>
+                                        channelOut.put(chunkFiber).andThen:
```
hearnadam: this unfortunately limits concurrency to the chunk size.

I think it may be better to essentially write:

```
channel.putBatch(chunk)
...
channel.drainUpTo(n).map(Async.foreach(_, n)(v => f(v))
```

Do you think that structure could work?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -21,18 +21,14 @@ object StreamCoreExtensions:
         Abort.run(emit).unit
     end emitMaybeChunksFromChannel
 
-    private def emitMaybeElementsFromChannel[V](channel: Channel[Maybe[V]])(using Tag[Emit[Chunk[V]]], Frame) =
+    private def emitElementsFromChannel[V](channel: Channel[V])(using Tag[Emit[Chunk[V]]], Frame) =
         val emit = Loop.foreach:
```
steinybot: Should this be `Loop.forever` since there is never a `Loop.done` and it only stops with an `Abort[Closed]`?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -21,18 +21,14 @@ object StreamCoreExtensions:
...
-                    Emit.valueWith(publishedChunk):
-                        if publishedChunk.size == fullChunk.size then
-                            Loop.continue
-                        else Loop.done
-        Abort.run(emit).unit
-    end emitMaybeElementsFromChannel
+                Abort.recover[Closed](_ => Chunk.empty)(channel.drain).map: chunk =>
+                    val fullChunk = Chunk(v).concat(chunk)
+                    Emit.valueWith(fullChunk)(Loop.continue)
+        Abort.run[Closed](emit).unit
```
steinybot, line 30: Not directly related to this PR but I wonder if there should be an `Abort.ignore[Closed]` or something

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
steinybot, line 378: Is there a reason not to use channel.stream?
Oh no that doesn't work does it because then it'd be inside the Channel.use and the channel would be closed before the stream in consumed. The way this works with Emit seemingly "escaping" the use still blows my mind 🤯. Doing Java and C# for my day job is dumbing me down 😅.
In that case, is there a reason not to expose channel.emitChunks and use that instead of emitElementsFromChannel?

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
steinybot: That would only run in parallel bursts though right? `Async.foreach` will only complete once they are all done so if there is one slow computation then there isn't other work being done.

Having a channel for the work seems like the way to go although I haven't fully groked how this should work yet.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: Directly handling cleanup means wrapping everything in `Sync.ensure` (or some equivalent, like `Abort.run`), which `use` does anyways. How much are we getting out of the unsafe API in a case like this? My sense is you're reducing the overhead by the equivalent of two or three `map`s.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: This doesn't limit the concurrency to the chunk size. While the element transformations are processed in parallel, constrained by `semaphore`, the chunk transformation is also forked. Note that `chunkFiber` is put into `channelOut`, not the chunk itself.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: In most of these methods (including `mapPar`) I'm not actually streaming from a channel of elements, but a channel of element chunks or even (in this case) a channel of fibers of element chunks. I think `mapParUnordered` is the only case where I could reuse `channel.emitChunks`, which I could do by using `Stream.unwrap`. This would provide minor overhead with an additional allocation, however, and I didn't want to expose `channel.emitChunks` since it would bloat the `Channel` API. (Might be worth making a public `channel.emitElements` though.)

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: Multiple chunk transformations can run concurrently, with each chunk having one or more element transformations running concurrently within them

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
hearnadam: I just was hoping to avoid deeply nested code, less than avoiding a few maps.

fwbrasil: Unsafe always comes with some risk. For example, the `AllowUnsafe` evidence of `Sync.Unsafe` could end up hiding unsafe operations that should require suspensions in the rest body. I'd only do it only when optimizing code.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -380,54 +376,55 @@ object StreamCoreExtensions:
...
-                Channel.use[Maybe[Chunk[V2]]](bufferSize): channelOut =>
-                    // Staging channel acts as concurrency limiter: contains fibers that either contain a
-                    // chunk to be published, or a signal to continue or end
-                    Channel.use[Fiber[Boolean | Chunk[V2], Abort[E | Closed] & S & S2]](parallel - 1): stagingChannel =>
-                        // Handle original stream by running transformations in parallel, limiting concurrency by
-                        // using staging channel as a limiter
+                // Emit from channel of fibers to allow parallel transformations while preserving order
+                Channel.use[Fiber[Chunk[V2], Async & Abort[E | Closed] & S & S2]](bufferSize): channelOut =>
+                    // Concurrency limiter
+                    Meter.useSemaphore(parallel): semaphore =>
```
fwbrasil, line 382: these new `use` methods are great! I'll start using them as well

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -380,54 +376,55 @@ object StreamCoreExtensions:
...
-                                fiber.get.map:
-                                    case true             => Loop.continue
-                                    case false            => channelOut.put(Absent).andThen(Loop.done)
-                                    case chunk: Chunk[V2] => channelOut.put(Present(chunk)).andThen(Loop.continue)
-
-                        // Run stream and staging handlers in background, handling failures (end stream)
-                        val background = Fiber.initUnscoped:
+                                    // Fork async generation of chunks (with each transformation limited by semaphore)
+                                    // and publish fiber to output channel. Wait for concurrency using semaphore first to
+                                    // backpressure handler loop
```
steinybot, line 398: 🤩 Nice!

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -380,54 +376,55 @@ object StreamCoreExtensions:
...
             given CanEqual[Boolean | Chunk[V2], Boolean | Chunk[V2]] = CanEqual.derived
             Stream[V2, S & S2 & Abort[E] & Async]:
-                Channel.use[Maybe[Chunk[V2]]](bufferSize): channelOut =>
-                    // Staging channel acts as concurrency limiter: contains fibers that either contain a
-                    // chunk to be published, or a signal to continue or end
-                    Channel.use[Fiber[Boolean | Chunk[V2], Abort[E | Closed] & S & S2]](parallel - 1): stagingChannel =>
-                        // Handle original stream by running transformations in parallel, limiting concurrency by
-                        // using staging channel as a limiter
+                // Emit from channel of fibers to allow parallel transformations while preserving order
+                Channel.use[Fiber[Chunk[V2], Async & Abort[E | Closed] & S & S2]](bufferSize): channelOut =>
```
steinybot, line 380: Is there a need to bound this channel? Or if there is then couldn't it be the same as the parallelism?

My understanding is that if there are small chunks and the buffer size is less than the parallelism, then won't the handler loop backpressure before reaching max parallelism? In other words, the effective parallelism is `min(chunk.size * bufferSize, parallel)`.

As a (hypothetical) user of this API I don't really know how to choose this buffer size. I'd probably always set it to the same as `parallel`.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -380,54 +376,55 @@ object StreamCoreExtensions:
...
+                                onSuccess = _ => channelOut.closeAwaitEmpty.unit,
                                 onFail = {
                                     case _: Closed       => bug("buffer closed unexpectedly")
-                                    case e: E @unchecked => Abort.run(channelOut.put(Absent)).andThen(Abort.fail(e))
+                                    case e: E @unchecked => cleanup.andThen(Abort.fail(e))
                                 },
-                                onPanic = e => Abort.run(channelOut.put(Absent)).andThen(Abort.panic(e))
-                            )(Async.gather(handleEmit, handleStaging))
+                                onPanic = e => cleanup.andThen(Abort.panic(e))
+                            )(handleEmit)
```
steinybot, line 413: Is this the same (minus the redundant `cleanup`) as:
```scala
Abort.run[Closed](
    Sync.ensure(cleanup):
        handleEmit.andThen(channelOut.closeAwaitEmpty)
).unit
```

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -467,47 +464,62 @@ object StreamCoreExtensions:
...
+                            val cleanup = Abort.run[Closed]:
+                                Sync.ensure(channelPar.close.andThen(channelOut.close)):
+                                    Loop.foreach:
+                                        channelPar.drain.map: chunk =>
+                                            if chunk.isEmpty then Loop.done
+                                            else Kyo.foreach(chunk)(_.interrupt).andThen(Loop.continue)
+
+                            // Handle original stream, asynchronously transforming input and publishing output
+                            // using semaphore as rate limiter
+                            val handleEmit = ArrowEffect.handleLoop(t1, stream.emit)(
```
steinybot, line 484: Getting very nit picky here but it'd be nice if these were in the same order as the other but that's just me being OCD.

On a similar note, it is over the rule-of-three so perhaps there is enough duplication in the cleanup and background to warrant refactoring?

steinybot (approved):
🤩 This is really awesome! I finally get it now. Very impressive speed up too and such low error margin.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: No -- cleanup must only be executed in the error case.

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: We need to bound this channel because a single long-running transformation can tie up a single concurrency slot while allowing subsequent transformations can keep running indefinitely. That one long-running transformation will prevent the channel from being drained while it continues getting filled with other fibers. This was one of the issues with my previous implementation (see problem (1) in my description above).

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
johnhungerford: I updated the order and made a couple other things more consistent. There are subtle differences between the four implementations that make abstraction marginally less worth it, IMO.

PR #1379, opened 2025-07-14
Scoped Fiber init

### LINKED ISSUE #1167: [feature]: Scoped fibers
### Problem Statement

Kyo fibers in their current state are fairly dangerous. Processes started in the background can remain running long after the parent fiber that forked them has ended (successfully or otherwise). This creates a substantial risk for memory leaks and other undesirable behavior.

### Proposed Solution

My own preference is for the pattern adopted by ZIO:
- `Async.run` should scope forked fibers to the lifecycle of the parent fiber _by default_. I.e., when the parent fiber completes, forked fibers are interrupted. Presumably this can be accomplished easily by adding `IO.ensure(fiber.interrupt):`
- A separate `Async.runScoped` would use the `Resource` effect to manage the fiber, allowing the fiber to live past the lifetime of the parent fiber without sacrificing resource safety.
- A third `Async.runDaemon` method would produce the existing behavior, but force users to be more explicit about it.

### Alternative Solutions

At a minimum, we should add `Async.runScoped`, which is already in kyo-combinators. Since this adds overhead compared with `IO.ensure`, we would probably still a want a `Resource`-free variant as well. I guess the question is whether this should be the _default_.

### Current Workaround

Use kyo-combinators:

```scala
val effect2 = effect1.forkScoped
```

or use `Resource` directly:

```scala
val effect2 = Resource.acquireRelease(Async.run(effect1))(_.interrupt)
```

### Additional Context

_No response_

johnhungerford:
Fixes #1167 

### Problem

By default, `Fiber.init` forks a fiber without ever guaranteeing it will be interrupted. The result is that fibers can easily end up lost in an application in a permanently suspended or long-running state. It would be good to have a behavior somewhat like ZIO, that guarantees cleanup of fibers

### Solution

Following the pattern established by #1168 , we change `Fiber.init` to introduce the `Scope` effect, interrupting the forked fiber in the registered finalizer; we add a `Fiber.use` method to use a forked fiber within a function and clean it up afterwards; and finally, we add `Fiber.initUnscoped` to produce the previous behavior, which forks a fiber and never guarantees interruptions.

### Notes

While this change was inspired by ZIO's behavior, it departs from ZIO in that by default ZIO binds each forked fiber to the parent fiber and interrupts child fiber whenever its parent fiber completes. I chose not to try to reproduce this for two reasons:
1. It would require some deeper changes to the runtime
2. I don't actually like ZIO's behavior so much, because it means the behavior of async effects depends on how they are composed. See for instance:

```scala
import zio.stream._
import zio._

Unsafe.unsafe { implicit unsafe =>
  val eff1 = for
    fiberA <- (ZIO.sleep(1.second) *> Console.printLine("A")).fork
    fiberB <- (ZIO.sleep(1.second) *> Console.printLine("B")).fork
    _ <- Console.printLine("1")
  yield ()

  val eff2 = eff1 *> (ZIO.sleep(5.seconds) *> Console.printLine("2"))

  val eff3 = eff1.fork *> (ZIO.sleep(5.seconds) *> Console.printLine("3"))

  val eff4 = eff1.fork.flatMap(f => (ZIO.sleep(5.seconds) *> f.join *> Console.printLine("4")))

  Runtime.default.unsafe.run(eff1).getOrThrowFiberFailure()
  Runtime.default.unsafe.run(eff2).getOrThrowFiberFailure()
  Runtime.default.unsafe.run(eff3).getOrThrowFiberFailure()
  Runtime.default.unsafe.run(eff4).getOrThrowFiberFailure()
}
```

`eff1` on its own will just print `1`
`eff2` will print `1`, `A`, `B`, `2`
`eff3` will print `1`, `2`
`eff4` will print `1`, `4`

If you construct an effect that depends on ephemeral forked tasks, it will behave properly in a test or when used in an application in a forked handler, but when you sequence it other effects it may behave differently.

In conclusion, I think introducing the notion of parent fiber and child fiber makes it harder to reason about the fiber lifecycle than if fibers are just treated the same way as other resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2244,7 +2244,7 @@ import kyo.*
 // taken by reference and automatically
 // suspended with 'Sync'
 val a: Fiber[Int, Any] < Sync =
-    Fiber.init(Math.cos(42).toInt)
+    Fiber.initUnscoped(Math.cos(42).toInt)
 
 // It's possible to "extract" the value of a
 // 'Fiber' via the 'get' method. This is also
@@ -2384,11 +2384,11 @@ import kyo.*
 
 // An example computation with fibers
 val a: Int < Async =
-    Fiber.init(Math.cos(42).toInt).map(_.get)
+    Fiber.initUnscoped(Math.cos(42).toInt).map(_.get)
 
 // Avoid handling 'Async' directly
 val b: Fiber[Int, Any] < Sync =
-    Fiber.init(a)
+    Fiber.initUnscoped(a)
 
 // The 'runAndBlock' method accepts
 // arbitrary pending effects but relies
@@ -2415,7 +2415,7 @@ val b: Boolean < Sync =
 // Fullfil the promise with
 // another fiber
 val c: Boolean < Sync =
-    a.map(fiber => Fiber.init(1).map(fiber.become(_)))
+    a.map(fiber => Fiber.initUnscoped(1).map(fiber.become(_)))
 ```
 
 > A `Promise` is basically a `Fiber` with all the regular functionality plus the `complete` and `become` methods to manually fulfill the promise.
@@ -2796,12 +2796,12 @@ val d: Unit < Async =
 val e: Unit < Async =
     for
         barrier <- Barrier.init(3)
-        fiber1  <- Fiber.init(Async.sleep(1.second))
-        fiber2  <- Fiber.init(Async.sleep(2.seconds))
+        fiber1  <- Fiber.initUnscoped(Async.sleep(1.second))
+        fiber2  <- Fiber.initUnscoped(Async.sleep(2.seconds))
         _       <- Async.zip(
                      fiber1.get.map(_ => barrier.await),
                      fiber2.get.map(_ => barrier.await),
-                     Fiber.init(barrier.await).map(_.get)
+                     Fiber.initUnscoped(barrier.await).map(_.get)
                    )
     yield ()
 ```
@@ -3400,14 +3400,14 @@ val a: Int < (Abort[Nothing] & Async) =
     for
         v1 <- ZIOs.get(ZIO.succeed(21))
         v2 <- Sync.defer(21)
-        v3 <- Fiber.init(-42).map(_.get)
+        v3 <- Fiber.initUnscoped(-42).map(_.get)
     yield v1 + v2 + v3
 
 // Using fibers from both libraries
 val b: Int < (Abort[Nothing] & Async) =
     for
         f1 <- ZIOs.get(ZIO.succeed(21).fork)
-        f2 <- Fiber.init(Sync.defer(21))
+        f2 <- Fiber.initUnscoped(Sync.defer(21))
         v1 <- ZIOs.get(f1.join)
         v2 <- f2.get
     yield v1 + v2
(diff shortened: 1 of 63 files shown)


README.md
```diff
@@ -2384,11 +2384,11 @@ import kyo.*
 
 // An example computation with fibers
 val a: Int < Async =
-    Fiber.init(Math.cos(42).toInt).map(_.get)
+    Fiber.initUnscoped(Math.cos(42).toInt).map(_.get)
 
 // Avoid handling 'Async' directly
 val b: Fiber[Int, Any] < Sync =
-    Fiber.init(a)
+    Fiber.initUnscoped(a)
```
ahoy-jon, line 2391: @johnhungerford `Fiber.init` ? There are a lot of cases where it should be `Fiber.init`

ahoy-jon (approved):
README.md / tests could use `Fiber.init` more. However, good to go for me.

README.md
fwbrasil: +1 I think we should prefer the scoped versions in the readme

kyo-core/shared/src/main/scala/kyo/Clock.scala
```diff
@@ -537,7 +537,7 @@ object Clock:
         frame: Frame,
         reduce: Reducible[Abort[E]]
     ): Fiber[A, reduce.SReduced] < (Sync & S) =
-        Fiber.init {
+        Fiber.initUnscoped {
```
fwbrasil, line 540: I don't think we need to address everything in this PR but, in some cases like this one, it makes more sense to add the pending `Scope`

scalafix/rules/src/main/scala/kyo/rules/KyoUpdateToV1_0.scala
```diff
@@ -44,6 +44,13 @@ class KyoUpdateToV1_0 extends SemanticRule("KyoUpdateToV1_0") {
 
             case resource @ q"Resource" if resource matches "kyo.Resource" =>
                 Patch.replaceTree(resource, "Scope")
+
+            // Async combinator
+            case fork @ q"$eff.fork" =>
+                Patch.replaceTree(fork, s"$eff.forkUnscoped")
+
+            case forkScoped @ q"$eff.forkScoped" =>
+                Patch.replaceTree(forkScoped, s"$eff.fork")
```
ahoy-jon, line 53: Nice! I need to alter a bit the setup to test from 1.0-RC1 to main. (I will update in another PR)

scalafix/output/src/main/scala/fix/TestKyoUpdateToV1_0.scala
```diff
@@ -9,7 +9,7 @@ object TestKyoUpdateToV1_0:
         init.now
         Console.printLine("hello").now
 
-    val async: Fiber[Nothing, Unit] < Async = Fiber.initUnscoped(Async.defer(Async.defer(Kyo.unit)))
+    val async: Fiber[Nothing, Unit] < Async = /* Consider using Fiber.init or Fiber.use to guarantee termination */ Fiber.initUnscoped(Async.defer(Async.defer(Kyo.unit)))
```
ahoy-jon, line 12: I will try to rework the rule to be able to do something like
```scala
/* Consider using Fiber.init or Fiber.use to guarantee termination */
val async: Fiber[Nothing, Unit] < Async = Fiber.initUnscoped(Async.defer(Async.defer(Kyo.unit)))
```

fwbrasil (approved):

kyo-core/shared/src/main/scala/kyo/Clock.scala
johnhungerford: Agreed, I'll take a look through and see if there are some obvious other ones to update. I did that with the other two resource safety PRs.

scalafix/rules/src/main/scala/kyo/rules/KyoUpdateToV1_0.scala
johnhungerford: Yeah I was considering adding new rules just for RC1 -> 1.0 but would get a little complicated. Should we have separate rules for each RC?

ahoy-jon: Yes, it's better, and we can keep tests for the last two rules.

PR #1388, opened 2025-07-18
[kernel] isolate variance for restore + convenience isolate.nest method

fwbrasil:
### Problem

`Isolate`'s new `Restore` type parameter is currently invariant and prevents referring to an isolate with fewer restored effects as an isolate with more effects, which is ok given that the pending set is contravariant. Additionally, we currently don't have a convenient way to separate a specific effect out of the pending ones.

### Solution

Make `Restore` contravariant and add `isolate.nest`.

### Notes

- I had issues with the compiler inferring `Any` for `Restore`, which is too restrictive and fails in a few scenarios. I had to add a `Restore <: Remove` constraint in the `given` to help the compiler.
- I've renamed the second `IsolateTest` in `kyo-prelude` to `IsolatePreludeTest` to avoid confusion.

### DIFF
--- a/kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
@@ -124,18 +124,18 @@ abstract class Isolate[Remove, -Keep, -Restore]:
       */
     def restore[A, S](v: Transform[A] < S)(using Frame): A < (Restore & S)
 
-    /** Isolates Remove effects while exposing them as nested Restore effects.
+    /** Isolates 'Remove' effects while exposing them as nested 'Restore' effects.
       *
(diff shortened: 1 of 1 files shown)


kyo-kernel/shared/src/main/scala/kyo/kernel/Isolate.scala
```diff
@@ -124,6 +124,24 @@ abstract class Isolate[Remove, -Keep, Restore]:
       */
     def restore[A, S](v: Transform[A] < S)(using Frame): A < (Restore & S)
 
+    /** Isolates Remove effects while exposing them as nested Restore effects.
```
hearnadam: Might help readability in at least this initial sentence, but could be used throughout.
```suggestion
    /** Isolates 'Remove' effects while exposing them as nested 'Restore' effects.
```

hearnadam (approved):

PR #1389, opened 2025-07-18
[data][prelude] allow bypassing dynamic tag check + support dynamic tags in Aspect.init

fwbrasil:
### Problem

For performance reasons, we don't allow dynamic tags to be inferred within the `kyo` package. This is a good optimization in general but it isn't a strict requirement. For example, in https://github.com/getkyo/kyo/pull/1305 we had to inline methods that don't make sense to inline to work it around, which isn't ideal.

### Solution

Introduce `Tag.dynamic` to allow bypassing the check within the `kyo` package and use it in `Aspect.init` to avoid the unnecessary inlining in https://github.com/getkyo/kyo/pull/1305. It seems reasonable as an exceptional case.

### DIFF
--- a/kyo-prelude/shared/src/main/scala/kyo/Aspect.scala
+++ b/kyo-prelude/shared/src/main/scala/kyo/Aspect.scala
@@ -32,7 +32,7 @@ import Aspect.*
   * @tparam S
   *   The effect type - what effects the aspect can perform
   */
-final class Aspect[Input[_], Output[_], S] private[kyo] (
+final class Aspect[Input[_], Output[_], S](
     using
     tag: Tag[(Input[Any], Output[Any])],
     frame: Frame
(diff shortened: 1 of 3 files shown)


kyo-prelude/shared/src/main/scala/kyo/Aspect.scala
```diff
@@ -32,7 +32,7 @@ import Aspect.*
   * @tparam S
   *   The effect type - what effects the aspect can perform
   */
-final class Aspect[Input[_], Output[_], S] private[kyo] (
```
fwbrasil, line 35: the compiler started complaining about the constructor needing `@publicInBinary` but I couldn't find out a way to add the annotation to the constructor. I've made it public instead since it doesn't seem problematic.

hearnadam (approved):

PR #1393, opened 2025-07-21
Ensure parallel map `Closed` error propagation

### LINKED ISSUE #1387: [BUG]: `Closed` error propagation in `Stream` parallel map functions
### Version

1.0-RC1+10-2e5d5c39+20250716-0923-SNAPSHOT

### Scala Version

3.7.1

### Expected Behavior

The following test should pass:
```scala
"should propagate Abort[Closed] in f" in run {
    val failure = Closed("fail", summon[Frame])
    val stream  = Stream.init(1 to 4).concat(Stream.init(5 to 8)).concat(Stream.init(9 to 12))
    val stream2 = stream.mapPar(i => if i == 5 then Abort.fail(failure) else i + 1)
    Abort.run(stream2.run).map(res => Result.Failure(failure))
}
```

### Actual Behavior

The test hangs and times out

### Steps to Reproduce

Add test to `StreamCoreExtensionsTest.scala` and run `sbt kyo-core/testOnly **StreamCoreExtensionsTest -- -z "should propagate Abort[Closed] in f"`

### Current Workaround

When using any `map*Par*` methods, handle `Closed` errors in the transformation functions (e.g., map them to some other error type and then map back again after)

johnhungerford:
Fixes #1387 

### Problem

For concurrent stream transformations using `Channel` and/or `Meter`, there is a bug (#1387) where if the transformation effect `f` fails with `Abort[Closed]`, that failure is interpreted as end-of-stream.

### Solution

Added `Abort.withMask` which allows you to mask an `Abort` failure type that may or may not be within a generic type intersection to protect it from being handled accidentally.

### Notes

A couple significant changes to make this work:
1. `ConcreteTag` derivation now does an implicit search for `createSingle` so that the tag for a union can be derived from tags of single types
2. `Fiber.init` and other `Async` methods do not use `Reducible` to eliminate `Abort[Nothing]`. Instead, the `S` type parameter in `Fiber` is constrained to being a subtype of `Abort[Nothing]` to reflect the coupling of `Fiber` to failures.


kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -38,7 +38,7 @@ import scala.util.NotGiven
   * @see
   *   [[Fiber.Unsafe]] for low-level operations requiring [[AllowUnsafe]]
   */
-opaque type Fiber[+A, -S] = IOPromiseBase[Any, A < (Async & S)]
+opaque type Fiber[+A, -S <: Abort[Nothing]] = IOPromiseBase[Any, A < (Async & S)]
```
hearnadam: Maybe we should make this change separately? We probably need Scalafix rule? @ahoy-jon

README.md
```diff
@@ -1900,26 +1900,26 @@ val a: Unit < Sync =
 
 // Recurring task with a delay between
 // executions
-val b: Fiber[Unit, Any] < Sync =
+val b: Fiber[Unit, Abort[Nothing]] < Sync =
```
fwbrasil: I think It'd be important to not track `Abort[Nothing]` here. Sorry, I think I indicated that it'd be fine to remove `Reducible` from `Fiber.init` but I thought this would still be `Any` since the underlying impl is `IOPromiseBase[Any, A < (Async & S)]` but I'm not sure it's possible.

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -455,7 +455,6 @@ object Actor:
                     Scope.run,                  // Close used resources
                     Fiber.init                  // Start the actor's processing loop in an async context
                 )
-            _ <- Scope.ensure(mailbox.close) // Registers a finalizer in the outer scope to provide the actor hierarchy behavior
```
fwbrasil: why is this not necessary anymore?

kyo-data/shared/src/main/scala/kyo/internal/SafeClassTagMacro.scala
```diff
@@ -50,45 +50,50 @@ private[kyo] object SafeClassTagMacro:
                             case _             => Chunk(tpe)
```
fwbrasil, line 50: We need to rename from `SafeClassTagMacro` to `ConcreteTagMacro`

kyo-data/shared/src/main/scala/kyo/internal/SafeClassTagMacro.scala
```diff
@@ -50,45 +50,50 @@ private[kyo] object SafeClassTagMacro:
...
-                            case x: Int     => Expr(x)
-                            case x: Long    => Expr(x)
-                            case x: Float   => Expr(x)
-                            case x: Double  => Expr(x)
-                            case x: Boolean => Expr(x)
-                            case x: Char    => Expr(x)
-                            case x: String  => Expr(x)
-                            case x          => report.errorAndAbort(s"Unsupported literal type: $x")
-                    '{ LiteralTag($value) }
+            Implicits.search(TypeRepr.of[[A] =>> ConcreteTag[A]].appliedTo(tpe)) match
```
fwbrasil, line 58: The macro change looks ok to me, it should have been this way from the beginning. What do you think about separating this change in a new PR and adding tests for it?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -603,4 +603,61 @@ object Abort:
 
     end literal
 
+    sealed class CanMaskAbort
```
fwbrasil, line 650: I can't see the purpose of this class. Maybe `MaskAbort .apply` should take it to ensure the masking is scoped?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -603,4 +603,61 @@ object Abort:
...
+
+    /** A function to apply a mask to any failures of type [[E]] than may or may not be possible for a generic error effect [[Abort[EG]]].
+      */
+    sealed abstract class MaskAbort[E, EG]:
+        /* Masked error type */
+        type Masked <: Product
```
fwbrasil: why the `Product` restriction?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -603,4 +603,61 @@ object Abort:
...
+
+        /** @param effect
+          *   Effect including [[Abort[EG]]] which may or may not include error subtype [[E]]
+          * @return
+          *   An effect with any [[Abort[E]]] failures masked as [[Masked]]
+          */
+        def apply[A, S](effect: A < (Abort[EG] & S)): A < (Abort[EG | Masked] & S)
+
+        given tag: ConcreteTag[Masked]
```
fwbrasil, line 665: it seems this doesn't need to be in the abstract class?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
```diff
@@ -603,4 +603,61 @@ object Abort:
...
+        case class Wr(e: E)
+
+        val wrTag: ConcreteTag[Wr] = summon
+
+        val mask = new MaskAbort[E, EG]:
+            type Masked = Wr
+            def apply[A, S](effect: A < (Abort[EG] & S)): A < (Abort[EG | Masked] & S) =
+                Abort.runPartial[E](effect).map:
+                    case Result.Success(a) => a
+                    case Result.Failure(e) => Abort.fail(Wr(e))
```
fwbrasil, line 693: Nice insight :) I'm wondering if we could have a more generic version of this mechanism since it's essentially temporarily swapping the tag. I couldn't test it and there's some type safety challenges but I wonder if this could be a good path:

```scala
sealed trait Mask[E] extends ArrowEffect[Const[Any], Const[Any]]

object Mask:

    def apply[E <: ArrowEffect[?, ?]](using Frame)[A, S](v: A < (E & S))(using eTag: Tag[E], maskTag: Tag[Mask[E]]): A < (Mask[E] & S) =
        ArrowEffect.handle(eTag, v) {
            [C] => (input, cont) => ArrowEffect.suspendWith(maskTag, input)(cont)
        }

    def run[E <: ArrowEffect[?, ?]](using Frame)[A, S](v: A < (Mask[E] & S))(using eTag: Tag[E], maskTag: Tag[Mask[E]]): A < (E & S) =
        ArrowEffect.handle(maskTag, v) {
            [C] => (input: Any, cont) => ArrowEffect.suspendWith(eTag, input)(cont)
        }
```

It seems it should work to prevent `E` suspensions from being handled until `Mask.run`. Ah, but there's also the challenge that `Abort` still uses erased tags, though.

johnhungerford: I tried something like this too, but I actually need to preserve the `Abort` so that I can do error handling. So I can't mask `Abort[E]` as `Mask[Abort[E]]`, but it would have to be `Abort[E]` -> `Abort[Mask[E]]`. That's why we need a specialized approach for abort.

kyo-actor/shared/src/main/scala/kyo/Actor.scala
johnhungerford: Because `Channel.init` above already introduces `Scope`

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: `MaskAbort.apply` does take it. The signature starts with: `CanMaskAbort ?=> ...`

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: I was trying something and forgot to remove it. I'll get rid of it.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: I'll see

README.md
johnhungerford: I think it may still be possible to get rid of `Reducible` without tracking `Abort[Nothing]`. I'll give it a try

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: Or rather, `Abort.withMask` requires it

fwbrasil: but it seems only provided and never required?

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
fwbrasil: given that `Abort[E1] & Abort[E2]` is equivalent to `Abort[E1 | E2]`, it seems `Mask[Abort[E1]]` would work to mask the appropriate failures? That's considering a scenario where `Abort` doesn't use erased tags

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: Oh good point, `MaskAbort.apply` should require it -- you're right

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: `Mask[Abort[E1]]` would mask the failure, but I would not be able to then call `Abort.run` and handle the masked case separately from the unmasked case. The use case is that I'm given some effect `A < (Abort[E] & S)`, and I don't know whether `Closed` is a subtype of `E`. I need to compose this with an effect of type `Unit < Abort[Closed]` effect and then handle *all* errors on it. I need to do something different depending on whether the failure comes from the `Closed` abort on the second effect or whether it's a failure `E` from the first effect, which may very well include `Closed` from some other resource.

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: I couldn't make it work without it here. If you need to call `Abort.run[mask.Masked]` within a `withMask` effect, this is the only way I can figure out to make the `ConcreteTag` available to the compiler.

kyo-data/shared/src/main/scala/kyo/internal/SafeClassTagMacro.scala
johnhungerford: I'll do this in the separate PR

kyo-prelude/shared/src/main/scala/kyo/Abort.scala
johnhungerford: Added

README.md
johnhungerford: We can pull it off by adding a `reduce` method to `Fiber` and `Fiber.Unsafe`

kyo-data/shared/src/main/scala/kyo/internal/SafeClassTagMacro.scala
johnhungerford: See #1397

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -91,14 +90,14 @@ object Fiber:
...
-      * types as the Result, with the error type reduced according to the Reducible instance.
+      * types as the Result.
       *
       * @param result
       *   The Result to create the Fiber from
       * @return
       *   A Fiber that is immediately completed with the provided Result
       */
-    def fromResult[E, A, S](result: Result[E, A < S])(using reduce: Reducible[Abort[E]]): Fiber[A, reduce.SReduced & S] =
+    def fromResult[E, A, S](result: Result[E, A < S]): Fiber[A, Abort[E] & S] =
```
hearnadam, line 100: Does this need a frame?

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -128,10 +127,8 @@ object Fiber:
     )(
         v: => A < (Abort[E] & Async & S)
     )(
-        using
-        reduce: Reducible[Abort[E]],
-        frame: Frame
-    ): Fiber[A, reduce.SReduced & S2] < (Sync & S & Scope) =
+        using frame: Frame
```
hearnadam, line 130: these don't need to be named anymore, since you are not using `reduce` anymore.

kyo-core/shared/src/main/scala/kyo/Fiber.scala
hearnadam: I think this is quite old 🙂

fwbrasil:
@copilot resolve the merge conflicts in this pull request

ahoy-jon:
@fwbrasil fixed the conflict with main

fwbrasil:
closing to avoid noise in the PR list

PR #1396, opened 2025-07-24
[core] Take platform logging module out

hygt:
### Problem

- `kyo-core` should not depend on SLF4J
- Log implementation should be pluggable
- Common façades should be provided

### Solution

- Remove SLF4J dependency in core
- Add `kyo-logging-slf4j` module with SLF4J bridge
- Add `kyo-logging-jpl` module with `System.Logger` bridge


kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -398,7 +398,7 @@ private[kyo] object IOPromise:
             case ex if NonFatal(ex) =>
                 given Frame = Frame.internal
                 import AllowUnsafe.embrace.danger
-                Log.live.unsafe.error("uncaught exception", ex)
+                Log.error("uncaught exception", ex).unsafeGet
```
hygt: Not sure about this. 🙂

kyo-logging-slf4j/jvm/src/main/scala/kyo/logging/LogPlatformSpecific.scala
```diff
@@ -1,17 +1,14 @@
...
-trait LogPlatformSpecific:
-    val live: Log = Log(LogPlatformSpecific.Unsafe.SLF4J("kyo.logs"))
-
 object LogPlatformSpecific:
 
-    /* WARNING: Low-level API meant for integrations, libraries, and performance-sensitive code. See AllowUnsafe for more details. */
+    /** WARNING: Low-level API meant for integrations, libraries, and performance-sensitive code. See AllowUnsafe for more details. */
     object Unsafe:
 
         object SLF4J:
```
hygt: Do we keep the namespacing `kyo.logging.LogPlatformSpecific.Unsafe.SLF4J` or go to something simpler?

johnhungerford: I don't think we need this to be `LogPlatformSpecific` because slf4j is already jvm only. This module should not include and should not publish `js` artifacts. I assuming it's not supported for native either?

I would make the top level object `SLF4JLog` and and add an apply method returning a `Log` instance (see my comment on the `JPL` implementation)

kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -31,7 +29,7 @@ object Log extends LogPlatformSpecific:
         case silent extends Level(60)
     end Level
 
-    private val local = Local.init(live)
+    private val local = Local.init(Log(Unsafe.ConsoleLogger("kyo.logs", Level.debug)))
```
johnhungerford: maybe include `private[kyo] val live: Log = Log(Unsafe.ConsoleLogger("kyo.logs", Level.debug))` so that you can still access it from `IOPromise` and revert your change to `local`?

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
johnhungerford: See other comment -- you can still use `live`, just put it directly in `Log` rather than `PlatformSpecific`

kyo-core/shared/src/test/scala/kyo/LogTest.scala
```diff
@@ -22,21 +22,6 @@ class LogTest extends Test:
...
-        Log.live.unsafe.warn("warn")
-        Log.live.unsafe.error("error")
-        Log.live.unsafe.trace("trace", ex)
-        Log.live.unsafe.debug("debug", ex)
-        Log.live.unsafe.info("info", ex)
-        Log.live.unsafe.warn("warn", ex)
-        Log.live.unsafe.error("error", ex)
-        succeed
-    }
-
```
johnhungerford: You can keep this if you keep live. Otherwise you should add an equivalent using the unsafe API

kyo-logging-jpl/jvm/src/main/scala/kyo/logging/LogPlatformSpecific.scala
```diff
@@ -0,0 +1,59 @@
+package kyo.logging
+
+import System.Logger.Level
+import kyo.AllowUnsafe
+import kyo.Frame
+import kyo.Log
+import kyo.Text
+
+object LogPlatformSpecific:
```
johnhungerford: I'd call this file `JPLLog.scala`, rename `LogPlatformSpecific` -> `JPLLog`, and add
- `apply(name: String): Log = Log(Unsafe.JPL(name))`, and
- `apply(logger: System.Logger): Log = Log(Unsafe.JPL(logger))`

build.sbt
```diff
@@ -386,6 +385,28 @@ lazy val `kyo-actor` =
         .nativeSettings(`native-settings`)
         .jsSettings(`js-settings`)
 
+lazy val `kyo-logging-jpl` =
+    crossProject(JVMPlatform)
+        .withoutSuffixFor(JVMPlatform)
+        .crossType(CrossType.Full)
```
johnhungerford, line 405: This should probably either not be `crossProject` or at least not `CrossType.Full`

build.sbt
```diff
@@ -386,6 +385,28 @@ lazy val `kyo-actor` =
...
+        .in(file("kyo-logging-jpl"))
+        .dependsOn(`kyo-core`)
+        .settings(`kyo-settings`)
+        .jvmSettings(mimaCheck(false))
+
+lazy val `kyo-logging-slf4j` =
+    crossProject(JVMPlatform)
+        .withoutSuffixFor(JVMPlatform)
+        .crossType(CrossType.Full)
```
johnhungerford, line 414: This should probably either not be crossProject or at least not CrossType.Full

johnhungerford (dismissed):
Mostly there -- thanks for the work on this!

kyo-core/shared/src/main/scala/kyo/Log.scala
hearnadam: I think it's fine if it's public, it is today. But otherwise, I agree!

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
hearnadam: For context, since `Log.error` is suspended, the `unsafeGet` would throw (`ClassCastException`).

It's used when the invariant `A | Kyo[A, S]` is known to be `A`.

https://github.com/getkyo/kyo/blob/a287ac733246505010dfed4816fe373578eb7f00/kyo-kernel/shared/src/main/scala/kyo/kernel/Pending.scala#L345-L348

kyo-logging-jpl/jvm/src/main/scala/kyo/logging/LogPlatformSpecific.scala
hearnadam: Or `JavaLogging`? maybe less obtuse.

kyo-logging-jpl/jvm/src/main/scala/kyo/logging/LogPlatformSpecific.scala
```diff
@@ -0,0 +1,59 @@
...
+    /** WARNING: Low-level API meant for integrations, libraries, and performance-sensitive code. See AllowUnsafe for more details. */
+    object Unsafe:
+
+        object JPL:
+            def apply(name: String) = new JPL(System.getLogger(name))
+
+        final class JPL(logger: System.Logger) extends Log.Unsafe:
+
+            val level: Log.Level =
+                if logger.isLoggable(Level.TRACE) then Log.Level.trace
```
hearnadam: Should we have a function to convert between the two level types here? It likely can be encoded as a switch.

Does `logger.level` or equivalent method exist?

kyo-logging-slf4j/jvm/src/main/scala/kyo/logging/LogPlatformSpecific.scala
hearnadam: Yes, `object SLF4J` seems good to me.

build.sbt
hearnadam: I think it can be, as we probably depend on `java-logging` in each of the scala-native/js variants.

hearnadam (commented):
Thank you for your contribution!

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
hygt: Right, I made the change to make scalac happy and not forget to come back to this.
Package private `Log.live` was my first idea. 👍

fwbrasil: Yeah, we should change `unsafeGet` to `private[kernel]` and fix use outside of the package it there's any

build.sbt
fwbrasil: I think having consistent paths is more important than restricting the level of cross compilation to what's needed. I'd say we should always use `CrossType.Full` to help navigation in the codebase

kyo-core/shared/src/main/scala/kyo/Log.scala
fwbrasil: All similar implementations provide a public `live` version: `Console`, `Clock`, etc. It's convenient especially in unsafe code

kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -61,6 +59,16 @@ object Log extends LogPlatformSpecific:
...
     def use[A, S](f: Log => A < S)(using Frame): A < S = local.use(f)
 
+    /** Executes an effect with a custom logger.
+      *
+      * @param v
+      *   The effect to execute with the given logger
+      * @return
+      *   The result of executing the effect with the given logger
+      */
+    def withLogger[A, S](logger: Log)(v: A < S)(using Frame): A < S =
```
fwbrasil: isn't this the same as `let`?

kyo-logging-jpl/jvm/src/main/scala/kyo/logging/LogPlatformSpecific.scala
fwbrasil: I'd suggest `JavaLog`

kyo-logging-jpl/jvm/src/main/scala/kyo/logging/LogPlatformSpecific.scala
```diff
@@ -0,0 +1,59 @@
...
+import kyo.Frame
+import kyo.Log
+import kyo.Text
+
+object LogPlatformSpecific:
+
+    /** WARNING: Low-level API meant for integrations, libraries, and performance-sensitive code. See AllowUnsafe for more details. */
+    object Unsafe:
+
+        object JPL:
```
fwbrasil: I think the `JPL` indirection isn't very valuable. We could have `class Unsafe` and `object Unsafe` directly under `LogPlatformSpecific` (or its new name if we change it)

kyo-logging-slf4j/jvm/src/main/scala/kyo/logging/LogPlatformSpecific.scala
fwbrasil: I vote for `SLF4JLog`

kyo-logging-slf4j/jvm/src/test/scala/kyo/logging/LogPlatformSpecificTest.scala
```diff
@@ -0,0 +1,92 @@
+package kyo.logging
```
fwbrasil: We typically put all APIs under the `kyo` package directly. Any particular reason to move these to a separate package?

fwbrasil:
@hygt nice first contribution! 👏

kyo-logging-jpl/jvm/src/main/scala/kyo/logging/LogPlatformSpecific.scala
hygt: No to both questions unfortunately, you can't get the logger level from the System.Logger façade, and these Java enums won't compile into a `@switch` anyway, I tried in a different project.

kyo-core/shared/src/main/scala/kyo/Log.scala
hygt: Right, I was deceived by the Scaladoc for `let` which is outdated. 😅 
I think keeping `withLogger` as an alias makes sense, but I can remove it if unnecessary.

kyo-logging-jpl/jvm/src/main/scala/kyo/logging/LogPlatformSpecific.scala
hygt: I agree, I will re-organize the constructors as suggested by @johnhungerford

kyo-logging-slf4j/jvm/src/test/scala/kyo/logging/LogPlatformSpecificTest.scala
hygt: I didn't put much thought into it, I just felt the bridges could benefit from the namespacing for discoverability, as they should be instantiated only once in a typical app. That said you should have only one in your dependency tree so this is a bit moot. I can refactor that. 👍

build.sbt
hearnadam: Oh, I just realized this is currently on `crossProject(JVMPlatform)` - I believe it should compile for all 3 platforms.

kyo-logging-jpl/jvm/src/main/scala/kyo/logging/LogPlatformSpecific.scala
hearnadam: Sounds good

kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -31,18 +29,20 @@ object Log extends LogPlatformSpecific:
...
-      * @param u
-      *   The Unsafe logger to use
+      * @param log
+      *   The custom logger to use
       * @param f
       *   The function to execute
       * @return
       *   The result of the function execution
       */
-    def let[A, S](log: Log)(f: => A < S)(using Frame): A < S =
```
hygt, line 45: To be consistent with `withLogger`. I think the by-name `f` was useless anyway?

hearnadam: Looks like `Local` is does not accept by-name. Outside the scope of this change, but maybe it should.

https://github.com/getkyo/kyo/blob/main/kyo-prelude/shared/src/main/scala/kyo/Local.scala#L68

kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -61,6 +61,18 @@ object Log extends LogPlatformSpecific:
...
+    /** Executes an effect with a custom logger. Alias for [[Log.let]].
+      *
+      * @param log
+      *   The custom logger to use
+      * @param v
+      *   The effect to execute with the given logger
+      * @return
+      *   The result of executing the effect with the given logger
+      */
+    def withLogger[A, S](log: Log)(v: A < S)(using Frame): A < S =
```
hearnadam: I don't think we need the alias

build.sbt
```diff
@@ -386,6 +385,28 @@ lazy val `kyo-actor` =
         .nativeSettings(`native-settings`)
         .jsSettings(`js-settings`)
 
+lazy val `kyo-logging-jpl` =
+    crossProject(JVMPlatform)
```
hearnadam, line 403: You may need to add the dependency, but this should work for JS and Native:

`libraryDependencies += "org.scala-native" %%% "scala-native-java-logging" % "1.0.0"`

https://github.com/getkyo/kyo/blob/main/build.sbt#L188

kyo-logging-jpl/jvm/src/main/scala/kyo/JavaLog.scala
```diff
@@ -0,0 +1,61 @@
...
+            val level: Log.Level =
+                if logger.isLoggable(Level.TRACE) then Log.Level.trace
+                else if logger.isLoggable(Level.DEBUG) then Log.Level.debug
+                else if logger.isLoggable(Level.INFO) then Log.Level.info
+                else if logger.isLoggable(Level.WARNING) then Log.Level.warn
+                else if logger.isLoggable(Level.ERROR) then Log.Level.error
+                else Log.Level.silent
+
+            inline def trace(msg: => Text)(using frame: Frame, allow: AllowUnsafe): Unit =
+                if logger.isLoggable(Level.TRACE) then logger.log(Level.TRACE, s"[${frame.position.show}] $msg")
```
hearnadam: Given we have `level` - shouldn't we just use that for consistency with other Unsafe loggers?

`if Level.trace.enabled(level)`?

hygt: Right, I forgot why I didn't simply copy/paste that from the SLF4J implementation. 🤔 

But maybe even better, remove these checks from both bridges? Anyway:

- In the outer `Log` we perform the check already, for instance here it'd be called by `logWhen(Level.trace)(_.trace(msg))`
- The backend will always check too, so even if you call the Unsafe instance directly, it won't matter.

build.sbt
hygt: I don't think it will, this provides JUL but not JPL, I could maybe try to contribute and add the new API...

I don't really see the value of this bridge for Scala.JS and Native users anyway. They'd probably better be served by a `Log` backed by a concrete library (such as Scribe).

kyo-logging-jpl/jvm/src/main/scala/kyo/JavaLog.scala
hearnadam: Sounds fine

fwbrasil (approved):
LGTM. Do you mind adding an entry to CHANGELOG.md with a brief description of the breaking changes? Thanks!

fwbrasil:
Thanks for the updating the changelog. It seems there's a merge issue `kyo.kernel.Platform` was moved to `kyo.internal.Platform`

https://github.com/getkyo/kyo/actions/runs/16607184022/job/46982157001?pr=1396#step:7:3808

kyo-logging-jpl/jvm/src/main/scala/kyo/JavaLog.scala
```diff
@@ -0,0 +1,61 @@
...
+            val level: Log.Level =
+                if logger.isLoggable(Level.TRACE) then Log.Level.trace
+                else if logger.isLoggable(Level.DEBUG) then Log.Level.debug
+                else if logger.isLoggable(Level.INFO) then Log.Level.info
+                else if logger.isLoggable(Level.WARNING) then Log.Level.warn
+                else if logger.isLoggable(Level.ERROR) then Log.Level.error
+                else Log.Level.silent
+
+            inline def trace(msg: => Text)(using frame: Frame, allow: AllowUnsafe): Unit =
+                logger.log(Level.TRACE, s"[${frame.position.show}] $msg")
```
hearnadam, line 30: Can you update the `Console.Unsafe` to avoid the branching on `level`? I think all loggers should follow the same contract of not checking level if we expect the Local to perform that behavior.

hearnadam (approved):

PR #1397, opened 2025-07-24
Derive ConcreteTag unions/intersections from given component tags

johnhungerford:
Prerequisite for #1393. See [discussion](https://github.com/getkyo/kyo/pull/1393#discussion_r2227227541).

### Problem

The derivation macro for `ConcreteTag` currently does not search for given `ConcreteTag` instances for components `A` and `B` in composite types like `A | B` or `A & B`. This forces users to require implicit evidence for the full composite type in cases where only one of the types is generic (e.g., `def someMethod[E](A < Abort[E | Closed])(using ConcreteTag[E | Closed]) ...`). This creates a more severe problem when one of the types in the intersection isn't known in the type signature, like when one of the types is a dependent type.

### Solution

Add an implicit search in the `createSingle` method of the `ConcreteTag` derivation macro to look for a given instance prior to trying to construct it from the `TypeRepr`.

### Notes

Also updated some names which hadn't been updated from the previous name `SafeClassTag`.


kyo-data/shared/src/main/scala/kyo/internal/ConcreteTagMacro.scala
```diff
@@ -0,0 +1,102 @@
...
+                    val exprs = flatten(tpe).map(create)
+                    '{ Union(Set(${ Varargs(exprs) }*)) }
+                case AndType(_, _) =>
+                    def flatten(tpe: TypeRepr): Chunk[TypeRepr] =
+                        tpe match
+                            case AndType(a, b) => flatten(a).concat(flatten(b))
+                            case _             => Chunk(tpe)
+                    val exprs = flatten(tpe).map(create)
+                    '{ Intersection(Set(${ Varargs(exprs) }*)) }
+                case _ => createSingle(tpe).asInstanceOf[Expr[ConcreteTag[Any]]]
```
fwbrasil: not sure if pre-existing but the `asInstanceOf` doesn't seem necessary here

johnhungerford: ah good catch

PR #1400, opened 2025-07-29
[kyo-parse] introduce module

fwbrasil:
### Problem

Parsing is a broad topic and will likely require new effects and APIs, which is better accommodated in a new module instead of prelude.

### Solution

Introduce the new `kyo-prelude` module and move `Parse` to it.


ahoy-jon:
Definitely, and @Iltotore might be working on a new version of `Parse` as well (AST recovery, ...)

ahoy-jon (approved):

Issue #1185, opened 2025-05-12
Explore the possibility of Span having covariant type parameter


sideeffffect:
The type parameter in ordinary Scala arrays is _invariant_. That is because the elements can be both read from the array and stored into the array.
However, with an immutable array, like `KArray`, elements can be only read from the array, storing to the array is not possible. That should allow the type parameter to be _covairant_.
```scala
opaque type KArray[+A] = Array[A]
```
Having it _covairant_ would make working with `KArray`s friendlier, e.g. you can pass a `KArray[Dog]` to a function that expects a `KArray[Animal]`.

Explore if this is indeed possible or if there are some hurdles which would make it problematic or impossible.

Note that Scala 3's `IArray` is defined as
```scala
opaque type IArray[+T] = Array[_ <: T]
```
https://github.com/scala/scala3/blob/3.3.6/library/src/scala/IArray.scala#L8

Maybe `KArray` should use `_ <: T` too then?

Context:
 * https://github.com/getkyo/kyo/pull/1180#discussion_r2083294499

PR #1401, opened 2025-07-30
[data] Span improvements

fwbrasil:
I couldn't come up with a new effect this time but this change is in honor of one of the last start the repo [when it got to 700](https://x.com/fbrasisil/status/1944722949715509346)! Thank you @evelinemguimaraes  🙏

### Problem

`Span`'s API is still quite limited. I introduced it to avoid boxing in `Tag` but never got back to expanding the API.

### Solution

Introduce common APIs offered by Scala collections.

### Notes

- I didn't include common collection methods that end up boxing elements. More details on the scaladoc
- I attempted to make `Span` covariant as suggested by https://github.com/getkyo/kyo/issues/1185 but it doesn't seem possible with specialized arrays
- I also expanded most methods to their complete code without calling other `inline` methods. The goal is to more easily visualize methods that are too large to inline
- There are methods that are already large but most can be optimized if we have a specialized `SpanBuilder`. I started working on it but it's not trivial to ensure it'll specialize. I'll open an issue to follow up.
- I'm also planning to introduce a new `Slice` collection that uses similar optimization techniques to `Chunk` but is unboxed like `Span`.


kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -77,16 +85,6 @@ object Span:
...
         array.unsafeArray.asInstanceOf[Array[A]]
 
-    /** Creates a Span from an IterableOnce.
-      *
-      * @param seq
-      *   the IterableOnce to create the Span from
-      * @return
-      *   a new Span containing the elements from the IterableOnce
-      */
-    def fromUnsafe[A: ClassTag](seq: IterableOnce[A]): Span[A] =
```
fwbrasil, line 87: this method wasn't necessary, there's already a `from` for `IterableOnce`

kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -117,16 +115,21 @@ object Span:
           * @return
           *   the number of elements in this Span
           */
-        def size: Int = self.length
-
-        inline def length: Int = size
```
fwbrasil, line 122: removed `length` since there's already `size`

kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -137,18 +140,182 @@ object Span:
           */
         inline def apply(idx: Int): A = self(idx)
 
+        /** Selects the first element.
+          *
+          * @return
+          *   the first element if the Span is not empty, Absent otherwise
+          */
+        inline def head: Maybe[A] =
```
fwbrasil, line 258: The APIs are also designed to be safer than the counterparts in the stdlib

fwbrasil: `Maybe` unfortunately will box for primitive values due to the fact that it is a union type under the hood but at least there isn't an extra wrapper object like with `Option`

kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -235,13 +818,460 @@ object Span:
...
+        /** Returns a new Span without duplicates.
+          *
+          * @return
+          *   a new Span which contains the first occurrence of every element of this Span
+          */
+        def distinct(using ClassTag[A]): Span[A] =
+            val size = self.length
+            if size == 0 then Span.empty[A]
+            else
+                val seen = scala.collection.mutable.Set.empty[A]
```
fwbrasil, line 1053: we'll need unboxed sets as well

kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -137,18 +140,182 @@ object Span:
...
+          *   the last element if the Span is not empty, Absent otherwise
+          */
+        inline def last: Maybe[A] =
+            Maybe.when(nonEmpty)(self(size - 1))
+
+        /** The rest of the Span without its first element.
+          *
+          * @return
+          *   a new Span containing all elements except the first
+          * @throws UnsupportedOperationException
```
hearnadam: ?

kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -137,18 +140,182 @@ object Span:
...
+
         /** Checks if this Span is equal to another Span.
           *
           * @param other
           *   the Span to compare with
           * @return
-          *   true if the KArrays contain the same elements in the same order, false otherwise
+          *   true if the Spans contain the same elements in the same order, false otherwise
           */
         inline def is(other: Span[A])(using CanEqual[A, A]): Boolean =
```
hearnadam, line 420: Can you check for reference equality first?

kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -197,6 +364,400 @@ object Span:
...
+
+        /** Selects all elements of this Span which do not satisfy a predicate.
+          *
+          * @param p
+          *   the predicate to test elements with
+          * @return
+          *   a new Span containing all elements that do not satisfy the predicate
+          */
+        inline def filterNot(inline p: A => Boolean)(using ct: ClassTag[A]): Span[A] =
+            self.filter(!p(_))(using ct)
```
hearnadam: Does the explicit using clause change the bytecode?

kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -197,6 +364,400 @@ object Span:
...
+        /** Tests whether this Span starts with the given Span.
+          *
+          * @param that
+          *   the Span to test
+          * @param offset
+          *   the index where the Span is searched
+          * @return
+          *   true if this Span has that as a prefix starting at offset, false otherwise
+          */
+        inline def startsWith[B >: A](that: Span[B], offset: Int = 0)(using CanEqual[A, B]): Boolean =
```
hearnadam, line 785: Should this also support 'drop' on the input span?

kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -235,13 +818,460 @@ object Span:
...
+            concat(suffix)
+
+        /** Returns a new Span with an element appended.
+          *
+          * @param x
+          *   the element to append
+          * @return
+          *   a new Span with x appended
+          */
+        inline def append(x: A)(using ClassTag[A]): Span[A] =
```
hearnadam, line 960: `appended`? Same for `prepended`

kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -235,13 +818,460 @@ object Span:
...
+          * @return
+          *   an option value containing pf applied to the first value for which it is defined, or None if none found
+          */
+        inline def collectFirst[B](pf: PartialFunction[A, B]): Maybe[B] =
+            val size = self.length
+            @tailrec def loop(idx: Int): Maybe[B] =
+                if idx >= size then Absent
+                else
+                    val elem = self(idx)
+                    if pf.isDefinedAt(elem) then Present(pf(elem))
```
hearnadam, line 1039: Is it better to use a marker object?

kyo-data/shared/src/main/scala/kyo/Span.scala
fwbrasil: no, it was a leftover from trying to add variance to `Span`

kyo-data/shared/src/main/scala/kyo/Span.scala
fwbrasil: isn't it the `offset` param?

kyo-data/shared/src/main/scala/kyo/Span.scala
fwbrasil: The naming is actually intentional. I find it odd that most methods are verbs: map, filter, fold, partition, etc but specific modification methods are nouns. I know it's the pattern in Scala collections but `append` seems to fit much better with the rest of the methods

kyo-data/shared/src/main/scala/kyo/Span.scala
fwbrasil: I'm not sure I follow, can you elaborate?

kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -135,20 +248,184 @@ object Span:
...
+        inline def last: Maybe[A] =
+            Maybe.when(nonEmpty)(Span(self)(size - 1))
+
+        /** The rest of the Span without its first element.
+          *
+          * @return
+          *   a new Span containing all elements except the first
+          */
+        inline def tail(using ClassTag[A]): Maybe[Span[A]] =
+            Maybe.when(nonEmpty)(Span.slice(self)(1, size))
```
fwbrasil, line 275: I gave variance another try and was successful this time. The trick was referring to other methods via the `Span` companion instead of calling directly, which fixed the compiler crashes I had previously

kyo-data/shared/src/main/scala/kyo/Span.scala
```diff
@@ -21,16 +21,141 @@ import scala.reflect.ClassTag
...
+  * Span is covariant, meaning that if Dog <: Animal, then Span[Dog] <: Span[Animal]. This enables flexible subtyping relationships while
+  * maintaining type safety through careful method design. However, covariance may impact primitive specialization in some cases. Consider
+  * using Chunk instead when you need compatibility with Scala's collection library or when you require efficient structural operations like
+  * slicing without copying data.
+  *
   * @tparam A
   *   the type of elements in this Span
   */
-opaque type Span[A] = Array[A]
+opaque type Span[+A] = Array[? <: A]
```
hearnadam, line 36: 🥳

PR #1404, opened 2025-08-03
dev: Rework Kyo Parse

Iltotore:
Following #1400 and #1305

### Problem

In its current state, `Parse` does pretty much what other parser combinator libraries do. However it is not powerful enough for many real-world use cases such as production-ready programming languages, LSPs etc. Such projects require features like error accumulation and AST recovery.

### Solution

This PR reworks the entire `Parse` effect, providing the mentioned features as well as the already existing ones in the `main` branch.

As suggested by @ahoy-jon, I encoded `Parse` as its own `ArrowEffect`. I found it much easier to implement the new `Parse` this way, especially error accumulation.

### Notes

- I still need adapt the tests
- I did not implement `spaced` at the moment. Not sure which way is the best to implement it.
- I did not add `char` and String-specific friends that overlapped with the generic methods e.g `any`, `literal`... Should I add them?
- I did not add `anyOf` (well, another method has the same name but I can still rename it). I'm not sure if it was really used. Should I add it?
- I struggled to support additional effects in `runWith` and resorted to `@unchecked` two times. Is there a way to avoid that?

### DIFF
--- a/kyo-parse/shared/src/main/scala/kyo/Parse.scala
+++ b/kyo-parse/shared/src/main/scala/kyo/Parse.scala
@@ -1,14 +1,11 @@
-package kyo
+package kyo.parse
 
-import kyo.Ansi.*
-import kyo.kernel.*
-import scala.util.matching.Regex
+import kyo.*
+import kyo.kernel.ArrowEffect
+import scala.annotation.tailrec
+import scala.annotation.targetName
 
 /** Parser combinator effect for compositional text parsing.
-  *
-  * The Parse effect combines three fundamental capabilities needed for powerful parsing: state management through `Var[Parse.State]` to
-  * track input position, alternative exploration via `Choice` for backtracking and ambiguity handling, and structured error reporting with
-  * `Abort[ParseFailed]` for failures.
   *
   * This design enables building sophisticated parsers that can handle complex grammars while maintaining readability and composition of
   * parsing logic. The core parsing model is based on consuming text incrementally and producing typed values with precise control over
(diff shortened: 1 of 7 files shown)


Iltotore:
All tests pass. I commented out the tests linked to `anyOf` because I'm still not sure whether we should keep it or not. I also still need to implement `spaced`.

I also had some strange stack overflows when executing the tests after changing them:

```scala
at kyo.parse.Parse$.kyo$parse$Parse$$$_$handleLoopLoop$1(Parse.scala:861)
at kyo.parse.Parse$.runWith(Parse.scala:888)
at kyo.parse.Parse$.runState(Parse.scala:895)
at kyo.parse.Parse$.kyo$parse$Parse$$$_$handleLoopLoop$1(Parse.scala:861)
at kyo.parse.Parse$.runWith(Parse.scala:888)
at kyo.parse.Parse$.runState(Parse.scala:895)
at kyo.parse.Parse$.kyo$parse$Parse$$$_$handleLoopLoop$1(Parse.scala:861)
```

Line 861 is the recursive `runState` call in `runWith` (case for `Op.Attempt`). It only happens sometimes when recompiling incrementally the tests after changing something but never happen again after a `clean` :shrug:.

Iltotore:
@fwbrasil do you have any idea on how to implement `spaced` here? I think this is the last blocker to finish this PR.

Iltotore:
> @Iltotore is there some limitation to introduce the read aspect in the new implementation? 

Actually no. I wasn't sure it was the prefered way with the new implementation 👍

Iltotore:
Only streaming remains to reimplement and general code cleanup then the PR should be ready!

ahoy-jon:
@Iltotore it doesn't pass tests in JS.

Maybe a tag or else on the test to check for stack size.
```
kyo-parseJS/test
```
```
[info] ParseTest:
[info] combinators
[info]   firstOf
[info]   - takes first match (7 milliseconds)
[info]   - tries alternatives (0 milliseconds)
[info]   - handles recursive parsers (4 milliseconds)
[info]   - handles mutually recursive parsers (4 milliseconds)
[info]   - lazy evaluation prevents stack overflow *** FAILED *** (5 seconds, 368 milliseconds)
[info]     scala.scalajs.js.JavaScriptException: RangeError: Maximum call stack size exceeded
[info]     at kyo.parse.Parse$$anon$61.apply(/Users/jon/Documents/GitHub/kyo/kyo-parse/js/target/scala-3.7.1/kyo-parse-test-fastopt/main.js:72271:124)
[info]     at kyo.parse.Parse$$anon$61.apply(/Users/jon/Documents/GitHub/kyo/kyo-parse/js/target/scala-3.7.1/kyo-parse-test-fastopt/main.js:72275:15)
[info]     at kyo.parse.Parse$$anon$63.apply(/Users/jon/Documents/GitHub/kyo/kyo-parse/js/target/scala-3.7.1/kyo-parse-test-fastopt/main.js:72310:458)
[info]     at kyo.parse.Parse$$anon$63.apply(/Users/jon/Documents/GitHub/kyo/kyo-parse/js/target/scala-3.7.1/kyo-parse-test-fastopt/main.js:72313:15)
[info]     at kyo.parse.Parse$$anon$61.apply(/Users/jon/Documents/GitHub/kyo/kyo-parse/js/target/scala-3.7.1/kyo-parse-test-fastopt/main.js:72272:573)
[info]     at kyo.parse.Parse$$anon$61.apply(/Users/jon/Documents/GitHub/kyo/kyo-parse/js/target/scala-3.7.1/kyo-parse-test-fastopt/main.js:72275:15)
[info]     at kyo.parse.Parse$$anon$63.apply(/Users/jon/Documents/GitHub/kyo/kyo-parse/js/target/scala-3.7.1/kyo-parse-test-fastopt/main.js:72310:458)
[info]     at kyo.parse.Parse$$anon$63.apply(/Users/jon/Documents/GitHub/kyo/kyo-parse/js/target/scala-3.7.1/kyo-parse-test-fastopt/main.js:72313:15)
[info]     at kyo.parse.Parse$$anon$61.apply(/Users/jon/Documents/GitHub/kyo/kyo-parse/js/target/scala-3.7.1/kyo-parse-test-fastopt/main.js:72272:573)
[info]     at kyo.parse.Parse$$anon$61.apply(/Users/jon/Documents/GitHub/kyo/kyo-parse/js/target/scala-3.7.1/kyo-parse-test-fastopt/main.js:72275:15)
[info]     ...
```

kyo-parse/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -29,157 +28,66 @@ import scala.util.matching.Regex
...
+        case Attempt[In, A, S](parser: A < (Parse[In] & S))                                              extends Op[In, Maybe[A] < S]
+        case RecoverWith[In, A, S](parser: A < (Parse[In] & S), recoverStrategy: RecoverStrategy[In, A]) extends Op[In, A < S]
+        case Discard[In, A, S](parser: A < (Parse[In] & S), isDiscarded: In => Boolean)                  extends Op[In, A < S]
+    end Op
 
-    private def fail[In](states: Seq[State[In]], message: String)(using frame: Frame): Nothing < Abort[ParseFailed] =
-        Abort.fail(ParseFailed(states, message))
+    inline def modifyState[Out](using
+        inline frame: Frame
+    )[In](modify: ParseState[In] => (ParseState[In], Maybe[Out]))(using inline tag: Tag[Parse[In]]): Out < Parse[In] =
```
ahoy-jon, line 50: it's not `[In]` then `[Out]`? (not sure)

kyo-parse/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -29,157 +28,66 @@ import scala.util.matching.Regex
...
-        Loop(seq) {
-            case Seq() => Choice.drop
-            case head +: tail =>
-                attempt(head()).map {
-                    case Present(value) => Loop.done(value)
-                    case Absent         => Loop.continue(tail)
-                }
-        }
+    def firstOf[In, Out, S](parsers: Seq[() => Out < (Parse[In] & S)])(using Tag[Parse[In]], Frame): Out < (Parse[In] & S) =
+        kyo.kernel.Effect.defer:
```
ahoy-jon:
```suggestion
        Effect.defer:
```

kyo-parse/shared/src/main/scala/kyo/Parse.scala
```diff
@@ -29,157 +28,66 @@ import scala.util.matching.Regex
...
   * @see
   *   [[kyo.Parse.attempt]], [[kyo.Parse.peek]] for parsers with look-ahead and backtracking
   * @see
   *   [[kyo.Parse.run]] for executing parsers against input text
-  * @see
-  *   [[kyo.Var]], [[kyo.Choice]], [[kyo.Abort]] for the underlying effects that Parse composes
   */
-opaque type Parse[A] <: (Var[Parse.State[A]] & Choice & Abort[ParseFailed]) =
-    Var[Parse.State[A]] & Choice & Abort[ParseFailed]
+sealed trait Parse[In] extends ArrowEffect[[in] =>> Parse.Op[In, in], Id]
```
ahoy-jon, line 36: Variance?

kyo-parse/shared/src/main/scala/kyo/Parse.scala
Iltotore: We currently chose `Out` then `In` in the last PR/current main state: https://github.com/getkyo/kyo/blob/9ddc95465c8dd90a2c3ba786f798caf6e11614be/kyo-parse/shared/src/main/scala/kyo/Parse.scala#L563

kyo-parse/shared/src/main/scala/kyo/Parse.scala
Iltotore: It does not seem to typecheck:

```scala
sealed trait Parse[-In] extends ArrowEffect[[in] =>> Parse.Op[In, in], Id]

object Parse:

    enum Op[-In, +Out]:
      //...
```
```scala
sealed trait Parse[-In] extends ArrowEffect[[in] =>> Parse.Op[In, in], Id]
[error]    |^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]    |contravariant type In occurs in covariant position in type kyo.kernel.ArrowEffect[[in] =>> kyo.parse.Parse.Op[In, in], kyo.Id] {...} of trait
```

not sure why tho since `ArrowEffect`'s `Input[_]` is contravariant as well as `Op`'s `In`.

kyo-bench/src/main/scala/kyo/bench/ParseBench.scala
```diff
@@ -0,0 +1,194 @@
...
+import org.openjdk.jmh.infra.Blackhole
+import scala.io.Source
+import scala.util.Using
+import zio.Chunk as ZChunk
+import zio.parser.*
+
+/*
+Inspired by ZIO-Parser's benchmarks: https://github.com/zio/zio-parser/tree/master/benchmarks/
+ */
+class ParseBench extends BaseBench:
```
fwbrasil, line 14: nice! thanks for adding benchs

kyo-parse/shared/src/main/scala/kyo/parse/Parse.scala
```diff
@@ -1,14 +1,13 @@
-package kyo
+package kyo.parse
```
fwbrasil: We typically put all main APIs directly in the `kyo` package to improve usability. Any downsides of doing the same here?

kyo-parse/shared/src/main/scala/kyo/parse/Parse.scala
```diff
@@ -973,47 +869,196 @@ object Parse:
...
+                                    // Emit the parsed value and continue with unconsumed text
+                                    Emit.valueWith(Chunk(result.out.get))(Text(state.input.remaining.mkString))
+                            )
+                        end if
+                }
+            }.map { (text, _) => runOrAbort(text)(Parse.entireInput(repeat(v))).map(Emit.value(_)) }
+        }
 
-end Parse
+    // TODO Rework
```
fwbrasil: maybe remove for now?

kyo-parse/shared/src/main/scala/kyo/parse/Parse.scala
```diff
@@ -973,47 +869,196 @@ object Parse:
...
+                        case Op.ModifyState(modify) =>
+                            val (newState, optOut) = modify(state.copy(input = state.input.advanceWhile(state.isDiscarded)))
+                            optOut match
+                                case Absent => Loop.done((newState, ParseResult.failure(newState.failures)))
+                                case Present(out) =>
+                                    Loop.continue(newState.copy(input = newState.input.advanceWhile(newState.isDiscarded)), cont(out))
+                            end match
+
+                        case Op.Attempt(parser: (Out < Parse[In]) @unchecked) =>
+                            runState(state)(parser)
```
fwbrasil: I think the cause of the stack overflows could be a missing safepoint check in `ArrowEffect.handleLoop`. It seems missing. Example from another method: https://github.com/getkyo/kyo/blob/249ae345357d566784bf9c2912caf6eb514f2d87/kyo-kernel/shared/src/main/scala/kyo/kernel/ArrowEffect.scala#L245

kyo-parse/shared/src/main/scala/kyo/parse/Parse.scala
```diff
@@ -973,47 +869,196 @@ object Parse:
...
-      */
-    def charIn(chars: String)(using Frame): Char < Parse[Char] =
-        charIf(c => chars.contains(c))
+    private[kyo] def runWith[In, Out, S, Out2, S2](state: ParseState[In])(parser: Out < (Parse[In] & S))(f: Out => Out2 < S2)(using
+        inTag: Tag[In],
+        tag: Tag[Parse[In]],
+        frame: Frame
+    ): (ParseState[In], ParseResult[Out2]) < (S & S2) =
+        extension [X, S1](v: X < (S1 & Parse[In]))
+            def castS: X < Parse[In] = v.asInstanceOf
```
fwbrasil: unused?

kyo-parse/shared/src/test/scala/kyo/parse/ParseTest.scala
```diff
@@ -1378,29 +1366,16 @@ class ParseTest extends Test:
             val parser = Parse.int
             val input  = Stream.init(Seq[Text]("1", "abc", "3"))
 
-            Abort.run(Parse.run(input)(parser).run).map { result =>
-                assert(result.isFailure)
-            }
-        }
-
-        "ambiguous parse" in run {
```
fwbrasil: I imagine that's not a commonly used feature, it seems ok. I added it as a way to validate `Choice` with other effects in the initial `Parse` version but it's good to simplify if it's possible

kyo-parse/shared/src/main/scala/kyo/parse/Parse.scala
Iltotore: I guess I could also do that for `recoverWith` (even tho it's not really supposed to be used recursively)?

kyo-parse/shared/src/main/scala/kyo/parse/Parse.scala
Iltotore: My bad. I removed it when trying to debug the stackoverflow and I guess I probably forgot to remove it after restting :sweat_smile:

kyo-parse/shared/src/main/scala/kyo/parse/Parse.scala
Iltotore: I looked at `kyo-scheduler` which puts things in `kyo.scheduler` but no problem in moving that to the root package :+1:

kyo-parse/shared/src/test/scala/kyo/parse/ParseTest.scala
Iltotore: > I imagine that's not a commonly used feature

Yes I think so. AFAIK other parsing libraries (ZIO Parse, FastParse, Parsley...) just don't have `Choice`-like mechanism. I'm also not sure the feature was really useful in real-world cases. Usually, syntax is disambiguated explicitly (e.g using `firstOf`+`require` instead of `anyOf`).

kyo-parse/shared/src/main/scala/kyo/parse/Parse.scala
Iltotore: Also I'm not sure how to use `Safepoint#handle` in this case. What would be passed in this case for `suspend` and `continue` arguments? `parser` for both?

Iltotore: I tried this without success:

```scala
SafepointAccess.handle(parser)(
    suspend = runState(state)(parser),
    continue = runState(state)(parser)
)
    .map((parseState, result) =>
```

I still get the stack overflow when removing `firstOf`'s `Effect.defer`:

ahoy-jon (approved):

Iltotore:
Sure.

PR #1405, opened 2025-08-06
test for catching panic

ahoy-jon:
add a test for
```scala
Abort.catching[E](Abort.panic(new E))
```
I do think we should transform the panic into a failure in this case.


ahoy-jon:
> @hearnadam @johnhungerford suggested removing the behavior.

So we have to go back to the old behavior fully or just on `catching`?

fwbrasil (approved):

fwbrasil:
> So we have to go back to the old behavior fully or just on catching?

I think the old behavior might be reasonable but it'd be nice to more details on how it could be problematic

PR #1408, opened 2025-08-17
add: OsSignal for scala-native

### LINKED ISSUE #1240: [feature]: add OsSignal for Scala Native
### Problem Statement

OsSignal will be used to trigger interruption in KyoApp on signals. Currently, Signal handling is only supported on the JVM.

### Proposed Solution

I tried an initial implementation, but had issues allocating a C function from an arbitrary Thunk.
https://github.com/scala-native/scala-native/blob/main/posixlib/src/main/scala/scala/scalanative/posix/signal.scala

### Alternative Solutions

_No response_

### Current Workaround

_No response_

### Additional Context

_No response_

hearnadam:
### Problem
Native binaries currently don't respect signal interrupts.
/fixes #1240

### Solution
Register native signal handlers.

### Notes
I didn't find a good way to test this. Assuming it gets merged, we can use the snapshots to build a test app.


kyo-core/native/src/main/scala/kyo/internal/OSSignalPlatformSpecific.scala
```diff
@@ -0,0 +1,34 @@
+package kyo.internal
+
+import OsSignal.Handler
+import scala.scalanative.posix.signal as posix
+import scala.scalanative.unsafe.*
+
+/** Internal global dispatcher to avoid capturing local state in C callbacks. */
+private object NativeSignalDispatch:
```
hearnadam, line 8: This is to get around `CFuncPtr.fromScalaFunction` introducing undefined behavior when capturing lambdas. Not sure about the correctness.

ahoy-jon (approved):
Should just be tested with a snapshot build

PR #1421, opened 2025-09-16
docs: Add section for `Parse` effect

Iltotore:
Add `Parse` documentation to the README, showing a simple example and another one on error recovery introduced in #1404.


Iltotore:
> > Should I just repeat the code in the cell or is there a way to maybe hide code on site generation (like in Scaladoc) or something like that?
> 
> @Iltotore Can you try `scala mdoc:shared`?

Looks like it is a sjs-only modifier but I used `nest`. Seems to work :+1:

hearnadam (approved):

PR #1423, opened 2025-09-21
fix: Layer.using + add test for a failing case

ahoy-jon:


kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -80,7 +80,10 @@ abstract class Layer[+Out, -S] extends Serializable:
       * @return
       *   A new layer producing both outputs
       */
-    final infix def using[Out2, S2, In2](that: Layer[Out2, Env[In2] & S2]): Layer[Out & Out2, S & S2] = self and (self to that)
+    final infix def provide[Out2, S2](that: Layer[Out2, Env[Out] & S2]): Layer[Out & Out2, S & S2] = self and (self to that)
```
ahoy-jon: `In2` was been removed without checks

kyo-prelude/shared/src/test/scala/kyo/LayerTest.scala
```diff
@@ -40,14 +40,51 @@ class LayerTest extends Test:
...
+        "a using b with Env2" in {
+            trait Dummy
+            case class Guppy(name: String)
+            case class Shark(belly: Guppy, dummy: Dummy)
+
+            val guppyLayer: Layer[Guppy, Any] = Layer(Guppy("Tiny Guppy"))
+            val sharkLayer                    = Layer.from(Shark.apply)
+            val dummy: Layer[Dummy, Any]      = Layer(new Dummy {})
+
+            val guppyAndShart: Layer[Guppy & Shark, Env[Dummy]] = guppyLayer provide sharkLayer
```
hearnadam: awkward typo 😅

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -288,34 +287,30 @@ object Layer:
         kyo.internal.LayerMacros.make[Target](layers*)
 
     private[kyo] object internal:
-        case class And[Out1, Out2, S1, S2](lhs: Layer[Out1, S1], rhs: Layer[Out2, S2])                   extends Layer[Out1 & Out2, S1 & S2]
-        case class To[Out1, Out2, S1, S2](lhs: Layer[?, ?], rhs: Layer[?, ?])                            extends Layer[Out1 & Out2, S1 & S2]
-        case class FromKyo[In, Out, S](kyo: () => TypeMap[Out] < (Env[In] & S))(using val tag: Tag[Out]) extends Layer[Out, S]
+        case class And[Out1, Out2, S1, S2](first: Layer[Out1, S1], second: Layer[Out2, S2 & Env[Out1]])  extends Layer[Out1 & Out2, S1 & S2]
+        case class To[Out1, Out2, S1, S2](first: Layer[Out1, S1], second: Layer[Out2, S2 & Env[Out1]])   extends Layer[Out2, S1 & S2]
```
ahoy-jon: @hearnadam the main problem was here. The previous definition of `To` was ditching other Envs from the type definition.

kyo-prelude/shared/src/test/scala/kyo/LayerTest.scala
```diff
@@ -40,14 +40,51 @@ class LayerTest extends Test:
...
+            // note: circular deps, not usable
+            val _: Layer[Shark & Guppy, Env[Guppy & Dummy]] = sharkLayer using guppyLayer
```
ahoy-jon: TODO: write a pendingUntilFixed, that composition should have been detected, and should not compile

README.md
```diff
@@ -828,7 +828,7 @@ The `Layer` type provides instance methods for manually composing layers:
 
 1. `to`: Combines two layers sequentially, where the output of the first layer is used as input for the second layer.
 2. `and`: Combines two layers in parallel, producing a layer that provides both outputs.
-3. `using`: Combines a layer with another layer that depends on its output, similar to `to` but keeps both outputs.
+3. `provide`: Combines a layer with another layer that depends on its output, similar to `to` but keeps both outputs.
```
steinybot: 🧶 `provide` still doesn't quite seem right. I don't get the intuition that it keeps the output from the first. Not sold on these but how about `branch`, `branchTo` or perhaps even `andTo`?

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -288,34 +287,30 @@ object Layer:
...
-                                leftResult  <- doRun(lhs)
-                                rightResult <- Env.runAll(leftResult)(doRun(rhs))
-                            yield rightResult
-                        }.asInstanceOf[Expected]
+            private type Expected = TypeMap[Out] < (S & Memo)
+            private val memo = Memo[Layer[Out, S], TypeMap[Out], S & Memo] {
+                case And(lhs, rhs) =>
+                    for
+                        leftResult  <- doRun(lhs)
+                        rightResult <- Env.runAll(leftResult)(doRun(rhs))
```
steinybot: ❓ Is this right? For `and` aren't the layers independent?

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -47,12 +47,10 @@ abstract class Layer[+Out, -S] extends Serializable:
       *   The output type of the composed layer
       * @tparam S2
       *   Additional effects of the composed layer
-      * @tparam In2
-      *   The input type required by the second layer
       * @return
       *   A new layer representing the composition of both layers
       */
-    final infix def to[Out2, S2, In2](that: Layer[Out2, Env[In2] & S2]): Layer[Out2, S & S2] = To(self, that)
+    final infix def to[Out2, S2, In2](that: Layer[Out2, Env[Out] & S2]): Layer[Out2, S & S2] = To(self, that)
```
steinybot: 🛠️ `In2` is unused.

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
steinybot, line 1: 📝 `using` is also mentioned in the main `Layer` ScalaDocs.

kyo-prelude/shared/src/test/scala/kyo/LayerTest.scala
```diff
@@ -40,14 +40,60 @@ class LayerTest extends Test:
...
+
+        "provide should not allow circular deps" in pendingUntilFixed {
+            trait Dummy
+            case class Guppy(name: String)
+            case class Shark(belly: Guppy, dummy: Dummy)
+
+            val guppyLayer: Layer[Guppy, Any]                = Layer(Guppy("Tiny Guppy"))
+            val sharkLayer: Layer[Shark, Env[Guppy & Dummy]] = Layer.from(Shark.apply)
+
+            typeCheckFailure("""sharkLayer provide guppyLayer""")("???")
```
steinybot: ❓ Are there no cases where this would be wanted? Ignoring the type/variable names, couldn't the second layer be some kind of augmentation layer or is that not allowed?

README.md
ahoy-jon: @steinybot I 100% agree on that. I think I would go for `andTo`

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
ahoy-jon: good catch!

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
ahoy-jon: it's not wrong. It avoids creating `AndTo`, to complete `And` and `To`.
Do you think there is a case where that would lead to an issue?

kyo-prelude/shared/src/test/scala/kyo/LayerTest.scala
ahoy-jon: from, on `provide`: 
- `Layer[B, Env[A & C]]`
- `Layer[A, Any]`
the output is `Layer[A & B, Env[A & C]]`

I do think that is confusing for the User, and we should help to put args in the right order.
If it's on the `and` operator, it would behave as expected.

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
steinybot: I'm not too sure. What is looking strange to me is that the `And` and `To` cases are the same. I would only expect the left to be available to the right for `to` but not for `and`.

kyo-prelude/shared/src/test/scala/kyo/LayerTest.scala
steinybot: I was thinking of something more like this:
```
val feedSharkLayer: Layer[Shark, Env[Shark & Guppy]] = Layer.from((s: Shark, g: Guppy) => s.copy(belly = Maybe(g)))
```

ahoy-jon: 🤔  that's not the case that would pose a problem. I would just restrict that on `andTo` / `to`, where the argument order matters a lot more!

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
ahoy-jon: `AndTo` could be enough for all of it, performance wise, it's better to specialise with `And`, `To`, `AndTo`

kyo-prelude/shared/src/test/scala/kyo/LayerTest.scala
ahoy-jon: @steinybot did some changes in that area

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
ahoy-jon: @steinybot went for `And`, `To` and `AndTo`

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -75,12 +73,13 @@ abstract class Layer[+Out, -S] extends Serializable:
...
-      * @tparam In2
-      *   The input type required by the second layer
       * @return
       *   A new layer producing both outputs
       */
-    final infix def using[Out2, S2, In2](that: Layer[Out2, Env[In2] & S2]): Layer[Out & Out2, S & S2] = self and (self to that)
+    final infix def andTo[Out2, S2](that: Layer[Out2, Env[Out] & S2]): Layer[Out & Out2, S & S2] = AndTo(self, that)
+
+    @deprecated("Use `andTo` instead, which is more consistent considering the argument order.")
+    final infix def using[Out2, S2](that: Layer[Out2, S2]): Layer[Out & Out2, S & S2] = self andTo that
```
hearnadam: I don't think we need a deprecation if this will catch bugs.

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -288,34 +287,38 @@ object Layer:
...
+                    yield leftResult.union(rightResult)
+
+                case To(lhs, rhs) =>
+                    for
+                        leftResult  <- doRun(lhs)
+                        rightResult <- Env.runAll(leftResult)(doRun(rhs))
+                    yield rightResult
+
+                case FromKyo(kyo)   => kyo()
+                case FromKyo_0(kyo) => kyo()
```
hearnadam: Since these will be materialized to the same expected type - I don't think we need a distinct variant. Can we keep it as just `FromKyo`?

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -288,34 +287,38 @@ object Layer:
...
+
+            private type Expected = TypeMap[Out] < (S & Memo)
+            private val memo = Memo[Layer[Out, S], TypeMap[Out], S & Memo] {
+                case And(lhs, rhs) =>
+                    for
+                        leftResult  <- doRun(lhs)
+                        rightResult <- (doRun(rhs))
+                    yield leftResult.union(rightResult)
+
+                case AndTo(lhs, rhs) =>
```
hearnadam: The `AndTo` case class isn't necessary due to memoization.

  The old implementation `self and (self to that)` works fine because:
  - It creates `And(self, To(self, that))`
  - When running, `doRun(self)` gets called twice (once in `And`, once in nested `To`)
  - But since all `doRun` calls share the same `memo` instance, the second call hits the cache

kyo-data/shared/src/main/scala/kyo/TypeMap.scala
```diff
@@ -66,7 +66,9 @@ object TypeMap:
           *   A new TypeMap containing all key-value pairs from both TypeMaps
           */
         inline def union[B](that: TypeMap[B]): TypeMap[A & B] =
-            self ++ that
+            if that.isEmpty then self
+            else if self.isEmpty then that
+            else self ++ that
```
hearnadam, line 71: Nice!

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
hearnadam: As in, we should just keep the same name with a new signature if necessary.

kyo-prelude/shared/src/test/scala/kyo/LayerTest.scala
ahoy-jon: It means nothing in French, I have no excuses! 🥇

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
ahoy-jon: That would be nice, but `Env[Nothing]` will not be helpful.

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -75,12 +73,12 @@ abstract class Layer[+Out, -S] extends Serializable:
...
       *   Additional effects of the other layer
-      * @tparam In2
-      *   The input type required by the second layer
       * @return
       *   A new layer producing both outputs
       */
-    final infix def using[Out2, S2, In2](that: Layer[Out2, Env[In2] & S2]): Layer[Out & Out2, S & S2] = self and (self to that)
+    final infix def andTo[Out2, S2](that: Layer[Out2, Env[Out] & S2]): Layer[Out & Out2, S & S2] = self and (self to that)
+
+    final infix def using[Out2, S2](that: Layer[Out2, Env[Out] & S2]): Layer[Out & Out2, S & S2] = self andTo that
```
hearnadam: `@deprecated`?

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
hearnadam: I think we can just cast. There's no difference in the underlying representation due to type widening.

kyo-prelude/shared/src/main/scala/kyo/Layer.scala
```diff
@@ -288,34 +286,29 @@ object Layer:
         kyo.internal.LayerMacros.make[Target](layers*)
 
     private[kyo] object internal:
-        case class And[Out1, Out2, S1, S2](lhs: Layer[Out1, S1], rhs: Layer[Out2, S2])                   extends Layer[Out1 & Out2, S1 & S2]
-        case class To[Out1, Out2, S1, S2](lhs: Layer[?, ?], rhs: Layer[?, ?])                            extends Layer[Out1 & Out2, S1 & S2]
-        case class FromKyo[In, Out, S](kyo: () => TypeMap[Out] < (Env[In] & S))(using val tag: Tag[Out]) extends Layer[Out, S]
+
+        case class And[Out1, Out2, S1, S2](_1: Layer[Out1, S1], _2: Layer[Out2, S2])            extends Layer[Out1 & Out2, S1 & S2]
+        case class To[Out1, Out2, S1, S2](_1: Layer[Out1, S1], _2: Layer[Out2, S2 & Env[Out1]]) extends Layer[Out2, S1 & S2]
+        case class FromKyo[Out, S](kyo: () => TypeMap[Out] < S)(using val tag: Tag[Out])        extends Layer[Out, S]
```
ahoy-jon, line 290: @hearnadam remove Kyo_0, it's just that at the end, it removes the initial drop of `In` in `Layer`

PR #1427, opened 2025-10-17
Update scala-native to 0.5.9

hearnadam:
Scala Native 0.5.9 introduced DoubleAdder: https://github.com/scala-native/scala-native/releases/tag/v0.5.9

This PR:
- Removes stub adders
- re-enables tests on PR/main CI
- fixes flipped conditional in BaseKyoKernelTest for `runNotNative`


hearnadam:
@steinybot there seems to be an issue with the current test 'allowed following ops when not empty' when correcting the bug in `runNotNative`. I am not convinced any of the queue tests need to be `runNotNative`, but that it was essentially skipping them before.

Do you mind taking look?

PR #1430, opened 2025-10-18
Queue: make subsequent closeAwaitEmpty calls wait for drain

hearnadam:
Problem
- Only the first closeAwaitEmpty awaited drain; later calls returned immediately (false).

Change
- If HalfOpen, subsequent calls adopt the existing promise (join the same wait) and map the result to false.
- If Closed, they still complete immediately with false.
- First caller still observes true; later callers observe false but now wait for the drain.

Before
```
  Open ├─ closeAwaitEmpty #1 ──> HalfOpen(P) ── waits for drain ──> Closed → true
       └─ closeAwaitEmpty #2..N ── immediate → false
```

After
```
  Open ├─ closeAwaitEmpty #1 ──> HalfOpen(P) ── waits for drain ──> Closed → true
       └─ closeAwaitEmpty #2..N ── join P ───── waits with #1 ───────┘     → false
```

### DIFF
--- a/kyo-kernel/shared/src/main/scala/kyo/internal/BaseKyoKernelTest.scala
+++ b/kyo-kernel/shared/src/main/scala/kyo/internal/BaseKyoKernelTest.scala
@@ -26,7 +26,7 @@ private[kyo] trait BaseKyoKernelTest[S] extends BaseKyoDataTest:
     def runNative(v: => Assertion < S)(using Frame): Future[Assertion] = runNative(v.map(Future.successful(_)))
 
     @targetName("runNotNativeAssertion")
-    def runNotNative(v: => Assertion < S)(using Frame): Future[Assertion] = runNative(v.map(Future.successful(_)))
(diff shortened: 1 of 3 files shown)


kyo-kernel/shared/src/main/scala/kyo/internal/BaseKyoKernelTest.scala
```diff
@@ -26,7 +26,7 @@ private[kyo] trait BaseKyoKernelTest[S] extends BaseKyoDataTest:
     def runNative(v: => Assertion < S)(using Frame): Future[Assertion] = runNative(v.map(Future.successful(_)))
 
     @targetName("runNotNativeAssertion")
-    def runNotNative(v: => Assertion < S)(using Frame): Future[Assertion] = runNative(v.map(Future.successful(_)))
+    def runNotNative(v: => Assertion < S)(using Frame): Future[Assertion] = runNotNative(v.map(Future.successful(_)))
```
hearnadam, line 29: this was masking the QueueTest which was not running.

johnhungerford (approved):
LGTM

PR #1441, opened 2025-11-20
Test Scala 3.8.0 on Kyo

ahoy-jon:
- [ ] Check experimental flags (CC, explicitNull), when using Kyo
- [ ] Fix compile errors
  - [ ] `kyo-data`
- [ ] Fix tests
  - [ ] `kyo-kernel`
  
 --- 
 kyo-data
 ```scala
-- [E007] Type Mismatch Error: SpanTest.scala:543:25 
val result = arr.flatMap(x => Span(x, x * 2))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        Found:    scala.reflect.ClassTag[Array[Int]]
        Required: scala.reflect.ClassTag[Span$package$_this.Span[Int]]

-- [E007] Type Mismatch Error: SpanTest.scala:1331:29 
val result = arr.flatten
             ^^^^^^^^^^^
      Found:    scala.reflect.ClassTag[Array[Int]]
      Required: scala.reflect.ClassTag[Span$package$_this.Span[Int]]
```

kyo-stm
```scala
java.util.NoSuchElementException: key not found: Field(name,*8:C:java.lang.String:0:::1:5:6:7:9
4:A
9:C:java.io.Serializable:0:::4
5:C:java.lang.constant.Constable:0:::2
0:C:java.lang.String:0:::1:5:6:7:9
2:C:java.lang.Object:0:::3
7:C:java.lang.Comparable:1:0:8:2
3:C:scala.Matchable:0:::4
6:C:java.lang.CharSequence:0:::2
1:C:java.lang.constant.ConstantDesc:0:::2)
  at scala.collection.immutable.Map$Map2.apply(Map.scala:320)
  at kyo.Record$.selectDynamic$extension(Record.scala:93)
  at kyo.TTableTest.kyo$TTableTest$$_$mapLoop$1(TTableTest.scala:83)
```

kyo-kernel (byte code size is sometimes smaller with 3.8)
```scala
-- map test failure:
Map("test" -> 26, "anonfun" -> 11, "mapLoop" -> 162) did not equal 
Map("test" -> 26, "anonfun" -> 11, "mapLoop" -> 158)
Analysis: Map("mapLoop": 162 -> 158) - 4 bytes difference

-- handle test failure:
Map("test" -> 26, "anonfun" -> 8, "handleLoop" -> 291) did not equal 
Map("test" -> 26, "anonfun" -> 8, "handleLoop" -> 283)  
[...]

PR #1445, opened 2026-01-02
Fix issue with a test caused by scala 3.8 type inferrence

jchyb:
Hello! I'm part of the Scala 3 compiler team, and we've noticed an issue with running certain tests in Kyo with scala 3.8 in the [open community build](https://github.com/VirtusLab/community-build3). We usually use this information to pinpoint regressions in the compiler, but this time I've determined that an adjustment in the library would be more applicable.

### Problem
The issue stems from a difference between how certain types are inferred and represented when using the Quotes reflect API. More context/information here: https://github.com/scala/scala3/issues/24596

In the minimization there, one Tag macro had `TypeRef(ThisType(TypeRef(NoPrefix,module class lang)),class String)` passed to it and another had `TypeRef(TermRef(TermRef(ThisType(TypeRef(NoPrefix,module class <root>)),object java),object lang),class String)`, leading to the generation of different `Tag`s (while the expectation was that they would be the same).

### Solution
Using `=:=` instead of `equals` when tagging seen types. This makes sure that the same, but differently represented types produce the same `Tag`s (also helps users for cases outside of the test and in the previous scala versions, e.g. if they use an inferred type for one `Tag`, and explicit for another, they may still look be represented differently).

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/internal/TagMacro.scala
+++ b/kyo-data/shared/src/main/scala/kyo/internal/TagMacro.scala
@@ -53,9 +53,9 @@ private[kyo] object TagMacro:
                     case true => tpe
                     case false =>
                         seen.get(tpe.typeSymbol) match
-                            case None                          => tpe.typeSymbol
-                            case Some((t, _)) if t.equals(tpe) => tpe.typeSymbol
-                            case _                             => tpe
+                            case None                      => tpe.typeSymbol
+                            case Some((t, _)) if t =:= tpe => tpe.typeSymbol
+                            case _                         => tpe
             if seen.contains(key) then
                 seen(key)._2
             else


kyo-data/shared/src/main/scala/kyo/internal/TagMacro.scala
```diff
@@ -53,9 +53,9 @@ private[kyo] object TagMacro:
                     case true => tpe
                     case false =>
                         seen.get(tpe.typeSymbol) match
-                            case None                          => tpe.typeSymbol
-                            case Some((t, _)) if t.equals(tpe) => tpe.typeSymbol
-                            case _                             => tpe
+                            case None                      => tpe.typeSymbol
+                            case Some((t, _)) if t =:= tpe => tpe.typeSymbol
```
hearnadam, line 57: I believe this is correct, but I am slightly concerned that Tag representation will be dependent on compilation order. I am not sure if there is a good way to isolate that.

I will take a closer look at the macro tomorrow.

ahoy-jon: that's the fix I have found : https://github.com/getkyo/kyo/pull/1441/files#r2696523640

ahoy-jon: I do think a mix of both would make it work. (sorting, and better equality). Thanks a lot @jchyb for helping

PR #1455, opened 2026-01-26
[stm] fix opacity bug

fwbrasil:
I finally read the TL2 paper again and understood that I should have used two timestamps for transactions. Funny thing: I did that in my first STM impl (don't look at the [code](https://github.com/fwbrasil/radon-stm/blob/master/src/main/scala/net/fwbrasil/radon/transaction/time/TransactionStopWatch.scala)!). It's nice to see how much simpler and powerful Kyo's impl is :)

### DIFF
--- a/kyo-stm/shared/src/test/scala/kyo/STMTest.scala
+++ b/kyo-stm/shared/src/test/scala/kyo/STMTest.scala
@@ -139,7 +139,9 @@ class STMTest extends Test:
                         }
                     }
                 a <- attempts.get
-            yield assert(v.isFailure && a == 1)
+            yield 
+                assert(v.isFailure)
+                assert(a == 1)
         }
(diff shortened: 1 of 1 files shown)


kyo-stm/shared/src/main/scala/kyo/TID.scala
```diff
@@ -7,7 +7,7 @@ private[kyo] object TID:
 
     private val tidLocal = Local.initNoninheritable[java.lang.Long](-1L)
 
-    def next(using AllowUnsafe): Long = nextTid.incrementAndGet()
+    def next()(using AllowUnsafe): Long = nextTid.incrementAndGet()
```
fwbrasil, line 10: I forgot to add `()` here since it's side effecting

PR #1456, opened 2026-01-27
[stm] optimizations

fwbrasil:
A few optimizations of the STM impl. Please check the comments for details.


kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -93,15 +94,15 @@ object STM:
     private def run[E: ConcreteTag, A](retrySchedule: Schedule)(v: A < (STM & Abort[E] & Async))(
         using Frame
     ): A < (Async & Abort[E | FailedTransaction]) =
-        TID.useIO {
-            case -1L =>
-                TID.useNew { tid =>
+        Tick.withCurrent(
```
fwbrasil, line 97: I've renamed `TID` to `Tick` for clarity since after https://github.com/getkyo/kyo/pull/1455, the transaction has two timestamps. I've also improved the API to avoid leaking the `-1` case.

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -133,11 +134,14 @@ object STM:
...
                 // STM effect in the parent transaction's `run`.
                 result.asInstanceOf[A < (Async & Abort[E | FailedTransaction])]
-        }
+        )
 
     end run
 
-    private def commit[A, S](tid: Long, log: TRefLog, probe: Boolean = false)(using AllowUnsafe): Boolean =
+    // Thread-local cache for the commit buffer to avoid repeated allocations
+    private val bufferCache = new ThreadLocal[ArrayList[Any]]
```
fwbrasil, line 142: cache the buffer used used to traverse the ref log to avoid allocations

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -146,54 +150,62 @@ object STM:
...
-                // there's a single ref
-                val ok = ref.lock(entry)
-                if ok then
-                    if !probe then ref.commit(TID.next(), entry)
-                    ref.unlock(entry)
-                ok
+                entry match
+                    case _: TRefLog.Read[?] =>
+                        // Read-only: just validate, no locking needed
+                        ref.validate(entry)
```
fwbrasil, line 156: optimization for read only transactions with a single ref to avoid locking

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -146,54 +150,62 @@ object STM:
...
-                        array(i) = ref
-                        array(i + 1) = entry
-                        i += 2
+                        hasWrites |= entry.isInstanceOf[TRefLog.Write[?]]
+                        discard(buffer.add(ref))
+                        discard(buffer.add(entry))
                     }
 
+                    // Read-only transaction: already validated, no locking needed
+                    if !hasWrites then boundary.break(true)
```
fwbrasil, line 186: optimization for read only transactions with multiple refs, avoiding locking as well

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -203,27 +215,30 @@ object STM:
...
                     // Successfully locked all references - commit changes
-                    if !probe then commit(TID.next(), 0)
+                    if !probe then commit(Tick.next(), 0)
 
                     // Release all locks
                     unlock(0, size)
                     true
                 }
+                // No try/finally needed - boundary.break returns from the block, not the method
+                buffer.clear()
```
fwbrasil, line 225: `try`/`catch` would be safer but there's some cost to installing the interrupt handler and the code is safe

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -41,43 +41,56 @@ sealed trait TRef[A]:
...
   *
-  * @param initialState
-  *   The initial value and transaction ID for this reference
+  * @param initEntry
+  *   The initial value and tick for this reference
   */
-final private class TRefImpl[A] private[kyo] (initialState: Write[A])
-    extends AtomicInteger(0) // Atomic super class to keep the lock state
+final private class TRefImpl[A] private[kyo] (initEntry: Write[A])
+    extends TRef.State.Owner
```
fwbrasil, line 58: Instead of extending from `AtomicInteger`, extend from `TRef.State.Owner`, which then extends from `AtomicLong` to pack both the read tick and the lock state in a single long. The bit packing is abstracted away via `TRef.State`.

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -41,43 +41,56 @@ sealed trait TRef[A]:
...
     with TRef[A]
     with Serializable:
 
-    @volatile private var currentState = initialState
+    import TRef.State
+    import TRef.State.*
 
-    private[kyo] def state(using AllowUnsafe): Write[A] = currentState
+    @volatile private var _entry = initEntry
```
fwbrasil, line 65: renamed the previous `state` methods/parameter to `entry` to avoid confusion

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -87,74 +100,79 @@ final private class TRefImpl[A] private[kyo] (initialState: Write[A])
...
-                    Var.setDiscard(log.put(this, entry))
+                    Var.setDiscard(log.put(this, Write(prev.tick, v)))
                 case Absent =>
-                    TID.useIORequired { tid =>
-                        val state = currentState
-                        if state.tid > tid then
-                            // Early retry if the TRef is concurrently modified
+                    Tick.withCurrent { tick =>
+                        val e = _entry
+                        if e.tick.value > tick.value || getState().readTick > tick.value then
```
fwbrasil, line 107: early retry optimization if the ref being written has a more recent read

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -87,74 +100,79 @@ final private class TRefImpl[A] private[kyo] (initialState: Write[A])
...
 
     private[kyo] def validate(entry: Entry[A])(using AllowUnsafe): Boolean =
-        currentState.tid == entry.tid
+        val current = _entry
+        current.tick == entry.tick || (
+            entry match
+                // Value-based fallback only for reads: if the same reference was written
+                // back, the read is still valid (reduces spurious aborts). Not safe for
+                // writes since two transactions computing the same value must not both commit.
+                case read: Read[?] => current.value.asInstanceOf[AnyRef].eq(read.value.asInstanceOf[AnyRef])
```
fwbrasil, line 125: if the entry is a read, it doesn't matter if the tick is different in case the value is the same. I'm using `eq` since `==` could be expensive and have to calculate the hashcode

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -87,74 +100,79 @@ final private class TRefImpl[A] private[kyo] (initialState: Write[A])
...
                 entry match
-                    case Read(tid, value) =>
-                        // Read locks can stack if no write lock
-                        lockState != Int.MaxValue && (super.compareAndSet(lockState, lockState + 1) || loop())
-                    case Write(tid, value) =>
-                        // Write lock requires no existing locks
-                        lockState == 0 && (super.compareAndSet(lockState, Int.MaxValue) || loop())
-                end match
+                    case _: Read[?]  => s.acquireReader.exists(next => casState(s, next) || loop())
+                    case _: Write[?] => s.acquireWriter(tick.value).exists(next => casState(s, next) || loop())
```
fwbrasil, line 136: I love how Scala 3 is able to keep `@tailrec` with the lamba. This happens because `Maybe.exists` is `inline`

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -146,54 +150,62 @@ object STM:
...
-
-                boundary {
-
-                    var i = 0
-                    // Pre-validate and dump the log to the flat array
+                // Commit multiple refs using cached buffer
+                var buffer = bufferCache.get()
+                if buffer == null then
+                    buffer = new ArrayList[Any]
+                    bufferCache.set(buffer)
```
hearnadam, line 170: It might be nice to push the getting/refilling this into a method in the `ThreadLocal` itself.

kyo-stm/shared/src/main/scala/kyo/Tick.scala
```diff
@@ -0,0 +1,54 @@
+package kyo
+
+/** Monotonic tick value for STM conflict detection */
+private[kyo] opaque type Tick = Long
```
hearnadam, line 4: ?
```suggestion
private[kyo] opaque type Tick <: Long = Long
```

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -183,20 +201,63 @@ object TRef:
...
+            inline def acquireReader: Maybe[State] =
+                Maybe.when((self & LockMask) < MaxReaders)(self + 1)
+
+            inline def releaseReader: State = self - 1
+
+            inline def acquireWriter(tick: Long): Maybe[State] =
+                Maybe.when((self & LockMask) == 0 && self.readTick <= tick)((self & ~LockMask) | WriteLock)
+
+            // Display
+            inline def asString: String =
```
hearnadam, line 256: ?
```suggestion
            inline def render: String =
```

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -87,74 +100,79 @@ final private class TRefImpl[A] private[kyo] (initialState: Write[A])
...
+                val s = getState()
                 entry match
-                    case Read(tid, value) =>
-                        // Read locks can stack if no write lock
-                        lockState != Int.MaxValue && (super.compareAndSet(lockState, lockState + 1) || loop())
-                    case Write(tid, value) =>
-                        // Write lock requires no existing locks
-                        lockState == 0 && (super.compareAndSet(lockState, Int.MaxValue) || loop())
-                end match
+                    case _: Read[?]  => s.acquireReader.exists(next => casState(s, next) || loop())
```
hearnadam, line 135: Should the `loop()` be inside `exists`?

fwbrasil: Yes, this is the cas retry loop. If `State` allows the `acquireReader` transition by returning a `Present` but `casState` fails, which indicates a concurrent write, it loops

kyo-stm/shared/src/main/scala/kyo/TRef.scala
fwbrasil: I'm including this in a follow up PR

kyo-stm/shared/src/main/scala/kyo/Tick.scala
fwbrasil: Good idea! I'm including in the next PR

kyo-stm/shared/src/main/scala/kyo/STM.scala
fwbrasil: There's `ThreadLocal.withInitial` on the JVM but JS doesn't have the stub for it

PR #1457, opened 2026-01-28
fix(kyo-stats-registry): runnable not given in the new thread resulting on metrics not exported periodically 

Salim-belkhir:
### Problem
The `ThreadFactory` in `StatsRefresh` creates threads without passing the `Runnable`:

```scala
val threadFactory = new ThreadFactory {
    def newThread(r: Runnable) = {
        val thread = new Thread  // ❌ Runnable `r` is ignored
        thread.setName("kyo-stats-refresh")
        thread.setDaemon(true)
        thread
    }
}
```
This causes the scheduled refresh() task to never execute, meaning metrics are never exported to registered StatsExporter implementations (including kyo-stats-otel).
Without this fix, kyo-stats-otel users see no metrics in their OpenTelemetry backend, even though counters/histograms are correctly registered. The refresh thread is created but sits idle because it has no runnable to execute

### Solution
Pass the Runnable to the Thread constructor:
```scala
val thread = new Thread(r)  // ✅ Runnable is now executed
```

### Notes
Tested with a Kyo app exporting metrics to SigNoz via OTLP. Before the fix: only direct OTel API calls produced metrics. After the fix: kyo-stats counters and histograms appear correctly

### DIFF
--- a/kyo-stats-registry/jvm/src/test/scala/kyo/stats/internal/StatsRegistryTest.scala
+++ b/kyo-stats-registry/jvm/src/test/scala/kyo/stats/internal/StatsRegistryTest.scala
@@ -158,10 +158,11 @@ class StatsRegistryTest extends AnyFreeSpec {
             
             val testExporter = new StatsExporter {
                 override def counter(path: List[String], description: String, delta: Long): Unit =
-                    if path.headOption.contains(testId) then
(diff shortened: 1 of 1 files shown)


ahoy-jon:
Nice catch! 🙏

PR #1459, opened 2026-01-29
[stm] cleanups and reduce allocations

fwbrasil:
please see the comments

### DIFF
--- a/kyo-stm/shared/src/main/scala/kyo/STM.scala
+++ b/kyo-stm/shared/src/main/scala/kyo/STM.scala
@@ -57,13 +57,13 @@ object STM:
     private val currentTransaction = Local.initNoninheritable[Maybe[Tick]](Absent)
 
     /** Use current transaction's tick (fails if not in transaction) */
-    private[kyo] inline def withCurrentTransaction[A, S](inline f: Tick => A < S)(using inline frame: Frame): A < (S & Sync) =
+    private[kyo] inline def withCurrentTransaction[A, S](inline f: Tick => A < S)(using inline frame: Frame): A < (S & STM & Sync) =
         Sync.withLocal(currentTransaction) {
             case Absent        => bug("STM operation attempted outside of STM.run")
             case Present(tick) => f(tick)
         }
 
-    /** Use current transaction's tick, or generate a new one if not in a transaction */
+    /** Use current transaction, or commit in a new one if not in a transaction */
     private[kyo] inline def withCurrentTransactionOrNew[A, S](inline f: AllowUnsafe ?=> Tick => A < S)(using
         inline frame: Frame
     ): A < (S & Sync) =


kyo-stm/shared/src/main/scala/kyo/internal/CommitBuffer.scala
```diff
@@ -0,0 +1,126 @@
...
+import java.util.ArrayList
+import scala.annotation.tailrec
+
+/** Thread-local buffer used during the STM commit protocol to lock, validate, and commit TRef entries.
+  *
+  * Pairs are stored as consecutive elements [ref0, entry0, ref1, entry1, ...] to avoid tuple allocations. All index parameters refer to
+  * logical pair indices; the stride encoding is handled internally. Entries are sorted by TRef id before lock acquisition to prevent
+  * deadlocks.
+  */
+opaque type CommitBuffer = ArrayList[Any]
```
fwbrasil, line 12: I've moved the thread local flat buffer optimization to this new opaque type to isolate the optimization from `STM`

kyo-stm/shared/src/main/scala/kyo/internal/CommitBuffer.scala
```diff
@@ -0,0 +1,126 @@
...
+        inline def entry(idx: Int)(using AllowUnsafe): TRefLog.Entry[Any] =
+            self.get(idx * 2 + 1).asInstanceOf[TRefLog.Entry[Any]]
+
+        inline def append(inline ref: TRef[Any], inline entry: TRefLog.Entry[Any])(using AllowUnsafe): Unit =
+            discard(self.add(ref))
+            discard(self.add(entry))
+
+        def sort(size: Int)(using AllowUnsafe): Unit =
+            if size <= insertionSortThreshold then
+                insertionSort(self, size)
```
fwbrasil, line 47: new optimization to use insertion sort instead of quick sort for small collections, which is more efficient

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -40,6 +38,40 @@ opaque type STM <: (Var[TRefLog] & Abort[FailedTransaction] & Async) =
 
 object STM:
 
+    /** Monotonic tick value for STM conflict detection */
+    private[kyo] opaque type Tick <: Long = Long
```
fwbrasil, line 42: Moved from a separate file to the `STM` companion since it's small

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -40,6 +38,40 @@ opaque type STM <: (Var[TRefLog] & Abort[FailedTransaction] & Async) =
...
+
+        /** Generate a new tick value */
+        def next()(using AllowUnsafe): Tick = counter.incrementAndGet()
+
+        /** Set counter value. For testing only. */
+        def testOnlySet(value: Long)(using AllowUnsafe): Unit = counter.set(value)
+
+    end Tick
+
+    private val currentTransaction = Local.initNoninheritable[Maybe[Tick]](Absent)
```
fwbrasil, line 57: The functionality here is the same but I've moved to outside of `Tick` since it's transaction-specific and renamed the APIs for clarity

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -94,52 +126,54 @@ object STM:
     private def run[E: ConcreteTag, A](retrySchedule: Schedule)(v: A < (STM & Abort[E] & Async))(
         using Frame
     ): A < (Async & Abort[E | FailedTransaction]) =
-        Tick.withCurrent(
-            ifAbsent =
-                Tick.withNext { tick =>
+        Sync.Unsafe.withLocal(currentTransaction) {
+            case Absent =>
+                // Optimistic retry loop: execute the transaction body, attempt commit, retry on conflict
+                def loop(schedule: Schedule)(using AllowUnsafe): A < (Async & Abort[E | FailedTransaction]) =
```
fwbrasil, line 132: Multiple optimizations to reduce suspensions and allocations here

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -1,36 +1,33 @@
...
 import kyo.TRefLog.*
 import scala.annotation.tailrec
 
 /** A transactional reference that can be modified within STM transactions. Provides atomic read and write operations with strong
   * consistency guarantees.
+  *
+  * @param initEntry
+  *   The initial value and tick for this reference
   */
-sealed trait TRef[A]:
```
fwbrasil, line 10: I had made `TRef` a `trait` because I had plans to eventually have alternative implementations like for durability but I think this kind of feature is best handled via the `TRefLog` and custom commit hooks if we ever get to it. Making it a `class` simplifies the code.

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -115,25 +85,39 @@ final private class TRefImpl[A] private[kyo] (initEntry: Write[A])
                     }
         }
 
+    // Atomically update readTick to max(current, tick)
+    @tailrec private def updateReadTick(tick: STM.Tick): Unit =
```
fwbrasil, line 89: just moved, unchanged

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -40,6 +38,40 @@ opaque type STM <: (Var[TRefLog] & Abort[FailedTransaction] & Async) =
...
+
+        /** Set counter value. For testing only. */
+        def testOnlySet(value: Long)(using AllowUnsafe): Unit = counter.set(value)
+
+    end Tick
+
+    private val currentTransaction = Local.initNoninheritable[Maybe[Tick]](Absent)
+
+    /** Use current transaction's tick (fails if not in transaction) */
+    private[kyo] inline def withCurrentTransaction[A, S](inline f: Tick => A < S)(using inline frame: Frame): A < (S & Sync) =
```
hearnadam: Should this return STM to prevent `Absent` case?

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -40,6 +38,40 @@ opaque type STM <: (Var[TRefLog] & Abort[FailedTransaction] & Async) =
...
+            case Absent        => bug("STM operation attempted outside of STM.run")
+            case Present(tick) => f(tick)
+        }
+
+    /** Use current transaction's tick, or generate a new one if not in a transaction */
+    private[kyo] inline def withCurrentTransactionOrNew[A, S](inline f: AllowUnsafe ?=> Tick => A < S)(using
+        inline frame: Frame
+    ): A < (S & Sync) =
+        Sync.Unsafe.withLocal(currentTransaction) {
+            case Absent        => f(Tick.next())
```
hearnadam, line 71: Does `current` need to be updated?

kyo-stm/shared/src/main/scala/kyo/STM.scala
```diff
@@ -94,52 +126,54 @@ object STM:
...
+            case Absent =>
+                // Optimistic retry loop: execute the transaction body, attempt commit, retry on conflict
+                def loop(schedule: Schedule)(using AllowUnsafe): A < (Async & Abort[E | FailedTransaction]) =
+                    val tick = Tick.next()
+                    // Consult the schedule for the next retry delay, or fail if exhausted
+                    def retry: A < (Async & Abort[E | FailedTransaction]) =
+                        schedule.next(Clock.live.unsafe.now()).map { (delay, next) =>
+                            Async.delay(delay)(Sync.Unsafe(loop(next)))
+                        }.getOrElse {
+                            Abort.fail(FailedTransaction())
```
hearnadam, line 139: This could be cached?

kyo-stm/shared/src/main/scala/kyo/STM.scala
fwbrasil: This method is used by `TMap` and `TRef` constructors to allow instantiation outside of a transaction. They execute as a single-ref transaction in case the `local` is `Absent`. I'll add a comment to clarify.

kyo-stm/shared/src/main/scala/kyo/STM.scala
fwbrasil: We'd lose the `Frame` information so it'd be more difficult to debug sources of retries

kyo-stm/shared/src/main/scala/kyo/STM.scala
fwbrasil: Good point about returning `STM`. I've updated to include it but we'll still need the `Absent` case since this is handled via a `local` instead of the `STM` effect itself

kyo-stm/shared/src/main/scala/kyo/TRef.scala
```diff
@@ -1,36 +1,33 @@
...
+    import TRef.State
+    import TRef.State.*
 
-    /** Sets a new value for the reference within a transaction.
-      *
-      * @param v
-      *   The new value to set
-      */
-    def set(v: A)(using Frame): Unit < STM
+    private[kyo] val id: Int = TRef.idCounter.incrementAndGet()
```
fwbrasil, line 21: I forgot to mention this other optimization. The commit needs to sort entries to avoid deadlocks when acquiring locks and was using the `hashCode` of `TRef`. In hotspot JVM, the default `hashCode` uses the identity hash code that is assigned to objects on demand, which has some overhead and contention (typically very little but can be significant for this code). Now `TRef` has a sequencial id that is used for the sorting at commit time

PR #1498, opened 2026-03-30
[flow] durable workflows module

fwbrasil:
This new module uses the same technique as `kyo-http` using `Record`s to track fields. Please [see the README](https://github.com/getkyo/kyo/blob/7a9676eac136f11b1d8bfdf6f73e4057c7ac7bbe/kyo-flow/README.md) for more info. A major omission in this initial version is no persistent `FlowStore`. I want to have a good solution for relational db support first.

The new module is build on top of `kyo-http` and supports JVM, JS, and Native.


kyo-data/shared/src/main/scala/kyo/Dict.scala
```diff
@@ -263,7 +263,7 @@ object Dict:
         infix def ++(other: Dict[K, V]): Dict[K, V] = concat(other)
 
         /** Applies the given function to each key-value pair in this Dict. */
-        inline def foreach(inline fn: (K, V) => Unit): Unit =
```
fwbrasil, line 266: I had compilation crashes. I'll follow up separately on this.

ahoy-jon:
@fwbrasil thanks a lot, reviewing!

ahoy-jon:
still at it, so far it's good. Even tested cycle in definitions, it behaves as it should (not easy to express, clear stackoverflow at the right place).

Will do some expressivity tests tomorrow!

ahoy-jon (approved):

PR #1511, opened 2026-04-05
[config] static and dynamic flags + rollouts

### LINKED ISSUE #1178: [feature]: Introduce `Flag` in kyo-data
### Problem Statement

We have a few places in the codebase that have configs like in [Async](https://github.com/getkyo/kyo/blob/909a9ffb52e03d9c7e962e440c8a012d9c4adba5/kyo-core/shared/src/main/scala/kyo/Async.scala#L66), which allows overrides, and [Tag](https://github.com/getkyo/kyo/blob/909a9ffb52e03d9c7e962e440c8a012d9c4adba5/kyo-data/shared/src/main/scala/kyo/Tag.scala#L296), which uses source code constants. 

### Proposed Solution

Introduce a `Flag` in `kyo-data` to handle these configs more properly. [This](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/Flag.scala) might be a good inspiration.

### Alternative Solutions

_No response_

### Current Workaround

_No response_

### Additional Context

_No response_

fwbrasil:
Fixes #1178

Please see README.md for context


kyo-config/shared/src/main/scala/kyo/Flag.scala
```diff
@@ -0,0 +1,335 @@
...
+                sb += '\n'
+            }
+            sb ++= line("\u2514", "\u2534", "\u2518", "\u2500")
+            sb.toString()
+        }
+    }
+
+    // --- Internal ---
+
+    private var registry = Map.empty[String, Flag[?]]
```
hearnadam, line 287: Couldn't this be atomic?

kyo-config/shared/src/main/scala/kyo/Flag.scala
```diff
@@ -0,0 +1,335 @@
...
+            "Built-in readers exist for Int, Long, Double, Boolean, String, and List[A]. " +
+            "For custom types, implement Flag.Reader[${A}]."
+    )
+    abstract class Reader[A] {
+
+        /** Parse a string value into type A. */
+        def apply(s: String): Either[Throwable, A]
+
+        /** Human-readable type name for error messages. */
+        def typeName: String
```
hearnadam, line 132: Could this be derived from a Tag or ConcreteTag?

Or are those not available on Scala 2?

kyo-config/shared/src/main/scala/kyo/Flag.scala
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    /** Environment variable name: flag name with dots replaced by underscores, uppercased. */
+    val envName: String = name.replace('.', '_').toUpperCase
+
+    /** The source that provided the resolved value. */
+    val source: Flag.Source =
+        if (java.lang.System.getProperty(name) ne null) Flag.Source.SystemProperty
+        else if (java.lang.System.getenv(envName) ne null) Flag.Source.EnvironmentVariable
+        else Flag.Source.Default
```
hearnadam: Should this be computed at the same time as the actual read? Because these sources could be modified?

kyo-config/shared/src/main/scala/kyo/Flag.scala
```diff
@@ -0,0 +1,335 @@
...
+  */
+abstract class Flag[A] private[kyo] (val default: A, val validate: A => Either[Throwable, A] = (a: A) => Right(a))(implicit
+    val reader: Flag.Reader[A]
+) {
+
+    // --- Public API ---
+
+    /** Fully-qualified flag name, derived from the JVM class name. */
+    val name: String = {
+        val className = getClass.getName.stripSuffix("$")
```
hearnadam, line 45: Do we want explicitly named Flags?

kyo-config/shared/src/main/scala/kyo/Flag.scala
```diff
@@ -0,0 +1,335 @@
...
+                    }
+                }
+            def typeName: String = s"Seq[${r.typeName}]"
+        }
+    }
+
+    // --- Public API ---
+
+    /** Returns all registered flags. */
+    def all: List[Flag[?]] = synchronized {
```
hearnadam, line 212: Chunk?

kyo-config/shared/src/main/scala/kyo/Flag.scala
```diff
@@ -0,0 +1,335 @@
...
+    /** Whether this flag supports runtime updates. */
+    def isDynamic: Boolean
+
+    // --- Internal ---
+
+    /** Raw expression string from system property or environment variable. Empty string when source is Default. */
+    private[kyo] val initialExpression: String =
+        if (source.eq(Flag.Source.SystemProperty)) java.lang.System.getProperty(name)
+        else if (source.eq(Flag.Source.EnvironmentVariable)) java.lang.System.getenv(envName)
+        else ""
```
hearnadam: Same here, assuming we have a method to return the Source + the original value, then you could avoid the duplicate lookup.

```
val source: Flag.Source = SystemProperty("Original")
```

kyo-scheduler/shared/src/main/scala/kyo/scheduler/Flags.scala
```diff
@@ -0,0 +1,15 @@
+package kyo.scheduler
+
+import kyo.StaticFlag
+
+private[kyo] object coreWorkers    extends StaticFlag[Int](Runtime.getRuntime().availableProcessors(), n => Right(Math.max(1, n)))
```
hearnadam, line 5: It seems somewhat limiting that flags need to be objects. If a method in the constructor throws, then you get a Class instantiation issue.

kyo-config/shared/src/main/scala/kyo/StaticFlag.scala
```diff
@@ -0,0 +1,107 @@
...
+                        case Right(Some(s)) =>
+                            try reader(s) match {
+                                    case Right(a) => Some(a)
+                                    case Left(e)  => throw FlagValueParseException(name, s, reader.typeName, e)
+                                }
+                            catch {
+                                case e: FlagException => throw e
+                                case e: Exception =>
+                                    throw FlagValueParseException(name, s, reader.typeName, e)
+                            }
```
hearnadam: Maybe this should be a helper method on readers?

kyo-config/shared/src/main/scala/kyo/Flag.scala
```diff
@@ -0,0 +1,335 @@
...
+  *   [[Rollout]] for the expression DSL used in conditional rollouts
+  */
+abstract class Flag[A] private[kyo] (val default: A, val validate: A => Either[Throwable, A] = (a: A) => Right(a))(implicit
+    val reader: Flag.Reader[A]
+) {
+
+    // --- Public API ---
+
+    /** Fully-qualified flag name, derived from the JVM class name. */
+    val name: String = {
```
hearnadam: Should the `val`s all be final?

kyo-config/shared/src/main/scala/kyo/Flag.scala
fwbrasil: yeah, not available

kyo-config/shared/src/main/scala/kyo/Flag.scala
fwbrasil: We could use a concurrent collection but I had trouble to cross build. Class loading is synchronized anyway in hotspot so it seems fine

kyo-config/shared/src/main/scala/kyo/Flag.scala
fwbrasil: Not sure, I decided to go with the approach in finagle. I think the restriction is good to help finding where configs are used and avoid name conflicts

kyo-scheduler/shared/src/main/scala/kyo/scheduler/Flags.scala
fwbrasil: the idea is fail fast and not even start the system if configs aren't correct. I think it's safer and avoid users getting confused thinking that a config is in effect when it's actually incorrect

kyo-config/shared/src/main/scala/kyo/Flag.scala
fwbrasil: the scheduler depends on this module so no kyo-data

PR #1514, opened 2026-04-08
[core] fix Async.foreach and derived methods

fwbrasil:
### Problem

The bounded concurrency in `Async.foreach*` and derived methods had two issues:

1. Index bug: `f(idx + idx2, v)` used the group index instead of the item offset, producing wrong indices for all groups after the first (e.g. 0,1,2,3,1,2,3,4,2,3 instead of 0..9). See tests.

2. Static batching: items were split into equal-sized groups with sequential execution within each group. Workers that finished fast batches sat idle while slow batches continued, wasting concurrency.

### Solution

Replace implementation with a new `Fiber.internal.foreachIndexedBounded`: spawns `min(size, concurrency)` worker fibers that pull items dynamically using a shared `AtomicInt` counter.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -408,7 +408,7 @@ object Async extends AsyncPlatformSpecific:
                 case 1 => f(0, iterable.head).map(Chunk(_))
                 case size =>
                     isolate.capture { state =>
-                        val items = Chunk.from(iterable).toIndexed
+                        val items = Chunk.Indexed.from(iterable)
                         Fiber.internal.foreachIndexed(items, concurrency) { (idx, v) =>
                             isolate.isolate(state, f(idx, v))
                         }.map(_.use(r => Kyo.foreach(r)(isolate.restore)))
--- a/kyo-core/shared/src/main/scala/kyo/Async.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Async.scala
@@ -406,17 +406,12 @@ object Async extends AsyncPlatformSpecific:
             iterable.size match
                 case 0 => Chunk.empty
                 case 1 => f(0, iterable.head).map(Chunk(_))
-                case size if size <= concurrency =>
-                    isolate.capture { state =>
-                        Fiber.internal.foreachIndexed(iterable)((idx, v) => isolate.isolate(state, f(idx, v)))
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -406,17 +406,12 @@ object Async extends AsyncPlatformSpecific:
...
-                        Fiber.internal.foreachIndexed(iterable)((idx, v) => isolate.isolate(state, f(idx, v)))
-                            .map(_.use(r => Kyo.foreach(r)(isolate.restore)))
-                    }
                 case size =>
                     isolate.capture { state =>
-                        val groupSize = Math.ceil(size.toDouble / Math.max(1, concurrency)).toInt
-                        Fiber.internal.foreachIndexed(Chunk.from(iterable.grouped(groupSize)))((idx, group) =>
-                            Kyo.foreachIndexed(Chunk.from(group))((idx2, v) => isolate.isolate(state, f(idx + idx2, v)))
-                        ).map(_.use(r => Kyo.foreach(r.flattenChunk)(isolate.restore)))
+                        val items = Chunk.from(iterable).toIndexed
```
hearnadam:
```suggestion
                        val items = Chunk.Indexed.from(iterable)
```

kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -681,45 +681,54 @@ object Fiber:
...
-                            val safepoint = Safepoint.get
-                            foreach(iterable) { (idx, v) =>
-                                val fiber = IOTask(f(idx, v), safepoint.copyTrace(trace), context)
+        def foreachIndexed[E, A, B](items: Chunk.Indexed[A], concurrency: Int)(
+            f: (Int, A) => B < (Abort[E] & Async)
+        )(using frame: Frame): Fiber[Chunk[B], Abort[E]] < Sync =
+            val size = items.size
+            if size == 0 then Fiber.succeed(Chunk.empty)
+            else
+                val numWorkers = Math.min(size, concurrency)
```
hearnadam, line 692: If this is 1, should we dispatch to `Kyo.foreach`?

hearnadam (approved):
How did I not notice this?

ahoy-jon:
> How did I not notice this?

Me neither

ahoy-jon (approved):

PR #1517, opened 2026-04-16
[schema] introduce module

fwbrasil:
From the module's README:

# kyo-schema

Define a case class and get JSON serialization, Protobuf encoding, field validation, type-safe lenses, structural diffs, and more, all derived from the type's structure. No annotations, no boilerplate. Works across JVM, JavaScript, and Scala Native.

The module depends only on `kyo-data` (pure data structures) and has no dependency on Kyo's effect runtime, so it can be adopted as a standalone library.

```scala
// Schema-based JSON and Protobuf
Json.encode(alice)      // {"id":1,"name":"Alice","password":"secret","address":{"city":"Portland","zip":"97201"}}
Protobuf.encode(alice)  // Span[Byte] (binary)

// Type-safe lenses reach any depth
Schema[User].focus(_.address.city).update(alice)(_.toUpperCase)
// User(1, "Alice", "secret", Address("PORTLAND", "97201"))

// Type-level reshaping: drop sensitive fields, rename, add computed
val publicView = Schema[User]
    .drop(_.password)
    .rename(_.name, "displayName")
// Encodes alice as: {"id":1,"displayName":"Alice","address":{"city":"Portland","zip":"97201"}}

// Diffs as data: capture just what changed, ship, replay
val renamed = alice.copy(name = "Alicia")
val cs = Changeset(alice, renamed) // Produces a changeset with the new name patch
cs.applyTo(alice)                  // Result.Success(renamed)
```

Everything flows from `Schema[A]`, the central type that captures a type's structure at compile time. It's the single source of truth that powers serialization, validation, navigation, and conversion.

The module provides these primitives:

| Primitive | Purpose |
|-----------|---------|
| `Json` / `Protobuf` | Serialize to JSON strings or Protocol Buffers bytes |
| `Focus` | Type-safe lens for reading, writing, and modifying fields at any depth |
| `Compare` | Read-only field-by-field comparison of two values |
| `Modify` | Batched field mutations applied as a single unit |
| `Changeset` | Serializable diff that can be stored, transmitted, and replayed |
| `Builder` | Incremental, type-safe case class construction |
| `Convert` | Bidirectional conversion between structurally compatible types |
| `Structure` | Runtime type description and untyped value trees |


kyo-schema/shared/src/main/scala/kyo/internal/Ryu.scala
```diff
@@ -0,0 +1,721 @@
...
+        else if v < 10000000000L then 10
+        else if v < 100000000000L then 11
+        else if v < 1000000000000L then 12
+        else if v < 10000000000000L then 13
+        else if v < 100000000000000L then 14
+        else if v < 1000000000000000L then 15
+        else if v < 10000000000000000L then 16
+        else if v < 100000000000000000L then 17
+        else if v < 1000000000000000000L then 18
+        else 19
```
hearnadam, line 44: It's beautiful

kyo-schema/shared/src/main/scala/kyo/Protobuf.scala
```diff
@@ -0,0 +1,193 @@
...
+                    case "Int"     => "sint32"
+                    case "Long"    => "sint64"
+                    case "Double"  => "double"
+                    case "Float"   => "float"
+                    case "String"  => "string"
+                    case "Boolean" => "bool"
+                    case "Short"   => "sint32"
+                    case "Byte"    => "sint32"
+                    case "Char"    => "sint32"
+                    case _         => "string"
```
hearnadam: Should this be reachable?

kyo-schema/shared/src/main/scala/kyo/internal/BuilderAt.scala
```diff
@@ -0,0 +1,25 @@
...
+  * @tparam Remaining
+  *   The current remaining fields
+  * @tparam Focus
+  *   The type of the field being set
+  * @tparam Name
+  *   The singleton string type of the field name
+  */
+final class BuilderAt[Root, Remaining, Focus, Name <: String](
+    private[kyo] val builder: Builder[Root, Remaining]
+):
```
hearnadam: Could this be an opaque type?

kyo-schema/shared/src/main/scala/kyo/Changeset.scala
```diff
@@ -0,0 +1,363 @@
...
+      *   - [[Op.SetField]]: replace a field with an entirely new value
+      *   - [[Op.RemoveField]]: delete a field from a record
+      *   - [[Op.SetNull]]: clear an optional field to null
+      *   - [[Op.NumericDelta]]: store a numeric delta rather than the absolute value
+      *   - [[Op.StringPatch]]: insert/delete/replace a character range within a string
+      *   - [[Op.Nested]]: recursively patch a nested record
+      *   - [[Op.SequencePatch]]: add, remove, or reorder sequence elements
+      *   - [[Op.MapPatch]]: add, remove, or update map entries
+      */
+    sealed abstract class Op derives CanEqual, Schema:
```
hearnadam: I would prefer this name:
```suggestion
    sealed abstract class Patch derives CanEqual, Schema:
```

kyo-schema/shared/src/main/scala/kyo/internal/JsonWriter.scala
```diff
@@ -0,0 +1,404 @@
...
+/** High-performance JSON writer.
+  *
+  * Escape lookup table technique and digit-pair int/long writing inspired by jsoniter-scala by Andriy Plokhotnyuk. Double/float formatting
+  * uses the Ryu algorithm by Ulf Adams.
+  */
+final class JsonWriter private (
+    private var buf: Array[Byte],
+    private var pos: Int,
+    private var depth: Int,
+    private var needsComma: Array[Boolean]
```
hearnadam: Could this be a Long in most cases? Or a bitpacked representation of the array?

kyo-schema/shared/src/main/scala/kyo/Focus.scala
```diff
@@ -0,0 +1,419 @@
...
+      *   transformation function for the focused value
+      * @return
+      *   a new root with the focused value transformed
+      * @see
+      *   [[Modify.update]] for batched, multi-field mutations
+      */
+    def update(root: Root)(f: Value => Value): Root = updateFn(root, f)
+
+    /** Returns the type tag for the focused value type. Summoned at compile time. */
+    def tag(using t: Tag[Value]): Tag[Value] = t
```
hearnadam, line 87: Should this be retained in the class?

kyo-schema/shared/src/main/scala/kyo/Focus.scala
```diff
@@ -0,0 +1,419 @@
...
+        else Maybe.empty
+
+    /** Returns whether this focused field is optional (Maybe/Option typed). */
+    def optional: Boolean =
+        path.nonEmpty && {
+            val name = path.last
+            schema.sourceFields.exists { sf =>
+                sf.name == name && {
+                    val s = sf.tag.show
+                    s.startsWith("scala.Option[") || s.startsWith("kyo.Maybe[")
```
hearnadam: Why not compare with tag literals?

kyo-schema/shared/src/main/scala/kyo/Json.scala
```diff
@@ -0,0 +1,257 @@
...
+      *   maximum nesting depth for objects/arrays (default 512)
+      * @param maxCollectionSize
+      *   maximum number of entries in maps, sets, or arrays (default 100000)
+      * @return
+      *   the decoded value, or a DecodeException if the input is malformed or does not match the schema
+      */
+    def decodeBytes[A](
+        input: Span[Byte],
+        maxDepth: Int = 512,
+        maxCollectionSize: Int = 100000
```
hearnadam: nit: might be useful to have a constant somewhere.

kyo-schema/shared/src/main/scala/kyo/Json.scala
```diff
@@ -0,0 +1,257 @@
...
+      *
+      * This overload does not require a Schema[A] in scope and does not incorporate runtime metadata such as validation constraints,
+      * descriptions, or examples. Use the `Schema[A]`-enriched overload when those are needed.
+      *
+      * @tparam A
+      *   the type to describe
+      * @return
+      *   a JsonSchema derived purely from the compile-time structure of A
+      */
+    inline def jsonSchema[A]: JsonSchema = JsonSchema.from[A]
```
hearnadam, line 109: This is really useful for LLM interactions.

kyo-schema/shared/src/main/scala/kyo/Json.scala
```diff
@@ -0,0 +1,257 @@
...
+
+        private def fromStructure(rt: Structure.Type, seen: Set[String]): JsonSchema =
+            rt match
+                case Structure.Type.Primitive(name, _) =>
+                    name match
+                        case "Int" | "Long" | "Short" | "Byte" | "BigInt" => Integer()
+                        case "Double" | "Float" | "BigDecimal"            => Num()
+                        case "String"                                     => Str()
+                        case "Boolean"                                    => Bool
+                        case _                                            => Str()
```
hearnadam: These seem like they should be enumerated?

kyo-schema/shared/src/main/scala/kyo/Protobuf.scala
```diff
@@ -0,0 +1,193 @@
...
+                            }
+                            sb.append("}\n")
+                            afterFields.copy(messages = sb.toString :: afterFields.messages)
+
+                        case _ => visited
+                    end match
+                end if
+            end collect
+
+            def protoFieldDecl(name: String, tpe: Structure.Type, fieldNumber: Int, state: State): (String, State) =
```
hearnadam, line 128: Private?

kyo-schema/shared/src/main/scala/kyo/Protobuf.scala
```diff
@@ -0,0 +1,193 @@
...
+      *   - Sealed traits -> `message` with `oneof`
+      *   - Primitives: Int -> sint32, Long -> sint64, Double -> double, Float -> float, String -> string, Boolean -> bool, Short -> sint32,
+      *     Byte -> sint32, Char -> sint32
+      *   - List[A] / Vector[A] / Seq[A] / Chunk[A] -> `repeated`
+      *   - Option[A] -> `optional`
+      *   - Nested case classes -> referenced by message name (definitions collected)
+      */
+    object ProtoSchema:
+        /** Generates a `.proto` schema string for type `A`. */
+        inline def from[A]: String = fromStructure(Structure.of[A])
```
hearnadam, line 83: We should add a variant that can accept a path to write to disk during compilation?

kyo-schema/shared/src/main/scala/kyo/Focus.scala
fwbrasil: do you mean for untyped runtime access? Not sure, `Focus` is meant as a lightweight navigator DSL. Runtime reflection should use other APIs I'd say

kyo-schema/shared/src/main/scala/kyo/Protobuf.scala
fwbrasil: yeah, we need a good way to generate the files. It's not trivial, though. A runtime approach with tastyquery for example might have better usability and completeness including libraries. I can open a ticket to follow up on this. I've also been considering creating a kyo sbt plugin that could handle this kind of need.

kyo-schema/shared/src/main/scala/kyo/internal/Ryu.scala
fwbrasil: 🎄

fwbrasil:
I've pushed some fixes, optimizations, and migrated the codebase from `zio-json` to `kyo-schema`. [Benchmark results](https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/e92aa6082e6034000282b9466fb7be9e/raw/412c30247ef5cf00ff957bed417f9189d91d7532/jmh-result.json):

image
image

PR #1518, opened 2026-04-18
[http] Pure Scala HTTP transport + TLS + WebSocket + Unix sockets

fwbrasil:
**IMPORTANT**: I'm planning to remove `kyo-sttp` and `kyo-tapir` once this PR is merged.

Evolving and optimizing the `kyo-http` native integration with `h2o` and `curl` has proven quite challenging. This PR adds a pure Scala backend that can be shared across platforms with only the network interaction now being platform-specific. The JVM is using NIO in this PR, which doesn't have great performance, so I'm working on a new `kyo-ffi` module that will enable sharing the same `epoll` and `kqueue` impls across all platforms (using Panama on the JVM and koffi in JS/Node). We should explore adding `io_uring` eventually.

In addition, this PR introduces support for Websockets, Unix sockets, and TLS.

Even though the module is still using NIO, `kyo-http` significantly outperforms competitors. The combination of Kyo's high-performance async primitives, its adaptive scheduler, and diligent allocation avoidance in the backend are the main factors.

[Benchmark results](https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/d10597cca43b6dd5b43876e72a190fbb/raw/7300645661a29342635fac32f3e1fad24d5d4236/jmh-result.json):

image


kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -91,26 +91,29 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
         removeInterruptLoop(this)
     end removeInterrupt
 
+    def preInterrupt(): Boolean = true
+
     final def mask(): IOPromise[E, A] =
         val p = new IOPromise[E, A]:
-            override def interrupt(error: Error[E]): Boolean = false
+            override def preInterrupt() = false
```
fwbrasil, line 98: I made an optimization to avoid allocating the `Error` of interrupt in case the promise is already completed, which is common in hot paths. When linking promises, Kyo always dispatches interrupts to ensure consistency. 

The new `preInterrupt` method enables methods that used to override `interrupt` to mask promises without requiring the error. Only when an actual interruption happens the error is allocated.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -180,6 +183,16 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
 
     protected def onComplete(): Unit = {}
 
+    /** Resets a completed promise back to Pending.Empty for reuse. Returns true if successful, false if not completed or CAS fails. Only
+      * valid when no callbacks are registered (i.e., the promise was used as a single-owner notification, not with onComplete listeners).
+      * Protected — only subclasses (e.g., ReadPump, WritePump) should call this.
+      */
+    protected def becomeAvailable(): Boolean =
```
fwbrasil, line 190: I'm reusing `IOPromise` instances in `ReadPump` and `WritePump` to avoid allocations. I haven't exposed it in user-facing APIs but I think we should explore doing so, maybe as a new `kyo-core` primitive.

kyo-core/shared/src/main/scala/kyo/Log.scala
```diff
@@ -30,7 +30,7 @@ object Log:
         case silent extends Level(60)
     end Level
 
-    val live: Log = Log(Unsafe.ConsoleLogger("kyo.logs", Level.debug))
+    val live: Log = Log(Unsafe.ConsoleLogger("kyo.logs", Level.warn))
```
fwbrasil, line 33: I'll follow up with a separate PR fully reviewing `Log` and adding async logging support

kyo-data/shared/src/main/scala/kyo/Duration.scala
```diff
@@ -116,7 +116,10 @@ object Duration:
     end Units
 
     object Units:
-        private val byChronoUnit: Map[ChronoUnit, Units] = Units.values.map(u => (u.chronoUnit, u)).toMap
+
+        val all = Units.values
```
fwbrasil, line 120: I thought I had pushed this optimization to main already. Calling `Units.values` allocates a collection on each call

kyo-data/shared/src/main/scala/kyo/KyoException.scala
```diff
@@ -20,14 +20,9 @@ import scala.util.control.NoStackTrace
   *   Implicit Frame providing context about where the exception was created
   */
 class KyoException private[kyo] (
-    message: Text = "",
+    message: => Text = "",
     cause: Text | Throwable = ""
-)(using val frame: Frame) extends Exception(
```
fwbrasil, line 25: Since we override `getMessage` and `getCause`, there's no need to initialize the super class with them

kyo-data/shared/src/main/scala/kyo/KyoException.scala
```diff
@@ -20,14 +20,9 @@ import scala.util.control.NoStackTrace
   *   Implicit Frame providing context about where the exception was created
   */
 class KyoException private[kyo] (
-    message: Text = "",
+    message: => Text = "",
```
fwbrasil, line 23: Lazy message to reduce allocations to produce error messages

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
hearnadam: Perhaps this should be `canInterrupt`?

kyo-http/jvm/src/main/scala/kyo/internal/NioHandle.scala
```diff
@@ -0,0 +1,58 @@
...
+        val appInBuf  = ByteBuffer.allocate(session.getApplicationBufferSize)
+        val tlsState  = NioTlsState(engine, netInBuf, netOutBuf, appInBuf)
+        new NioHandle(channel, bufferSize, Present(tlsState))
+    end initTls
+
+    /** Close the handle: close channel, close SSLEngine if TLS (buffer is GC'd). */
+    def close(handle: NioHandle)(using AllowUnsafe): Unit =
+        handle.tls.foreach { tls =>
+            try tls.engine.closeOutbound()
+            catch case _: Exception => ()
```
hearnadam:
```suggestion
            catch case e if NonFatal(e) = ()
```

PR #1528, opened 2026-05-01
[core] complete IOPromise even when Result.Panic.apply rethrows fatals

gcsolaroli:
### Problem
`IOTask.eval`'s `catch case ex => completeDiscard(Result.panic(ex))` goes through the `Result.Panic.apply` factory, which rethrows fatal Throwables (LinkageError, OOM, ...). The rethrow escapes the catch handler, the worker thread dies via the JVM uncaught-exception path, and the IOPromise is never completed — any awaiter hangs forever.

### Solution
Wrap the call in an inner try/catch that falls back to `new Result.Panic(ex)` (case-class constructor, bypasses the `NonFatal` guard in `Panic.apply`) so the IOPromise is always completed and the panic surfaces through the normal fiber result channel.

Adds a regression test that throws a synthetic `LinkageError` from a fiber and verifies completion via `unsafe.poll`, deliberately bypassing `Async.use` / `Fiber.Unsafe.toFuture` — both have the same `Result.Panic.apply`-rethrow shape and are out of scope for this change; the patch closes the worker-death leak at the scheduler boundary only.

### Notes
'Problem' and 'Solution' sections where written by Claude. The problem showed up as failing tests in another project I am working on, and Claude suggested this solution.
The solution seems legit, even if I am not sure the are best ways to fix it.


kyo-core/shared/src/test/scala/kyo/FiberTest.scala
```diff
@@ -322,6 +322,30 @@ class FiberTest extends Test:
         }
     }
 
+    "fatal throwable from a fiber computation" - {
+        "LinkageError surfaces as Panic on the IOPromise instead of killing the worker" in run {
+            val fatal = new LinkageError("simulated NoClassDefFoundError")
```
hearnadam: We may want to define our own NonFatal, as I don't know if LinkageError should be considered as such.

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -90,7 +90,8 @@ sealed private[kyo] class IOTask[Ctx, E, A] private (
             end if
         catch
             case ex =>
-                completeDiscard(Result.panic(ex))
+                try completeDiscard(Result.panic(ex))
+                catch case _: Throwable => completeDiscard(new Result.Panic(ex))
```
hearnadam: I think the correct solution here is to both complete the promise and to re-throw.

kyo-core/shared/src/test/scala/kyo/FiberTest.scala
gcsolaroli: Everything started from this issue I had traced to zio-test:
- https://github.com/zio/zio/pull/10843

hearnadam: Yeah, this is one of the reasons I expect we should not treat it as fatal. Linkage errors or class not found (reflection) can cause 'fatals' when really they could be potentially handled by the caller.

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
gcsolaroli: Does the implementation in the new commit look better?

kyo-core/shared/src/test/scala/kyo/FiberTest.scala
fwbrasil: Regardless of the `NonFatal` logic, the fix seems beneficial. The fiber should not stay hanging on panics

gcsolaroli:
~~I have updated the test to make it run only for the JVM platform.
I was actually trying to avoid it running on the JS platform, but made the "filter" too strict. 😬~~
I should have managed to set the test for just the JVM and Native platform. 🤞

kyo-core/jvm-native/src/test/scala/kyo/FatalFiberTest.scala
```diff
@@ -0,0 +1,24 @@
+package kyo
+
+class FatalFiberTest extends Test:
+
+    "fatal throwable from a fiber computation" - {
+        "LinkageError completes the IOPromise with a Panic before the worker rethrows" in run {
```
hearnadam:
```suggestion
        "Fatal Error completes the IOPromise with a Panic before the worker rethrows" in run {
```

hearnadam (approved):

kyo-core/jvm-native/src/test/scala/kyo/FatalFiberTest.scala
```diff
@@ -0,0 +1,24 @@
...
+            val fatal                                 = new LinkageError("simulated NoClassDefFoundError")
+            val body: Int < (Sync & Abort[Throwable]) = Sync.defer { throw fatal; 0 }
+            for
+                signal <- Fiber.Promise.init[Int, Abort[Throwable]]
+                fiber  <- Fiber.initUnscoped(body)
+                _ <- Sync.defer {
+                    import AllowUnsafe.embrace.danger
```
fwbrasil: Use `Sync.Unsafe.defer` please, it provides the `AllowUnsafe` evidence

gcsolaroli: I have updated the test with both your suggestion and a different arrangement to try avoid long execution times that fail the CI builds.

gcsolaroli:
I don't understand why some tests still fail; they don't seem related to the change I have submitted. 😭

fwbrasil:
@gcsolaroli `main` had a merge issue that is fixed now and I've been attempting to run Kyo's build in free runners. We used to use paid runners that are much more stable. The limited capacity of free runners sometimes make tests flaky. I've been fixing issues as they appear but we seem close to stable builds again

PR #1529, opened 2026-05-01
[zio-test] force TestResult arrows inside the test body's scope

gcsolaroli:
### Problem
zio-test's `assertTrue` returns a `TestResult` whose `result: TestTrace` is a lazy val that re-evaluates the captured `TestArrow` chain on first access. With `KyoSpecDefault.run = ZIOs.run(Scope.run(v))`, `Scope.run` closes finalizers before zio-test inspects the result, so any captured expression that references a `Scope.acquireRelease`-managed resource observes torn-down state — e.g. `assertTrue(fileExists(file))` reports "Result was false" against a temp directory just deleted by the scope's finalizer.

### Solution
Splice a `.map { tr.result; tr }` step into both `KyoTestConstructor` and `KyoCheckConstructor` so the arrow chain is forced inside the test body's effect scope, ahead of finalizer close.

Adds `ScopeAssertEvalSpec` with one regression case (assertion against a file inside a `Scope.acquireRelease` temp dir) plus two controls (eagerly-bound boolean, `Sync.defer` immediately before the yield) that isolate the property to TestResult eval ordering.

### Notes
'Problem' and 'Solution' sections where written by Claude. The problem showed up as failing tests in another project I am working on, and Claude suggested this solution.
The solution seems legit, even if I am not sure the are best ways to fix it.

### DIFF
--- a/kyo-zio-test/shared/src/main/scala/kyo/test/KyoSpecAbstract.scala
+++ b/kyo-zio-test/shared/src/main/scala/kyo/test/KyoSpecAbstract.scala
@@ -46,7 +46,16 @@ abstract class KyoSpecAbstract[S] extends ZIOSpecAbstract:
                 Spec.labeled(
                     label,
                     Spec
-                        .test(ZTest(label, run[A](assertion)), TestAnnotationMap.empty)
+                        .test(
+                            ZTest(
+                                label,
+                                run[A](assertion.map { result =>
+                                    val _ = result.result
+                                    result
+                                })
+                            ),
+                            TestAnnotationMap.empty
+                        )
                         .annotate(trace, sl :: Nil)
                 )
 
@@ -54,6 +63,10 @@ abstract class KyoSpecAbstract[S] extends ZIOSpecAbstract:
         new CheckConstructor[Any, A < S1]:
             type OutEnvironment = Any
             type OutError       = Throwable
-            def apply(input: => A < S1)(using Trace): ZIO[OutEnvironment, OutError, TestResult] = run(input)
+            def apply(input: => A < S1)(using Trace): ZIO[OutEnvironment, OutError, TestResult] =
+                run(input.map { result =>
+                    val _ = result.result
+                    result
+                })
 
 end KyoSpecAbstract
--- a/kyo-zio-test/shared/src/test/scala/kyo/test/ScopeAssertEvalSpec.scala
+++ b/kyo-zio-test/shared/src/test/scala/kyo/test/ScopeAssertEvalSpec.scala
@@ -1,54 +1,32 @@
 package kyo.test
 
-import java.nio.file.Files
-import java.nio.file.Path
+import java.util.concurrent.atomic.AtomicBoolean
 import kyo.*
 import zio.test.*
 
 object ScopeAssertEvalSpec extends KyoSpecDefault:
 
-    private def tempDir(using Frame): Path < (Scope & Sync) =
+    private def aliveFlag(using Frame): AtomicBoolean < (Scope & Sync) =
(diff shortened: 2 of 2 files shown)


kyo-zio-test/shared/src/main/scala/kyo/test/KyoSpecAbstract.scala
```diff
@@ -46,14 +46,27 @@ abstract class KyoSpecAbstract[S] extends ZIOSpecAbstract:
                 Spec.labeled(
                     label,
                     Spec
-                        .test(ZTest(label, run[A](assertion)), TestAnnotationMap.empty)
+                        .test(
+                            ZTest(
+                                label,
+                                run[A](assertion.map { result =>
+                                    val _ = result.result
```
fwbrasil, line 53: What does this do?

gcsolaroli: It's a way to force the evaluation of TestResult.result, that otherwise may not be evaluated, letting other assertions fail.
The accompanying `ScopeAssertEvalSpec` regression test exhibits this directly: `assertTrue(fileExists(file))` against a file inside a `Scope.acquireRelease` temp dir — fails without this `.map`.

gcsolaroli:
I have updated the test in order to pass also for the JS platform.

kyo-zio-test/shared/src/test/scala/kyo/test/ScopeAssertEvalSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+    private def aliveFlag(using Frame): AtomicBoolean < (Scope & Sync) =
+        Scope.acquireRelease(
+            Sync.defer(new AtomicBoolean(true))
+        )(flag => Sync.defer(flag.set(false)))
+
+    def spec = suite("ScopeAssertEvalSpec")(
+        test("assertTrue captures inside Scope see live resources"):
+            for
+                alive <- aliveFlag
+            yield assertTrue(alive.get)
```
hearnadam, line 18: Sorry for the additional work, but would it be possible to write the same test in ZIO? I am wondering if this is a framework level issue.

Another possibility is that we need to bind the closure of `kyo.Scope` to the `zio.Scope`...

gcsolaroli: I'll give it a shot.

gcsolaroli: @hearnadam I have reproduced the issue on zio-test as well, but I am not fully sure about the fix: https://github.com/zio/zio/pull/10844

gcsolaroli: I have pushed a "better" fix that presumably now doesn't break all other tests 🤦‍♂️: https://github.com/zio/zio/pull/10844/commits/1005ef3d77c62ec3db52bcc93fb992f26299ad7e

gcsolaroli: The third try looks much better: https://github.com/zio/zio/pull/10844/commits/18b4a3c2533f13989d15dbbcede4e27ec976f4bd

fwbrasil (approved):

hearnadam:
TBH I don't know if we need this, but I suppose it's alright. As seen in the ZIO repo, this seems like it will always be an issue with deferring the capturing.

PR #1530, opened 2026-05-01
[zio-test] regression test for zio/zio#10843

gcsolaroli:
### Problem
zio-test's smart-assertion macro reified the lhs of a method call as a value via `TestArrow.succeed(lhs)`. For Scala companion objects this worked, but for Java classes used as a static-method qualifier (e.g. `java.lang.Math.abs(-1)`) Scala 3 emitted a load of `Math$.MODULE$` at runtime — a class that doesn't exist for Java statics — throwing `NoClassDefFoundError: java/lang/Math$` from a kyo scheduler worker and hanging the test runner.

Adds `AssertJavaStaticSpec` exercising the upstream fix: each case is a `java.lang.*` static call directly inside `assertTrue`. Compiles unconditionally; passes only against a zio-test release containing zio/zio#10843.

### Solution
This is just a canary-test to verify when the zio-test fix will be available bumping `zioVersion` in `build.sbt`, once the above pull-request will be accepted.


gcsolaroli:
https://github.com/zio/zio/pull/10843 has been merged

fwbrasil:
@gcsolaroli it seems this requires the new ZIO release? Can we close it for now?

gcsolaroli:
Handle how you deem best, @fwbrasil .

PR #1531, opened 2026-05-03
[data] add Base64 encoder/decoder (RFC 4648)

fwbrasil:
Pure-Scala implementation of standard ("basic") Base64 encoding/decoding that cross-compiles to JVM, JS, and Native without depending on java.util.Base64 from shared/src/main.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Base64.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Base64.scala
@@ -1,0 +1,133 @@
+package kyo
+
+/** Cross-platform Base64 (RFC 4648) encoder/decoder.
+  *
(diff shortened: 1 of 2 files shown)


kyo-data/shared/src/main/scala/kyo/Base64.scala
```diff
@@ -0,0 +1,133 @@
...
+  */
+object Base64:
+
+    private val Alphabet: Array[Char] =
+        ("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/").toCharArray
+
+    private val Decode: Array[Int] =
+        val arr = Array.fill[Int](128)(-1)
+        var i   = 0
+        while i < Alphabet.length do
```
hearnadam, line 20: Doesn't this require a load? where a local val wouldn't?

PR #1543, opened 2026-05-13
feat(grpc): Kyo gRPC Support — fixes #390

frank-ih:
/claim #390

## Summary

This PR implements native gRPC support for the Kyo effect system (fixes #390).

## Deliverables

- [x] gRPC server works within Kyo concurrency model
- [x] gRPC client works within Kyo
- [x] Tests pass (207 core + 52 code-gen + 26 e2e = 285 tests)
- [x] Documentation added (README.md with quick start, streaming, error handling)
- [x] Example code provided (helloworld client/server)

## Features

- **Unary calls**: Single request, single response
- **Server streaming**: Single request, stream of responses  
- **Client streaming**: Stream of requests, single response
- **Bidirectional streaming**: Full-duplex streaming
- **Fiber-native**: Built on Kyo's `Async` and `Abort` effects
- **Backpressure-aware**: Respects gRPC flow control
- **Cross-platform**: JVM and Scala.js support

## Modules

- `kyo-grpc-core`: Runtime library with server/client utilities and call handlers
- `kyo-grpc-code-gen`: Protoc plugin for generating Kyo service bindings
- `kyo-grpc-e2e`: Integration tests covering all four gRPC call patterns
- `kyo-grpc/examples/helloworld`: Example project

## Validation

```bash
sbt -batch 'kyo-grpc-core/test; kyo-grpc-code-gen/test; kyo-grpc-e2e/test'
```

- `kyo-grpc-core`: 207 tests passed
- `kyo-grpc-code-gen`: 52 tests passed
- `kyo-grpc-e2e`: 26 tests passed

## Attribution

This work builds on the excellent gRPC implementation efforts from the Kyo community, particularly:
- @steinybot's foundational gRPC support in PR #1334
- @webhop123's refresh work in PR #1526
- @zxzok's client-streaming readiness race fix
- @akamabu's request deadline support

PR #1615, opened 2026-05-16
fix(grpc): stabilize bidi gRPC support

mt211211:
/claim #390

This continues the closed gRPC support work from #1543 and adds fixes targeted at the latest validation feedback.

Changes in this branch:
- simplify bidi request sending to mirror the existing client-streaming sender: send each outbound request directly, halfClose after outbound completion, and leave response/completion propagation to the listener path
- keep the delayed-response-headers path from deadlocking the first outbound bidi request
- fix the delayed-headers e2e test server so it binds the full TestService descriptor instead of only ManyToMany

Validation:
- git diff --check passes locally after the latest update
- I could not run sbt locally because Java and sbt are not available on this machine PATH
- @lymeow233-wq reported local validation on the previous head plus this focused patch: bidi ServiceTest 11 passed, full ServiceTest/e2e 27 passed, core 207 passed, code-gen 52 passed

Attribution: this branch is based on the implementation discussed in #1543, with follow-up fixes added to address the reported e2e failures. The latest bidi sender simplification folds in the locally validated patch suggested by @lymeow233-wq.


mt211211:
Follow-up from #1543 after the Windows Java 21 validation report: this branch keeps the closed gRPC implementation path but fixes the reported bidi e2e issues by binding the full delayed-header test service and by preserving outbound progress before waiting for readiness. cc @fwbrasil @steinybot, this is intended as the reviewable continuation path for #390.

lymeow233-wq:
I was able to run the e2e suite locally on `ca50c707` with Java/sbt. The remaining bidi failures appear to come from the readiness wait in `bidiStreaming.sendAndClose`: after sending an outbound message it can wait on readiness/completion in a way that either interrupts the shared completion wait or prevents the later failing request from reaching the server in the delayed-header cases.

The smallest local patch I found is to make bidi request sending mirror the existing client-streaming sender: send each request directly, `halfClose` after the outbound stream completes, and leave response/completion propagation to the existing listener path. That keeps the change scoped to `kyo-grpc/core/shared/src/main/scala/kyo/grpc/ClientCall.scala` and removes the extra request-side readiness/`CallClosed` race.

Local validation on top of this PR head:

- `kyo-grpc-e2e/testOnly kyo.grpc.ServiceTest -- -z bidirectional` -> 11 passed
- `kyo-grpc-e2e/testOnly kyo.grpc.ServiceTest` -> 27 passed
- `kyo-grpc-e2e/test` -> 27 passed
- `;kyo-grpc-core/test;kyo-grpc-code-gen/test` -> core 207 passed, code-gen 52 passed
- `git diff --check` passes

I have not opened a competing PR; this looks like a small follow-up patch that could be folded into this PR.

mt211211:
Thanks @lymeow233-wq, I folded this into the PR in 880362af. `bidiStreaming.sendAndClose` now mirrors the client-streaming sender: it sends each outbound request directly, half-closes after the outbound stream completes, and leaves response/completion propagation to the existing listener path. I also updated the PR body to reflect this follow-up and attribute your local validation. `git diff --check` passes locally; I still cannot rerun the sbt suite on this machine because Java/sbt are not available on PATH.

PR #1605, opened 2026-05-16
Support non-Throwable aborts in KyoApp

cerredz:
Closes #1186.

## Summary
- widen KyoApp's accepted effect boundary from Abort[Throwable] to Abort[Any]
- convert non-Throwable abort failures to a KyoApp.FailureException at the app boundary
- keep Throwable abort failures and panic handling on the existing Throwable path
- add KyoApp regression coverage for non-Throwable aborts through both main and Unsafe.runAndBlock

## Verification
- java "-Dsbt.server.forcestart=true" -Xms2G -Xmx4G -Xss10M -jar ..\tools\sbt-launch-1.10.11.jar "kyo-core/testOnly kyo.KyoAppTest"
- java "-Dsbt.server.forcestart=true" -Xms2G -Xmx4G -Xss10M -jar ..\tools\sbt-launch-1.10.11.jar "kyo-coreJS/Test/compile"
- git diff --check

PR #1619, opened 2026-05-17
Add kyo-case-app module for case-app CLI integration

### LINKED ISSUE #1620: Add kyo-case-app module for case-app CLI integration
## Summary

Add a new integration module, **`kyo-case-app`**, that connects [case-app](https://github.com/alexarchambault/case-app) command-line parsing with Kyo application entrypoints.

## Motivation

Many Kyo CLI apps need structured option parsing, help text, and subcommands. [case-app](https://github.com/alexarchambault/case-app) already provides this on JVM, Scala.js, and Scala Native. Today, authors must wire case-app's `CaseApp` / `Command` `run(options, remainingArgs)` hook to Kyo effects manually.

`KyoApp` already offers a familiar pattern: register effectful work in `run { ... }` blocks and read `args` inside them after `main` runs. We want the same ergonomics for case-app apps, with parsed options and remaining positional arguments available in scope.

### `KyoCaseApp[T]`

Extends `caseapp.CaseApp[T]`. Authors define a options case class (via case-app) and register Kyo effects:

```scala
object Greet extends KyoCaseApp[GreetOptions]:
  run {
    Console.printLine(s"Hello, ${options.name}!")
  }
```

`Greet.main` comes from case-app; `run` registers what executes after parsing.

### `KyoCommand[T]`

Extends `caseapp.core.app.Command[T]` for subcommands, with the same `run` machinery and access to `options` / `remainingArgs`.

Subcommands can be grouped with case-app's `CommandsEntryPoint` (e.g. a todo app with `create`, `complete`, `list`, `delete`, `start`).

### Effect scope

`run` blocks use the same effect bound as `KyoApp`: `Async & Scope & Abort[Throwable]`, with interrupt handling aligned to `KyoApp` on JVM/Native.

## Platform support

- JVM
- Scala.js
- Scala Native

(case-app 2.1.x)

## Documentation

Module README should cover:

- Dependency setup
- `KyoCaseApp` greeter example
- `KyoCommand` + `CommandsEntryPoint` todo-style example
- Link to case-app docs for option definitions (out of scope for kyo-case-app)
- sbt `run` / `runMain` and packaged binary invocation

## Implementation notes

- New cross-project `kyo-case-app` depending on `kyo-core`
- Mirror `KyoApp` platform-specific `run` registration (JVM/Native blocking vs JS fiber loop)
- Override case-app's `run(options, remainingArgs)` to stash parsed values before executing registered `run` blocks
- Avoid naming clash with case-app's `exit` (use `exitApp` for non-throwable error exits)

## Related

Implementation PR: https://github.com/getkyo/kyo/pull/1619

DamianReeves:
Closes #1620

## Summary

This PR adds **`kyo-case-app`**, a tri-platform module that connects [case-app](https://github.com/alexarchambault/case-app) to Kyo application entrypoints, and introduces **`KyoAppRunner`** in **`kyo-core`** so that runner logic (interrupts, result handling, JVM/JS execution) is shared rather than duplicated per integration.

Authors define case-app options and subcommands as usual, then register effectful work with **`run`** overloads after parsing. Parsed CLI data is passed **explicitly** (not via implicit `options` / `remainingArgs` accessors like `KyoApp`’s `args`). **`run { options => ... }`** is the recommended default; full-parse and no-arg overloads are available when needed.

## Motivation

`KyoApp` covers programs that only need raw `args`. Many CLIs need declarative options, positionals, and subcommands — [case-app](https://github.com/alexarchambault/case-app) handles that on JVM, Scala.js, and Scala Native.

Without this work, authors would either bridge case-app’s `run(options, remainingArgs)` hook to Kyo by hand, or copy `KyoApp`’s interrupt / `onResult` / platform execution for every new entrypoint style. **`KyoAppRunner`** is the shared integration API in `kyo-core` for “run Kyo effects after your framework has started” (case-app today; decline, zio-cli, Spark-style mains, etc. later without a new `kyo-*` module each time). Design rationale: [ADR on #1620](https://github.com/getkyo/kyo/issues/1620#issuecomment-4499126782).

### 1. New module: `kyo-case-app`

| Platform | Artifact |
|----------|----------|
| JVM | `io.getkyo %% kyo-case-app` |
| Scala.js | `io.getkyo %%% kyo-case-app` |
| Scala Native | `io.getkyo %%% kyo-case-app` |

- **Dependency:** `kyo-core`, [case-app 2.1.0](https://github.com/alexarchambault/case-app)
- **Layout:** `CrossType.Full` — `shared/`, `jvm-native/`, `js/`
- **Aggregates:** `kyoJVM`, `kyoJS`, `kyoNative`
- **Root README:** integrations table entry
- **Docs:** [`kyo-case-app/README.md`](kyo-case-app/README.md)

#### `KyoCaseApp[T]` / `KyoCommand[T]`

Extend `caseapp.CaseApp[T]` / `caseapp.core.app.Command[T]`. `main` comes from case-app; `run` registers Kyo effects after parsing.

#### Three `run` overloads (same name, resolved by block shape)

| Form | When to use |
|------|-------------|
| **`run { options => ... }`** | **Recommended** — typical CLI logic (flags/options only) |
| `run { (options, remainingArgs) => ... }` | Options plus leftover positionals |
| `run { ... }` | Effect does not use parsed CLI data (startup hooks, etc.) |

All overloads share one **`registerRun`** queue (via [`KyoAppRunner`](kyo-core/shared/src/main/scala/kyo/KyoAppRunner.scala)). Multiple blocks run in **object-initialization order** with the **same parse snapshot** per `main`.

```scala
object Greet extends KyoCaseApp[GreetOptions]:
  run { Console.printLine("starting") }
  run { options => Console.printLine(s"Hello, ${options.name}!") }
  run { (options, remainingArgs) =>
    Console.printLine(s"extras: ${remainingArgs.remaining.mkString(" ")}")
  }
```

#### Other

- **`exitHook`** — process exit on non-throwable failures (case-app reserves `exit`; case-app types override `exitHook` → `Platform.exit`)
- **`runTimeout`** — per-block timeout (default `Duration.Infinity`)

### 3. `kyo-core`: `KyoAppRunner` integration API

Refactors **`KyoApp`** onto shared runner types (addresses review feedback to deduplicate interrupt / execution / `onResult` logic).

| Type | Location | Role |
|------|----------|------|
| [`KyoAppRunner`](kyo-core/shared/src/main/scala/kyo/KyoAppRunner.scala) | shared | Public integration API: `initCode`, `runTimeout`, `exitHook`, `onResult`, `runInitCode` |
| [`KyoAppRunnerWithInterrupts`](kyo-core/shared/src/main/scala/kyo/KyoAppRunner.scala) | shared | SIGINT/SIGTERM + interrupt-aware `handle` |
| [`KyoAppInterrupts`](kyo-core/shared/src/main/scala/kyo/KyoAppInterrupts.scala) | shared | `private[kyo]` signal wiring |
| [`KyoAppRunnerPlatform`](kyo-core/jvm-native/src/main/scala/kyo/KyoAppRunnerPlatform.scala) | jvm-native | JVM/Native: `registerEffect` + `KyoApp.runAndBlock` |
| [`KyoAppRunnerPlatform`](kyo-core/js/src/main/scala/kyo/KyoAppRunnerPlatform.scala) | js | JS: deferred runs + long-lived fiber |

**`KyoApp`** is now a thin `extends KyoAppPlatformSpecific` (no inline interrupt copy). **`kyo-case-app`** mixes the same runner traits — **no** `KyoCaseAppInterrupts` or case-app-specific platform files.

**Why public, not `private[kyo]` only:** `private[kyo]` suffices for monorepo dedup in `package kyo`, but external entrypoints (decline, zio-cli, framework-owned `main` such as Spark) need a documented API on `kyo-core` without copying runner code or adding another full `kyo-*` module per library. See [ADR on #1620](https://github.com/getkyo/kyo/issues/1620#issuecomment-4499126782) (Alternatives 3 / 4).

### Architecture

```
case-app main(args)
  → parse → CaseApp.run(options, remainingArgs)
      → KyoCaseAppSupport (cliArgs, run overloads → registerRun)
          → KyoAppRunner.registerEffect (JVM: runAndBlock | JS: fiber loop)
              → handleWithInterrupts → onResult → exitHook
```

### Tests

| Suite | JVM | JS | Native |
|-------|-----|-----|--------|
| `KyoCaseAppTest` | ✓ | ✓ | ✓ (some `main` repetition tests canceled: too slow) |
| `KyoCaseAppRunOverloadTest` | ✓ | ✓ | — |
| `casetest.TodoAppFixtureTest` | ✓ | ✓ | ✓ |
| `KyoAppTest` (regression) | ✓ | — | — |

**21** tests on JVM/JS for `kyo-case-app`; **9** `KyoAppTest` on JVM after runner refactor.

Fixture todo app: [`casetest.TodoAppFixture`](kyo-case-app/shared/src/test/scala/casetest/TodoAppFixture.scala) (package `casetest` for `Frame` / strict enum equality).

## Review feedback

| Topic | Status |
|-------|--------|
| Explicit parsed CLI (not implicit `options` / `remainingArgs`) | Done |
| Multiple `run` blocks / registration order | Done — shared `registerRun` / `registerEffect` queue + tests |
| `run { options => ... }` as primary shape | Done — recommended in README |
| `run { ... }` when CLI data unused | Done |
| Deduplicate interrupt / execution / `onResult` with `KyoApp` | Done — **`KyoAppRunner`** ([thread reply](https://github.com/getkyo/kyo/pull/1619#discussion_r3261720529), [ADR on #1620](https://github.com/getkyo/kyo/issues/1620#issuecomment-4499126782)) |

## Example usage

```scala
libraryDependencies += "io.getkyo" %%% "kyo-case-app" % "<version>"
libraryDependencies += "com.github.alexarchambault" %%% "case-app" % "2.1.0"
```

```scala
object Greet extends KyoCaseApp[GreetOptions]:
  run { options => Console.printLine(s"Hello, ${options.name}!") }
```

```bash
sbt run -- --name Alice
./greet --name Alice
```

## Out of scope (follow-ups)

- `KyoCommandsEntryPoint` wrapper (use case-app’s `CommandsEntryPoint` directly)
- case-app completion / argument-file helpers
- Additional third-party adapters (decline, zio-cli, etc.) built on `KyoAppRunner`

## Test plan

- [x] `sbt 'kyo-core/testOnly kyo.KyoAppTest'`
- [x] `sbt 'kyo-case-app/test'` (JVM) — 21 tests
- [x] `sbt 'kyo-case-appJS/test'` (JS) — 21 tests
- [x] `sbt 'kyo-case-appNative/test'` (Native) — 18 run (3 canceled: repeated `main` too slow on Native)


kyo-core/shared/src/main/scala/kyo/KyoAppInterrupts.scala
```diff
@@ -0,0 +1,28 @@
+package kyo
+
+import kyo.internal.OsSignal
```
hearnadam, line 3: All of the logic for interruption, execution, and behaviors on the Result should ideally be merged/replaced by the logic already in our `KyoApp`. It should be possible to use most of this via extension? if not, we can make some methods `private[kyo]`?

kyo-case-app/jvm-native/src/main/scala/kyo/KyoCaseAppPlatformSpecific.scala
```diff
@@ -0,0 +1,14 @@
+package kyo
+
+trait KyoCaseAppPlatformSpecific:
+    self: KyoCaseAppSupport[?, Async & Scope & Abort[Throwable]] =>
+
+    final override protected def run[A](v: => A < (Async & Scope & Abort[Throwable]))(using Frame, Render[A]): Unit =
```
hearnadam: Does it make sense to support multiple `run` blocks with args being parsed? I somewhat envision this as becoming `A => B < S` instead of implicitly allowing the invocation of `options`?

What do you think?

DamianReeves: I had considered that as well, but was following the `KyoApp` behavior.
I'm open to it, but do we think `KyoApp` should eventually change to do the same?

DamianReeves: Updated to require explicit parsed inputs: `run { (options, remainingArgs) => ... }`. We dropped the implicit `options` / `remainingArgs` accessors so case-app does not mirror `KyoApp`'s `args` pattern — parsing is a single step, so the data flow is clearer.

**Question:** Do you want a second overload for apps that do not need CLI data at all (e.g. `run { ... }` with no parameters)? Users can already write `run { (_, _) => ... }` when they do not use the values. Happy to add a zero-arg overload if you think that is worth the extra API surface, or keep a single explicit shape.

DamianReeves: Follow-up: added a second `run { ... }` overload (no parameters) for ergonomics when a block does not use parsed CLI data — equivalent to `run { (_, _) => ... }` but nicer to write. Both overloads delegate to a single `registerRun` queue, so mixed registration keeps object-init order and every block still sees the same parse snapshot.

DamianReeves: Follow-up: added a third overload `run { options => ... }` for when only parsed options are needed (no `remainingArgs`). Together with the existing `run { (options, remainingArgs) => ... }` and the no-arg `run { ... }` form, all three delegate to the same `registerRun` queue so mixed registration keeps object-init order and the same parse snapshot. Overload resolution is covered in `KyoCaseAppRunOverloadTest` — no explicit type annotations required at call sites.

DamianReeves: To clarify intent: **`run { options => ... }` is the recommended default** for typical commands (flags/options only). Use `run { (options, remainingArgs) => ... }` when leftover positionals matter, and `run { ... }` for effects that skip parsed CLI data entirely. README and scaladoc are updated accordingly.

kyo-core/shared/src/main/scala/kyo/KyoAppInterrupts.scala
DamianReeves: Thanks for the review — agreed we shouldn't duplicate this logic.

I wrote up the options in an ADR on this branch: [kyo-case-app/docs/adr/0001-kyo-app-runner-integration-entrypoints.md](https://github.com/getkyo/kyo/blob/feat/case-app-support/kyo-case-app/docs/adr/0001-kyo-app-runner-integration-entrypoints.md)

**Direction:** I went with **Alternative 3** — a public **`KyoAppRunner`** integration API in `kyo-core` (with **`KyoAppInterrupts`** kept `private[kyo]` as in the ADR's hybrid note). `KyoApp` and `kyo-case-app` both mix `KyoAppRunnerWithInterrupts` + `KyoAppRunnerPlatform`; case-app no longer has its own interrupt/platform copies.

**Why public, not `private[kyo]` only:** `private[kyo]` is enough for in-repo dedup, but it makes **`kyo-core` harder to integrate with other entrypoint models** without another bespoke `kyo-*` module each time — e.g. decline, zio-cli, or frameworks that own their own `main` (Spark and similar). A documented `KyoAppRunner` lets those adapters depend on `kyo-core` and hook in after parse/bootstrap, rather than copy-pasting interrupts / `onResult` / platform execution again.

**What that gives us:** one place for interrupts, `onResult` / `exitHook`, and JVM/JS `registerEffect`, while case-app stays focused on parsing + the `run` overloads.

I'm open to feedback on naming, package placement, or narrowing the public API — happy to adjust before merge.

Latest: `017c4b9b` on this PR.

kyo-case-app/docs/adr/0001-kyo-app-runner-integration-entrypoints.md
```diff
@@ -0,0 +1,372 @@
+# ADR 0001: Shared application runner for CLI and framework entrypoints
```
fwbrasil: Let's avoid pushing files related to the development process to avoid polluting the repo. It can be confusing for agents as well.

DamianReeves: Agreed. There actually was a decision here that I wanted to be sure we were aligned on.

DamianReeves: I figured once we were aligned I can remove the doc

fwbrasil:
I merged https://github.com/getkyo/kyo/pull/1605. It touches KyoApp and is conflicting, sorry

DamianReeves:
## Architecture decision summary (`KyoAppRunner`)

Full ADR record is on the issue: https://github.com/getkyo/kyo/issues/1620#issuecomment-4499126782

### Problem
`kyo-case-app` initially duplicated ~55–80 lines of `KyoApp` runner logic (interrupts, `onResult`, JVM `runAndBlock`, JS fiber loop). Review asked to deduplicate without blocking future CLI/framework entrypoints.

### Decision
**Alternative 4 (hybrid):** public **`KyoAppRunner`** + **`KyoAppRunnerPlatform`** in `kyo-core`, with **`private[kyo]`** internals (`KyoAppInterrupts`). `KyoApp` and `kyo-case-app` both mix the runner; case-app keeps only parsing + `run` overloads.

### Why not only `private[kyo]`?
Enough for monorepo dedup in `package kyo`, but external adapters (decline, zio-cli, framework-owned `main`) need a documented `kyo-core` API without copy-paste or a new `kyo-*` module per library.

### Rejected alternatives (short)
- Status quo duplication
- Inherit `KyoApp.Base` from case-app (`run` / `main` clashes)
- Embed `KyoApp` instance / extension-only sugar / `kyo.internal` free functions

### What shipped
- `KyoAppRunner`, `KyoAppRunnerWithInterrupts`, `KyoAppRunnerPlatform` (JVM + JS)
- `KyoApp` refactored onto runner; upstream `Abort[Any]` support merged
- `kyo-case-app`: explicit `run` overloads (`options` recommended), `registerRun` → `registerEffect`; platform/interrupt copies removed

ADR markdown removed from the repo; decision lives in the issue comment above.

kyo-case-app/docs/adr/0001-kyo-app-runner-integration-entrypoints.md
DamianReeves: Moved the ADR info to the GitHub issue and included a summary in comments on this PR.

kyo-case-app/jvm-native/src/main/scala/kyo/KyoCaseAppPlatformSpecific.scala
DamianReeves: This has been addressed

kyo-core/shared/src/main/scala/kyo/KyoAppInterrupts.scala
DamianReeves: Added the ADR to the related GitHub issue as a comment: https://github.com/getkyo/kyo/issues/1620#issuecomment-4499126782

fwbrasil (approved):

PR #1624, opened 2026-05-18
Fix TMap initWith and contains

### LINKED ISSUE #1621: kyo TMap.contains ignores the key parameter
## What

`TMap.contains(key)` in kyo `0.19.0` ignores its `key` parameter and returns `!_.isEmpty` of the underlying map. For any non-empty map, `tmap.contains(any_key)` returns `true` regardless of whether the key is present.

## Source

`kyo-stm/shared/src/main/scala/kyo/TMap.scala` — `contains` is implemented as

```scala
def contains(key: K)(using Frame): Boolean < STM =
    self.use(!_.isEmpty)
```

The `key` parameter is bound but never referenced; the body collapses to "is this map non-empty".

## Minimal runnable reproducer

> Correction to the original report: the module is `io.getkyo::kyo-stm` (not `kyo-core`), latest published version is `0.19.0` (not `0.20.0`), and Scala `3.7.0` is required by the TASTy version.

Save as `repro.scala` and run `scala-cli run --server=false repro.scala`:

```scala
//> using scala 3.7.0
//> using dep io.getkyo::kyo-stm:0.19.0

import kyo.*

object Main extends KyoApp:
  run {
    val program =
      for
        map  <- TMap.init("a" -> 1, "b" -> 2)
        hasA <- map.contains("a")
        hasZ <- map.contains("z")
      yield (hasA, hasZ)

    STM.run(program).map { case (hasA, hasZ) =>
      println(s"""contains("a") = $hasA  (expected: true)""")
      println(s"""contains("z") = $hasZ  (expected: false)""")
      if hasA == hasZ then
        println("BUG REPRODUCED: contains ignores key parameter")
      else
        println("NOT REPRODUCED: contains works correctly")
    }
  }
```

Observed output:

```
contains("a") = true   (expected: true)
contains("z") = true   (expected: false)
BUG REPRODUCED: contains ignores key parameter
```

## Expected

`contains(k)` returns `true` iff `k` is one of the map's keys; otherwise `false`.

## Actual

For any non-empty TMap, every key — whether present or not — yields `true`.

## Suggested fix

```scala
def contains(key: K)(using Frame): Boolean < STM =
    self.use(_.contains(key))
```

---

Verified locally:
1. cross-vendor adversarial gate (codex prover + claude skeptic + 2 independent judges, distinct `judge_id` by vendor) returned `accepted`
2. `scala-cli run --server=false` on the repro above reproduces the symptom

AI-assisted report. **Apology**: my originally-posted body referenced a non-existent `kyo-core:0.20.0` artifact — this update corrects the coordinates and Scala version after I re-ran the repro against the actually-published `kyo-stm:0.19.0`. Bug is real; coordinates were wrong.

### LINKED ISSUE #1622: kyo TMap.initWith ignores the function parameter
## What

`TMap.initWith(entries*)(f)` in kyo `0.19.0` silently ignores the function `f` and returns the freshly-built `TMap` instead of `f(tmap)`. The return-type signature also drops the `A` type parameter, so `initWith` is observably broken at both compile-time (wrong return type) and runtime (`f` never executes).

## Source

`kyo-stm/shared/src/main/scala/kyo/TMap.scala` lines 62-72 (kyo `0.19.0`):

```scala
inline def initWith[K, V](inline entries: (K, V)*)[A, S](inline f: TMap[K, V] => A < S)(
    using inline frame: Frame
): TMap[K, V] < (IO & S) =                       // <-- return type is TMap[K, V], not A
    TID.useIOUnsafe { tid =>
        val trefs = entries.foldLeft(Map.empty[K, TRef[V]]) {
            case (acc, (k, v)) => acc.updated(k, TRef.Unsafe.init(tid, v))
        }
        TRef.Unsafe.init(tid, trefs)             // <-- builds the TMap; `f` is never used
    }
```

Compare with `TRef.initWith` and `TChunk.initWith` which both correctly thread `f`.

## Minimal runnable reproducer

> Correction to the original report: the module is `io.getkyo::kyo-stm` (with `kyo-core` for the basics), latest published `0.19.0`, Scala `3.7.0` required.

Save as `repro.scala` and run `scala-cli run --server=false repro.scala`:

```scala
//> using scala 3.7.0
//> using dep io.getkyo::kyo-core:0.19.0
//> using dep io.getkyo::kyo-stm:0.19.0

import kyo.*

@main def repro(): Unit =
    var fWasCalled = false

    val program =
        TMap.initWith("a" -> 1, "b" -> 2) { tmap =>
            fWasCalled = true
            tmap.size.map(n => s"INNER_F_RAN size=$n")
        }

    // If `f` were respected the inferred type of `program` would be
    //   String < (IO & STM)
    // In kyo 0.19.0 it is instead
    //   TMap[String, Int] < (IO & STM).
    // Compile-time ascription proves the drift:
    //   val expected: String < (IO & STM) = program   // does NOT compile

    val result =
        IO.run(STM.run(program)).eval

    println(s"fWasCalled after run = $fWasCalled")
    println(s"runtime result class = ${result.getClass.getSimpleName}")
    println(s"runtime result value = $result")
    if !fWasCalled then
        println("BUG REPRODUCED: TMap.initWith ignored the function parameter `f`.")
    else
        println("NOT REPRODUCED: f executed.")
```

Observed output:

```
fWasCalled after run = false
runtime result class = TRefImpl
runtime result value = TRef(state=Write(1,Map(a -> TRef(state=Write(1,1), lock=free), b -> TRef(state=Write(1,2), lock=free))), lock=free)
BUG REPRODUCED: TMap.initWith ignored the function parameter `f`.
```

## Expected

`initWith` returns `A < (IO & S)` (where `A` is the result of `f(tmap)`), and `f` is invoked exactly once with the freshly-built `TMap`.

## Actual

`initWith` returns `TMap[K, V] < (IO & S)` — wrong type — and `f` is never invoked.

## Suggested fix

```scala
inline def initWith[K, V](inline entries: (K, V)*)[A, S](inline f: TMap[K, V] => A < S)(
    using inline frame: Frame
): A < (IO & S) =
    TID.useIOUnsafe { tid =>
        val trefs = entries.foldLeft(Map.empty[K, TRef[V]]) {
            case (acc, (k, v)) => acc.updated(k, TRef.Unsafe.init(tid, v))
        }
        val tmap = TRef.Unsafe.init(tid, trefs)
        f(tmap)
    }
```

---

Verified locally:
1. cross-vendor adversarial gate (codex prover + claude skeptic + 2 independent judges, distinct `judge_id` by vendor) returned `accepted`
2. `scala-cli run --server=false` on the repro above reproduces the symptom

AI-assisted report. **Apology**: the originally-posted body referenced a non-existent `kyo-core:0.20.0` artifact — this update corrects the coordinates and Scala version after re-running the repro against the actually-published `kyo-stm:0.19.0`. Bug is real; coordinates were wrong.

NguyenCong2k:
## Summary
- fix `TMap.initWith` so it applies and returns the provided function result
- fix `TMap.contains(key)` so it checks the requested key instead of only map non-emptiness
- add regression coverage for both behaviors

Fixes #1621
Fixes #1622

## Test
- `kyo-stm/testOnly kyo.TMapTest` -> 29 passed

Note: `kyo-stm/test` currently fails in unrelated `STMTest` concurrency cases on this machine (`concurrent modifications` / `concurrent updates`).

### DIFF
--- a/kyo-stm/shared/src/main/scala/kyo/TMap.scala
+++ b/kyo-stm/shared/src/main/scala/kyo/TMap.scala
@@ -61,14 +61,14 @@ object TMap:
       */
     inline def initWith[K, V](inline entries: (K, V)*)[A, S](inline f: TMap[K, V] => A < S)(
         using inline frame: Frame
-    ): TMap[K, V] < (Sync & S) =
+    ): A < (Sync & S) =
         STM.withCurrentTransactionOrNew { tick =>
             val trefs =
                 entries.foldLeft(Map.empty[K, TRef[V]]) {
                     case (acc, (k, v)) =>
                         acc.updated(k, TRef.Unsafe.init(tick, v))
                 }
-            TRef.Unsafe.init(tick, trefs)
+            f(TRef.Unsafe.init(tick, trefs))
         }
 
     extension [K, V](self: TMap[K, V])
@@ -165,7 +165,7 @@ object TMap:
           *   true if the key exists, false otherwise
           */
         def contains(key: K)(using Frame): Boolean < STM =
-            self.use(!_.isEmpty)
+            self.use(_.contains(key))
 
         /** Updates the value associated with a key based on its current value.
           *
--- a/kyo-stm/shared/src/test/scala/kyo/TMapTest.scala
+++ b/kyo-stm/shared/src/test/scala/kyo/TMapTest.scala
@@ -26,6 +26,19 @@ class TMapTest extends Test:
             }
         }
 
+        "initWith applies function" in run {
+            STM.run {
+                TMap.initWith("a" -> 1, "b" -> 2) { map =>
+                    for
+                        _        <- map.put("c", 3)
+                        snapshot <- map.snapshot
+                    yield snapshot
+                }
+            }.map { snapshot =>
+                assert(snapshot == Map("a" -> 1, "b" -> 2, "c" -> 3))
+            }
+        }
+
         "add and contains" in run {
             STM.run {
                 for
@@ -36,7 +49,7 @@ class TMapTest extends Test:
                     value   <- map.get("key")
                 yield (exists, missing, value)
             }.map { case (exists, missing, value) =>
-                assert(exists && missing && value == Maybe(42))
+                assert(exists && !missing && value == Maybe(42))
             }
         }
 


NguyenCong2k:
Thanks for the review. This return-type change is intentional and matches the existing `initWith` convention used by nearby APIs such as `TRef.initWith`, `TChunk.initWith`, and `Memory.initWith`: `initWith` initializes the resource, applies the callback, and returns the callback result.

The previous `TMap.initWith` implementation accepted a callback but never invoked it, which contradicted its own Scaladoc (`The result of applying the function...`) and made `TMap.init(entries*) = initWith(entries*)(identity)` the only path that behaved as intended.

The map-return convenience remains available through `TMap.init(entries*)`, which still delegates to `initWith(entries*)(identity)`. This PR fixes `initWith` to match the documented behavior and the surrounding API pattern.

PR #1625, opened 2026-05-18
Enforce WebSocket max frame size

NguyenCong2k:
## Summary
- enforce `HttpWebSocket.Config.maxFrameSize` while decoding inbound WebSocket frames
- apply the configured limit on both server and client WebSocket read loops
- reject oversized 64-bit frame lengths before converting to `Int`
- add codec and server-dispatch regression coverage

## Test
```powershell
kyo-http/testOnly kyo.internal.WebSocketCodecTest
kyo-http/testOnly kyo.internal.WebSocketCodecTest kyo.internal.UnsafeServerDispatchTest -- -z maxFrameSize
kyo-http/testOnly kyo.internal.WebSocketCodecTest kyo.internal.UnsafeServerDispatchTest -- -z HttpWebSocket
```


fwbrasil:
thank you! I just spent a couple of hours debugging test failures and hadn't noticed this could be the issue 🙏

PR #1627, opened 2026-05-19
[compat] seamless cross-stack libraries

fwbrasil:
## Problem

Developing libraries in Scala is cumbersome because the ecosystem has multiple effect stacks people might want to use a library with. The common solution is typeclasses or hardcoding to a specific stack. Typeclasses add dispatch overhead and leak `F[_]` into signatures, still miss stacks without a lawful instance (Ox, plain `Future`), and degrade features like tracing. Hardcoding cuts off every consumer on a different runtime.

## Solution

`kyo-compat`: write a library once against the `kyo.compat.*` surface and ship it to 6 backends (Kyo, ZIO, Cats Effect, Future, Ox, Twitter Future), selected by the consumer at deploy time.

- `CIO[+A]` is a per-backend opaque type; every method is `inline def` and lowers to the backend primitive (no dispatch, no adapters).
- Fully isolated from Kyo's runtime. Each `kyo-compat-<backend>` artifact depends only on its target library and nothing else: no kyo-core, no Kyo runtime. A non-Kyo consumer (e.g. a ZIO user) pulls zero Kyo code.
- Cross-platform JVM/JS/Native, per each backend's supported set.
- An sbt plugin (`compatLibrary`) cross-publishes every `(backend x platform x scala-version)` cell from a single source tree.

See `kyo-compat/README.md` for the full surface and usage.

## Notes

- Two commits: a kyo-core `Channel` fix (cancelled producers no longer deliver their value), a prerequisite for the Kyo binding; and the `kyo-compat` module itself.
- Backend/platform support is not uniform: Future/Ox/Twitter Future are JVM-only, Cats Effect is JVM+JS, Kyo/ZIO are all three.
- Every binding runs the same shared test suite, verifying consistent behavior across backends and platforms.
- Out of scope by design: streams, scoped-resource types, and the Kyo/ZIO defect channel. Reach them via `c.lower`.

### DIFF
--- a/kyo-compat/bindings/ce/jvm/src/main/scala/kyo/compat/FromCompletionStage.scala
+++ b/kyo-compat/bindings/ce/jvm/src/main/scala/kyo/compat/FromCompletionStage.scala
@@ -3,17 +3,11 @@ package kyo.compat
 import cats.effect.IO
 import java.util.concurrent.CompletionStage
 
-/** JVM-only. CIO interruption propagates back to the source `CompletionStage` via `cf.cancel(false)` attached as an `onCancel` finalizer.
-  */
+/** JVM-only. Delegates to `IO.fromCompletionStage`, which propagates CIO interruption to the source via `cf.cancel(true)`. */
 object CompatFromCompletionStage:
 
     inline def fromCompletionStage[A](inline cs: CompletionStage[A]): CIO[A] =
-        CIO.lift {
-            IO.delay(cs.toCompletableFuture).flatMap { cf =>
-                IO.fromCompletableFuture(IO.pure(cf))
-                    .onCancel(IO.delay { val _ = cf.cancel(false) })
-            }
-        }
+        CIO.lift(IO.fromCompletionStage(IO.delay(cs)))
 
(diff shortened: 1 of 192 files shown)


kyo-compat/bindings/ce/jvm/src/main/scala/kyo/compat/FromCompletionStage.scala
```diff
@@ -0,0 +1,23 @@
...
+
+/** JVM-only. CIO interruption propagates back to the source `CompletionStage` via `cf.cancel(false)` attached as an `onCancel` finalizer.
+  */
+object CompatFromCompletionStage:
+
+    inline def fromCompletionStage[A](inline cs: CompletionStage[A]): CIO[A] =
+        CIO.lift {
+            IO.delay(cs.toCompletableFuture).flatMap { cf =>
+                IO.fromCompletableFuture(IO.pure(cf))
+                    .onCancel(IO.delay { val _ = cf.cancel(false) })
```
hearnadam: Perhaps I am missing something, but shouldn't the `fromCompletableFuture` handle this?

PR #1628, opened 2026-05-19
[scheduler] deliver fiber interrupts reliably

fwbrasil:
## Problem

`AsyncTest`'s `interrupt` tests ("one fiber", "multiple fibers") intermittently hung to the 60s test timeout — an intermittent CI failure on JVM and Native.

Root cause: a lost-update race on `Task.state`. That single `@volatile Int` packs the interrupt flag, the preemption flag, and accumulated runtime. `requestInterrupt` (called from the interrupting thread) and `addRuntime` (called from the worker thread in `Worker.runTask`'s finally block) both perform non-atomic read-modify-writes of it. `@volatile` provides visibility, not read-modify-write atomicity: a worker that read `state` *before* an interrupt landed and then wrote back `addRuntime(staleSnapshot)` silently clobbered the just-set interrupt and preemption bits. The interrupt was lost, `shouldPreempt()` stayed `false`, `IOTask.eval`'s evaluation loop never returned, the `Scope` finalizer never ran, and the fiber awaiting completion hung.

## Solution

Interruption is now sourced from `IOPromise`, whose state is CAS-completed to `Error` on interrupt and therefore cannot be lost. `IOTask.eval`'s stop check and `IOTask.needsInterrupt()` read it directly (`!isPending()`). `Task.requestInterrupt` is removed and `Task.state` packs only runtime and preemption — there is no longer a cross-thread interrupt write that can be raced. A looped regression test is added to `AsyncTest`.

## Notes

- A stress harness reproduced the hang at iteration 3 on unmodified code; with the fix it runs 1000+ consecutive iterations cleanly.
- Verified on JVM (`kyo-core` 1875, `kyo-scheduler` 170, `AsyncTest` 130), JS, and Native (`kyo-coreNative AsyncTest` 130 — the fix and regression test pass on Native).
- Scheduler-level interrupt unit tests are updated to the new model (interruption derived from the promise rather than a `Task` flag) — no coverage removed.
- `Task.state`'s remaining `doPreempt`/`addRuntime` race (runtime + preemption bits only) is benign: at worst a dropped best-effort time-slice preemption, retried on the next `checkStalling` pass.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
+++ b/kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
@@ -31,12 +31,16 @@ sealed private[kyo] class IOTask[Ctx, E, A] private (
     final override def removeFinalizer(f: Maybe[Error[Any]] => Unit) =
         finalizers = finalizers.remove(f)
 
-    // Bridges fiber-level interruption to thread-level interruption. When the promise is
-    // successfully interrupted (CAS from Pending to Error), sets the needsInterrupt flag
-    // and wakes the BlockingMonitor for immediate scan. If the worker thread is blocked
-    // (flat CPU time), Thread.interrupt() is dispatched within ~100μs.
+    // Fiber interruption is recorded by IOPromise.interrupt's CAS of the promise state to
+    // Error — the single source of truth. needsInterrupt and eval's stop check both read
+    // it, so an interrupt can never be lost to a racing scheduler-level state update.
+    final override def needsInterrupt(): Boolean =
+        !isPending()
+
+    // Wakes the BlockingMonitor for an immediate scan. If the worker thread is blocked
+    // (flat CPU time), Thread.interrupt() is dispatched once the monitor observes the
+    // interrupted promise.
     final override def preInterrupt(): Boolean =
-        requestInterrupt()
         Scheduler.get.notifyInterrupt()
         true
     end preInterrupt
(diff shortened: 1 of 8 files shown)


kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -51,7 +55,11 @@ sealed private[kyo] class IOTask[Ctx, E, A] private (
                 Isolate.internal.restoring(trace, this) {
                     ArrowEffect.handlePartial(erasedAbortTag, Tag[Async.Join], curr, context)(
                         stop =
-                            shouldPreempt() || (deadline != Long.MaxValue && clock.currentMillis() > deadline),
+                            // !isPending() is the authoritative interrupt signal — IOPromise.interrupt
+                            // CAS-completes the promise, so checking it here stops an interrupted fiber even if
+                            // the racing scheduler preemption flag was lost. Ordered after shouldPreempt() and
+                            // the deadline check so a step that stops for either of those skips the read.
```
hearnadam, line 61: This feels a bit duplicative of the above?

kyo-core/shared/src/main/scala/kyo/scheduler/IOTask.scala
```diff
@@ -51,7 +55,11 @@ sealed private[kyo] class IOTask[Ctx, E, A] private (
...
+                            shouldPreempt() || (deadline != Long.MaxValue && clock.currentMillis() > deadline) || !isPending(),
```
hearnadam: ?
```suggestion
                            shouldPreempt() || (deadline != Long.MaxValue && clock.currentMillis() > deadline) || needsInterrupt(),
```

hearnadam (approved):

PR #1629, opened 2026-05-19
[stm] expand test coverage and fix concurrency bugs

fwbrasil:
### Problem

The kyo-stm module shipped with thin test coverage and a few latent concurrency bugs that thorough testing surfaced:

1. Opacity could be violated: `TRef.use` did not reject reads of a write-locked ref, so a concurrent transaction could observe a half-applied multi-ref commit.
2. Writers could be starved indefinitely: under sustained reader pressure a commit that politely yielded the write lock on `readTick` would never win.
3. `TRef` holding `null` crashed `STM.run`'s commit path because the success matcher relied on `Result.Success.unapply`, which collapses `Present(null)` to `Absent` and fell through to MatchError.
4. `TTable.insert` and `TTable.upsert` shared a transactional `nextId` counter, which contended on every insert and could let an auto-assigned insert overwrite a caller-upserted record at the same id.
5. `TTable.Indexed.upsert` leaked stale index entries when an indexed field changed between the old and new record.

### Solution

Engine fixes:
- `TRef.use` now consults the write lock and aborts the transaction when the ref is locked, so an in-flight transaction never observes a half-applied multi-ref commit.
- Bounded barging: after a configurable number of polite attempts a commit barges past the `readTick` fairness yield, so a writer under sustained reader pressure always makes progress. The lock is never taken from a transaction that physically holds it and commit-time validation is unchanged. The threshold is a `StaticFlag[Int]` (`kyo.bargeThreshold`, default 4) at package level.
- The default retry schedule is sized to `Async.defaultConcurrency * 16` so a legitimately contended transaction has tail margin proportional to actual parallelism, while the budget stays bounded.
- The `STM.run` commit loop uses `Result.foldError` for the success/error split, which handles null payloads correctly without an `asInstanceOf`.
- `TTable.insert` now draws ids from a lock-free `AtomicInt` and skips any id `upsert` already occupies, so concurrent inserts no longer contend on a shared `TRef` and an auto-assigned insert never overwrites a caller-upserted record. `upsert` is a plain `store.put`. The contract is monotonic, unique ids; failed transactions consume ids (same as SQL sequences).
- `TTable.Indexed.upsert` drops the previous record's index entries before indexing the new record, so a changed indexed field no longer leaves a stale entry.

Tests:
- Spec expansion across `CommitBuffer`, `STM`, `TChunk`, `TMap`, `TRef`, `TRefLog`, and `TTable`.
- New property and stress suites covering opacity, writer-starvation, retry-budget exhaustion, and id-uniqueness under interleaved insert/upsert.
- The Java-serialization round-trip lives in a JVM-only source set; the module compiles and its suite passes on JVM, JS, and Native.

### Notes

- The barge path only ignores the `readTick` fairness yield. It still respects `LockMask == 0` and never steals a lock from a transaction that physically holds it. Opacity is unchanged.
- `bargeThreshold` is a top-level `StaticFlag[Int]` in `package kyo`, configurable via the `kyo.bargeThreshold` system property; it clamps negative values to 0.
- Auto-assigned ids are monotonic and unique but not dense: a failed or retried insert consumes an id. This matches Postgres/MySQL sequence semantics and is the price of avoiding a transactional counter that would serialize every insert.

### DIFF
--- a/kyo-stm/jvm/src/test/scala/kyo/STMSerializationJvmTest.scala
+++ b/kyo-stm/jvm/src/test/scala/kyo/STMSerializationJvmTest.scala
@@ -1,0 +1,27 @@
+package kyo
+
+// JVM-only: Java object serialization (java.io.ObjectOutputStream / ObjectInputStream)
+// has no equivalent on Scala.js or Scala Native.
+class STMSerializationJvmTest extends Test:
+
+    "TRef" - {
+
+        "serializes and deserializes preserving the committed value" in run {
+            for
+                ref0 <- TRef.init(42)
+                bytes <- Sync.defer {
+                    val baos = new java.io.ByteArrayOutputStream()
+                    val oos  = new java.io.ObjectOutputStream(baos)
+                    oos.writeObject(ref0)
+                    oos.close()
+                    baos.toByteArray
+                }
+                ref1 <- Sync.defer {
+                    val ois = new java.io.ObjectInputStream(new java.io.ByteArrayInputStream(bytes))
+                    ois.readObject().asInstanceOf[TRef[Int]]
+                }
+                v <- STM.run(ref1.get)
+            yield assert(v == 42, "deserialized TRef must yield original value via STM.run(ref.get)")
+        }
+    }
+end STMSerializationJvmTest
--- a/kyo-stm/shared/src/main/scala/kyo/STM.scala
+++ b/kyo-stm/shared/src/main/scala/kyo/STM.scala
@@ -72,8 +72,15 @@ object STM:
             case Present(tick) => f(tick)
         }
 
-    /** The default retry schedule for failed transactions */
-    val defaultRetrySchedule = Schedule.fixed(1.millis).jitter(0.5).take(20)
+    /** The default retry schedule for failed transactions.
+      *
+      * The retry budget scales with [[Async.defaultConcurrency]] because that is the level of concurrent contention a transaction faces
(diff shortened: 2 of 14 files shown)


kyo-stm/jvm/src/test/scala/kyo/STMSerializationJvmTest.scala
```diff
@@ -0,0 +1,27 @@
...
+                    val baos = new java.io.ByteArrayOutputStream()
+                    val oos  = new java.io.ObjectOutputStream(baos)
+                    oos.writeObject(ref0)
+                    oos.close()
+                    baos.toByteArray
+                }
+                ref1 <- Sync.defer {
+                    val ois = new java.io.ObjectInputStream(new java.io.ByteArrayInputStream(bytes))
+                    ois.readObject().asInstanceOf[TRef[Int]]
+                }
```
hearnadam, line 22: I think you can make a platform specific method in `BaseKyoKernelTest` that can be used to serialize `[A]` across a boundary. then we can test this with many different classes.

On JS it can just be `identity`

hearnadam (approved):

kyo-stm/jvm/src/test/scala/kyo/STMSerializationJvmTest.scala
fwbrasil: I don't know the JS runtimes well but I think both JS and Native don't have similar mechanisms to serialize objects and restore them like Java does

Issue #217, opened 2024-03-25
Wasm Support


fwbrasil:
Would it possible to somehow use Kyo in Wasm? I imagine we could try going via scalajs or [scala native](https://github.com/getkyo/kyo/issues/210). Would it be possible to have a mapping between Kyo effects and Wasm's mechanism to track capabilities?

hearnadam:
https://github.com/scala-js/scala-js/issues/4928

PR #1632, opened 2026-05-22
[data] Schedule fixes and coverage improvements

### LINKED ISSUE #1623: kyo Schedule.linear produces exponential (doubling) growth, not linear
## What

`Schedule.linear(1.second)` is documented/named as producing a **linear** delay sequence (1s, 2s, 3s, 4s, 5s …) but in fact produces **exponential / powers-of-two** growth (1s, 2s, 4s, 8s, 16s …) because of a recursive doubling bug in the `next` step.

## Source

`kyo-data/shared/src/main/scala/kyo/Schedule.scala:326-328` (kyo `0.19.0`):

```scala
case Linear(base) =>
  Maybe(base -> linear(base + base))
//                       ^^^^^^^^^^^^ — doubles the base every step;
//                       linear `base + delta` would be correct
```

## Minimal runnable reproducer

> Correction to the original report: the artifact is `io.getkyo::kyo-data` (not `kyo-core`), real published version is `0.19.0`, and Scala `3.7.0` is required because of TASTy `28.7`.

Save as `repro.scala` and run `scala-cli run --server=false repro.scala`:

```scala
//> using scala 3.7.0
//> using dep io.getkyo::kyo-data:0.19.0

import kyo.*
import scala.concurrent.duration.*

@main def repro(): Unit =
  val base = 1.second.asInstanceOf[Duration]
  var s = Schedule.linear(base)
  val delays = scala.collection.mutable.ListBuffer[Long]()
  for _ <- 1 to 5 do
    s.next match
      case Maybe.Present((d, next)) =>
        delays += d.toMillis
        s = next
      case _ => ()
  println(s"delays in ms = ${delays.toList}")
  val expected = List(1000L, 2000L, 3000L, 4000L, 5000L)
  val actual = delays.toList
  println(s"expected = $expected")
  println(s"actual   = $actual")
  assert(actual != expected, "bug not reproduced")
```

Observed output:

```
delays in ms = List(1000, 2000, 4000, 8000, 16000)
expected     = List(1000, 2000, 3000, 4000, 5000)
actual       = List(1000, 2000, 4000, 8000, 16000)
```

## Expected

Schedule named `linear(base)` should produce delays `base, 2·base, 3·base, 4·base, …`.

## Actual

Delays are powers of two of `base`: `base, 2·base, 4·base, 8·base, 16·base, …` — this is exponential, not linear.

## Suggested fix

```scala
case Linear(base) =>
  Maybe(base -> linear(base + initial))   // need to remember the original base
```

or refactor `Linear` to carry both `step` and `current` separately.

---

Verified locally:
1. cross-vendor adversarial gate (codex prover + claude skeptic + 2 independent judges, distinct `judge_id` by vendor) returned `accepted`
2. `scala-cli run --server=false` on the repro above runs to completion and confirms the doubling.

AI-assisted report. **Apology**: the originally-posted body referenced a non-existent `io.getkyo::kyo-core:0.20.0` artifact — this update corrects the coordinates and Scala version after I re-ran the repro against the actually-published `kyo-data:0.19.0`. Bug is real; coordinates were wrong.

fwbrasil:
Fixes #1623

## Problem

`Schedule.linear(base)` produces a doubling sequence (1s, 2s, 4s, 8s, ...) instead of a linear one (1s, 2s, 3s, ...). The single-field `Linear(base)` was recursing as `linear(base + base)`.

Deep audit of every combinator and combinator pairing surfaced 8 more bugs of similar shape (smart-constructor reductions that quietly lose information, asymmetric handling of terminal states, or runtime behavior that drifts from the docstring).

## Solution

Nine fixes in `kyo-data/shared/src/main/scala/kyo/Schedule.scala`:

1. `Linear` now carries `(current, step)` so the recursion is `Linear(current + step, step)`. Sequence is arithmetic as documented.
2. `Jitter` clamps the jitter factor into `[-1, 1]` before multiplying. Previously `Duration.*` silently floored negative results to Zero, biasing the mean upward by ~12.5% at factor 2.0.
3. `Repeat` redesigned to `Repeat(original, current, remaining)` and counts emissions, restarting `current` from `original` when it exhausts. `Schedule.fixed(X).repeat(N)` now emits N items instead of looping forever on infinite inners. The `repeat(1) == this` reduction is gone since it was wrong for multi-emission inners.
4. `MaxDuration` reduces `maxDuration(Zero)` to `Done` and the runtime detects no-progress self-loops (`d == Zero && s == schedule`), emitting once and stopping. `Schedule.fixed(Zero).maxDuration(d)` no longer loops infinitely.
5. `exponentialBackoff(initial, 1.0, maxBackoff)` reduces to `fixed(initial.min(maxBackoff))` instead of `fixed(initial)`, so the cap is respected when `initial > maxBackoff`.
6. `Anchored` normalizes the offset to `offset % period` for subsequent calls. `anchored(1.h, 2.h)` now emits 2.h then 1.h, 1.h, 1.h, instead of 2.h, 2.h, 2.h. First-call semantics unchanged so the initial offset still applies.
7. `andThen` no longer reduces `s.andThen(Schedule.immediate)` to `s`. Immediate emits one zero (unlike Done/Never which emit nothing), so for finite `s` the post-`s` zero was silently dropped.
8. `exponential(Zero, _)` and `exponentialBackoff(Zero, _, _)` reduce to `immediate.forever` to match `linear(Zero)` and `fibonacci(Zero, Zero)`. Previously the first two reduced to `immediate` (one-shot), the others to `immediate.forever`.
9. `max` now reduces `Done.max(s) = Done` and `s.max(Done) = Done`, symmetric with `min` and matching the docstring "produces the maximum delay of both schedules". Previously `max` continued with the surviving side after either became Done, contradicting the doc.

## Notes

Test suite grew from 110 to 170 (all green). Six existing tests asserted the now-fixed behavior and were updated: exponential zero-initial reduces to `immediate.forever`, andThen with immediate does not reduce, max of Done equals Done, and the three Repeat tests now assert N-total-emission semantics.

Audit also added passing coverage for combinator interactions that previously had no tests (Min runtime-Absent branch, AndThen orElse branch, Linear overflow clamping to Infinity, nested Take/MaxDuration/Repeat, Fibonacci(0, 1.s) starting sequence) and ports of related ZIO bugs: #7214 sub-millisecond fixed precision, #3943 andThen no-duplicate-on-handoff, #217 jitter range bounds, #2957 maxDuration over-fire.

Build passes JVM, JS, and Native compilation. Branch contains a single commit on top of main.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/Schedule.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Schedule.scala
@@ -26,9 +26,8 @@ sealed abstract class Schedule extends Serializable derives CanEqual:
       */
     final infix def max(that: Schedule): Schedule =
         this match
-            case Never     => this
-            case Done      => this
-            case Immediate => that
+            case Never | Done => this
+            case Immediate    => that
             case _ =>
                 that match
                     case Never     => that
--- a/kyo-data/shared/src/main/scala/kyo/Schedule.scala
+++ b/kyo-data/shared/src/main/scala/kyo/Schedule.scala
@@ -26,13 +26,15 @@ sealed abstract class Schedule extends Serializable derives CanEqual:
       */
     final infix def max(that: Schedule): Schedule =
         this match
-            case Never            => this
-            case Done | Immediate => that
+            case Never     => this
+            case Done      => this
+            case Immediate => that
             case _ =>
                 that match
-                    case Never            => that
-                    case Done | Immediate => this
-                    case _                => Max(this, that)
+                    case Never     => that
+                    case Done      => that
+                    case Immediate => this
+                    case _         => Max(this, that)
 
     /** Combines this schedule with another, taking the minimum delay of both.
       *
@@ -78,8 +80,8 @@ sealed abstract class Schedule extends Serializable derives CanEqual:
             case Done  => that
             case _ =>
                 that match
-                    case Done | Never | Immediate => this
-                    case _                        => AndThen(this, that)
+                    case Done | Never => this
+                    case _            => AndThen(this, that)
 
     /** Repeats this schedule a specified number of times.
       *
@@ -90,11 +92,10 @@ sealed abstract class Schedule extends Serializable derives CanEqual:
       */
     final def repeat(n: Int): Schedule =
         if n <= 0 then Schedule.done
-        else if n == 1 then this
         else
             this match
                 case Never | Done => this
-                case _            => Repeat(this, n)
+                case _            => Repeat(this, this, n)
 
     /** Limits the total duration of this schedule.
       *
@@ -105,6 +106,7 @@ sealed abstract class Schedule extends Serializable derives CanEqual:
       */
     final def maxDuration(maxDuration: Duration): Schedule =
         if !maxDuration.isFinite then this
+        else if maxDuration == Duration.Zero then Schedule.done
         else
             this match
                 case Never | Done | Immediate => this
@@ -218,7 +220,7 @@ object Schedule:
       */
     def linear(base: Duration): Schedule =
         if base == Duration.Zero then immediate.forever
-        else Linear(base)
+        else Linear(base, base)
 
     /** Creates a schedule with intervals following the Fibonacci sequence.
       *
@@ -243,7 +245,7 @@ object Schedule:
       *   a new schedule with exponentially increasing intervals
       */
     def exponential(initial: Duration, factor: Double): Schedule =
-        if initial == Duration.Zero then immediate
+        if initial == Duration.Zero then immediate.forever
         else if factor == 1.0 then fixed(initial)
         else Exponential(initial, factor)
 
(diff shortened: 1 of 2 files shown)


kyo-data/shared/src/main/scala/kyo/Schedule.scala
```diff
@@ -26,13 +26,15 @@ sealed abstract class Schedule extends Serializable derives CanEqual:
       */
     final infix def max(that: Schedule): Schedule =
         this match
-            case Never            => this
-            case Done | Immediate => that
+            case Never     => this
+            case Done      => this
+            case Immediate => that
```
hearnadam:
```suggestion
            case Never | Done => this
            case Immediate    => that
```

kyo-data/shared/src/main/scala/kyo/Schedule.scala
```diff
@@ -26,13 +26,15 @@ sealed abstract class Schedule extends Serializable derives CanEqual:
...
             case _ =>
                 that match
-                    case Never            => that
-                    case Done | Immediate => this
-                    case _                => Max(this, that)
+                    case Never     => that
+                    case Done      => that
+                    case Immediate => this
+                    case _         => Max(this, that)
```
hearnadam:
```suggestion
                    case Never | Done => that
                    case Immediate    => this
                    case _            => Max(this, that)
```

hearnadam (approved):

PR #1637, opened 2026-05-24
[compat] Stream support + scaladocs

### LINKED ISSUE #1630: [feature]: Support for streams in kyo-compat
### Problem Statement

`kyo-compat` is a great fit for libraries that want one source against many runtimes. I'm exploring it as a way to collapse the four backends in [Proteus](https://github.com/ghostdogpr/proteus) (Future, ZIO/ZStream, fs2, Ox/Flow) into one implementation.

The current `CIO` surface covers everything I need for unary work, but the proteus gRPC backends are mostly streaming — server-streaming, client-streaming, and bidi RPCs all push a stream-shaped value through user code. The README explicitly lists streams as out of scope:

> `fs2.Stream`, `zio.stream.ZStream`, and `kyo.Stream` are out of scope

I'd like to propose a **separate companion module** — `kyo-compat-stream` — so the core `CIO` surface stays as minimal as it is today, and only users who need streams pay the design/maintenance cost.

### Proposed Solution

I'd suggest leaving `scala.concurrent.Future` (and likely Twitter Future too) out of this module entirely. There is no widely-adopted stream type in those ecosystems, and trying to invent a pull-based encoding for them is exactly the tarpit that makes "unify streams" hard. The four runtimes that ship a stream library cover the realistic deployment targets:

| Backend | `CStream[A]` |
|---|---|
| Kyo | `kyo.Stream[A, Abort[Throwable] & Async]` |
| ZIO | `zio.stream.ZStream[Any, Throwable, A]` |
| Cats Effect | `fs2.Stream[cats.effect.IO, A]` |
| Ox | `ox.flow.Flow[A]` |

Libraries that target the streams module simply restrict to these four backends. Users on a Future-only deployment keep using `CIO` from core `kyo-compat`.

### Alternative Solutions

_No response_

### Current Workaround

_No response_

### Additional Context

_No response_

fwbrasil:
Fixes #1630


kyo-compat/README.md
```diff
@@ -342,6 +342,57 @@ def fetchBounded(urls: Seq[String]): CIO[CChunk[String]] =
...
+```
+
+This compiles and runs against every binding × supported platform.
+
+Constructors and terminals not in the surface compose from what is:
+
+- Failure stream: `CStream.init(CIO.fail(e))` — `init(c: CIO[Seq[A]])` propagates `c`'s failure.
+- Count: `s.foldPure(0L)((c, _) => c + 1L)`.
+
+### Known divergences (kyo binding)
```
fwbrasil, line 378: These are issues in Kyo's `Stream`. I'll follow up on them cc/ @johnhungerford

PR #1641, opened 2026-05-26
[schema] derivation fix: generic case-class defaults and nested transform routing

fwbrasil:
## Problem

Two macro-level bugs in `Schema.derived`:

1. `Schema.derived[F[A]]` crashed at macro expansion ("Expected an expression. This is a partially applied Term") for any generic case class with at least one default-valued field, because `MacroUtils.getDefault` emitted the companion's default-method reference without applying the case class's type arguments.

2. Discriminator-tagged sum types nested inside another case class lost their discriminator on both JSON and Protobuf round-trips, because `SerializationMacro` emitted direct `$schema.serializeWrite/Read` calls for nested fields, bypassing the transform-aware dispatch layer.

## Solution

`MacroUtils.getDefault` now takes a `TypeRepr` and applies `tpe.typeArgs` to the method reference before lifting to an `Expr`. Five call sites (`BuilderMacro`, `CodecMacro`, `FocusMacro`, `SchemaConvertMacro`, `StructureMacro`) pass the local `TypeRepr` already in scope.

`SerializationMacro`'s four nested-field call sites now route through `kyo.internal.SchemaSerializer.writeTo/readFrom` instead of calling `$schema.serializeWrite/Read` directly. The dispatch helper resolves the field's transform-aware Schema. For the Protobuf path specifically, the discriminator reader phase 0 now matches field tags via `inner.fieldParse() + matchField(nameBytes)` (a wire-format-agnostic match that compares parsed field tags rather than UTF-8 names), and `matchField` falls back to `CodecMacro.fieldId(expected)` for wire formats that report fields by numeric tag.

## Notes

`NestedTransformTest` (new) pins six scenarios: one-level discriminator nesting, two-level deep, `.drop`, `.rename`, `.add` (computed field), and discriminator survival through Protobuf round-trip. A seventh scenario (`discriminator + drop` combined) is `pending` as a documented design follow-up.

Eight new `JsonTest` scenarios + three fixtures in `SchemaTestData` (`MTGenericDefault[A]`, `MTGenericMaybe[A]`, `MTGenericTwoParam[A, B]`) cover the generic-default fix.

### DIFF
--- a/kyo-schema/shared/src/test/scala/kyo/NestedTransformTest.scala
+++ b/kyo-schema/shared/src/test/scala/kyo/NestedTransformTest.scala
@@ -1,0 +1,82 @@
+package kyo
+
+// --- Reporter's repro: discriminator on a nested sealed-trait field ---
+sealed trait NestedRO derives CanEqual
+object NestedRO:
+    final case class `string`(value: String) extends NestedRO derives CanEqual, Schema
+    final case class `number`(value: Int)    extends NestedRO derives CanEqual, Schema
+end NestedRO
+
+given Schema[NestedRO] = Schema.derived[NestedRO].discriminator("type")
+final case class NestedEnvelope(result: NestedRO) derives CanEqual, Schema
+
+// --- Two-deep nesting of the same discriminator ---
+final case class NestedTwoDeepMiddle(payload: NestedRO) derives CanEqual, Schema
+final case class NestedTwoDeepOuter(middle: NestedTwoDeepMiddle) derives CanEqual, Schema
+
+// --- .drop on nested schema ---
+final case class NestedDropInner(visible: String, secret: String) derives CanEqual
+given Schema[NestedDropInner] = Schema[NestedDropInner].drop("secret")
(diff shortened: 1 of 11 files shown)


kyo-schema/shared/src/test/scala/kyo/NestedTransformTest.scala
```diff
@@ -0,0 +1,79 @@
...
+    "add (computed field) on nested schema emits the computed field at the inner level" in {
+        val js = Json.encode(NestedAddOuter(NestedAddInner(3)))
+        assert(js.contains("\"derived\":6"), js)
+    }
+
+    "discriminator + drop combine on a nested schema" in {
+        // TODO Phase 4 follow-up: needs RO-variant fixture where each variant carries
+        // a shared `metadata: String` field so the drop has something concrete to remove
+        // at the nested level while the discriminator still applies; not required for the
+        // bug-unblock baseline.
```
hearnadam: Leftover?

PR #1644, opened 2026-05-27
[browser] cross-platform browser access with automatic settlement

fwbrasil:
## Problem

`kyo-playwright` is a thin wrapper over Microsoft's Playwright Java SDK. Two structural limits make it not a great primitive:

**1. JVM-only.** The Playwright Java SDK launches a per-platform Playwright Node.js driver subprocess and proxies CDP through it. JS and Native targets have no browser-automation story at all.

**2. No post-action settlement.** Playwright auto-waits for element prerequisites before each action (visible, enabled, stable, hit-testable) but does not wait for the page to settle after. The caller has to chain `waitForLoadState`, `waitForSelector`, `waitForResponse` for each action whose aftermath might matter, and has to know in advance which actions trigger which side effects. AI agents are where this breaks hardest, my attempts with kyo-playwright weren't very successful for more ephemeral/quick browser access code.

## Solution

`kyo-browser`: a Chrome DevTools Protocol (CDP) client built directly on kyo-http's WebSocket transport. Not a Playwright wrapper, not a Node-driver shim, just CDP over WS. Two main design choices drive the work:

**Cross-platform by construction.** Because the only transport is a WebSocket and the only protocol is CDP, the implementation compiles on JVM, JS, and Native. Per-platform code is confined to `BrowserLauncherPlatform` in each target's `{jvm,js,native}/src/main/scala/`. The CDP client, exchange, event dispatchers, DOM resolution, screenshot/PDF pipelines, dialog/popup/iframe handling, and the entire user-facing `Browser` API live in `shared/`.

**Action-aftermath settlement is the default.** Every page-mutating action (`Browser.click`, `fill`, `press`, `goto`, `eval` when it triggers nav) is routed through a settlement pipeline that combines:

- a `NavigationWatcher` tracking frame-level lifecycle (committed -> DOMContentLoaded -> load -> networkIdle, with degrade paths so a stuck `networkIdle` falls back to `load`),
- a `MutationSettlement` observer installed on `document.body` that watches DOM mutations to quiesce,
- a `StabilitySampler` that confirms repeated samples have converged before returning,
- per-element actionability checks (visibility, hit-testing, in-viewport, not-covered) gating clicks and fills with structured `BrowserElementNotActionableException.Reason` on failure.

The call returns when the page is ready for the next action, not when the action's event fired into the void. This is the property that makes the module workable as the substrate for AI-agent driving, and it is the reason it exists in this shape rather than as a thinner CDP binding.

A second-order benefit of owning the transport: typed exception hierarchy (`BrowserConnectionLostException`, `BrowserElementNotActionableException`, `BrowserDecodingException`, `BrowserNavigationException`, `BrowserProtocolErrorException`, etc.) so handlers pattern-match instead of grepping `e.getMessage`.

## Notes

The 9 `demo/*.scala` files under `kyo-browser/shared/src/test/scala/demo` are runnable end-to-end scenarios (httpbin form submission, GitHub trending, Wikipedia search, HackerNews scrape, registry race, cookie dance, GitHub 404 recovery, etc.). They live under test scope so they don't end up in the published jar, and stay executable via `sbt 'kyo-browser/Test/runMain demo.X'`.


kyo-browser/native/src/main/scala/kyo/internal/BrowserLauncherPlatform.scala
```diff
@@ -0,0 +1,11 @@
+package kyo.internal
+
+import kyo.*
+
+private[kyo] object BrowserLauncherPlatform:
+
+    /** No-op on Native: process lifecycle is managed by the native runtime; no JVM shutdown hooks are available. */
```
hearnadam: Can't the JVM file be shared with native?

PR #1645, opened 2026-05-27
[docs] kyo-doctest + readmes for all modules

fwbrasil:
## Problem

Module-level documentation in the repo was uneven. The root README was the only validated doc, checked by sbt-mdoc against a handful of fenced blocks. A few module READMEs (kyo-config, kyo-flow, kyo-http, kyo-pod, kyo-schema) existed and had drifted against the API they were meant to describe. The other 29 modules had no README at all.

Closing that gap meant validating hundreds of fenced Scala blocks across 30+ modules on every commit. mdoc was already a bottleneck at the much smaller scale we were running it at. There was also no shared mechanism for writing the next README or for catching when an API change broke an existing one, so any module added later would land into the same drift the existing ones had.

## Solution

- **kyo-doctest.** A new kyo-agnostic sbt plugin and JVM-side library that compiles (but does not run) every fenced Scala block in a Markdown file against its module's classpath, and validates every internal link inside the file. A block-level DSL in HTML-comment modifiers controls scope, expectation, target platforms, and setup-prelude injection. A content-hash cache keeps the per-block cost near zero. The aggregate `sbt doctest` task runs the whole repo's READMEs in 14 seconds on a warm cache, and CI runs it on every PR. mdoc is replaced.

- **The /readme skill.** A multi-stage agent pipeline at `.claude/skills/readme` that produced these READMEs and is what maintainers run to add or rewrite one. It walks a module's source, picks a use-case-first or type-first spine, drafts (or cleans up) the README, runs a whole-document critique pass that catches catalog-style writing where the README enumerates APIs instead of teaching them, and verifies coverage against the source inventory. Conventions (running-domain example, cluster headings, opener style) live inside the sub-skills rather than as prose to follow. AGENTS.md and CONTRIBUTING.md trim their "writing module READMEs" sections to a one-line pointer at the skill. CLAUDE.md is added as a symlink so Claude and other agents converge on the same instructions.

- **29 new module READMEs.** One per module that previously had none, plus a refreshed root index with an "Introduction to Kyo" section covering the `import kyo.*` convention, the pending type, composition, the shape of an effect (init / get / run lifecycle), effect ordering, and the idiomatic monadic style. Every fenced block compiles under `sbt <module>/doctest`.

## Notes

- The plugin auto-enables on every JVM project via `trigger = allRequirements`, so a new module gets doctest for free and opts out with `.disablePlugins(KyoDoctestPlugin)`. Non-JVM crosses are filtered out of the aggregate automatically.
- The link validator covers same-document anchors, relative paths to other files in the repo, and anchors into those files. External URLs (`http`, `https`, `mailto`, etc.) are skipped because network reachability is out of scope. Inline-code spans inside prose are masked before extracting links so Scala signatures like `Abort.run[E](v)` are not mistaken for Markdown links.
- The default predef is just `import kyo.*`. READMEs that need other imports keep them in their code blocks where the reader can see them. `Actor.Subject`, `Browser.Image`, `caliban.RootResolver`, `kyo.zioSchema` and friends are spelled out the same way a user copying the example would have to spell them. The three READMEs that genuinely need an `AllowUnsafe` witness (root, kyo-stats-registry, kyo-stats-otlp) import it explicitly in their code blocks and call out the risk.
- kyo-kernel sets `doctestFreshDriver := true`. Reusing the dotty driver across blocks trips a "denotation invalid in run N" assertion on macro-heavy code (kyo-kernel is the only current case, but the same toggle is what a dotty-cps-async user would need).
- `OTLPConfig.loadIfEnabled` widened from `private[otlp]` to public. The README documents it as the user-facing entry point and it was unreachable from user code.


README.md
```diff
@@ -1,3692 +1,357 @@
...
 
-We can still see the pure value (23) in the output, but now we can also see that it is a `Kyo`. This will work similarly for other unboxed types like `Maybe` and `Result` (see below). 
+Mental model: reach for `Kyo.*` when sequential processing is sufficient, `Async.*` when concurrent execution would help. The Kyo companion lives in kyo-prelude alongside the core effects, not on `Sync` or `Async` themselves, because it works for any effect row.
 
-Note that `Render` does not convert to string but to `Text`--an enriched `String` alternative provided and used internally by Kyo. Kyo methods for displaying strings all accept `Text` values (see `Console` and `Log`, below). Converting values using `Render` directly can be cumbersome, however, so Kyo also provides a string interpolator to construct properly formatted `Text`s automatically. To use this interpolator, prefix your interpolated strings with `t` instead of `s`.
+## Idiomatic style: monadic by default
```
fwbrasil: @ahoy-jon @hearnadam this is a departure from our recommendations so far. I think kyo-direct isn't mature enough and suggesting it can end up actually hurting adoption. We should explore a compiler plugin for direct syntax instead

PR #1646, opened 2026-05-27
[data] remove Text

### LINKED ISSUE #798: Optimize Text
The new `Text` implementation in `kyo-data` is using regular loops + `charAt`. Explore optimizations using direct memory access or Java's new vector APIs to avoid the overhead to read each character.

fwbrasil:
Fixes #798

## Problem

`Text` was introduced as an API placeholder (`opaque type Text >: String = String | Op` with lazy `Cut` / `Concat` / `Reverse`) on the understanding that it would later be optimized with SIMD or direct-memory backing (the premise of https://github.com/getkyo/kyo/issues/798). With 1.0 approaching, that decision needs to be made: invest in the optimization and migrate the rest of the String surface to Text so the lazy structure can amortize, or remove it.

The current footprint is small: ~30 files, almost all exception messages, `Render`, ANSI, and a few Parse signatures. Values are built from a String, lightly concatenated, then toString'd. The amortisation the lazy rope was designed for never shows up.

## Solution

Remove Text. All Text signatures become String. `Render[A].asText` becomes `asString`; `t""` returns String. The Text overloads in `Ansi` go away (the String overloads already cover everything). 

## Notes

A future Text backed by Vector API or MemorySegment cannot meaningfully beat String on the JVM. HotSpot's `StringLatin1` / `StringUTF16` intrinsics already emit SIMD (AVX2/AVX-512, NEON/SVE) for `indexOf` / `equals` / `hashCode` / `compareTo`.

The one place a custom type could earn its keep is substring-heavy workloads (parsers, HTTP header slicing) via zero-copy `asSlice`, where the downstream doesn't need a String. That's a different design than the current Text and a different scope than 1.0 needs.


regiskuckaertz:
Just wondered if you ever came across this string implementation: https://cedardb.com/blog/german_strings/

hearnadam:
Do we need the `t` interpolator now?

fwbrasil:
@regiskuckaertz thanks, that's interesting! but I don't think the investment is worth it at least for now with the backlog for 1.0. We can revisit after the release

fwbrasil:
> Do we need the `t` interpolator now?

It's still useful since it uses `Render`. I've renamed it to `render"text .."`

PR #1647, opened 2026-05-27
Add shared Kyo companion operations

hearnadam:
### Problem
Kyo collection/pure operations from the `Kyo` object are not available on the companion objects of all effects.

### Solution
Add abstract class to support sharing these methods throughout other effects.


fwbrasil:
why provide it in all effects? `Sync` and `Async` seem enough and intuitive

hearnadam:
> why provide it in all effects? `Sync` and `Async` seem enough and intuitive

It's just lift and unit for all effects.

fwbrasil:
I don't see a good motive to add those methods to all effects. We typically prefer canonical APIs and avoid duplication. I think this can be an exception given the discoverability issue but we should limit the duplication. An idea: 

```scala
// this can be in kyo.internal but public
class KyoOps[S] private[kyo]():
   def lift[A](v: A): A < S = ...
   def collectAll(v: Chunk[A < S]): Chunk[A] < S

// rename the Kyo companion to Pure in kyo-kernel
object Pure extends KyoOps[Any]

object Sync extends KyoOps[Sync]
```

The use would be more intuitive:

- If  only `kyo-prelude` is in the classpath, which isn't a common case, `Pure` can be used matching the no side effects restriction of the module
- Most users will be using `kyo-core` and, given the patterns in similar libraries, they'll try to find these methods in the equivalent of `IO`, landing on `Sync`
- The pairing of `Sync` and `Async` with a similar set of methods seems much more intuitive already and `Pure` would follow a similar naming

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -46,7 +46,7 @@ import scala.util.control.NonFatal
   */
 opaque type Async <: (Sync & Async.Join) = Async.Join & Sync
 
-object Async extends AsyncPlatformSpecific:
+object Async extends KyoOps[Sync], AsyncPlatformSpecific:
```
fwbrasil: I think we don't need it here. Async has separate collection APIs with similar naming

PR #1648, opened 2026-05-28
[core] Signal: reactive combinator API

fwbrasil:
## Problem

`Signal[A]` currently exposes `map` and the streaming helpers. Standard reactive combinators are missing: `switchMap`, `combineLatest`, the N-ary variants, `awaitAny`. They're the primitives any non-trivial reactive layer needs, and kyo-ui has been hitting the gap.

## Solution

Adds `switchMap`, `zip`, `combineLatest` to the instance, and `Signal.awaitAny`, `zipAll`, `combineLatestAll` to the companion. Named after Rx, not after the monadic operators. These don't satisfy monad laws and calling one `flatMap` would lie.

## Notes

Also fixes a pre-existing bug in `SignalRef.Unsafe.init`: the initial `nextPromise` was a plain `init()` while every rotation in `onUpdate` is `initMasked()`. The first subscriber to a fresh ref ended up with different interrupt semantics than every subsequent one. The initial promise now uses `initMasked()` to match.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Signal.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Signal.scala
@@ -220,13 +220,15 @@ object Signal:
         CanEqual[A, A],
         CanEqual[Chunk[A], Chunk[A]]
     ): Signal[Chunk[A]] =
-        if signals.isEmpty then initConst(Chunk.empty[A])
-        else
-            val sigs = Chunk.from(signals)
-            initRaw(
-                currentWith = [B, S] => f => Kyo.foreach(sigs)(_.current).map(f),
-                nextWith = [B, S] => f => Async.foreachDiscard(sigs, sigs.size)(_.next).andThen(Kyo.foreach(sigs)(_.current).map(f))
-            )
+        signals.size match
+            case 0 => initConst(Chunk.empty[A])
+            case 1 => signals.head.map(Chunk(_))
+            case n =>
+                val sigs = Chunk.from(signals, n)
+                initRaw(
+                    currentWith = [B, S] => f => Kyo.foreach(sigs)(_.current).map(f),
+                    nextWith = [B, S] => f => Async.foreachDiscard(sigs, sigs.size)(_.next).andThen(Kyo.foreach(sigs)(_.current).map(f))
+                )
 
     /** Zips a sequence of signals, emitting when any changes.
       *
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/Signal.scala
```diff
@@ -138,6 +197,58 @@ export Signal.SignalRef
...
 object Signal:
 
+    /** Waits for any of the given signals to change.
+      *
+      * @param signals
+      *   The signals to watch
+      */
+    def awaitAny(signals: Seq[Signal[?]])(using Frame): Unit < Async =
+        if signals.isEmpty then ()
+        else Async.race(signals.map(_.next)).unit
```
hearnadam, line 207: I suppose it doesn't matter, but shouldn't this be `raceFirst`?

kyo-core/shared/src/main/scala/kyo/Signal.scala
```diff
@@ -138,6 +197,58 @@ export Signal.SignalRef
...
+      * @return
+      *   A signal of Chunk that updates when any input changes
+      */
+    @nowarn("msg=anonymous")
+    inline def combineLatestAll[A](signals: Seq[Signal[A]])(using Frame, CanEqual[A, A]): Signal[Chunk[A]] =
+        signals.size match
+            case 0 => initConst(Chunk.empty[A])
+            case 1 => signals.head.map(Chunk(_))
+            case _ =>
+                val sigs = Chunk.from(signals)
```
hearnadam: It might be a nice optimization to accept a size in the from method as an optional arg to pre-allocate the buffer if a copy is needed.

kyo-core/shared/src/main/scala/kyo/Signal.scala
```diff
@@ -138,6 +197,58 @@ export Signal.SignalRef
...
+    @nowarn("msg=anonymous")
+    inline def zipAll[A](signals: Seq[Signal[A]])(
+        using
+        Frame,
+        CanEqual[A, A],
+        CanEqual[Chunk[A], Chunk[A]]
+    ): Signal[Chunk[A]] =
+        if signals.isEmpty then initConst(Chunk.empty[A])
+        else
+            val sigs = Chunk.from(signals)
```
hearnadam: Should we also optimize the unary case?

hearnadam (approved):

kyo-core/shared/src/main/scala/kyo/Signal.scala
fwbrasil: I think the semantics with failures not propagating immediately makes more sense. The main use case for `Signal` is UIs and more stable behaviors help mitigate temporary inconsistencies due to settlement of signals in processing.

PR #1649, opened 2026-05-29
[ui] kyo-ui: a pure-value web UI framework

fwbrasil:
## Problem

Kyo can serve HTTP and drive browsers, but it has no way to build an UI.

## Solution

kyo-ui describes web UIs as pure values that run unchanged in three places: a Scala.js client mounting to the DOM, a server-driven HTML-over-SSE deployment, or an HTML stream for SSR and tests. Signals are first-class values throughout the API, and the API itself is typed tightly enough that broad categories of HTML and state mistakes do not compile.

**A `UI` is a pure value, runnable on any of three targets.** `UI.runMount(ui)` mounts to the DOM on Scala.js. `UI.runHandlers(basePath)(ui)` exposes the same value as an HTTP triple (GET page + POST events + SSE diff stream). `UI.runRender(ui)` returns a `Stream[String, Async]` of full-page HTML. The same source becomes a single-page app or a server-push app; only the runner changes.

**A `Signal[A]` is a value of type `A` everywhere kyo-ui takes one.** Setters and child slots that accept `A` equally accept `Signal[A]` (read-only, re-renders on change) or `SignalRef[A]` (read-write, two-way binding); conditional rendering is `when(cond: Signal[Boolean])(ui)`. There is no wrapper type, no binder operator, no hooks; reactivity is the API, not a layer on top of it.

**Invalid states do not compile.** Container factories restrict their children at the type level (`ul(div(...))` is a compile error). `Length` variants are method-typed (`.padding` rejects `Auto`, `.gap` rejects `Pct`). Domain enums (`Target`, `Href`, `Keyboard`, `FileAccept`, `Color`) replace stringly-typed attributes. Capability traits gate setters (`.checked` only on `BooleanInput`, `.disabled` only on `HasDisabled`).

**Updates are fine-grained.** No virtual DOM, no component re-execution. The `UI` is built once; a subscription is registered at every point a signal appears, so when a signal emits only the subtree bound to it re-renders and only the DOM nodes inside it are patched. You pick the granularity at the call site.

Around that core: a full element and forms vocabulary, a typed `Style` DSL with pseudo-states and clamped values, typed event payloads (`MouseEvent`/`KeyboardEvent` with modifier chords), ARIA and `data-*` attributes, the `UI.Ast.*` case-class tree for pattern-matching, and the JS-only `UIWindow` (size/visibility/key signals) and `UILocation` (history routing with same-origin anchor interception) namespaces.

## Notes

- Reactivity reuses the kyo-core `Signal` combinators (`awaitAny`, `switchMap`, ...) added in #1648; kyo-ui carries no kyo-core changes of its own.
- Tests drive a real Chrome through kyo-browser rather than a DOM mock, so the server-push path is exercised end to end. The JS-only pieces cannot run in the Node-based Scala.js test runtime, so `kyo-ui-spa-harness` links a Chrome-loadable ES-module bundle (classpath: kyo-ui only) and tests invoke real Scala.js through a `@JSExportTopLevel` registry. That subproject exists so the browser is never faked.
- Web-only (no TUI backend). `runRender` re-emits the whole document per change; diff-pushing lives in the server-push handlers. The custom `dropdown` is a styleable `div` overlay, not a native `<select>`; gradients are linear. These are scoping choices.


kyo-ui/js/src/main/scala/kyo/internal/DomBackend.scala
```diff
@@ -0,0 +1,212 @@
...
+
+    /** Mount a UI into the page body. */
+    def mount(ui: UI)(using Frame): Unit < (Async & Scope) =
+        mountInto(ui, document.body)
+
+    /** Mount a UI into a specific DOM element selected by CSS selector. */
+    def mount(ui: UI, selector: String)(using Frame): Unit < (Async & Scope) =
+        Sync.defer {
+            val target = document.querySelector(selector)
+            if target == null then Abort.panic(KyoException(s"Element not found: $selector"))
```
hearnadam: Do we want raw KyoExceptions? Ie not subclasses

kyo-ui/js/src/main/scala/kyo/internal/DomBackend.scala
```diff
@@ -0,0 +1,212 @@
...
+            end if
+        end handler
+
+        Seq("click", "input", "change", "submit", "keydown", "keyup", "focus", "blur").foreach { t =>
+            document.body.addEventListener(t, handler, true)
+        }
+    }
+    end setupEventDelegation
+
+    private def findPathElement(el: dom.Element): dom.Element =
```
hearnadam: Should this return a Maybe?

kyo-ui/js/src/main/scala/kyo/internal/DomStyleSheet.scala
```diff
@@ -0,0 +1,89 @@
...
+          |nav, li, span, button, a { display: flex; flex-direction: row; align-items: center; }
+          |ul, ol { list-style: none; padding: 0; margin: 0; }
+          |h1, h2, h3, h4, h5, h6, p { margin: 0; }
+          |a { color: inherit; text-decoration: none; }
+          |table { border-collapse: collapse; width: 100%; }
+          |""".stripMargin
+    )
+
+    /** Forces class loading so the base CSS reset is injected. No-op after the first invocation. */
+    def injectBase()(using Frame): Unit < Sync = Sync.defer(())
```
hearnadam: How does this work?

kyo-ui/js/src/main/scala/kyo/UILocation.scala
```diff
@@ -0,0 +1,120 @@
...
+  *
+  * A capture-phase click interceptor installed on `document` rewrites same-origin no-modifier anchor clicks into `pushState`
+  * navigations, so plain `<a href="/foo">` links participate in client-side routing without explicit handler wiring. Modifier-key
+  * clicks (ctrl/meta/shift/alt) and non-primary mouse buttons are NOT intercepted; the browser handles those natively (new tab, etc.).
+  */
+object UILocation:
+
+    private lazy val currentRef: SignalRef[String] =
+        import AllowUnsafe.embrace.danger
+        SignalRef.Unsafe.init(readWindowPath()).safe
```
hearnadam, line 20: Might be nice to have a `def apply[A](a: AllowUnsafe ?=> A): A` method to avoid the import keyword, just 

`AllowUnsafe.embrace.danger:`

kyo-ui/jvm/src/test/scala/kyo/internal/spa/SpaHarnessLocation.scala
```diff
@@ -0,0 +1,10 @@
+package kyo.internal.spa
```
hearnadam: This file can be shared with native

kyo-ui/shared/src/main/scala/kyo/internal/CssStyleRenderer.scala
```diff
@@ -0,0 +1,215 @@
...
+    private def renderFilterFragment(prop: Prop): String = prop match
+        case BrightnessProp(v) => s"brightness(${fmt(v)})"
+        case ContrastProp(v)   => s"contrast(${fmt(v)})"
+        case GrayscaleProp(v)  => s"grayscale(${fmt(v)})"
+        case SepiaProp(v)      => s"sepia(${fmt(v)})"
+        case InvertProp(v)     => s"invert(${fmt(v)})"
+        case SaturateProp(v)   => s"saturate(${fmt(v)})"
+        case HueRotateProp(v)  => s"hue-rotate(${fmt(v)}deg)"
+        case BlurProp(v)       => s"blur(${size(v)})"
+        case _                 => ""
```
hearnadam: What's missing?

fwbrasil:
@hearnadam I've just pushed the demos as we discussed: https://github.com/getkyo/kyo/pull/1649/changes/0a4a4ac72c3f730f3cd54ab3eac69400fcd1b7dd

kyo-ui/js/src/main/scala/kyo/UILocation.scala
fwbrasil: good point. I'll work on better AllowUnsafe usability as a follow up

He-Pin:
Is there any preview/demo, or product that shows what it can do:)?

fwbrasil:
@He-Pin take a look at https://github.com/getkyo/kyo/commit/0a4a4ac72c3f730f3cd54ab3eac69400fcd1b7dd

PR #1650, opened 2026-05-29
[compat] kyo-compat CStream: tag-free, uniformly invariant portable surface

fwbrasil:
### Problem

The streams surface promises portability: code written against `CStream` compiles unchanged against every binding. The kyo binding silently broke that. `kyo.Stream` keys its emit channel by a `Tag[Emit[Chunk[A]]]` re-summoned on every operator, and our inline delegations forwarded that requirement to callers, so generic element-abstract pipelines compiled against ce/zio/future/ox/twitter but not against the binding wrapping Kyo's own stream. Separately, the bindings disagreed on variance (`CStream[+A]` everywhere, but kyo cannot be covariant; see below), which breaks portability the other way.

### Solution

Replace the kyo opaque alias with a small invariant `final class CStream[A]` that captures the `Tag` once at construction (where `A` is concrete) and threads it into each delegated `kyo.Stream` call. Element-preserving operators (`take`/`drop`/`filter`/`run`/`fold`/`concat`/...) need no `Tag`; element-changing ones (`map`/`mapPure`/`flatMap`/`collectPure`) take the output `Tag[Emit[Chunk[B]]]`, which derives automatically for concrete `B`. Make all other bindings invariant too (`CStream[A]`) so the portable surface agrees on variance.

### Notes

- The kyo carrier is richer than the others: a lowered `kyo.Stream` is a raw computation that can carry several independent effect channels keyed by element type. The stored tag must stay matched to that carrier, which is why the binding is invariant and `lower` needs no re-tagging and no casts. The other bindings' carriers are covariant, so dropping `+` is a pure annotation change; all 38 shared `CStreamTest` cases still pass on every binding.
- A root fix in `kyo.Stream` (dropping the per-operator `Tag`) was rejected: it exposes that raw multi-channel emit computation, where the element-typed tag is load-bearing across `Emit`/`Poll`/`Pipe`/`Sink`.
- `CStreamTagFreeTest` (kyo binding) covers it: previously-uncompilable generic pipelines, channel isolation across flatMap chains, faithful `lower` after an element change, plus `assertCompiles`/`assertDoesNotCompile` for the tag boundary and invariance.

### DIFF
--- a/kyo-compat/bindings/kyo/shared/src/main/scala/kyo/compat/CStream.scala
+++ b/kyo-compat/bindings/kyo/shared/src/main/scala/kyo/compat/CStream.scala
@@ -2,102 +2,140 @@ package kyo.compat
 
 import kyo.*
 
-/** Underlying carrier is `kyo.Stream[A, Abort[Throwable] & Async]`. Operations preserve kyo `Frame`, `Tag`, and chunking semantics by
-  * delegating to the native `kyo.Stream` methods directly. `lift` and `lower` are identity since the carrier is already a kyo-native
-  * stream. Method names mirror `kyo.Stream` (`mapPure`/`filterPure`/`takeWhilePure`/`collectPure`/`foldPure`/`discard`); the pure/effectful
-  * split tracks the kyo convention.
+/** Underlying carrier is `kyo.Stream[A, Abort[Throwable] & Async]`, paired with the `Tag[Emit[Chunk[A]]]` that keys its emit channel.
+  *
+  * `kyo.Stream` re-summons `Tag[Emit[Chunk[A]]]` on every operator, so generic code over an abstract `A` cannot call it. `CStream` captures
+  * that tag once, at construction (where `A` is concrete), and threads the stored tag into each delegated `kyo.Stream` call. The result is a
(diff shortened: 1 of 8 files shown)


kyo-compat/bindings/kyo/shared/src/main/scala/kyo/compat/CStream.scala
```diff
@@ -2,102 +2,140 @@ package kyo.compat
...
+    /** Maps each element with an effectful function. */
+    def map[B](f: A => CIO[B])(using outTag: Tag[Emit[Chunk[B]]], frame: Frame): CStream[B] =
+        given tagA: Tag[Emit[Chunk[A]]] = tag
+        given tagB: Tag[Emit[Chunk[B]]] = outTag
+        new CStream(outTag, stream.map(a => f(a).lower))
+    end map
+
+    /** Flat-maps each element to another stream and concatenates the results. */
+    def flatMap[B](f: A => CStream[B])(using outTag: Tag[Emit[Chunk[B]]], frame: Frame): CStream[B] =
+        given tagA: Tag[Emit[Chunk[A]]] = tag
```
hearnadam, line 51: Could we make this tag given in the class?

fwbrasil:
@ghostdogpr closing this since I think it didn't solve the issue

PR #1658, opened 2026-06-03
[test] kyo-native test framework + codebase migration

### LINKED ISSUE #214: Test Support
Kyo still doesn't have a story for implementing tests. We could build a solution from scratch, like in ZIO, or implement an integration with an existing framework.

fwbrasil:
Fixes #214

## Problem

Kyo's tests ran on ScalaTest, which never fit the model. A Kyo test is an effectful value (`Unit < (S & Async & Abort[Any] & Scope)`), not a `Future` or an `Assertion`: ScalaTest forced an async bridge, gave synchronous and asynchronous tests different shapes, and could not express the effects a test actually needs (an `Env`, a seeded `Random`, a scoped resource) without ceremony. It also has to behave identically on JVM, JS, and Native, where ScalaTest's support is uneven. We were bending the test layer to the tool instead of writing tests in the same language as the rest of the codebase.

## Solution

A purpose-built framework, kyo-test, plus the migration of most module suites onto it.

You write a suite by extending `Test[S]` and registering cases with two operators: `"name" - { ... }` is a group, `"name" in { ... }` is a leaf whose body is a Kyo computation that the runner captures and runs. Async and sync tests share one shape; an extra effect is declared in `S` and discharged with `.handle`, so the runner only ever sees a baseline leaf. The parts worth knowing:

- A single power-`assert` that renders a subexpression diagram on failure, alongside `intercept`/`fail`/`cancel`/`assume`, `assertEventually`, and compile-time `typeCheck`. Decorators (`retry`, `timeout`, `flaky`, `pending`, platform gates, tags, `focus`) chain onto the name; `aroundLeaf` is the per-suite fixture hook (the equivalent of ScalaTest's `withFixture`).
- `AssertScope`, a per-leaf value that records failures even from detached or leaked fibers, so an assertion that fires on a fiber the runner never joins still fails the right test.
- No-assertion enforcement: a leaf that runs to completion having evaluated zero assertions now fails. ScalaTest got this for free through its `Assertion` return type; an effectful leaf body cannot, so kyo-test counts evaluations on `AssertScope` and checks after the body joins (`succeed` opts out a genuinely assertion-free leaf).
- Property-based testing (`PropertyTest`, `Gen` with automatic shrinking, edge-case bias, applicative `zip`, seeded replay) and snapshot testing (`SnapshotTest`).
- A runner with a CLI, console/TAP/JUnit-XML reporters, and a programmatic `RunConfig`; an sbt plugin wires external projects in.

## Notes

- kyo-test is built on the effect system (`kyo-test-api` depends on `kyo-core`), so a normal `dependsOn(kyo-test-runner % Test)` from a foundation module would form a project cycle, since `kyo-data`, `kyo-kernel`, and `kyo-prelude` sit below `kyo-core`. The `.withKyoTest` sbt helper avoids it by adding the runner's test classpath through `LocalProject` references (one per platform), which sbt resolves at task-evaluation time rather than during project-graph construction, so the cycle detector never sees it. That is what lets even the modules below `kyo-core` test with kyo-test.
- Every module not migrated is excluded by design, not deferred: `kyo-scheduler` (with its `-zio`/`-cats`/`-pekko`/`-finagle` integrations), `kyo-stats-registry`, and `kyo-config` are plain infrastructure below the effect system, tested as ordinary code; the `kyo-compat-*` modules exercise other effect systems (ZIO, cats-effect, ox, Twitter, Future) with those libraries' own idioms; `kyo-zio-test` runs on ZIO Test by definition; `kyo-reactive-streams` is migrated but keeps a ScalaTestPlus TestNG harness for the Reactive Streams TCK; `kyo-bench` is JMH and `kyo-examples` has no tests; and `kyo-test-api`/`runner`/`prop`/`snapshot` keep a small ScalaTest bootstrap, since a framework cannot test itself without a circular dependency.
- Cross-platform is the point: the framework and the migrated suites run on JVM, JS, and Native. `kyo-ui`, `kyo-browser`, and `kyo-pod` are compile-verified here, with runtime left to the Chrome / k8s CI gates.
- No-assertion enforcement is on by default; a suite that asserts only through a throwing domain helper opts out with `super.config.failOnNoAssertion(false)`.


kyo-test/api/shared/src/main/scala/kyo/test/internal/TestBase.scala
```diff
@@ -0,0 +1,574 @@
...
+            h: [A] => (A < (S1 & Async & Abort[Any] & Scope)) => A < (Async & Abort[Any] & Scope)
+        ): EnrichedTestBuilder[S1] =
+            new EnrichedTestBuilder[S1](TestBuilder(name), h)
+
+        def focus: TestBuilder  = TestBuilder(name).copy(focus = true)
+        def ignore: TestBuilder = TestBuilder(name).copy(ignore = true)
+        def slow: TestBuilder   = TestBuilder(name).copy(tags = Set("slow"))
+
+        /** Non-terminal decorator: marks the leaf as pending; the body is registered but not run. */
+        def pending: TestBuilder                 = TestBuilder(name).copy(pending = Maybe(""))
```
hearnadam: `def pending = pending("")`?

kyo-test/README.md
```diff
@@ -0,0 +1,597 @@
...
+
+Because the body is an ordinary Kyo value, an asynchronous test and a synchronous one share the same shape: a bare `()` or `assert(...)` auto-lifts into the effectful type, with no `Future` or `toFuture` shim. Inside a leaf you assert with a single power-`assert` that renders a subexpression diagram on failure.
+
+The type parameter `S` is an additive extra effect row. The baseline `Async & Abort[Any] & Scope` is always present and cannot be dropped; `Test[Any]` is the common case (baseline only), spelled explicitly because Scala 3 has no default type arguments. A suite that needs an extra effect (a database `Env[Db]`, a `Var`, a seeded `Random`) declares it in `S` and discharges it with `.handle` at the leaf or group level, so the runner only ever sees a baseline-shaped leaf. Decorators chain off the name before the operator (`"slow op".retry(3).timeout(5.seconds) in { ... }`). Property-based and snapshot testing are opt-in by extending `PropertyTest[S]` or `SnapshotTest[S]` instead of `Test[S]`. The whole surface compiles and runs on JVM, JS, and Scala Native.
```
hearnadam: Random is no longer an effect?

kyo-test/README.md
```diff
@@ -0,0 +1,597 @@
+# kyo-test
+
+kyo-test is the Kyo project's own test framework. You write a suite by extending `Test[S]` and registering cases with two infix operators on string names. `"name" - { ... }` declares a GROUP whose body runs immediately at registration, so its nested `-`/`in` calls fire and build the tree (loops, helper defs, and conditionals inside a group all register their children). `"name" in { ... }` declares a LEAF, and the leaf body is the test: a Kyo computation of type `Unit < (S & Async & Abort[Any] & Scope)` that is captured, not run, until the runner executes it.
+
+Because the body is an ordinary Kyo value, an asynchronous test and a synchronous one share the same shape: a bare `()` or `assert(...)` auto-lifts into the effectful type, with no `Future` or `toFuture` shim. Inside a leaf you assert with a single power-`assert` that renders a subexpression diagram on failure.
```
hearnadam, line 5: nit: this is the expectation 😁

kyo-test/README.md
```diff
@@ -0,0 +1,597 @@
...
+        case _ => Result.fail(s"malformed url: $raw")
+```
+-->
+
+```scala
+import kyo.*
+import kyo.test.*
+
+class BankTest extends Test[Any]:
+    "a fresh account has zero balance" in {
```
hearnadam, line 41: Do we want to follow the scalatest format? i guess it's useful so users can migrate trivially, but personally I don't love it.

kyo-test/README.md
```diff
@@ -0,0 +1,597 @@
...
+        val account = Account(1, "ada", balanceCents = 0L)
+        assert(account.balanceCents == 0L)
```
hearnadam, line 43: The one thing I would argue to change is regarding the assertion syntax.

If `assert` throws directly, it's not possible to bind conditionals or negations onto asserts. This also means that multiple assertion failures cannot all be collected together.

Perhaps we could have an implicit conversion to `unit` that throws? then users could write:

```
assert(a)
assert(b)
```
which would be semantically equivalent to

```
assert(a) && assert(b)
```

kyo-test/README.md
```diff
@@ -0,0 +1,597 @@
...
+        assert(Url("https", host, path).host == host)
+    }
+end UrlPropertyTest
+```
+
+> **Note:** inside `PropertyTest`, the body row is `Abort[Throwable]` (not the `Abort[Any]` of a plain `Test` leaf). The property name is auto-synthesized from the source `Frame` (`forAll @ file:line`), so two `forAll` calls on the same source line collide on name; keep one `forAll` per line.
+
+> **Note:** `forAll`'s RNG is a pure splittable seed derived from `nonRandomSeed` (default 42L) unless the suite sets `randomize = true`. It is unrelated to kyo-`Random`. A deterministic kyo-`Random` scenario must flow through `S` via `.handle(Random.withSeed(...))`. Override `numSamples`, `maxShrinks` (default 100), or `nonRandomSeed` per suite to tune the run.
+
+### `Gen`: generators with shrinking
```
hearnadam, line 386: We should make sure we don't have the same issue as ZIO: 9101

kyo-test/README.md
```diff
@@ -0,0 +1,597 @@
...
+        assert(true)
+    }
+end ConnectedSuite
+```
+
+Other per-suite hooks you can override: `name` (the suite display name), `timeout` (the default per-leaf bound), `randomize` / `randomSeed` (shuffle leaf order), and `config` (the suite's own `RunConfig`, e.g. `override def config = super.config.sequential`).
+
+> **Note:** the default per-leaf `timeout` is 60 seconds, but `Duration.Infinity` (no timeout) when a debugger is attached, so a breakpoint does not trip the bound. The effective limit therefore differs between a debug run and a normal run.
+
+## Property-based testing
```
hearnadam, line 352: I think we should make 'all tests' property tests if possible. No need for the distinction.

kyo-test/README.md
```diff
@@ -0,0 +1,597 @@
...
+```scala
+class SelectionTest extends Test[Any]:
+    "wip".focus in { assert(1 == 1) }
+    "broken".ignore in { assert(1 == 2) }
+    "ci-only".only(sys.env.contains("CI")) in { assert(true) }
+    "heavy".tagged("slow", "db") in { assert(true) }
+end SelectionTest
+```
+
+### Lifecycle: `pending` vs `pendingUntilFixed`
```
hearnadam: `pending` is just the same as `ignore`?

kyo-test/README.md
```diff
@@ -0,0 +1,597 @@
...
+        assert(a.id > 0)
+        assert(a.owner.nonEmpty)
+
+    "a fresh account is valid" in {
+        assertValidAccount(Account(1, "ada", balanceCents = 0L))
+    }
+end HelperAssertTest
+```
+
+## Decorating tests
```
hearnadam, line 213: Can we get nonFlaky and flaky like ZIO?

kyo-test/README.md
fwbrasil: I actually like it given how lightweight and readable it is. Is mutability in registration the main wart you see?

kyo-test/README.md
fwbrasil: The module already has a "smart assert" that shows details of the values for example in `assert(a & b)` the error message would show which failed. We could not throw in asserts but accumulating multiple failures can be more confusing than useful. If a test reaches an invalid state, further failures are poisoned by it and can produce odd behaviors. Throwing in conversions to `Unit` also seems weak given that the assert can be in a position that doesn't require `Unit` and we have several methods that discard results of functions by accepting `Any` as the return type. As a escape hatch, we also have `Check` in `kyo-prelude` to accumulate failures.

kyo-test/README.md
fwbrasil: I've added new tests to verify, the issue isn't present in kyo-test

kyo-test/README.md
fwbrasil: I'd prefer to keep it separate and be intentional about introducing property-based test files. I haven't had good experiences with similar solutions in the past so I'm being conservative

kyo-test/README.md
fwbrasil: fixed, only `ignore` kept

kyo-test/README.md
fwbrasil: added

fwbrasil:
I'm merging this to try to stabilize the build. I added a new execution mechanism with better concurrency control that seems to have solved some flakiness with the limited free runners

PR #1661, opened 2026-06-05
[compat] Resolve plugin-generated paths against the build root

ghostdogpr:
`compatLibrary` captures `m.base` — relative when the matrix is declared as `(projectMatrix in file("my-lib"))` — and assigns the derived `File`s directly into `baseDirectory` and the unmanaged source/resource directories of every generated row. sbt derives globs from those settings, and sbt-io prints `Warning: Tried to extract the base path for relative glob ...` for each one on every build load (~34 lines for a 4-backend JVM matrix) unless the consumer sets `-Dsbt.io.implicit.relative.glob.conversion=allow`.

This resolves each captured path against `LocalRootProject / baseDirectory` at setting-evaluation time. `IO.resolve` is a no-op for already-absolute paths, so builds passing absolute base dirs are unaffected.

It also creates the per-row `baseDirectory` eagerly: it is the forked JVM's working directory under `Test / fork := true`, and tests fail with `Cannot run program "java" (in directory ...)` when only `shared/` exists on disk — previously consumers had to pre-create the `<lib>/<backend>/<platform>/` dirs.

Found while building a new lib on kyo-compat.

### DIFF
--- a/project/TestKyo.scala
+++ b/project/TestKyo.scala
@@ -198,6 +198,8 @@ object TestKyo {
     private def fileToProjects(file: String, allProjectNames: Set[String]): Set[String] = {
         val parts = file.split("/").toList
         parts match {
+            case module :: "plugin" :: _ if allProjectNames.contains(s"$module-plugin") =>
+                Set(s"$module-plugin")
             case module :: sub :: _ =>
                 val affectedPlatforms = sub match {
                     case "shared" => Seq("JVM", "JS", "Native")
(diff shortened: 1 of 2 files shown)


ghostdogpr:
You can see the problem in [kyo CI logs](https://github.com/getkyo/kyo/actions/runs/26901005218/job/79353070881) as well.

Screenshot 2026-06-05 at 16 32 15

project/TestKyo.scala
```diff
@@ -198,6 +198,8 @@ object TestKyo {
     private def fileToProjects(file: String, allProjectNames: Set[String]): Set[String] = {
         val parts = file.split("/").toList
         parts match {
+            case module :: "plugin" :: _ if allProjectNames.contains(s"$module-plugin") =>
```
ghostdogpr, line 201: CI was silently skipping the plugin's scripted tests — `fileToProjects` didn't map `<module>/plugin/**` to `<module>-plugin`.

ghostdogpr:
The test failure is unrelated.

PR #1662, opened 2026-06-05
[http-client] Add client filter scopes

### LINKED ISSUE #1660: Add scoped and config-driven HttpClient middleware
## Summary

`kyo-http` already has a strong foundation for HTTP middleware through `HttpFilter`: route-level filters, client-side helpers such as `bearerAuth`, `basicAuth`, `addHeader`, and `logging`, and a `HttpFilter.Factory` ServiceLoader SPI for globally discovered server and client filters.

This proposal is to add first-class programmatic client middleware configuration so users can apply client-wide behavior without attaching a filter to every route and without needing to publish a ServiceLoader factory. The main supported scenarios are auth, logging, tracing, metrics, request IDs, and other cross-cutting policies that should apply to all outgoing HTTP requests in a scoped block or reusable client config.

## Motivation

A common HTTP client workflow is to build a client once with a policy pipeline:

```scala
HttpClient.withConfig(
    HttpClientConfig()
        .baseUrl("https://api.example.com")
        .filter(HttpFilter.client.bearerAuth(token))
        .filter(HttpFilter.client.addHeader("X-App", "kyo"))
) {
    HttpClient.getText("/users")
}
```

or to add a temporary policy around a block:

```scala
HttpClient.withFilter(HttpFilter.client.addHeader("X-Request-Id", requestId)) {
    HttpClient.postJson[CreateUser]("/users", body)
}
```

Today this is possible at the route level:

```scala
val route = HttpRoute
    .getText("/users")
    .filter(HttpFilter.client.bearerAuth(token))
```

and possible globally through `HttpFilter.Factory.clientFilter`, but there is no ergonomic middle ground for application code that wants scoped or config-owned client middleware.

## Existing behavior

Current `kyo-http` support includes:

- `HttpRoute.filter(...)`, which applies filters per route.
- `HttpFilter.client`, which includes client helpers for auth, custom headers, and logging.
- `HttpFilter.Factory.clientFilter`, which lets libraries contribute globally discovered client filters through ServiceLoader.
- Server-side equivalents via `HttpFilter.server` and `HttpFilter.Factory.serverFilter`.

The client backend already composes discovered client filters with route filters before sending the request. The proposed change extends that pipeline with programmatic filters.

## Proposed API

Add a filter field to `HttpClientConfig`:

```scala
case class HttpClientConfig(
    baseUrl: Maybe[HttpUrl] = Absent,
    timeout: Duration = 5.seconds,
    connectTimeout: Duration = 30.seconds,
    followRedirects: Boolean = true,
    maxRedirects: Int = 10,
    retrySchedule: Maybe[Schedule] = Absent,
    retryOn: HttpStatus => Boolean = _.isServerError,
    transportConfig: HttpTransportConfig = HttpTransportConfig.default,
    tls: HttpTlsConfig = HttpTlsConfig.default,
    clientFilter: HttpFilter.Passthrough[Nothing] = HttpFilter.noop
)
```

Add builder methods:

```scala
def filter(f: HttpFilter.Passthrough[Nothing]): HttpClientConfig =
    copy(clientFilter = clientFilter.andThen(f))

def filters(fs: Seq[HttpFilter.Passthrough[Nothing]]): HttpClientConfig =
    copy(clientFilter = fs.foldLeft(clientFilter)(_.andThen(_)))
```

Add scoped helpers on `HttpClient`:

```scala
def withFilter[A, S](filter: HttpFilter.Passthrough[Nothing])(v: A < S)(using Frame): A < S

def withFilters[A, S](filters: Seq[HttpFilter.Passthrough[Nothing]])(v: A < S)(using Frame): A < S
```

The intended composition order for normal HTTP requests is:

```scala
HttpFilter.Factory.composedClient
    .andThen(config.clientFilter)
    .andThen(scopedClientFilter)
    .andThen(route.filter)
    .andThen(pool/send)
```

This preserves the existing SPI behavior while giving application code two programmatic attachment points:

- config filters, for reusable client policy
- scoped filters, for block-local additions

Route filters remain the most specific layer.

## WebSocket handshakes

WebSocket connections currently bypass the normal `HttpClient` send path. The upgrade request is built directly in `WebSocketCodec.requestUpgradeWith`, so route-level and client-level HTTP filters do not naturally apply to the handshake.

This proposal should make client filters apply to the WebSocket HTTP upgrade handshake as well:

```scala
HttpClient.withConfig(_.filter(HttpFilter.client.bearerAuth(token))) {
    HttpClient.webSocket("wss://api.example.com/events") { ws =>
        // authenticated handshake
    }
}
```

For WebSockets, the pipeline would be:

```scala
HttpFilter.Factory.composedClient
    .andThen(config.clientFilter)
    .andThen(scopedClientFilter)
    .andThen(upgrade/send)
```

Route filters would not apply unless typed WebSocket routes are introduced later.

This should be limited to the handshake. Message-level WebSocket middleware is a different abstraction because, after the 101 response, there is no normal `HttpResponse[Out]`; there is a long-lived `HttpWebSocket`. If message interception becomes useful, it should probably be modeled separately, for example as `HttpWebSocketMiddleware`.

## Supported scenarios

This would support:

- Auth headers applied to every HTTP request and WebSocket handshake.
- Request IDs or correlation IDs applied at a scoped boundary.
- Logging and tracing that wraps outbound requests without route boilerplate.
- Metrics and timing instrumentation from libraries through the existing SPI and from application code through config.
- Reusable API clients that carry base URL, timeout, TLS, retry, and middleware policy in one config value.
- Tests that install a temporary client filter without modifying route definitions or global ServiceLoader state.

## Pros

- Reuses the existing `HttpFilter` model instead of introducing a parallel middleware type for basic request/response interception.
- Keeps ServiceLoader SPI immutable and cacheable while adding programmatic configuration.
- Avoids repeating `.filter(...)` on every route when the behavior is client-wide.
- Gives clear precedence: SPI, config, scoped block, route.
- Makes WebSocket authentication and tracing work through the same client policy used for HTTP requests.
- Preserves the current fast path when all filters are `noop`.

## Possible cons and design questions

- `HttpClientConfig` would contain a function-like value, so equality and case-class ergonomics may become less useful. If this is a concern, the filter could live in the private `HttpClient` local context instead, but then middleware would not travel with reusable config values.
- Applying filters to WebSocket handshakes requires representing the handshake as a request-like value before `WebSocketCodec` writes bytes. Care is needed so required WebSocket headers such as `Upgrade`, `Connection`, `Sec-WebSocket-Key`, and `Sec-WebSocket-Version` remain correct.
- The exact ordering should be documented. SPI first is useful for library instrumentation, while route last keeps endpoint-specific behavior closest to the send.
- Retrying and redirect behavior should be documented. Existing client filters run at the pool/send layer, so they apply per wire attempt rather than only once per logical request.
- `HttpFilter.Passthrough[Nothing]` is intentionally conservative for config and scoped filters. If users need typed field additions or custom error types at client-config scope, that likely needs a separate design.

## Testing expectations

The change should include shared `kyo-http` tests covering:

- Config filters apply to `HttpClient` convenience methods.
- Config filters apply to explicit `client.sendWith`.
- Scoped `HttpClient.withFilter` applies only inside the block.
- Config and scoped filters compose in the documented order.
- Route filters still compose after config and scoped filters.
- `HttpFilter.noop` preserves existing behavior and fast path.
- WebSocket handshakes receive headers from SPI, config, and scoped filters.
- WebSocket frame messages are not intercepted by HTTP client filters.

## Documentation

The README filter section should distinguish four layers:

1. Route filters: endpoint-specific behavior via `HttpRoute.filter`.
2. Config filters: reusable client policy via `HttpClientConfig.filter`.
3. Scoped filters: temporary block-local policy via `HttpClient.withFilter`.
4. SPI filters: library-level global instrumentation via `HttpFilter.Factory`.

It should also call out that WebSocket support applies to the HTTP upgrade handshake only, not to messages after the connection is established.

DamianReeves:
## Summary

This PR adds programmatic and scoped client filter configuration for `kyo-http`, building on the existing `HttpFilter` model and the existing `HttpFilter.Factory` SPI.

The goal is to make cross-cutting client concerns first-class without requiring every typed route or convenience call to repeat the same filter setup. Supported scenarios include auth headers, request IDs, logging, tracing, metrics, and other request/response middleware that should apply consistently across outgoing HTTP requests.

Closes #1660

## What changed

- Adds `clientFilter` to `HttpClientConfig`, defaulting to `HttpFilter.noop`.
- Adds `HttpClientConfig.filter(...)` and `HttpClientConfig.filters(...)` builder methods that append filters in order.
- Adds `HttpClient.withFilter(...)` and `HttpClient.withFilters(...)` for dynamic, scoped filter configuration.
- Composes client filters in this order:
  1. `HttpFilter.Factory` ServiceLoader filters
  2. `HttpClientConfig` filters
  3. scoped `HttpClient.withFilter` filters
  4. typed route filters
- Applies the same config and scoped client filters to WebSocket HTTP upgrade handshakes.
- Preserves WebSocket query strings in the client upgrade path and server dispatch path.
- Prevents user/filter-provided headers from duplicating required WebSocket upgrade headers such as `Host`, `Upgrade`, `Connection`, `Sec-WebSocket-Version`, and `Sec-WebSocket-Key`.
- Updates the `kyo-http` README with the new client filter configuration options and composition order.

## Why

`kyo-http` already has a solid filter abstraction and an SPI path through `HttpFilter.Factory`, but the client side had two important gaps:

- Programmatic factory-level configuration was awkward for application code that wants to construct one client config and pass it through a scope.
- Reusable client policy could be attached to typed routes, but not consistently applied to all convenience calls and WebSocket handshakes from a configured scope.

This PR keeps the existing route-level filter support, keeps SPI support for library-provided filters such as tracing, and adds config plus scoped layers for application-controlled middleware.

## Behavior notes

- Filters configured in `HttpClientConfig` apply to outgoing HTTP requests and WebSocket upgrade handshakes.
- Scoped filters apply only inside the `HttpClient.withFilter` or `HttpClient.withFilters` block.
- Route filters remain the most local layer and run after SPI, config, and scoped filters.
- WebSocket filters apply to the HTTP upgrade request only. They do not intercept WebSocket messages after the protocol upgrade.
- `HttpClientConfig` remains a case class. Its equality now includes the `clientFilter` field by reference, similar to the existing function-valued `retryOn` field.

## Pros

- Centralizes client auth, logging, tracing, and metrics setup.
- Lets applications choose SPI, config, scoped, or route-level filter installation based on ownership and lifetime.
- Reuses the existing `HttpFilter.Passthrough` abstraction instead of introducing a separate middleware type.
- Keeps the no-filter fast path in the backend.
- Makes HTTP and WebSocket handshake configuration consistent.

## Tradeoffs

- `HttpClientConfig` equality now includes the filter reference, so two configs with behaviorally equivalent but separately allocated filters will not compare equal.
- The WebSocket path has to bridge the HTTP filter abstraction around the upgrade handshake, which is narrower than full message-level WebSocket middleware.
- Filters that try to override required WebSocket upgrade headers are ignored for those specific headers to keep the handshake valid.

## Validation

- `sbt 'kyo-http/doctest'`: passed, 64 doctest blocks, 0 failures
- `sbt 'kyo-http/test'`: passed, 2,184 tests, 0 failed, 30 pending
- `git diff --check`: passed

## Related issue

- #1660

### DIFF
--- a/kyo-http/shared/src/main/scala/kyo/HttpClient.scala
+++ b/kyo-http/shared/src/main/scala/kyo/HttpClient.scala
@@ -143,7 +143,7 @@ object HttpClient:
 
     /** Clears client filters for the given computation. ServiceLoader and route filters still apply. */
     def withoutFilters[A, S](v: A < S)(using Frame): A < S =
-        withConfig(_.withoutFilters)(v)
+        withConfig(_.clearFilters)(v)
 
     // --- Factory methods ---
 
--- a/kyo-http/shared/src/main/scala/kyo/HttpClient.scala
+++ b/kyo-http/shared/src/main/scala/kyo/HttpClient.scala
@@ -16,6 +16,7 @@ import kyo.internal.client.HttpClientBackend
   *   - `HttpClient.withConfig(_.timeout(10.seconds)) { ... }` — transform the config (stacks with the current config)
   *   - `HttpClient.withConfig(config) { ... }` — replace the config entirely (discards current config)
   *   - `HttpClient.withFilter(filter) { ... }`: add a scoped client filter for outgoing requests
+  *   - `HttpClient.withoutFilters { ... }`: clear client filters for a nested scope
   *
   * The request lifecycle chains through `retryWith → redirectsWith → timeoutWith → poolWith`, each layer wrapping the next. Retries only
   * activate when a `Schedule` is set in the config. Redirects follow up to `maxRedirects` hops and switch to GET on 303 See Other per RFC
@@ -50,8 +51,8 @@ object HttpClient:
         initUnsafe(HttpPlatformTransport.transport, 100, 60.seconds)
     end defaultClient
 
-    private val local: Local[(HttpClient, HttpClientConfig, HttpFilter.Passthrough[Nothing])] =
-        Local.init((defaultClient, HttpClientConfig(), HttpFilter.noop))
+    private val local: Local[(HttpClient, HttpClientConfig)] =
+        Local.init((defaultClient, HttpClientConfig()))
 
     // Cached non-generic routes — avoids per-request allocation for convenience methods.
     private val routeGetText      = HttpRoute.getText("")
@@ -80,8 +81,8 @@ object HttpClient:
         )(
             f: HttpResponse[Out] => A < (Async & Abort[HttpException])
         )(using Frame): A < (Async & Abort[HttpException]) =
-            local.use { (_, config, filter) =>
-                self.sendWithConfig(route, request, config, filter)(f)
+            local.use { (_, config) =>
+                self.sendWithConfig(route, request, config)(f)
             }
 
         /** Closes the client with a grace period for in-flight requests. */
@@ -102,35 +103,47 @@ object HttpClient:
 
     /** Sets the client instance for all `HttpClient` calls within the given computation. The current config is preserved. */
     def let[A, S](client: HttpClient)(v: A < S)(using Frame): A < S =
-        local.use { (_, config, filter) => local.let((client, config, filter))(v) }
+        local.use { (_, config) => local.let((client, config))(v) }
 
     /** Accesses the current client. */
     def use[A, S](f: HttpClient => A < S)(using Frame): A < S =
-        local.use { (client, _, _) => f(client) }
+        local.use { (client, _) => f(client) }
+
+    /** Accesses the current client configuration. */
+    def useConfig[A, S](f: HttpClientConfig => A < S)(using Frame): A < S =
+        local.use { (_, config) => f(config) }
+
+    /** Accesses the current composed client filter. */
+    def useFilter[A, S](f: HttpFilter.Passthrough[Nothing] => A < S)(using Frame): A < S =
+        useConfig(config => f(config.clientFilter))
 
     /** Transforms the current client for the given computation. */
     def update[A, S](f: HttpClient => HttpClient)(v: A < S)(using Frame): A < S =
-        local.use { (client, config, filter) => local.let((f(client), config, filter))(v) }
+        local.use { (client, config) => local.let((f(client), config))(v) }
 
     /** Applies a config transformation for all `HttpClient` calls within the given computation. Stacks with the current config —
       * `withConfig(_.timeout(5.seconds)) { withConfig(_.retry(schedule)) { ... } }` results in a config with both timeout and retry set.
       */
     def withConfig[A, S](f: HttpClientConfig => HttpClientConfig)(v: A < S)(using Frame): A < S =
-        local.use { (client, config, filter) => local.let((client, f(config), filter))(v) }
+        local.use { (client, config) => local.let((client, f(config)))(v) }
 
     /** Replaces the config entirely for all `HttpClient` calls within the given computation. Unlike the function overload, this does not
       * stack — it discards the current config.
       */
     def withConfig[A, S](config: HttpClientConfig)(v: A < S)(using Frame): A < S =
-        local.use { (client, _, filter) => local.let((client, config, filter))(v) }
+        local.use { (client, _) => local.let((client, config))(v) }
 
     /** Applies a client filter for all `HttpClient` calls within the given computation. Stacks with the current scoped filter. */
     def withFilter[A, S](filter: HttpFilter.Passthrough[Nothing])(v: A < S)(using Frame): A < S =
-        local.use { (client, config, current) => local.let((client, config, current.andThen(filter)))(v) }
+        withConfig(_.filter(filter))(v)
 
     /** Applies multiple client filters for all `HttpClient` calls within the given computation. */
     def withFilters[A, S](filters: Seq[HttpFilter.Passthrough[Nothing]])(v: A < S)(using Frame): A < S =
-        withFilter(filters.foldLeft(HttpFilter.noop: HttpFilter.Passthrough[Nothing])(_.andThen(_)))(v)
+        withConfig(_.filters(filters))(v)
+
+    /** Clears client filters for the given computation. ServiceLoader and route filters still apply. */
+    def withoutFilters[A, S](v: A < S)(using Frame): A < S =
+        withConfig(_.withoutFilters)(v)
 
     // --- Factory methods ---
 
@@ -751,7 +764,7 @@ object HttpClient:
     )(
         f: HttpWebSocket => A < S
     )(using Frame): A < (S & Async & Abort[HttpException]) =
-        local.use { (client, clientConfig, filter) =>
+        local.use { (client, clientConfig) =>
             val resolved = clientConfig.baseUrl match
                 case Present(base) if !url.contains("://") =>
                     base.toString.stripSuffix("/") + url
@@ -762,7 +775,7 @@ object HttpClient:
                     resolveHeaders(headers),
                     config,
                     clientConfig.connectTimeout,
-                    clientConfig.clientFilter.andThen(filter)
+                    clientConfig.clientFilter
                 )(f)
             )
         }
@@ -777,8 +790,8 @@ object HttpClient:
     def webSocket[A, S](url: HttpUrl, headers: HttpHeaders, config: HttpWebSocket.Config)(
         f: HttpWebSocket => A < S
     )(using Frame): A < (S & Async & Abort[HttpException]) =
-        local.use { (client, clientConfig, filter) =>
-            client.connectWebSocket(url, headers, config, clientConfig.connectTimeout, clientConfig.clientFilter.andThen(filter))(f)
+        local.use { (client, clientConfig) =>
+            client.connectWebSocket(url, headers, config, clientConfig.connectTimeout, clientConfig.clientFilter)(f)
         }
 
     // ==================== Raw connection methods ====================
@@ -793,7 +806,7 @@ object HttpClient:
         body: Span[Byte] = Span.empty,
         headers: HttpHeaders | Seq[(String, String)] = HttpHeaders.empty
     )(using Frame): HttpRawConnection < (Async & Abort[HttpException] & Scope) =
-        local.use { (client, clientConfig, _) =>
+        local.use { (client, clientConfig) =>
             resolveUrl(url).map(parsed =>
                 client.connectRaw(parsed, method, body, resolveHeaders(headers), clientConfig.connectTimeout)
             )
@@ -836,9 +849,9 @@ object HttpClient:
     )(
         extract: HttpResponse[Out] => A
     )(using Frame): A < (Async & Abort[HttpException]) =
-        local.use { (client, config, filter) =>
+        local.use { (client, config) =>
             val req = HttpRequest(route.method, applyQuery(url, query), headers, Record.empty)
-            client.sendWithConfig(route, req, config, filter) { resp =>
+            client.sendWithConfig(route, req, config) { resp =>
                 if !resp.status.isSuccess then
                     resp.rawBody match
                         case Present(b) => Abort.fail(HttpStatusException(resp.status, route.method.name, url.baseUrl, b))
@@ -856,9 +869,9 @@ object HttpClient:
     )(
         extract: HttpResponse[Out] => A
     )(using Frame): A < (Async & Abort[HttpException]) =
-        local.use { (client, config, filter) =>
+        local.use { (client, config) =>
             val req = HttpRequest(route.method, applyQuery(url, query), headers, Record.empty).addField("body", body)
-            client.sendWithConfig(route, req, config, filter) { resp =>
+            client.sendWithConfig(route, req, config) { resp =>
                 if !resp.status.isSuccess then
                     resp.rawBody match
                         case Present(b) => Abort.fail(HttpStatusException(resp.status, route.method.name, url.baseUrl, b))
@@ -875,9 +888,9 @@ object HttpClient:
     )(
         extract: HttpResponse[Out] => A
     )(using Frame): A < (Async & Abort[HttpException]) =
-        local.use { (client, config, filter) =>
+        local.use { (client, config) =>
             val req = HttpRequest(route.method, applyQuery(url, query), headers, Record.empty)
-            client.sendWithConfig(route, req, config, filter)(extract)
+            client.sendWithConfig(route, req, config)(extract)
         }
 
     private def sendUrl[B, Out, A](
@@ -889,9 +902,9 @@ object HttpClient:
     )(
         extract: HttpResponse[Out] => A
     )(using Frame): A < (Async & Abort[HttpException]) =
-        local.use { (client, config, filter) =>
+        local.use { (client, config) =>
             val req = HttpRequest(route.method, applyQuery(url, query), headers, Record.empty).addField("body", body)
-            client.sendWithConfig(route, req, config, filter)(extract)
+            client.sendWithConfig(route, req, config)(extract)
         }
 
     // --- Private implementation ---
--- a/kyo-http/shared/src/main/scala/kyo/HttpClient.scala
+++ b/kyo-http/shared/src/main/scala/kyo/HttpClient.scala
@@ -16,7 +16,8 @@ import kyo.internal.client.HttpClientBackend
   *   - `HttpClient.withConfig(_.timeout(10.seconds)) { ... }` — transform the config (stacks with the current config)
   *   - `HttpClient.withConfig(config) { ... }` — replace the config entirely (discards current config)
   *   - `HttpClient.withFilter(filter) { ... }`: add a scoped client filter for outgoing requests
-  *   - `HttpClient.withoutFilters { ... }`: clear client filters for a nested scope
+  *   - `HttpClient.withoutFilters { ... }`: clear configured client filters for a nested scope
+  *   - `HttpClient.withoutAutoFilters { ... }`: disable ServiceLoader-discovered filters for a nested scope
   *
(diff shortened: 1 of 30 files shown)


kyo-http/shared/src/main/scala/kyo/internal/client/HttpClientBackend.scala
```diff
@@ -700,19 +701,46 @@ final private[kyo] class HttpClientBackend[Handle] private (
...
-        config: HttpWebSocket.Config
+        config: HttpWebSocket.Config,
+        clientFilter: HttpFilter.Passthrough[Nothing]
     )(
         f: HttpWebSocket => A < S
     )(using Frame): A < (S & Async & Abort[HttpException]) =
         val transportStream = new ConnectionBackedStream(connection)
+        val path = url.rawQuery match
+            case Present(q) => s"${url.path}?$q"
+            case Absent     => url.path
```
fwbrasil: let's please avoid URL manipulation outside of HttpUrl

kyo-http/shared/src/main/scala/kyo/internal/client/HttpClientBackend.scala
```diff
@@ -700,19 +701,46 @@ final private[kyo] class HttpClientBackend[Handle] private (
...
+                    serveWebSocketWith(transportStream, wsStream, config)(f)
+                }
+            else
+                val request = HttpRequest(HttpMethod.GET, url, headers, Record.empty)
+                filter[Any, "body" ~ A, HttpException](
+                    request,
+                    (filteredReq: HttpRequest[Any]) =>
+                        val filteredPath = filteredReq.url.rawQuery match
+                            case Present(q) => s"${filteredReq.url.path}?$q"
+                            case Absent     => filteredReq.url.path
```
fwbrasil: ditto

kyo-http/shared/src/main/scala/kyo/internal/client/HttpClientBackend.scala
```diff
@@ -700,19 +701,46 @@ final private[kyo] class HttpClientBackend[Handle] private (
...
+                            filteredReq.url.host,
+                            filteredPath,
+                            filteredReq.headers,
+                            config
+                        ) { wsStream =>
+                            serveWebSocketWith(transportStream, wsStream, config)(f).map { result =>
+                                HttpResponse(HttpStatus.SwitchingProtocols).addField("body", result)
+                            }
+                        }.asInstanceOf[HttpResponse["body" ~ A] < (Async & Abort[HttpException | HttpResponse.Halt])]
+                ).map(_.fields.body).asInstanceOf[A < (S & Async & Abort[HttpException])]
```
fwbrasil: why are these casts necessary? are they safe?

kyo-http/shared/src/main/scala/kyo/internal/server/UnsafeServerDispatch.scala
```diff
@@ -266,8 +266,10 @@ private[kyo] object UnsafeServerDispatch:
         parser: Http1Parser
     )(using AllowUnsafe, Frame): Unit =
         val headers = HttpHeaders.fromPacked(request.headersAsPacked)
-        val path    = request.pathAsString
-        val conn    = new ChannelBackedStream(streamCtx.inbound, streamCtx.outbound)
+        val path = request.queryRawString match
+            case Present(query) => s"${request.pathAsString}?$query"
+            case Absent         => request.pathAsString
```
fwbrasil: let's move to HttpUrl

kyo-http/shared/src/main/scala/kyo/HttpClient.scala
```diff
@@ -49,7 +50,8 @@ object HttpClient:
         initUnsafe(HttpPlatformTransport.transport, 100, 60.seconds)
     end defaultClient
 
-    private val local: Local[(HttpClient, HttpClientConfig)] = Local.init((defaultClient, HttpClientConfig()))
+    private val local: Local[(HttpClient, HttpClientConfig, HttpFilter.Passthrough[Nothing])] =
+        Local.init((defaultClient, HttpClientConfig(), HttpFilter.noop))
```
fwbrasil: could we move the filter to HttpClientConfig?

kyo-http/shared/src/main/scala/kyo/HttpClient.scala
```diff
@@ -100,27 +102,35 @@ object HttpClient:
...
-        local.use { (client, _) => local.let((client, config))(v) }
+        local.use { (client, _, filter) => local.let((client, config, filter))(v) }
+
+    /** Applies a client filter for all `HttpClient` calls within the given computation. Stacks with the current scoped filter. */
+    def withFilter[A, S](filter: HttpFilter.Passthrough[Nothing])(v: A < S)(using Frame): A < S =
+        local.use { (client, config, current) => local.let((client, config, current.andThen(filter)))(v) }
+
+    /** Applies multiple client filters for all `HttpClient` calls within the given computation. */
+    def withFilters[A, S](filters: Seq[HttpFilter.Passthrough[Nothing]])(v: A < S)(using Frame): A < S =
+        withFilter(filters.foldLeft(HttpFilter.noop: HttpFilter.Passthrough[Nothing])(_.andThen(_)))(v)
```
fwbrasil: Reasoning about the filters running in a computation becomes more difficult with the new features. Can we have methods to inspect the current enabled filters + to disable them in nested scopes?

kyo-http/shared/src/main/scala/kyo/internal/client/HttpClientBackend.scala
DamianReeves: Done. Extracted `HttpUrl.pathWithQuery` and call it here instead of concatenating the query inline. `HttpUrl.full` now reuses it too, so the path+query construction lives in one place.

kyo-http/shared/src/main/scala/kyo/internal/client/HttpClientBackend.scala
DamianReeves: Same fix here: the WebSocket path now uses `url.pathWithQuery` / `filteredReq.url.pathWithQuery` rather than rebuilding the target from `rawQuery`.

kyo-http/shared/src/main/scala/kyo/internal/server/UnsafeServerDispatch.scala
DamianReeves: Done. `serveWebSocket` now takes an `HttpUrl` built once as `HttpUrl(Absent, "", 0, request.pathAsString, request.queryRawString)` and threads it through, replacing both the hand-built path string and the later `HttpUrl.parse(path)` round-trip (which had a `parse("/").getOrThrow` fallback).

kyo-http/shared/src/main/scala/kyo/HttpClient.scala
DamianReeves: Done. The scoped filter is no longer a third slot in the `Local` tuple plus an extra param threaded through `sendWithConfig → retryWith → timeoutWith → poolWith`. It now lives in `config.clientFilter`: `withFilter`/`withFilters` delegate to `withConfig(_.filter(...))` / `withConfig(_.filters(...))`, and the `Local` is back to `(HttpClient, HttpClientConfig)`.

kyo-http/shared/src/main/scala/kyo/HttpClient.scala
DamianReeves: Added both. Inspect: `HttpClient.useConfig(config => ...)` and `HttpClient.useFilter(filter => ...)` expose the active config and composed client filter. Disable in nested scope: `HttpClient.withoutFilters { ... }` (backed by `HttpClientConfig.clearFilters`) clears the config/scoped client filters for that computation while leaving ServiceLoader and route filters in place. Tests cover the nested-disable + outer-scope-restore case.

One limitation to flag: `useFilter`/`withoutFilters` only see and clear the programmatic `config.clientFilter`; the ServiceLoader-discovered filters (`Factory.composedClient`) are not included, so a caller cannot inspect or fully disable those. Happy to widen the scope if you want those covered too.

kyo-http/shared/src/main/scala/kyo/internal/client/HttpClientBackend.scala
DamianReeves: Necessary given the current `HttpFilter.Passthrough` shape, and safe. The continuation type is fixed: `next: HttpRequest[In] => HttpResponse[Out] < (Async & Abort[E2 | Halt])`. It has no slot for the WebSocket user-handler effect row `S` (the WS callback is `f: HttpWebSocket => A < S` with arbitrary `S`), so to thread the session through a filter we erase `S` to match the continuation type and restore it after. I pulled that into two named helpers (`eraseWebSocketUserEffects` / `handleWebSocketFilterResult`) with comments so the intent is explicit.

Why it is safe: filters only forward the continuation result; they never inspect or handle `S`, and the suspended `S` effects are preserved unchanged at the runtime level across the erase/restore. This mirrors the existing non-WS pattern in `poolWith` (from #1518), which does the same erase-into-continuation + restore-result cast. The WS path is actually stricter on `Halt`: instead of casting `Halt` away like the non-WS path, it runs `Abort.run[Halt]` and converts a filter `Halt` into an `HttpStatusException`.

One residual assumption worth naming: this is sound as long as `S` is disjoint from what filters handle (`Async`, `Abort[Halt]`). The only realistic collision would be a user `S` containing `Abort[HttpResponse.Halt]`, which is an http-internal type and not something a WS handler row would carry.

If you would prefer the casts gone entirely, the clean route is to scope the filter to the upgrade handshake only and serve the session outside the filter (where `S` flows naturally). That needs splitting `WebSocketCodec.requestUpgradeWith` so the handshake and the `f(wsStream)` session are no longer in one resource bracket. Happy to do that as a follow-up if you want it.

kyo-http/shared/src/main/scala/kyo/HttpClient.scala
DamianReeves: Followed up on the ServiceLoader-filter limitation I flagged above with a read-only blast-radius exploration. Sketch of how we could cover the auto-discovered filters too, if you want it:

**Proposal (additive, non-breaking):** add `disableAutoFilters: Boolean = false` to `HttpClientConfig`, plus a dedicated `HttpClient.useAutoFilter` accessor (exposes the composed `Factory.composedClient`) and a `HttpClient.withoutAutoFilters { ... }` scope. Backend composition becomes `(if config.disableAutoFilters then noop else Factory.composedClient).andThen(config.clientFilter)` at the two client sites (`poolWith`, `runWsSessionWith`). This keeps `Factory.composedClient` as a memoized `lazy val` and preserves the `eq noop` fast path exactly.

**Why not just fold the auto filters into `config.clientFilter`:** it would force ServiceLoader discovery at every `HttpClientConfig()` construction (including the `Local.init` default and every doctest), defeat the lazy-val memoization, and break the no-filter hot path.

**Footgun to avoid:** I would NOT make the existing `withoutFilters` clear auto filters. Today the only auto client filter is OTLP W3C trace-context propagation; having `withoutFilters` silently drop it would break distributed traces with no compile error. Hence a separate, explicit `withoutAutoFilters` rather than overloading `withoutFilters`.

**Open question for you:** scope. Do you want server-side parity in the same change? The server path composes `composedServer.andThen(route.filter)` in `HttpHandler`, which would have the same inspect/disable gap. Happy to do client-only now and server as a follow-up, or both together. Size is roughly M client-only (field + conditional + accessor + scope + tests + README/scaladoc), M/L if server parity is included.

Can leave the current limitation as-is for this PR and land the auto-filter support separately, or fold it in here. Your call.

kyo-http/shared/src/main/scala/kyo/internal/client/HttpClientBackend.scala
fwbrasil: I don't think it's safe. Can you introduce tests that use effects like `Var` in the provided function? It seems we need an `S` type param in `HttpFilter.apply` or restrict the provided function to not have free `S` effects

kyo-http/shared/src/main/scala/kyo/HttpClient.scala
fwbrasil: I like the proposal to handle auto filters separately. We should also document the terminology and behavior. About parity, it's always ideal.

ahoy-jon (approved):

kyo-http/shared/src/main/scala/kyo/internal/client/HttpClientBackend.scala
DamianReeves: Implemented the effect-row fix instead of keeping the casts. `HttpFilter.apply` now carries an `S` type parameter through the downstream continuation, so WebSocket callbacks can preserve their free effects across client filters. The WebSocket helper casts were removed, and `HttpWebSocketTest` now covers a filtered client WebSocket callback that uses `Var` inside the session.

kyo-http/shared/src/main/scala/kyo/HttpClient.scala
DamianReeves: Implemented this with client and server parity. The API now distinguishes configured filters, auto filters, and route filters in docs/scaladoc. Client config and server config both have `autoFilters` plus `withoutAutoFilters`; `HttpClient` exposes `useAutoFilter`; and server auto-filter composition moved to server initialization so `HttpServerConfig.withoutAutoFilters` actually disables the SPI filters. Added tests for default auto filters, disabling them by config, and composition order.

PR #1664, opened 2026-06-08
[ui] Connection-bound reactive sessions: the WebSocket owns the session

fwbrasil:
**Problem**

This PR solves flaky tests in CI. A server-driven kyo-ui session (the live per-tab `Signal` graph) has no owner and no lifetime. It is a cookie-keyed entry in an in-memory `SessionStore`, fed by Server-Sent-Events down and HTTP POST up, with per-region subscription fibers tracked by hand. Nothing tears it down: the store is never evicted, so sessions accumulate for the process lifetime; and a nested region's re-render interrupts only its immediate child fibers, orphaning grandchildren, so the leak grows while the app is in use. The root cause is a missing ownership model: no scope defines when a session is alive or gone, so cleanup is manual and incomplete.

**Solution**

Add `Signal.observe`, a scoped switch combinator, and build the session on it. `f` runs for the current value and each change inside a fresh per-value `Scope`, held open until the next value, then closed (releasing whatever `f` forked, transitively) before the next runs. It is register-before-read, so a write racing the read never loses the latest value, and it never tears a still-current value's scope down on an idle timer.

The session then *is* the connection, owned by a `Scope`:

```scala
Scope.run {
  subscribe(root, wsExchange(ws))
  Async.race(ws.stream.foreach(dispatch), ws.onPeerClose)
}
```

Socket close completes the block, the `Scope` closes, and every reactive region (a `Fiber.init` in the scope tree, each driven by one `observe`) is interrupted by cascade. Teardown is structural and cannot be forgotten, and a value change releases the prior subtree with no manual fiber bookkeeping. WebSocket is the enabling choice: its handler runs for the connection's lifetime with a `Scope` that closes on disconnect, where SSE+POST had no per-connection lifecycle to attach cleanup to.

Transport is a single WebSocket. The page `GET` is pure SSR (initial HTML + a small WS client; no session created, nothing stored). `SessionStore`, the `kyo-sid` cookie, and the `/_kyo/event` + `/_kyo/sse` routes are gone; the inline client uses one `WebSocket` both directions (`HtmlOp` out, `UIEvent` in, shapes unchanged). `runHandlers` / `runMount` / `runRender` signatures are unchanged.

**Notes**

- **Trade-off to weigh: per-connection state resets on disconnect.** A dropped socket ends the session; reconnect starts a fresh one. Better than today (reconnect was silently broken), but state that must survive a drop has to live above the per-connection evaluation. A resume/grace layer can be added later without changing the ownership model.
- **WebSocket is required for interactivity** (no-WS clients get the SSR page only). It also pins a session to one server, removing the prior need to sticky-route SSE and POST together.


kyo-core/shared/src/main/scala/kyo/Signal.scala
```diff
@@ -426,6 +493,29 @@ object Signal:
         end new
     end _initRaw
 
+    // Like _initRaw but also supplies `observe`, letting structural combinators delegate observation to their sources'
+    // (register-before-read) loops instead of inheriting the default read/register repair loop. This keeps `map`-over-leaf
+    // chains exact (and idle-no-teardown) on the leaf.
+    @nowarn("msg=anonymous")
+    private inline def _initRawF[A](
```
hearnadam, line 500: Shouldn't `initRaw` be created with this?

fwbrasil: The new observe mechanism is derived from the other functions. We could expose it but it'd be a low-level feature that is very easy to get wrong. I think we can wait for concrete use cases to see if exposing makes sense

PR #1667, opened 2026-06-11
[scheduler] fix interrupt starvation via WorkerQueue rebuild and runtime reset

fwbrasil:
## Problem

A CPU-bound fiber that was preempted (sitting in a worker's run queue) and then interrupted while queued could be starved before it had a chance to observe the interrupt and run its finalizers. The root cause was a combination of two gaps.

First, when an `IOTask` completed (either normally or via interrupt), it did not reset its accumulated runtime. A fiber interrupted while mid-flight had a high runtime value and therefore low scheduling priority; when re-queued after the preempt, it sat behind newly-submitted, zero-runtime tasks. This held even for the interrupt case, where prompt scheduling is required to release the worker and run finalizers.

Second, the worker run queue was a generic `Queue[Task]` backed by a `scala.collection.mutable.PriorityQueue` using an explicit `Ordering[Task]`. Resetting a task's priority key in place (via `resetRuntime`) does not trigger a re-sift in a binary heap; the task stays at its old array position and its new key is invisible to subsequent polls. The starvation delay therefore scaled with queue depth and was unbounded in the general case.

## Solution

On completion, a task now calls `Task.resetRuntime()` to drop its accumulated runtime to zero, giving it top scheduling priority. Because a binary heap does not re-sift when a key changes in place, resetting a queued task's key would be invisible to subsequent polls without an explicit rebuild. Each worker therefore calls a worker-local `rebalance()` at the top of its run loop whenever the global interrupt epoch has advanced, triggering `WorkerQueue.rebuild()` to re-heapify the run queue (O(n) Floyd construction) so the reset task surfaces at the head within a bounded, load-independent delay.

The generic `Queue[Task]` backed by `scala.collection.mutable.PriorityQueue` is replaced by `WorkerQueue`, a `Task`-specialized array-backed min-heap. Inlining the priority comparison directly to `Task.runtime()` removes the `Ordering` indirection and all element casts from the hot poll path. `rebuild()` exists precisely because `WorkerQueue` owns the backing array and can re-heapify it in place, something the opaque `PriorityQueue` wrapper could not expose.

## Notes

`WorkerQueue` is specialized on `Task` rather than generic to allow the inline `runtime()` comparison and remove all `Ordering` indirection from the hot path. The 4-ary branching factor (d=4) gives shallower trees and better cache locality (four contiguous children per probe) than a binary heap, at the cost of a 4-way min scan per sift-down level. This tradeoff favors poll-heavy workloads like the scheduler's run loop.

`rebalance()` is single-thread-owned by the worker's own run loop; `lastRebuiltEpoch` and `lastRebuildMs` are plain `var`s because they are read and written only inside `rebalance()`, which is never called concurrently. The epoch gate ensures a rebuild fires at most once per `notifyInterrupt` call per worker, and the time gate (`minInterval`) bounds rebuild frequency under an interrupt storm.

`Task.taskOrdering` is removed. It was the only user of the generic `Ordering[Task]` surface and is no longer needed now that `WorkerQueue` inlines the comparison.


kyo-scheduler/jvm-native/src/main/scala/kyo/scheduler/WorkerQueue.scala
```diff
@@ -0,0 +1,254 @@
...
+      * Uses busy waiting rather than parking/blocking since:
+      *   - Effective concurrency is bounded by available cores
+      *   - Lock hold times are very short (optimized queue operations)
+      *   - Most contended operations (stealing) fail fast if lock unavailable
+      *
+      * The scheduler's worker model ensures the number of threads competing for any given queue is limited, making spinning more efficient
+      * than thread parking overhead.
+      */
+    private def lock(): Unit =
+        while (!compareAndSet(false, true)) {}
```
hearnadam, line 251:
```suggestion
        while (!tryLock()) {}
```

hearnadam (approved):

PR #1672, opened 2026-06-14
[parse] fix Parse.repeat quadratic complexity via indexed ParseInput

fwbrasil:
## Problem

`ParseInput.remaining` returned `tokens.drop(position)`, which materializes a copy of the entire remaining input on every call. Every per-token consumer (`readOne`, `anyMatch`, `literal`, `boolean`, `identifier`) called `remaining` to read the head or test a prefix, so running `Parse.repeat` over an n-token input made O(n) such calls, each copying O(n) data. The overall cost was O(n^2) in both time and allocation.

The `startsWith` check in `literal` and `boolean` shared the same problem: `in.remaining.startsWith(text)` routed through `Seq.head` after copying the tail, not through direct indexing.

## Solution

`ParseInput` now caches a `Chunk.Indexed[In]` (`indexed`) derived once from `tokens` at construction time. All per-token reads go through that field instead of re-deriving the tail.

Two targeted additions replace the problematic call sites:

- `headMaybe: Maybe[In]` reads `indexed(position)` in O(1) and returns `Absent` at end of input. `readOne`, `anyMatch`, and `end` are rewritten to use it.

- `startsWith(prefix: Seq[In])` and `startsWith(prefix: String)` walk `indexed` from `position` without any tail copy, both in O(prefix.length). `literal`, `boolean`, and `identifier` are rewritten to call `in.startsWith(...)` directly.

`advanceWhile` already indexed into `tokens` directly; it is updated to use `indexed` for consistency.

The `remaining` accessor is kept for callers that genuinely need a `Chunk[In]` slice (for example `identifier`, which calls `in.remaining.takeWhile` after the head check). Its cost is now paid only where a contiguous slice is actually required.

The performance test suite (`"performance"` suite in `ParseTest`) adds three regression guards at n=500000: `repeat(any)`, `readWhile`, and `repeat(literal/any)`. Each asserts correct output length and a 30s wall-clock budget. At that input size the pre-fix code takes minutes; the fixed code finishes in well under a second.

## Notes

`Chunk.Indexed` is constructed once at `ParseInput` construction, not lazily, so callers that discard `ParseInput` immediately (for example single-token parses) pay the indexing cost even if they never call `headMaybe`. That trade-off is acceptable because `ParseInput` is already a value object allocated once per `Parse.run` invocation; the `toIndexed` call is O(n) in the token count, which is dominated by the O(n) scan that `repeat` would perform anyway.

The `remaining` method signature is unchanged. Callers on `origin/main` that read `in.remaining.head` or `in.remaining.startsWith(...)` would still compile; the hot-path callers inside `Parse.scala` are the ones rewritten here.

### DIFF
--- a/kyo-parse/shared/src/main/scala/kyo/ParseInput.scala
+++ b/kyo-parse/shared/src/main/scala/kyo/ParseInput.scala
@@ -2,7 +2,47 @@ package kyo
 
 case class ParseInput[In](tokens: Chunk[In], position: Int):
 
-    def remaining: Chunk[In] = tokens.drop(position)
+    // The backing tokens as an indexed chunk, so per-token reads index in O(1) instead of copying the tail.
+    private val indexed: Chunk.Indexed[In] = tokens.toIndexed
+
+    def remaining: Chunk[In] = indexed.drop(position)
+
+    /** The next token, or `Absent` at end of input. O(1).
+      *
+      * Reads the token at `position` by direct indexing rather than `remaining.head`. The latter routes through the generic `Seq.head`, which
+      * materializes a fresh copy of the entire remaining input on every call; driving a `repeat` over an n-token input that way is O(n^2).
+      */
+    def headMaybe: Maybe[In] =
+        if position < indexed.length then Maybe(indexed(position))
+        else Maybe.empty
+
+    /** True if the tokens from `position` start with `prefix`, compared by `==`. O(prefix.length), no tail copy. */
+    def startsWith(prefix: Seq[In])(using CanEqual[In, In]): Boolean =
+        val n = prefix.length
+        if position + n > indexed.length then false
+        else
+            var i  = 0
+            var ok = true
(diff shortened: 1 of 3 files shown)


kyo-parse/shared/src/main/scala/kyo/ParseInput.scala
```diff
@@ -2,7 +2,47 @@ package kyo
 
 case class ParseInput[In](tokens: Chunk[In], position: Int):
 
-    def remaining: Chunk[In] = tokens.drop(position)
+    // The backing tokens as an indexed chunk, so per-token reads index in O(1) instead of copying the tail.
+    private val indexed: Chunk.Indexed[In] = tokens.toIndexed
```
hearnadam, line 6: Wouldn't it be better to require `ParseInput` to be a `Chunk.Indexed`? then we can just make an apply that does this?

PR #1675, opened 2026-06-15
[build] enable -Werror on Scala 3 and fix all resulting warnings

fwbrasil:
## Problem

The build had no `-Werror` gate on Scala 3, so compiler warnings accumulated silently across multiple modules. The Scala 2.13 cross-builds used by `kyo-scheduler` carry a noisier, different warning set, so a blanket flag would have broken them.

Three deprecated FFI types (`FfiUnsupported`, `FfiAbiMismatch`, `FfiKoffiVersionMismatch`) remained as subclasses of the canonical `FfiLoadError` hierarchy and were still referenced throughout the runtime, codegen emitters, and tests, triggering deprecation warnings on every Scala 3 compile.

`FfiLoadError.AbiMismatch` lacked a message-carrying constructor; `StructAbiCheck.verifyByteSize` had to instantiate the legacy `FfiAbiMismatch` (with four named fields) to preserve the structured diagnostic message, keeping the deprecated type alive.

`FfiLoadError.AbiMismatch` and `FfiLoadError.Unsupported` were `class` (open), not `final`, solely to allow the deprecated shims to inherit from them.

`kyo.IO` and `kyo.Resource` top-level deprecated aliases for `Sync` and `Scope` were unused internally and generated deprecation warnings.

`SelfCheck` extended `scala.App`, which is deprecated in Scala 3.

`CStream` passed `ExecutionContext` values positionally to `Future.map`/`flatMap` without `using`, triggering Scala 3's "implicit parameters should be provided with a `using` clause" migration warning.

Several pattern matches across `kyo-tasty`, `kyo-ui`, `kyo-browser`, and `kyo-test` were non-exhaustive or triggered unchecked-type-test warnings due to erasure on `Result.Failure[TastyError]` patterns and newly added ADT cases (`Style.Color.Var`, `Svg.PathCommand.Raw`, `Tasty.Symbol.EnumCase`, `TastyError` variants).

`TestBase`'s `TestBuilder` extension defined zero-argument `focus`, `ignore`, and `pendingUntilFixed` decorators that the same-named `TestBuilder` case-class fields shadowed, so they could never be selected (an E194 "extension method will never be selected" warning) and the chained decorator form was silently dead.

A leftover empty test file (`PathDispatchTest.scala`) contained only a comment and no test class.

## Solution

`kyo-settings` in `build.sbt` now appends `-Werror` only when `scalaVersion` starts with `"3"`, leaving the Scala 2.13 `kyo-scheduler` family unaffected.

The three deprecated FFI shim files (`FfiUnsupported.scala`, `FfiAbiMismatch.scala`, `FfiKoffiVersionMismatch.scala`) are deleted. Every throw site across the JVM, JS, and Native runtimes, the JS and JVM codegen emitters, and all tests now reference `FfiLoadError.Unsupported` or `FfiLoadError.AbiMismatch` directly. `FfiLoadError.AbiMismatch` gains a three-argument constructor `(expected, actual, msg)` so `StructAbiCheck.verifyByteSize` can pass the full struct-diagnostic message (binding FQN, struct name, sizes, remediation hint) without the removed shim's named fields. Both `AbiMismatch` and `Unsupported` are now `final`. Tests that previously asserted on `traitFqn`/`structName`/`expectedSize`/`actualSize` fields now assert on `expected`/`actual` strings and `getMessage.contains(...)`.

The `kyo.IO`/`kyo.Resource` top-level aliases are removed from `Sync.scala` and `Scope.scala`.

`SelfCheck` replaces `extends App` with an explicit `def main(args: Array[String]): Unit`.

All `CStream` `Future.map`/`flatMap` calls are changed to pass `parasiticEc` with an explicit `using` keyword.

Pattern matches in `kyo-tasty`'s `Subtyping` and `ClasspathOrchestrator` that match `Result.Failure[TastyError]` add `@unchecked` to suppress the erasure warning; the match is structurally correct because `Result` is a sealed hierarchy and only `Failure` and `Panic` remain after the `Success` arm. `IsSubtypeOfTest`'s `TastyError` exhaustive match gains the four missing ADT arms (`InvalidUuid`, `UnresolvedReference`, `UnknownType`, `MissingDeclaredType`). `SymbolExhaustiveMatchTest` and `SymbolMemberSearchTest` add the `EnumCase` arm to `ClassLike` matches. `AnnotationLikeBaseTest` and `InvariantsSpec` add a defensive catch-all arm for the phantom `AnnotationLike` subtype that `compiletime.testing.typeCheckErrors` injects into exhaustivity analysis. `Tasty.collisionReport` is rewritten using `Chunk.collect` to eliminate the non-exhaustive `flatMap` match. `DecoderFidelity5Wave2Test` replaces `Maybe` pattern matches with `.fold`. `SymbolResolutionTest` uses the overload of `classpath.parents` that accepts any `Tasty.Symbol` directly. The `kyo-ui` `colorComponents` and `renderPathDataStr` functions add arms for `Style.Color.Var` and `Svg.PathCommand.Raw` respectively, both introduced by recent UI additions.

`TestBase`'s `TestBuilder` extension drops the dead zero-argument `focus`, `ignore`, and `pendingUntilFixed` decorators (the field-shadowed, never-selected ones); the required-argument `ignore(reason)` and `pendingUntilFixed(reason)` decorators are kept. `PathDispatchTest.scala` is deleted.

## Notes

The `@unchecked` annotations on `Result.Failure[TastyError]` patterns in `Subtyping` and `ClasspathOrchestrator` suppress a true erasure gap: the JVM cannot distinguish `Result.Failure[TastyError]` from `Result.Failure[SomethingElse]` at runtime. The matches are safe because `Result` is sealed and the only failure subtype in these call paths is `Failure`; `@unchecked` is scoped to the individual pattern, not the whole match.

The `FfiLoadError.AbiMismatch` three-argument constructor accepts a pre-formatted message string so callers that need the full struct-diagnostic (binding FQN, struct name, size pair, remediation hint) can supply it via `FfiErrors.structAbiMismatch`, while the two-argument convenience constructor `(expected, actual)` still covers the common case.

`kyo-scheduler`'s Scala 2.13 builds are intentionally excluded from `-Werror`: their warning profile is different in character and volume and is tracked separately from the Scala 3 codebase.


kyo-core/shared/src/main/scala/kyo/Scope.scala
```diff
@@ -36,12 +36,6 @@ import kyo.kernel.ContextEffect
   */
 sealed trait Scope extends ContextEffect[Scope.Finalizer]
 
-@deprecated("Will be removed in 1.0. Use `Scope` instead.", "1.0-RC")
```
fwbrasil, line 39: I don't think these deprecated aliases for `IO` and `Resource` have much value

hearnadam (approved):

PR #1681, opened 2026-06-18
[schema] A type-safe foundation for wire protocols

fwbrasil:
## Problem

Protocols like JSON-RPC could not be built type-safe on the old derivation. A JSON-RPC message is a typed envelope (id, method, error) around params / result / data whose type depends on the method; kyo-jsonrpc, and kyo-mcp / kyo-lsp / the CDP client above it, all live on that shape. Two things made it unsafe: the open part had to be handled by reaching into raw JSON, the exact runtime-failing code a typed schema exists to remove; and a type's wire codec and its structure were derived by two separate macros that could silently disagree, so the JSON Schema a type advertised could not be trusted to match the bytes it wrote.

## Solution

Protocol code now stays in types end to end.

Structure.Value is the first-class open value (Record / Array / Str / Integer / Null / ...), and every Schema converts to and from it via fromStructureValue / toStructureValue. The envelope is one derived Schema[JsonRpcEnvelope] whose open slots are Structure.Value, and each method decodes its own params with its own derived Schema. Codec.IntrospectingReader makes "can read an open value" a type-level capability: JSON and YAML extend it, a binary codec without per-value type tags (Protobuf) does not, so decoding a Structure.Value through one is a compile error, not a runtime surprise. Structure.Type.Open is the structure those open-shaped schemas report.

Structure and codec are co-derived from one macro. FocusMacro walks the case fields or sealed children once and emits both, and Schema.init captures the structure by-name, so the two are produced at one site and cannot drift; the JSON Schema a type advertises (for example MCP tools/list inputSchema) is the exact shape it reads and writes. The standalone StructureMacro is deleted.

The codec body is static per field: the macro emits summonInline[Schema[ft]] per field and a kyo.writeField / kyo.readField call against the inline bridges in SchemaBridge.scala (primitives stay unboxed there, everything else goes through the field schema). There is no runtime Array[Schema[Any]] walk, and recursion closes by forward-reference to the in-flight derived$Schema, so the macro carries no per-type cases. Constant-sized per-method bytecode keeps a class with many derives Schema under the JVM class-file limit. SerializationMacro.scala and StructureMacro.scala (about 1450 lines together) are gone.

## Notes

Structure.Field holds its field type as a () => Structure.Type thunk so recursive structures are not forced eagerly; read it only through def fieldType, never the case-class productElement / unapply / copy that expose the raw thunk. The writeField / readField bridges are the single place JVM primitives are special-cased; moving that back into the emitted bodies reintroduces per-field branching and blows the class-file limit on wide types.

]

---
<a href="https://app.blacksmith.sh/getkyo/codesmith/kyo/pr/1681"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784338231&installation_id=140673115&pr_number=1681&repository=getkyo%2Fkyo&return_to=https%3A%2F%2Fgithub.com%2Fgetkyo%2Fkyo%2Fpull%2F1681&signature=7d80d875c3ca534cc19ad1a381566b8997aedf05d8c65a88f90e71ebeac37106"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>


fwbrasil:
This requires changes to other modules. I'll prepare another PR

PR #1687, opened 2026-06-20
[jsonrpc][browser] Add kyo-jsonrpc + kyo-jsonrpc-http; port browser CDP

fwbrasil:
### Problem

Kyo had no JSON-RPC 2.0 support, so every protocol built on it (MCP, LSP, and the Chrome DevTools Protocol the browser module speaks) had to hand-roll request/response correlation, notification dispatch, cancellation, and progress over its own ad-hoc wire handling. kyo-browser in particular carried a bespoke CdpClient that reimplemented JSON-RPC framing, id correlation, and error decoding by hand.

### Solution

- Add kyo-jsonrpc, a cross-platform (JVM, JS, Native) JSON-RPC 2.0 peer. One JsonRpcHandler owns the bidirectional dispatch loop and exposes typed call, notify, callWithProgress, callPartialResults, and cancel; routes answer inbound requests and notifications via JsonRpcRoute.request and .notification, with user-domain errors mapped to wire codes through .error[E]. The stack is layered: JsonRpcHandler over JsonRpcTransport (one envelope each way) over JsonRpcWireTransport plus JsonRpcFramer plus a Schema[JsonRpcEnvelope] (bytes to and from envelopes), so a custom transport plugs in at the byte level. Built-in transports: in-memory, line and content-length stdio, and a JVM Unix domain socket. Cancellation, progress, id allocation, and unknown-method handling are pluggable policies; failures use a sealed JsonRpcError hierarchy organized by pipeline stage. The envelope and id/error wire shapes derive from kyo-schema.
- Add kyo-jsonrpc-http, a JsonRpcTransport backed by a kyo-http WebSocket that lifts text frames to and from envelopes.
- Port kyo-browser onto kyo-jsonrpc: replace the hand-rolled CdpClient with a CdpBackend that runs the CDP session as a kyo-jsonrpc peer (id correlation, cancellation, and error decoding handled by the engine), deleting the bespoke wire code.

### Notes

- All three platforms share one source tree. The Unix-domain-socket transport is JVM-only (the NIO UDS APIs are absent on JS and Native) and aborts with UnsupportedOperationException there; unixDomain takes a cross-platform kyo.Path, not a java.nio.file.Path.
- Unsafe boundaries follow the kyo tier discipline: public factories are safe, internal hot-path state propagates AllowUnsafe rather than summoning it, and each Unsafe site is discharged at a Sync.Unsafe.defer boundary.


DamianReeves:
Curious if a general kyo-rpc makes sense as a foundational layer. With that folks could do simple things like Message Pack over rpc. Thoughts?

DamianReeves:
Also excited to see this land. We can build [ACP](https://agentcommunicationprotocol.dev/introduction/welcome) on-top of this.

PR #1689, opened 2026-06-20
Actor pub/sub: strand-safe ask, Hub bridge, PubSub, and Subject.contramap (closes #1688)

### LINKED ISSUE #1688: Actor ask/pub-sub publish acknowledgements should not strand callers
## Summary

Kymora tried to use `kyo-actor` as the serialization point for workflow telemetry publishing, but backed away because the request/ack shape can strand callers if a scoped actor closes, is interrupted, or panics before replying.

This is especially relevant to the actor pub/sub work being discussed: workflow telemetry is a concrete pub/sub use case where publish acknowledgement semantics matter. We need a publisher to know whether an event was accepted and projected, rejected/closed, or failed. It should never wait forever.

## Use case

Kymora workflow telemetry has two sinks that must stay ordered:

1. `Hub[WorkflowEvent]` for live subscribers.
2. `SignalRef[WorkflowRunState]` for snapshot/state projection.

Required semantics:

- **linearized publish:** subscribers and projected state observe events in the same order, even with concurrent publishers.
- **snapshot-after-publish:** if `publish(event)` returns successfully while telemetry is open, `snapshot.events` includes that event.
- **closed telemetry no-op/failure:** publishing after scope shutdown should complete deterministically, ideally as `Closed` or a documented no-op policy.
- **no stranded callers:** a caller waiting for an actor reply must complete with success, failure, closed, interruption, or panic.
- **panic propagation:** if the handler panics before replying, the caller should observe a panic/failure rather than waiting forever.

## Minimal reproduction shape

This is the reduced pattern that made us avoid the actor implementation:

```scala
import kyo.*

final case class Publish(value: Int, replyTo: Actor.Subject[Unit])

class Publisher(actor: Actor[Publish]):
  def publish(value: Int)(using Frame): Unit < Async =
    Abort.recover[Closed](_ => ())(actor.ask(Publish(value, _)))
```

Two cases are important for acceptance coverage.

### 1. Scoped shutdown with queued ask messages

If the actor closes after an `ask` has created a reply subject and the publish message is accepted/enqueued, but before the handler sends to `replyTo`, the waiting publisher should complete. It should not remain pending indefinitely.

Sketch:

```scala
"ask completes when scoped actor closes with queued messages" in {
  val result =
    for
      fiber <- Scope.run {
                 for
                   actor <- Actor.run(capacity = 1) {
                              Actor.receiveLoop[Publish] { msg =>
                                msg.replyTo.send(()).andThen(Loop.continue)
                              }
                            }
                   fiber <- Fiber.initUnscoped(actor.ask(Publish(1, _)))
                 yield fiber
               }
      result <- Abort.run[Closed](Async.timeout(2.seconds)(fiber.get))
    yield result

  assert(result.isSuccess || result.isFailure)
}
```

The exact test should use Kyo's preferred timeout/closed/interruption assertions, but the assertion is that there is no indefinitely pending fiber.

### 2. Handler panic before reply

If the actor handler panics before sending the reply, the `ask` caller should complete with the panic/failure instead of hanging.

Sketch:

```scala
"ask completes when handler panics before reply" in {
  val result =
    Scope.run {
      for
        actor <- Actor.run(capacity = 16) {
                   Actor.receiveLoop[Publish] { _ =>
                     Abort.panic(new RuntimeException("boom"))
                   }
                 }
        result <- Abort.run[Throwable](Async.timeout(2.seconds)(actor.ask(Publish(1, _))))
      yield result
    }

  assert(result.isPanic || result.isFailure)
}
```

## Why simple fire-and-forget pub/sub was insufficient

We also considered splitting the flow into:

1. `hub.put(event)`
2. enqueue a separate projector message

That avoided `ask`, but concurrent publishers could put to the hub in one order and enqueue projector updates in another. Then live subscribers and `snapshot.events` could disagree about event order.

For this use case, a useful actor/pub-sub API needs one linearized publish operation that couples event delivery/projection with caller completion.

## Workaround used in Kymora

Kymora currently ships this with a mutex around `Hub.put` and `SignalRef.updateAndGet`:

```scala
def publish(event: WorkflowEvent)(using Frame): Unit < Async =
  Abort.recover[Closed](_ => ())(serial.run {
    for
      published <- Abort.run[Closed](hub.put(event))
      _ <- published match
             case Result.Success(_) => stateRef.updateAndGet(_.applyEvent(event)).unit
             case Result.Failure(_) => Kyo.unit
             case Result.Panic(ex)  => Abort.panic(ex)
    yield ()
  })
```

This preserves ordering and snapshot-after-publish and has no per-message reply promise to strand. It is less actor-shaped, which is why we are filing this as feedback for actor/pub-sub design.

## Desired outcome

Any of these would unblock Kymora moving telemetry back to actors:

- `Actor.ask` completes with `Closed` or interruption if the actor closes after accepting/enqueuing a message but before processing it.
- actor shutdown drains or rejects queued ask messages and completes their reply subjects.
- a request/reply actor helper guarantees caller completion before actor termination.
- actor pub/sub documents publish acknowledgement semantics that satisfy ordered delivery/projection and no stranded publishers.

DamianReeves:
## Summary

Gives `kyo-actor` a complete publish/subscribe story and makes request/reply reliable. Three layers, all composing through the existing `Subject` abstraction:

1. **Strand-safe request/reply.** `ask` no longer hangs a caller when the actor terminates or panics before replying, and a `respond`/`respondLoop` combinator owns the reply so it cannot be forgotten. This is the direct fix for #1688.
2. **`Hub` bridge.** `Subject.init(hub)` publishes through the uniform `Subject` API, and `Actor.subscribe` funnels hub events into an actor's own mailbox.
3. **`PubSub[A]` and `Subject.contramap`.** A push-based pub/sub abstraction (the `ActorRef`-as-topic / EventStream analog) with a plain and a linearized variant, plus a contravariant sink adapter so any actor can subscribe.

25 commits, cross-platform (JVM/JS/Native), doctest-validated docs, a worked ATM/banking example as a test, a README tutorial, and a runnable `kyo-examples` demo.

Closes #1688.

## Motivation

`Subject[A]` is kyo's analog of Akka typed's `ActorRef[T]`: a uniform send/ask handle whose delivery topology is a property of the receiver, not the handle. `Hub` is kyo's pull-based broadcast primitive (buffered listeners drained at the consumer's rate). What was missing: an ergonomic, thread-safe path to publish to and subscribe from a `Hub`; a push-based `PubSub` (the EventStream analog) where subscribers are handles and publish delivers straight into them; and a request/reply path that never strands a caller.

The #1688 use case (workflow telemetry) needs an actor as a linearization point writing two ordered sinks, but backed away because `ask` could strand callers on shutdown or panic. A single actor mailbox is exactly that linearization point; the blocker was the `ask` defect, which this PR fixes.

## 1. Strand-safe request/reply

The defect: the reply promise was never coupled to the actor's liveness, so if the actor closed (scope shutdown) or panicked after a message was enqueued but before the reply was sent, `promise.get` hung forever.

Fix: a `private[kyo]` `awaitReply` hook on `Subject` that the actor mailbox subject overrides to complete when the actor terminates, backed by a `PendingReplies` registry. Reply promises are saved directly and failed by a single per-actor termination sweep installed once on the consumer fiber (no per-ask `onComplete`); each ask's waiter is removed on completion via `Sync.ensure`, so the set is bounded by concurrently in-flight asks rather than the actor's lifetime.

```scala
def ask[C](f: Subject[C] => A)(using Frame): C < (Async & Abort[Closed | E])
```

`Subject.ask` stays `Abort[Closed]` (an actor failure collapses to `Closed`); `Actor.ask` refines it to surface the actor's actual `E` or a panic. A caller now always completes: with the reply, `Closed`, the actor's `E`, or a panic. A typed `ask(request)` instance method (with an `A =:= Ask[Req, Resp]` evidence constraint) gives clear error messages when an actor's message type is not a request/reply envelope.

`respond` makes the framework own the reply, so a forgotten reply cannot strand a caller, and `respondLoop` is its stateful analogue (threads `State` across requests, returns the final state when the mailbox closes):

```scala
publisher <- Actor.run {
    Actor.respond[Event, Unit] { event =>
        hub.put(event).andThen(stateRef.updateAndGet(_.apply(event)).unit)
    }
}
publisher.ask(event)   // strand-safe; returns when the handler ran both sinks
```

## 2. Hub bridge

`Hub.put`/`offer` already match `Subject.send`/`trySend`, so publishing is one overload:

```scala
val log: Subject[LogEvent] = Subject.init(hub)   // EventStream.publish analog
```

`Actor.subscribe(hub)(adapt)` funnels hub events into the subscribing actor's own mailbox via an enqueue-only pump fiber (producer-only, so the actor keeps a single consumer and its serialized processing is unchanged). Overloads take an optional `bufferSize` and `filter`, mirroring `hub.listen`:

```scala
Actor.run {
    Actor.subscribe(hub)(identity).andThen(Actor.receiveAll[LogEvent](e => Console.printLine(e.render)))
}
```

## 3. PubSub and Subject.contramap

`PubSub[A]` is push-based: subscribers are `Subject`s, `publish` delivers straight into them (the subscriber's mailbox is the buffer), and a subscriber whose `send` fails `Closed` is pruned. Two constructors, each with an optional fan-out concurrency bound:

```scala
def init[A]: PubSub[A] < Sync                       // plain: per-subscriber FIFO, no cross-subscriber order
def linearized[A: Tag]: PubSub[A] < (Scope & Async) // actor-owned: total order across subscribers + ordered membership
```

`Subject.contramap` is the contravariant sink adapter (a sink is contravariant in its message type, so `contramap` is lawful and `map` is not), letting any actor subscribe by adapting its sum-typed subject:

```scala
pubsub.subscribe(actor.subject.contramap(Msg.Event(_)))
```

`PubSub.linearized` reuses the strand-safe `Actor.ask` internally, so it is strand-safe too. `PubSub` is push; `Hub` is pull. Use `PubSub` for actor/event fan-out, `Hub` for rate-isolated streaming consumers. (The type was named `Topic` in earlier commits; it was renamed to `PubSub` to avoid colliding with `kyo.Topic` in `kyo-aeron`, an unrelated Aeron transport type.)

## #1688 requirement coverage

| Requirement | How it is met |
|---|---|
| linearized publish (concurrent publishers, subscribers and state same order) | a single actor mailbox (or `PubSub.linearized`) is the serialization point |
| snapshot-after-publish | `ask`/`respond`/`publish` returns only after the handler ran its sinks |
| closed telemetry no-op/failure | `send`/`ask`/`publish` to a closed target returns `Abort[Closed]`, recoverable to a no-op |
| no stranded callers | the reply races the actor's termination signal |
| panic propagation | the actor fiber's panic is surfaced to the caller |

## Worked example: ATMs and a bank

A tangible domain demonstrates the whole feature, in three forms: an `ActorTest` simulation (the executable proof), a five-stage progressive README tutorial, and a runnable `kyo-examples` demo. Multiple ATMs hit accounts point-to-point via `ask`; each account publishes committed transactions to one `PubSub.linearized`; an audit observer (`subject`) and a fraud observer (`contramap`) subscribe. With concurrent publishers, both observers observe the identical ordered stream:

```scala
assert(auditObserved == fraudObserved.map(_.tx))   // the property PubSub.init would not guarantee
```

Money is modeled with an `Amount` type (opaque over `BigDecimal`), and the withdraw reply uses `Result[String, Amount]` (`Result.fail("Insufficient funds")`), following kyo conventions.

## Thread-safety

Subscription is producer-only: the pump (Hub bridge) and `publish` (PubSub) only `send` into a subscriber's mailbox; they never add a second consumer. The actor's single loop stays the sole reader/writer of its state. A regression test interleaves direct sends and subscribed events into a non-atomic accumulator and asserts the exact total, which would fail if a second consumer existed.

## Behavioral notes and breaking changes

- **`Actor.ask` widening.** `Actor.ask` widens from `B < (Async & Abort[Closed])` to `C < (Async & Abort[Closed | E])`. For `E = Nothing` actors (every actor in the repo, since per-request errors are modeled in reply types) this is `Closed | Nothing = Closed`, an identical type and zero source break. `E != Nothing` actors gain a residual `Abort[E]` to handle, which was previously the silently-unhandleable strand bug.
- **Close drains to end-of-stream.** Closing an actor now drives its receive loop to end-of-stream so the behavior completes with its final value `B`, and `await` returns it; previously a close made `await` fail `Closed`. A panic still propagates, and the strand-safe ask sweep still fails in-flight ask callers with `Closed`. This is what makes the stateful loops' return value (for example `respondLoop`'s final `State`) observable.
- **`close` keeps its name** (uniform with `Channel`/`Hub`/`Queue`/`PubSub.close`, and consistent with the `Closed` error); the actor-domain verbs `stop`/`Terminated` are reserved for a future death-watch feature.

## Tests and platforms

All green: JVM (`SubjectTest` 21, `ActorTest` 48, `PubSubTest` 19), JS (83 total), Native (21 / 44 / 19), doctest 35/35, and `kyo-examplesJVM` compiles and runs. New tests are folded into the existing `ActorTest`/`SubjectTest` plus a 1:1 `PubSubTest` (no orphan test files), and include the two #1688 reproductions, the strand-safety and no-accumulation regression guards, the single-consumer serialization probe, the linearized cross-subscriber ordering proof, the bounded-concurrency fan-out tests, and the ATM/banking fan-out integration. A module `CONTRIBUTING.md` documents the design and the ask/strand-safety entry points.


kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -95,9 +135,78 @@ end Actor
...
+            onTerminated: => C < (Async & Abort[Closed | E])
+        )(using frame: Frame): C < (Async & Abort[Closed | E]) =
+            Promise.init[Unit, Any].map { signal =>
+                val register: Unit < Sync =
+                    Sync.Unsafe.defer(discard(waiters.add(signal))).andThen(
+                        // Close the add-after-termination race: if the sweep already set the flag, complete
+                        // our own signal so the terminated branch fires instead of awaiting a reply that
+                        // never comes. Either the sweep sees this freshly added signal, or we self-complete.
+                        Sync.Unsafe.defer[Boolean, Any](terminated.get()).map(t => if t then signal.completeUnitDiscard else Kyo.unit)
+                    )
```
fwbrasil: This can be a single suspension. Since it's a hot path, it's worth optimizing:

```scala
Sync.Unsafe.defer {
    discard(waiters.add(signal))
    if terminated.get() then 
        signal.completeUnitDiscard 
    else 
        Kyo.unit
}
```

It's an idiom that I think we shouldn't recommend to users unless they're working on performance-sensitive code. It still properly defers side effects but, once the code reaches a Kyo computation, the user must be careful not to introduce side effects in continuations. For example, this would fail to defer a side effect:

```scala
Sync.Unsafe.defer {
    discard(waiters.add(signal))
    if terminated.get() then 
        // WRONG: unsuspended side effect
        signal.completeUnitDiscard.andThen(waiters.remove(signal))
    else 
        Kyo.unit
}
```

It compiles because the `AllowUnsafe` provided by `Sync.Unsafe.defer` is in scope.

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -95,9 +135,78 @@ end Actor
...
+                val raced: C < (Async & Abort[Closed | E]) =
+                    Sync.ensure(Sync.defer(discard(waiters.remove(signal)))) {
+                        // raceFirst (not race): complete on the first branch to finish, success or failure.
+                        // A plain race only completes on the first success, so a stranded caller whose actor
+                        // terminated without replying (a failure on the terminated branch) would hang forever.
+                        Async.raceFirst[Closed | E, C, Any](
+                            reply.get,
+                            signal.get.andThen(onTerminated)
+                        )
```
fwbrasil: It'd be nice if we can avoid this race in the hot path due to the overhead. Could we save the `reply` promises directly and close them when the actor closes instead? I might be missing something but `onTerminated` and `terminated` could be unnecessary

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -168,6 +277,42 @@ object Actor:
...
+      *   Maps a hub event to the actor's message type
+      * @tparam A
+      *   The actor's message type
+      * @tparam E
+      *   The hub's event type
+      */
+    def subscribe[A](using Frame, Tag[Subject[A]])[E](hub: Hub[E])(adapt: E => A): Unit < Context[A] =
+        for
+            self     <- Actor.self[A]
+            listener <- hub.listen
```
fwbrasil: Wdyt about also taking `bufferSize` and `filter`? Take a look at the `hub.listen` overloads

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -367,6 +512,44 @@ object Actor:
...
+      * the lifecycle-aware [[Actor.ask]]. Model expected per-request errors in `Resp` (for example a `Result`) rather than aborting.
+      *
+      * @param handler
+      *   Produces the reply for each request
+      * @tparam Req
+      *   The request type
+      * @tparam Resp
+      *   The single reply type for the whole actor
+      */
+    def respond[Req, Resp](using
```
fwbrasil, line 618: Can we have `respondLoop` methods like in the other factories? It'd give more freedom to use effects and maintain local state

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -367,6 +512,44 @@ object Actor:
...
+        Frame
+    )[S](handler: Req => Resp < S)(
+        using Tag[Poll[Ask[Req, Resp]]]
+    ): Unit < (Context[Ask[Req, Resp]] & S) =
+        receiveLoop[Ask[Req, Resp]](msg => handler(msg.request).map(resp => msg.replyTo.send(resp).andThen(Loop.continue)))
+
+    extension [E, Req, Resp, B](actor: Actor[E, Ask[Req, Resp], B])
```
fwbrasil: Can you check if the overload via extension methods produces good error messages? It's typically better to have an instance method with implicit evidence constraints (`=:=`) to produce good error messages

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -439,10 +622,28 @@ object Actor:
             mailbox <-
                 // Create a bounded channel to serve as the actor's mailbox
                 Channel.init[A](capacity, Access.MultiProducerSingleConsumer)
+            pending =
+                // Per-actor registry of in-flight ask waiters; entries are removed as each ask resolves so
+                // the set is bounded by concurrent in-flight asks, never by the actor's lifetime.
+                new PendingReplies
```
fwbrasil: Allocating a type that holds mutable state is itself a side effect, we need a suspension here

fwbrasil:
Nice improvements! There's some good info in the PR description, could you also create a `CONTRIBUTING.md` for the module to capture it? The `dev` skill I shared can be invoked to generate it

kyo-actor/shared/src/main/scala/kyo/Actor.scala
DamianReeves: Done in 9daa0e50f. `register` is now a single `Sync.Unsafe.defer` that adds the reply to the set and checks the terminated flag in one deferred block, no second suspension on the hot path. Kept it as the internal idiom only; the public-facing `respond`/`ask` paths stay on the safe tier.

kyo-actor/shared/src/main/scala/kyo/Actor.scala
DamianReeves: Reworked along these lines in 9daa0e50f. Reply promises are now saved directly in a per-actor `PendingReplies` registry and failed by a single termination sweep installed once on the consumer fiber (`_consumer.onComplete`), so the per-ask `onComplete` and its leak are gone. The set is bounded by concurrently in-flight asks, not by actor lifetime.

`terminated` stays for one narrow race that direct registration alone does not close: `Subject.ask` enqueues the request *before* the reply is registered, so the actor can accept the request, terminate, and run its sweep all before registration happens. The sweep would never see that reply and the caller would strand. The flag (set before the sweep iterates) lets a late registration self-fail immediately: either the sweep sees the freshly added reply, or the registration sees the flag and fails its own reply. `completeDiscard` is first-wins, so a recipient that already replied is unaffected.

`onTerminated` is doing a different job: it is the hook that lets `Actor.ask` refine the `Closed` into `E`/panic via the actor fiber, while the bare `Subject.ask` collapses it to `Closed`. Same sweep, two interpretations.

kyo-actor/shared/src/main/scala/kyo/Actor.scala
DamianReeves: Added in 1ff4c8698, mirroring the `hub.listen` overloads: `subscribe(bufferSize)`, `subscribe(filter)`, and the canonical `subscribe(bufferSize, filter)`; the no-arg form delegates to the canonical. Tests assert exact ordered output (`List(2,4,6)` / `List(3,4,5)` for filter, `List(10,20,30)` for bufferSize).

kyo-actor/shared/src/main/scala/kyo/Actor.scala
DamianReeves: Added `respondLoop` in a85121d62. Signature `respondLoop[Req, Resp, State](state: State)(handler: (Req, State) => (Resp, State) < S)`: the handler returns the reply plus the next state in an effectful context `S`, so callers keep local state across requests and use effects in the body. It runs as a `Poll` loop over `Ask[Req, Resp]`, same as the other factory loop methods.

kyo-actor/shared/src/main/scala/kyo/Actor.scala
DamianReeves: Good call, the extension-method overload produced poor errors (the mismatch surfaced as a failed implicit search on an unrelated type). Switched to an instance method in 8c2454e7f: `def ask[Req, Resp](request: Req)(using Frame, A =:= Actor.Ask[Req, Resp]): Resp < (Async & Abort[Closed | E])`. With the `=:=` evidence, calling `ask(req)` on an actor whose message type is not `Ask[Req, Resp]` now reports the actual message type vs the expected `Ask[...]` directly.

kyo-actor/shared/src/main/scala/kyo/Actor.scala
DamianReeves: Fixed in 9daa0e50f. The `PendingReplies` allocation is now suspended: `pending <- Sync.defer(new PendingReplies)`, since it holds mutable state (the `CopyOnWriteArraySet` and the `AtomicBoolean.Unsafe`).

DamianReeves:
@fwbrasil thanks for the review. Pushed `883a7e8e2..b181ea022` addressing all six comments; replies are inline on each thread.

**Strand-safe `ask` (comments 1, 2, 6)** ` 9daa0e50f`
- Reply promises are now saved directly in a per-actor `PendingReplies` registry and failed by a single termination sweep installed once on the consumer fiber (`_consumer.onComplete`). The per-ask `onComplete` and its leak are gone; the set is bounded by concurrently in-flight asks, not actor lifetime.
- `register` collapsed to a single `Sync.Unsafe.defer` (add + flag check in one deferred block); kept as an internal idiom only.
- `PendingReplies` allocation is suspended via `Sync.defer(new PendingReplies)` since it holds mutable state.
- `terminated` stays for one narrow race: `Subject.ask` enqueues the request before the reply is registered, so the actor can accept, terminate, and sweep before registration. The flag lets a late registration self-fail (`completeDiscard` is first-wins). `onTerminated` is the hook that refines `Closed` to `E`/panic for `Actor.ask` while the bare subject collapses to `Closed`.

**`subscribe` overloads (comment 3)** `1ff4c8698`
- Added `subscribe(bufferSize)`, `subscribe(filter)`, and canonical `subscribe(bufferSize, filter)`, mirroring `hub.listen`. Tests assert exact ordered output.

**`respondLoop` (comment 4)** `a85121d62`
- Added `respondLoop[Req, Resp, State](state)(handler: (Req, State) => (Resp, State) < S)`: local state across requests, effects in the handler, runs as a `Poll` loop over `Ask[Req, Resp]`.

**`ask(req)` error messages (comment 5)** `8c2454e7f`
- Switched from the extension-method overload to an instance method with `A =:= Actor.Ask[Req, Resp]` evidence; mismatches now report the actual message type vs the expected `Ask[...]`.

Also added a module `CONTRIBUTING.md` (`b181ea022`) documenting the design and the ask/strand-safety entry points.

Green on all platforms at `b181ea022`: JVM 21/48/13, JS 77, Native 44, doctest 35/35, examples compile.

kyo-actor/shared/src/main/scala/kyo/Actor.scala
```diff
@@ -367,6 +591,66 @@ object Actor:
...
+    def respondLoop[Req, Resp, State](using
+        Frame
+    )[S](state: State)(
+        handler: (Req, State) => (Resp, State) < S
+    )(using Tag[Poll[Ask[Req, Resp]]]): Unit < (Context[Ask[Req, Resp]] & S) =
+        receiveLoop[Ask[Req, Resp]](state) { (msg, st) =>
+            handler(msg.request, st).map { case (resp, next) =>
+                msg.replyTo.send(resp).andThen(Loop.continue(next))
+            }
+        }.unit
```
fwbrasil: Loops with state typically return the final state, not `Unit`. See the other overloads

kyo-actor/shared/src/main/scala/kyo/Topic.scala
```diff
@@ -0,0 +1,168 @@
...
+/** A push-based publish/subscribe handle.
+  *
+  * Subscribers are [[Actor.Subject]] sinks; `publish` delivers each value into every subscriber, awaiting delivery, and prunes any
+  * subscriber whose send fails with `Closed`. Unlike [[Hub]] (a pull primitive whose listeners buffer and drain at their own rate), a Topic
+  * pushes directly into subscriber sinks, so a subscriber's own mailbox is the buffer.
+  *
+  * @tparam A
+  *   The type of values published through this Topic
+  */
+sealed abstract class Topic[A]:
```
fwbrasil: I think I missed this in the initial review. We already have `kyo.Topic` in `kyo-aeron`. We should use a different name or unify if possible

fwbrasil: Actually, it doesn't seem possible to unify since aeron's `Topic` is actually an effect. Maybe this could be with `Hub`

kyo-actor/shared/src/main/scala/kyo/PubSub.scala
```diff
@@ -0,0 +1,168 @@
...
+      */
+    def close(using Frame): Unit < Sync
+
+end Topic
+
+object Topic:
+
+    // Sends `value` to each subscriber concurrently and returns the set of subscribers that failed with Closed (to be pruned).
+    private def fanOut[A](subscribers: Set[Subject[A]], value: A)(using Frame): Set[Subject[A]] < Async =
+        Async.foreach(subscribers.toSeq) { subscriber =>
```
fwbrasil: it'd be nice if the user could pass the concurrency in an `init` overload

kyo-actor/shared/src/main/scala/kyo/Actor.scala
DamianReeves: Done in abd4ec70b: `respondLoop` now returns the final `State` (dropped the `.unit`), matching the stateful `receiveLoop` overloads.

Flagging one thing, because making this meaningful required a change to the `Actor.run` consumer that reaches beyond `respondLoop`, so I want your read on it.

`respondLoop`'s handler always `Loop.continue(next)`, so the loop only ends on `Poll` -> `Absent` (end-of-stream). But the old consumer was `mailbox.take.map(v => cont(Maybe(v)))`: on close, `take` aborted `Closed` and the actor failed `Closed`; it only ever fed `Present`, never `Absent`. So `Absent` was never delivered and every `case Absent => Loop.done(state)` branch in the receive loops was effectively unreachable, which means a returned `State` could never be observed.

So the consumer now translates a graceful close into end-of-stream:

```scala
Abort.run[Closed](mailbox.take).map {
    case Result.Success(v) => Loop.continue(cont(Maybe(v)))
    case Result.Failure(_) => Loop.continue(cont(Absent))   // graceful close -> end-of-stream
    case Result.Panic(e)   => Abort.panic(e)
}
```

This mirrors the existing subscribe pump (`Failure => Loop.done // mailbox closed: stop cleanly`). The semantic change is actor-wide, not just `respondLoop`: closing an actor now drains its receive loop to end-of-stream and `await` returns the behavior's final value `B`, whereas before `await` failed `Closed` on close. A `Panic` still propagates, and the strand-safe ask sweep still fires on completion so in-flight ask callers are still failed with `Closed`. All actor tests stay green on JVM/JS/Native.

If you intended `await`-after-close to keep failing `Closed`, this is the design point to confirm. My read is that end-of-stream-on-close is the correct semantics (it is what makes the stateful loops return-value useful, and it matches the Poll/stream model), but it is broader than your comment so I did not want to slip it in silently.

kyo-actor/shared/src/main/scala/kyo/Topic.scala
DamianReeves: Renamed it to `PubSub` (`84d187bc3`): `PubSub.init` / `PubSub.linearized`, with the source/test renamed to `PubSub.scala` / `PubSubTest.scala` and the README, CONTRIBUTING, and bank example updated. The unrelated kyo-aeron `Topic` is left as-is.

On folding into `Hub`: I looked into it and it does not work cleanly, for two reasons.

1. Module layering blocks it. `PubSub` subscribers are `Actor.Subject` sinks, and `Subject` is defined in `kyo-actor`, which `dependsOn(kyo-core)`. `Hub` lives in `kyo-core`. A `Hub.subscribe(subject)` would make `kyo-core` depend on a `kyo-actor` type, inverting the layer order. Avoiding that means either moving `Subject` down into `kyo-core` or adding an abstract sink type in core, both large refactors.

2. Even setting layering aside, the two are duals rather than one thing. `Hub` buffers per-listener channels it owns and backpressures by stalling its fan-out fiber; `close` closes every listener buffer. `PubSub` pushes directly into each subscriber's own mailbox (the mailbox is the buffer), backpressures the publisher on a full mailbox, and `close` leaves subscriber sinks open. Subscribers are `Listener`s in one and `Subject`s in the other. Merging them would put two contradictory contracts under one name.

So I kept them separate: `Hub` is the pull-based broadcast in core, `PubSub` is the push-based one in kyo-actor next to `Subject`/`Actor`. Happy to revisit if you had a specific unification in mind.

kyo-actor/shared/src/main/scala/kyo/PubSub.scala
DamianReeves: Added in 2ef6dd9ba. `init(concurrency: Int)` and `linearized(concurrency: Int)` overloads let the caller bound how many subscriber sends a single `publish` runs in parallel; the no-arg forms delegate with `Async.defaultConcurrency`, so existing behavior is unchanged. The bound is threaded into `fanOut` via `Async.foreach(subscribers.toSeq, concurrency)`. For `linearized` it only bounds parallelism within one publish; the cross-publish total order (the single mailbox) is unaffected.

`concurrency` is validated eagerly with `require(concurrency >= 1, ...)`, matching `Async.race`. One thing worth noting from this: the previous `fanOut` was already calling `Async.foreach` with the default bound (`availableProcessors * 2`), not unbounded; this change just exposes that bound to callers. Tests cover `init(1)`/`init(2)` and `linearized(1)`/`linearized(2)` delivering to all subscribers, plus rejection of `0`. Green on JVM/JS/Native + doctest.

fwbrasil (approved):

PR #1700, opened 2026-06-22
[build] require JDK 25 for foreign modules, enable compact headers

fwbrasil:
## Problem

`java.lang.foreign` (the Foreign Function and Memory API) is final in JDK 22, so the modules that use it cannot compile against the project-wide `-release 17` flag set by `kyo-settings`. The previous floor guard required JDK 21, which is not sufficient to compile those modules at their required release target.

The build also made no use of compact object headers (JEP 519), a product flag available in JDK 25 that reduces the per-object header from 12-16 bytes to 8 bytes, leaving heap pressure on the table for the test forks that allocate most heavily (kyo-tasty decodes roughly 80k symbols per run).

## Solution

Add a `foreignRelease` setting in `build.sbt` that strips the inherited `-release 17` and replaces it with `-release 25`. Apply it to the seven modules that use `java.lang.foreign`: `kyo-data`, `kyo-offheap`, `kyo-ffi`, `kyo-ffi-it`, `kyo-ffi-codegen`, `kyo-ffi-bench`, and `kyo-tasty`. All other modules keep `-release 17`.

Because `-release 25` requires a JDK >= 25 present at compile time, the whole build now requires JDK 25. The `Global / onLoad` version guard is raised from 21 to 25 with an updated error message. The CI toolchain is moved to `corretto:25` across the setup action, all workflow files, and `scripts/build.sh`.

`-XX:+UseCompactObjectHeaders` is added to the test fork `javaOptions` in `kyo-settings`, to `.jvmopts` for local builds, to the `JAVA_OPTS`/`JVM_OPTS` lines in every workflow that carried `-XX:+UseG1GC`, and to `CI_DRIVER_OPTS` in `scripts/build.sh`.

## Notes

`foreignRelease` is applied via `.settings(...)` without platform scoping, so `-release 25` reaches the JS, Native, and Wasm variants of the affected modules as well. All seven modules compile on every platform they target under `-release 25`, so the setting does not need to be scoped to JVM only.

The `doctestScalacOptions` for `kyo-offheap` (previously `-release 22`) and `kyo-ffi` are updated to `-release 25` to match.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -98,5 +98,5 @@ runs:
 
     - uses: coursier/setup-action@v3.0.0
       with:
-        jvm: corretto:24
+        jvm: corretto:25
         apps: sbt
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -60,7 +60,7 @@ jobs:
       - uses: coursier/cache-action@v8.1.1
       - uses: coursier/setup-action@v3.0.0
         with:
-          jvm: corretto:21
+          jvm: corretto:25
           apps: sbt
 
       - name: install async-profiler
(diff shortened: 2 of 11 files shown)


hearnadam (approved):

He-Pin:
Java 25 is right

PR #1704, opened 2026-06-23
[schema] add Adjacent, Tuple, TupleFlat, and Untagged sum representat…

DamianReeves:
…ions

## Problem

kyo-schema serialized sum types (sealed traits / enums) under exactly two wire shapes: the default external wrapper `{"Circle":{"radius":10.0}}` and the flat discriminator `{"type":"Circle","radius":10.0}` set by `.discriminator`. That second form had a silent correctness hole: when a variant's payload was not a JSON object (a bare scalar, an array, null), the engine's `case _` arm in `flattenWithDiscriminator` returned only the discriminator field and discarded the payload entirely, with no error.

Beyond the data-loss bug, there was no way to interoperate with external producers that emit adjacently-tagged, positional-array, or untagged JSON. Any schema that needed one of those shapes had to pre- or post-process the wire bytes outside the schema layer.

## Solution

Adds a `Schema.SumRepresentation` enum with six cases (External, Internal, Adjacent, Tuple, TupleFlat, Untagged) and a `representation` constructor slot on `Schema[A]`, threaded through all five construction and copy sites and OR-ed into `hasTransforms`/`hasReadTransforms` so non-default representations take the transform-aware path automatically. The External default is inert; existing wire output is byte-identical.

Four new Focused-preserving builder methods select the alternate representations:

- `.adjacent(tagKey, contentKey)` -- two-field object `{"type":"Circle","content":{...}}`. The payload occupies the content position whole, so a scalar, array, or null payload survives (fixing the silent-drop the flat discriminator had). Missing `tagKey` on decode raises `MissingTagKeyException`.
- `.tupleTagged` -- two-element positional array `["Circle",{"radius":10.0}]`.
- `.tupleFlat` -- flattened positional array `["Triangle",10.0,10.0,10.0]`; payload fields spread in declaration order, field names dropped, a record field carried as one nested element (not deep-flattened). Decode re-assembles field names from the schema before handing off to the macro's `readBody`.
- `.untagged` -- bare payload `{"radius":10.0}` with no tag. Decode captures the payload once into a `Structure.Value` tree and tries each variant's decoder in declaration order over a fresh `StructureValueReader` per attempt (non-destructive); first clean parse wins. No match raises `NoVariantMatchException` with the attempted variant wire names.

`.discriminator` is reimplemented as sugar over `Internal(fieldName)` while continuing to set the legacy `discriminatorField` slot, preserving byte-identical output and requiring zero call-site changes.

All representation logic lives in `SchemaSerializer`, not the derivation macro. The macro emits one additional artifact at derive time: a `variantDecoders: Chunk[Codec.Reader => Any]` list of per-variant closures (one per variant, in declaration order) that the untagged decode path iterates.

Three new typed `SchemaException` leaves:

- `RepresentationUnsupportedException(codec, representation)` -- thrown (encode path convention) when Tuple, TupleFlat, or Untagged is used with a codec that cannot express a top-level array or bare scalar. Raised before any bytes are written.
- `NoVariantMatchException(path, variants)` -- `DecodeException` in the Result when untagged decode exhausts all variants.
- `MissingTagKeyException(path, tagKey)` -- `DecodeException with NavigationException` when adjacent decode input lacks the configured tag key.

A write-side capability signal, `Codec.Writer.canWriteTopLevelNonObject`, gates the Tuple/TupleFlat/Untagged encode path. Positive opt-in: Json, Yaml, Ion, and MsgPack writers declare `true`; Protobuf does not (it is field-number-driven and cannot express a top-level array or bare scalar without emitting an invalid stream). `Codec.Writer.codecName` is also added for use in error messages.

`IonReader` and `YamlReader` now extend `Codec.IntrospectingReader` with `readStructure()` implementations, bringing them on par with `JsonReader` and `MsgPackReader`. This is required for untagged decode on those codecs: the mechanism captures the payload once via `captureValue()`, materializes it to an immutable `Structure.Value` tree via `readStructure()`, and replays a fresh `StructureValueReader` per variant attempt. Protobuf remains a plain `Reader` (non-self-describing; untagged decode raises the existing self-describing-reader typed failure).

`StructureValueReader.short` and `.byte` now accept `Decimal` and `BigNum` `Structure.Value` variants in addition to `Integer`, fixing round-trips for `Short` and `Byte` fields through Ion (which materializes integer literals as `BigNum` in the value tree).

All representations compose with the variant naming and alias layer from PR #1694: tag resolution goes through `resolveVariantWire`/`variantReverse` in every tagged representation; `untagged` skips tag resolution entirely (there is no tag).

## Notes

`variantDecoders` is a `@publicInBinary private[kyo]` constructor slot emitted by the derivation macro. Schemas derived before this change will not carry the slot; the default `Chunk.empty` means untagged decode on a pre-change derived schema will exhaust all attempts and raise `NoVariantMatchException`. Re-deriving (recompiling the `derives Schema` site) populates the slot.

The field-level transform path in `SchemaSerializer.writeWithTransforms` now guards field-rename/drop/computed logic to `Structure.Type.Product` schemas only. Previously the same code ran unconditionally; for a sum schema the materialized `Record` is a single-field variant wrapper whose key is a variant name, not a data field, and running it through `applyFieldConvention` was incorrect. This fix is a prerequisite for correct Adjacent and Tuple encode, not a behavior change for schemas that did not combine field transforms with sum representations.

Annotation-driven representation selection (`@adjacent`, `@untagged`) and per-variant mixed representations are not in scope; the builders added here are the intended desugaring target for a companion annotations feature.


kyo-schema/shared/src/main/scala/kyo/internal/SchemaSerializer.scala
```diff
@@ -113,6 +133,84 @@ private[kyo] object SchemaSerializer:
                 other
     end flattenWithDiscriminator
 
+    /** Raises `RepresentationUnsupportedException` before any bytes are written when the active
+      * writer cannot express a top-level array / bare scalar (the shape Tuple/TupleFlat/Untagged
+      * require). Capability is a positive opt-in on the writer (Codec.Writer.canWriteTopLevelNonObject).
+      */
+    private def requireTopLevelCapable(writer: Writer, representation: String): Unit =
+        given Frame = Frame.internal
```
fwbrasil, line 141: Can we propagate the implicit isntead?

kyo-schema/shared/src/main/scala/kyo/internal/SchemaSerializer.scala
```diff
@@ -32,51 +32,71 @@ private[kyo] object SchemaSerializer:
...
+                            originalFields.find(_._1 == sourceName).map((_, v) => (targetName, v))
+                        })
+
+                        // Add computed fields
+                        val computedFieldValues = schema.computedFields.map { (name, compute) =>
+                            (name, anyToStructureValue(compute(value)))
+                        }
+
+                        val allFields = transformedFields ++ renamedFieldValues ++ computedFieldValues
+                        given Frame   = Frame.internal
```
fwbrasil, line 73: ditto

fwbrasil (approved):

kyo-schema/shared/src/main/scala/kyo/internal/SchemaSerializer.scala
DamianReeves: Have a fast follower PR that will add type union support and a few other things, I will fix it there.

kyo-schema/shared/src/main/scala/kyo/internal/SchemaSerializer.scala
DamianReeves: Have a fast follower PR that will add type union support and a few other things, I will fix it there.

PR #1706, opened 2026-06-24
[build] bump dependencies and plugins to latest; migrate to Scala 3.8.4

fwbrasil:
### Problem

Dependency, plugin, and toolchain versions had drifted behind their latest stable releases.

### Solution

- Bump the toolchain (Scala 3.8.3 to 3.8.4, LTS 3.3.7 to 3.3.8, sbt 1.12.5 to 1.12.13) and all library/plugin versions to latest stable: zio, ox, scalatest, caliban, jsoniter, fs2, aeron, pekko, slf4j, logback, scalameta, scalajs-dom, scala-java-time, the kyo-bench set, sbt-scalajs, sbt-scala-native, sbt-projectmatrix, sbt-mima, sbt-updates, scalac-options.
- Freeze the kyo-tasty real-world fixture jars (a deliberate spread of versions used to exercise TASTy decoding) with an explanatory comment.
- Adapt to Scala 3.8.4's stricter checks:
  - make Fields.Pin accessible (3.8.4 tightened implicit accessibility);
  - widen receivers in statically-decidable isInstanceOf assertions to genuine runtime checks (kyo-schema, kyo-ui, kyo-tasty);
  - hoist LogTest's MarkedBackend out of local scope so its type test is runtime-checkable;
  - drop inline on Sync.ensure's error-aware finalizer so a pure-value finalizer body adapts to the opaque `<` (3.8.4 infers an unfolded union for the inlined body that no longer conforms).

### Notes

- scalafmt left at 3.9.6; a 3.11.1 bump reformats 843 files for no functional change.
- The Sync.ensure inline drop is a workaround for a dotty 3.8.4 inference regression; restore if it is fixed upstream.
- Validated with a clean kyoJVM/Test/compile under -Werror plus the affected test suites.

### DIFF
--- a/project/WasmPlatform.scala
+++ b/project/WasmPlatform.scala
@@ -30,8 +30,7 @@ case object WasmPlatform extends Platform {
     def enable(project: Project): Project =
         project.enablePlugins(ScalaJSPlugin).settings(
             scalaJSLinkerConfig ~= {
-                // Scala.js 1.22.0 made the WebAssembly backend require an ES2022-or-later target and deprecated the old
-                // `withExperimentalUseWebAssembly` flag in favor of the `withESFeatures(_.withUseWebAssembly(...))` form.
+                // sbt-scalajs 1.22.0 requires ES2022 for the Wasm backend (withUseWebAssembly replaces the deprecated flag).
                 _.withESFeatures(_.withESVersion(ESVersion.ES2022).withUseWebAssembly(true))
                     .withModuleKind(ModuleKind.ESModule)
             },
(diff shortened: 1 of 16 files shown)


project/WasmPlatform.scala
```diff
@@ -29,7 +30,9 @@ case object WasmPlatform extends Platform {
     def enable(project: Project): Project =
         project.enablePlugins(ScalaJSPlugin).settings(
             scalaJSLinkerConfig ~= {
-                _.withExperimentalUseWebAssembly(true)
+                // Scala.js 1.22.0 made the WebAssembly backend require an ES2022-or-later target and deprecated the old
+                // `withExperimentalUseWebAssembly` flag in favor of the `withESFeatures(_.withUseWebAssembly(...))` form.
```
hearnadam: do we need this?

hearnadam (approved):

project/WasmPlatform.scala
fwbrasil: Yes, there were failures without it. I've updated the comment to not mention the old experimental setting

PR #1707, opened 2026-06-24
[ai] kyo-ai: a typed effect module for LLM generation

fwbrasil:
## Problem

kyo had no first-class way to call a language model. Using an LLM meant reaching for a mainstream SDK and writing the orchestration by hand: authoring JSON schemas, running the tool-call loop, threading the message list, parsing streaming deltas, and decoding replies.

## Solution

Add kyo-ai, a module where a call to a language model is a typed value you compose, not a request you orchestrate. You declare the result type and the capabilities the model may use; the module derives the schema, runs the loop, and decodes the reply.

- `AI.gen[A]` returns a typed `A`: the result schema is derived from `A derives Schema`, the model is forced to that shape, and the reply is decoded. `AI` is a first-class instance value (one conversation slot); `LLM` is a custom `ArrowEffect` whose single world-reaching op runs the eval loop.
- A composable generation surface, each enabled via `AI.enable`: `Tool` (a typed function the model may call, with an async body), `Prompt` (instructions), `Thought` (typed reasoning woven into the result schema), `Mode` (generation-interception middleware).
- `Agent`: an actor-backed conversational entity on kyo-actor that remembers across `ask`s.
- Streaming: `ai.stream[A]` in two forms inferred from `A`, incremental text for `String` (token by token) and object by object for any other type (each complete element emitted once).
- Providers: an OpenAI-compatible backend (OpenAI, DeepSeek, Gemini, Groq, Baseten, OpenRouter) and the Anthropic Messages backend, both on kyo-http, with auto-config from environment API keys. Temperature is opt-in.
- A typed failure hierarchy (`AIException`), the `Config`/`Context`/`Image` value types, and cross-run misuse detection.

Supporting kyo-schema change the module builds on: a `@doc` annotation for JSON Schema field descriptions, recorded on each field during derivation and surfaced in the generated schema.

## Notes

- Cross-platform: JVM, Scala.js, and Scala Native.
- Tested with an in-process server implementing both wire protocols, plus eight runnable `*Demo.scala` exemplars validated live against OpenAI and Anthropic.


kyo-ai/shared/src/main/scala/kyo/ai/completion/OpenAICompletion.scala
```diff
@@ -0,0 +1,215 @@
...
+        // derived Schema writes fields in declaration order, so the order here IS the wire order.
+        case class FunctionCall(name: String, arguments: String) derives Schema
+        case class ToolCall(id: String, `type`: String, function: FunctionCall) derives Schema
+        case class FunctionDef(description: String, name: String, strict: Boolean, parameters: JsonSchema) derives Schema
+        case class ToolDef(function: FunctionDef, `type`: String = "function") derives Schema
+
+        // The message `content` is polymorphic on the wire: a bare string for text messages, a parts array for
+        // vision messages. It is carried as a Structure.Value, whose identity Schema writes shape-aware (a string
+        // stays a string, a sequence becomes an array), reproducing both shapes from one field with no union.
+        case class MessageEntry(
```
hearnadam, line 128: Can this be `@CamelCase` or whatever? now that we have the schema control? @DamianReeves

kyo-ai/shared/src/main/scala/kyo/ai/Config.scala
```diff
@@ -0,0 +1,207 @@
...
+        yield Config(provider.baseUrl, key, org, provider, modelName, modelMaxTokens)
+
+    /** A purely-constructed config for a provider's catalog entry (key/org left absent; filled at use via
+      * the provider default path). The catalog values use this so a model literal is pure.
+      */
+    private[kyo] def catalog(provider: Provider, modelName: String, modelMaxTokens: Int): Config =
+        Config(provider.baseUrl, Absent, Absent, provider, modelName, modelMaxTokens)
+
+    /** A provider: its display name, base URL, the key env-var name, and the wire completion backend. */
+    trait Provider(
```
hearnadam: any reason to choose a trait over an abstract class?

kyo-ai/shared/src/main/scala/kyo/ai/completion/OpenAICompletion.scala
fwbrasil: I believe we'd need https://github.com/getkyo/kyo/issues/1713 and use an untagged union here.

PR #1715, opened 2026-06-25
[schema] codec-aware representation fallback and field serde controls

### LINKED ISSUE #1705: [schema] Sum representation fallback: hint-based codec-aware degradation and fallback chains
## Summary

Sum-type wire representations (#1695, shipped in #1704) are an all-or-nothing choice: a
`Schema[A]` carries exactly one `SumRepresentation`, and encoding with a codec that cannot
express it throws. This issue proposes treating a representation as a **hint** with
codec-aware degradation, and an optional **fallback chain** so a schema can declare an
ordered set of acceptable representations (highest priority first) that the encoder and
decoder negotiate against the active codec.

## Background: what exists today

`Schema[A]` holds a single `representation` slot, set by one of the builders:

```scala
val s = Schema[Shape].tupleFlat   // ["Triangle",10.0,10.0,10.0]
```

At encode time, representations that produce a top-level non-object value (`Tuple`,
`TupleFlat`, `Untagged`) require a codec whose writer can emit a top-level non-object.
Protobuf cannot, so the encoder hard-fails:

```scala
// SchemaSerializer.scala
private def requireTopLevelCapable(writer: Writer, representation: String): Unit =
    given Frame = Frame.internal
    if !writer.canWriteTopLevelNonObject then
        throw RepresentationUnsupportedException(writer.codecName, representation)
```

```scala
// SchemaException.scala
case class RepresentationUnsupportedException(codec: String, representation: String)(using Frame)
    extends SchemaException(
        s"Codec '$codec' cannot express the '$representation' sum representation. " +
        s"Use a self-describing codec (Json, Yaml, Ion, MsgPack) or a different representation."
    ) with TransformException
```

## Problem

A schema authored for a self-describing codec (Json/Yaml/Ion/MsgPack) cannot be reused with
Protobuf without rewriting the schema. The same `Schema[A]` value is not portable across the
codec set: the wire-shape choice that is ideal for one transport becomes a hard failure on
another. The user has no way to express "use TupleFlat where the codec supports it, otherwise
fall back to a representation that works."

```scala
// today: one schema, two codecs, one of them throws
val s = Schema[Shape].tupleFlat
s.encodeString[Json](Triangle(10, 10, 10))      // ["Triangle",10.0,10.0,10.0]
s.encode[Protobuf](Triangle(10, 10, 10))        // throws RepresentationUnsupportedException
```

## Proposed unlock

Two composable capabilities. (A) is the core ask; (B) generalizes it.

### (A) Representation as a hint with degradation

A builder variant marks a representation as preferred rather than required. When the active
codec cannot express it, the encoder degrades to a declared fallback instead of throwing, and
the decoder accepts both shapes.

```scala
// after: degrade to External when the codec cannot express TupleFlat
val s = Schema[Shape].tupleFlat.orElseRepresentation(SumRepresentation.External)

s.encodeString[Json](Triangle(10, 10, 10))   // ["Triangle",10.0,10.0,10.0]  (preferred)
s.encode[Protobuf](Triangle(10, 10, 10))     // {"Triangle":{...}}            (degraded, no throw)
```

### (B) Fallback chain (ordered, highest priority first)

A schema declares an ordered list of acceptable representations. Encode selects the
highest-priority representation the active codec can express; decode tries them in order until
one parses (the Untagged capture-replay machinery is prior art for try-in-order decode).

```scala
// last-configured / first-listed wins where supported; the rest are fallbacks
val s = Schema[Shape].representations(TupleFlat, Adjacent, External)

s.encodeString[Json](Triangle(10, 10, 10))   // TupleFlat  (first the codec supports)
s.encode[Protobuf](Triangle(10, 10, 10))     // Adjacent   (TupleFlat unsupported; Adjacent is a top-level object)
```

## API sketch (illustrative, exact surface decided at design)

```scala
// on Schema[A], Focused-preserving, consistent with the existing representation builders:
def orElseRepresentation(fallback: SumRepresentation): Schema[A] { type Focused = ... }
def representations(first: SumRepresentation, rest: SumRepresentation*): Schema[A] { type Focused = ... }

// the selected representation for a given codec is observable for tooling/tests:
def representationFor(codecCapabilities: Codec.Capabilities): SumRepresentation
```

## Acceptance criteria (EARS)

Ubiquitous:
- The schema serializer shall select, for each `(schema, codec)` pair, exactly one
  representation to encode with, deterministically.
- The schema serializer shall preserve byte-identical output for a schema configured with a
  single representation and no fallback (full back-compat with #1695).

Event-driven:
- When a schema declares a fallback or representation chain and the active codec can express
  the highest-priority representation, the encoder shall encode using that representation.
- When the active codec cannot express the highest-priority representation, the encoder shall
  encode using the highest-priority representation in the chain that the codec can express.
- When decoding a value produced by any representation in the schema's chain, the decoder shall
  reconstruct the original value regardless of which representation produced the wire.

State-driven:
- While a schema is configured with a single representation and no fallback, the serializer
  shall retain today's behavior exactly, including throwing `RepresentationUnsupportedException`
  on an incapable codec.

Unwanted behavior:
- If no representation in the chain can be expressed by the active codec, then the encoder shall
  throw `RepresentationUnsupportedException` naming the codec and the attempted chain.
- If two representations in a chain produce wire shapes a decoder cannot disambiguate, then the
  decode order shall be the declared chain order, and the first variant that parses to a valid
  value shall win (consistent with Untagged declaration-order semantics).
- If a fallback chain is empty or contains duplicates, then schema construction shall reject it
  with a typed exception rather than silently normalizing.

Optional:
- Where a schema declares a fallback chain, the selected representation for a given codec shall
  be observable through a public accessor for testing and tooling.
- Where naming/alias configuration (#1694) is present, each representation in the chain shall
  apply the same naming layer, so tag and field names are consistent across the chain.

## Design questions to resolve at the gate

1. Silent degradation vs opt-in: is automatic fallback acceptable as the default, or must it be
   explicitly requested (`.tupleFlat.orElseRepresentation(...)`) so a wire-shape change across
   codecs is never a surprise? A silent shift is a round-trip footgun if a producer and consumer
   assume different codecs.
2. Single hint vs full chain: ship (A) alone, or (A) as sugar over (B)?
3. Stacking DX: does an ordered chain help real workflows, or does codec-dependent wire shape make
   the contract harder to reason about than a hard failure? Weigh against the explicit
   "one schema, one shape" mental model.
4. Capability model: today capability is a single boolean (`canWriteTopLevelNonObject`). A chain
   may need a richer `Codec.Capabilities` so selection generalizes beyond the one axis.
5. Decode ambiguity bound: how to bound the cost of try-in-order decode for a long chain, and
   whether to cap chain length.

## Composition and constraints

- Must not change the default-`External` byte-identical guarantee from #1695.
- Must compose with the variant-naming and alias layer (#1694).
- Decode try-in-order reuses the Untagged capture-once / replay-per-variant approach already in
  `SchemaSerializer.readUntagged`.
- The failure model shifts from "hard fail" to "negotiated"; determinism per `(schema, codec)` is
  the load-bearing invariant and every acceptance test must pin a concrete wire shape per codec.

## Out of scope

- New representations beyond the five shipped in #1695.
- Cross-codec automatic transcoding (decode with codec X, re-encode with codec Y).
- Runtime/global codec capability registration outside the schema/codec types.

## References

- #1695 (sum representations: the feature this builds on)
- #1704 (the PR that shipped #1695)
- #1694 (variant naming / alias layer this must compose with)

### LINKED ISSUE #1712: [schema] Support omit-empty field policies for optional and collection fields
## Problem

`kyo-schema` currently has useful structural transforms for naming and sum-type discriminators, including the new variant-name configuration. That gets close to several external JSON contracts, but one common wire-shape requirement still forces hand-written codecs: omitting fields whose values are semantically absent.

The two most common cases are:

- optional fields: omit when the value is `None` / `Maybe.empty` instead of writing `null`
- collection fields: omit when the value is an empty `Chunk`, `List`, `Vector`, `Set`, `Seq`, or `Map` instead of writing `[]` / `{}`

Ascribe has this exact need in its ASG JSON contract. The ASG shape uses derived-looking product/sum data, but the wire contract omits many empty fields:

```json
{"name":"text","value":"hello","location":[{"line":1,"col":1},{"line":1,"col":10}],"type":"string"}
```

Instead of encoding optional fields as `null` or empty child arrays, the contract often omits them entirely. This is one of the remaining reasons Ascribe still needs an explicit `Structure.Value` wire layer even after the new `discriminator("name").variantNames(...)` support.

## Requested capability

Add schema-level omit policies for absent/empty fields. This should be available programmatically first, and should also fit naturally into the annotation-support work in #1696.

Possible builder API shape:

```scala
Schema[Node]
  .discriminator("name")
  .variantNames("CharRef" -> "charref", "DListItem" -> "dlistItem")
  .omitNone
  .omitEmptyCollections

Schema[Node]
  .omit(_.blocks).whenEmpty
  .omit(_.header).whenNone
```

Possible annotation shape as an extension to #1696:

```scala
import kyo.schema.*

case class Document(
  @omitWhenEmpty blocks: Chunk[Block],
  @omitWhenNone header: Option[Header]
) derives Schema
```

The exact API names are not important; the important part is that omission is represented as schema metadata/transforms rather than requiring a custom codec.

## Desired semantics

- Encode omits configured fields when their value is absent/empty.
- Decode treats missing omitted fields as the normal empty/default value when the Scala field type can supply one:
  - `Option[A]` -> `None`
  - `Maybe[A]` -> `Maybe.empty`
  - supported collections -> empty collection
  - maps -> empty map
- Field-level configuration should override broader schema-wide policy.
- Programmatic configuration should take precedence over annotation-derived configuration, consistent with #1696.
- Omission should compose with existing `rename`, `renameAllFields`, `discriminator`, and `variantNames` transforms.

## Relationship to existing issues

- Extends #1696: annotation capture/configuration should include omit hints such as `@omitWhenNone` and `@omitWhenEmpty`.
- Complements #1695: alternate sum-type wire representations help with the enclosing sum shape, while omit policies handle product fields inside those variants.

## Why this matters

Many public JSON contracts use omission rather than explicit nulls or empty arrays for absent data. Without omit policies, users can get close with derived schemas and naming transforms, then still have to fall back to an explicit codec for a small but contract-breaking difference.

For Ascribe specifically, this would make it more realistic to replace a large explicit ASG wire codec with derived kyo-schema configuration while preserving the existing ASG/TCK JSON contract.

## Acceptance criteria

- The schema library shall provide a programmatic way to omit optional fields when absent.
- The schema library shall provide a programmatic way to omit collection/map fields when empty.
- The schema library shall decode a missing configured-omitted field back to the corresponding empty/default Scala value.
- Omit policies shall compose with field renaming and discriminator/variant naming transforms.
- The annotation support design in #1696 shall have corresponding annotation hooks for per-field omission hints.
- An unconfigured schema shall preserve current encode/decode behavior.

### LINKED ISSUE #1713: kyo-schema: support native Schema derivation for Scala type unions
> *This was generated by AI during triage.*

## Problem

The recent sum-representation work gives `kyo-schema` a strong story for nominal sums: sealed traits, enums, and tagged or untagged ADTs. It does not provide the same native story for Scala type unions like `A | B`.

Kyo already has hand-rolled schema support for union-shaped data:

- `Maybe[A]` encodes as `null` or the inner `A`
- `Result[E, A]` encodes as a discriminated object with `"$type"` and `"value"`
- `Either[A, B]` encodes as a discriminated `Left` / `Right` object

Those are useful precedents, but users should not need bespoke schemas to represent ordinary type unions.

## Current kyo-schema behavior

Please verify this first with a compile test, but local inspection suggests there is no generic `Schema[A | B]` support today.

`Tag` / `ConcreteTag` already understand union types, which may help runtime branch selection. `Schema` currently has explicit instances for `Maybe`, `Result`, and `Either`, plus nominal sum derivation through sealed traits and enums.

## Request

Add native, out-of-the-box support for Scala type unions:

```scala
type Payload = Int | String | User

summon[Schema[Payload]]

Json.encode[Payload](42)
Json.decode[Payload]("42")
```

The unconfigured schema should be useful without per-project setup, and it should have a clear configuration story for cases where the default shape is not enough.

Because the recent sum-representation work has not landed and `kyo-schema` is not yet 1.0, this is also the right point to decide whether the public naming should shift from `SumRepresentation` to `UnionRepresentation`. The same representation taxonomy applies to both nominal sums and Scala type unions, so `UnionRepresentation` may be the better long-term name if the feature is meant to cover both.

## Proposed default behavior

A type union schema should flatten nested unions, preserve a deterministic branch order, and expose the union as a `Structure.Type.Sum`.

For self-describing codecs, the natural default should be an untagged branch-payload representation:

```scala
type Payload = Int | String

Json.encode[Payload](42)      // 42
Json.encode[Payload]("text")  // "text"
```

Decode should try the member schemas deterministically. If exactly one member decodes cleanly, return that value. If none match, return a typed decode failure listing the attempted members. If multiple members match, return a typed ambiguity failure unless the schema is explicitly configured to choose a branch order.

## Configuration story

Type unions should reuse the existing sum or union representation machinery where possible:

```scala
given Schema[Payload] =
  Schema[Payload]
    .adjacent("type", "value")
    .renameAllVariants(Schema.NameCase.SnakeCase)
```

The design should explicitly decide whether the shared representation descriptor is named:

```scala
Schema.SumRepresentation
```

or renamed before landing to:

```scala
Schema.UnionRepresentation
```

A rename is viable now because the API has not shipped in a stable release. If renamed, the representation should still cover sealed traits, enums, and Scala type unions. The goal is one vocabulary for a schema whose value may be one of several alternatives, not separate parallel APIs.

The design should also provide a way to name union members whose default type label is not suitable for a wire contract, especially primitive members and aliases:

```scala
given Schema[Int | String] =
  Schema[Int | String]
    .adjacent("type", "value")
    .unionMemberNames(
      "Int" -> "number",
      "String" -> "text"
    )
```

Exact API names can be decided during implementation, but the supported capabilities should include:

- choosing `External`, `Internal`, `Adjacent`, `Tuple`, `TupleFlat`, or `Untagged`
- deciding whether the representation type is called `SumRepresentation` or `UnionRepresentation`
- renaming union member tags
- accepting decode aliases for member tags
- configuring ambiguity behavior for untagged decode
- producing precise errors for codecs that cannot express the selected representation

## Acceptance criteria

- [x] A compile test confirms whether `summon[Schema[A | B]]` is unsupported today, or captures the existing behavior if support already exists.
- [x] `Schema[A | B]` derives or summons without user configuration when `Schema[A]` and `Schema[B]` exist.
- [x] Nested unions such as `A | (B | C)` are flattened into one logical union schema.
- [x] JSON, YAML, Ion, and MsgPack round-trip disjoint primitive and product unions with no configuration.
- [x] Untagged decode returns a typed no-match failure when no member schema can decode the input.
- [x] Untagged decode returns a typed ambiguity failure when multiple member schemas can decode the input, unless the schema is configured to resolve ambiguity by order.
- [x] Tagged representations compose with the existing sum or union representation machinery.
- [x] Union member names compose with the existing variant naming and alias layer.
- [x] Binary codecs either encode the configured representation correctly or fail before writing with a typed `SchemaException`.
- [x] `Maybe`, `Result`, and `Either` keep their current wire shapes and remain byte-compatible.
- [x] The design decides before landing whether the shared representation type should be named `SumRepresentation` or `UnionRepresentation`, with the chosen name applying consistently to sealed traits, enums, and Scala type unions.

## Design questions

1. Should unconfigured type unions default to untagged branch payloads, or to the same external wrapper form as sealed-trait sums?
2. What is the stable default member name for primitives, aliases, opaque types, and parameterized types?
3. Should ambiguity detection be strict by default for type unions even though existing untagged nominal sums use declaration-order matching?
4. Can branch selection on encode rely on `Tag` / `ConcreteTag`, or does schema need its own member classifier?
5. How should field-level schema generation, JSON Schema `oneOf`, and `Structure.Type.Sum` expose type-union members?
6. Should `SumRepresentation` be renamed to `UnionRepresentation` before the sum-representation API lands, given that the same concept now needs to cover both nominal sums and Scala type unions?

## References

- #1691 for variant naming and aliasing
- #1695 / #1704 for sum representations

### LINKED ISSUE #1714: [schema] Per-field custom serialization transforms (serialize_with / deserialize_with)
## Problem

kyo-schema can retype a whole schema with `Schema[A].transform[B](to)(from)`, but it cannot give a single field a bespoke wire form independent of the field type's own codec: a `Duration` written as integer seconds, a number read from a string, an enum written as a code, an epoch-millis timestamp. This is the most common real-world wire-contract need kyo-schema cannot express at field granularity. Users hit it for date formats, stringly-typed numbers, and legacy field encodings, and today must hand-roll an entire codec or restructure their types.

This came out of a capability audit against Rust serde's serialization-customization behaviors (the behaviors, not the attribute mechanism, which belongs to #1696). The other field-level gaps that audit surfaced (programmatic decode-time field defaults, predicate-driven omission, strict decode / reject-unknown-fields) are being delivered with the representation/omission/union work; per-field custom transforms are the remaining piece and are filed here because they are an independent new slot rather than an extension of the omit/decode-default machinery.

## Capability: per-field custom transform (serde `with` / `serialize_with` / `deserialize_with`)

Let a single field encode and decode through a configured function pair, leaving every other field on its derived codec. The escape hatch for "this one field has a bespoke wire form."

Shape: a per-field transform slot alongside `renamedFields` / `omitPolicies`, carrying a `(value, Writer) => Unit` / `Reader => Any` pair (or a `Schema[FieldType] => Schema[FieldType]` substitution). The encode gate in `SchemaSerializer.writeWithTransforms` (the loop the omit gate already hooks) routes the field's value through the override instead of its derived `Schema`; decode mirrors it in `readWithTransforms`. This reuses the field-transform threading the representation/omission work already builds, so the substrate cost is incremental.

A serialize-only or deserialize-only variant (serde `serialize_with` / `deserialize_with` independently) falls out of the same slot by leaving one direction at the derived codec.

## Scope notes

- Builds on the field-transform threading in `SchemaSerializer` introduced by the omission and representation-fallback work.
- Out of scope: annotation-driven spelling (see #1696); container-level conversion, which `Schema[Underlying].transform[B]` already covers; Rust-specific serde attributes with no wire-fidelity value (`borrow`, `remote`, `bound`, `crate`, the identifier codegen primitives).

## Prior art

Rust serde field attributes: https://serde.rs/field-attrs.html.

DamianReeves:
Closes #1705 (representation fallback chain): https://github.com/getkyo/kyo/issues/1705
Closes #1712 (field omission): https://github.com/getkyo/kyo/issues/1712
Closes #1713 (type unions): https://github.com/getkyo/kyo/issues/1713
Closes #1714 (per-field transform): https://github.com/getkyo/kyo/issues/1714

## Problem

A `Schema[A]` fixed its wire shape at derivation time with no awareness of the codec doing the encoding. A sum schema declaring a top-level non-object shape (`tupleFlat`, `tuple`, untagged) hard-failed with `RepresentationUnsupportedException` on a codec that cannot express it (Protobuf cannot write a top-level non-object), so a single schema could not serve a Json producer wanting a compact array AND a Protobuf consumer needing an object envelope. The only workaround was a separate schema per codec.

Beyond the wire shape of sums, the per-field serialization surface was thin. There was no way to omit empty or absent fields on encode, supply a decode-time default for an absent field, reject unknown input fields, or give a single field a custom wire form. Scala 3 type unions (`A | B`) had no derivation at all.

## Solution

Make representation selection a negotiation instead of a hard failure. A schema declares an ordered preference and the encoder picks the highest-priority shape the active codec can express, degrading rather than throwing.

- `Schema[A].representations(first, rest*)` declares the ordered chain; `.orElseRepresentation(fallback)` is sugar for a two-entry chain. `.representationFor(caps: Codec.Capabilities)` is the pure determinism witness: the same `(schema, codec)` pair always selects the same shape. Encode picks the first expressible entry highest-priority-first; decode tries entries in declared order over a wire captured once, first valid parse wins, re-throwing a `Panic` rather than advancing. An exhausted chain throws `RepresentationUnsupportedException`; an empty or duplicate chain is rejected at the builder with `DuplicateRepresentationException`.
- Capability is projected off the `Codec.Writer` type as `Codec.Capabilities`, so an external codec participates in selection with no kyo-schema source change.
- `SumRepresentation` is renamed `UnionRepresentation`: one taxonomy now covers sealed traits, enums, and Scala type unions.

Field-level serde customization, all focus-macro driven (`omit(_.field)`, etc.):

- `.omitNone` / `.omitEmptyCollections` / `.omit(_.f).whenEmpty|.whenNone|.when(pred)|.whenDefault`: predicate-driven encode omission, type-aware (empty-omit fires only for collection and map field types), decode-defaulted back.
- `.default(_.f)(supplier)`: by-name decode-time default for an absent field of any type, evaluated at most once and only when the field is absent. Precedence is supplier over Scala default over tag zero.
- `.denyUnknownFields`: strict decode raising `UnknownFieldException` on an unknown wire field, applied after naming, aliases, and transforms, on both self-describing (Json) and binary (Protobuf) codecs.
- `.transformField` / `.transformFieldWrite` / `.transformFieldRead`: per-field custom wire form composing with omit, rename, and default.

`Schema[A | B]` now derives directly, untagged by default. Decode multi-probes every member; under the default `Strict` policy more than one match raises `AmbiguousVariantMatchException` listing the matched members, while `.unionAmbiguity(FirstMatch)` resolves a known-ambiguous union by declaration order.

## Notes

- The no-customization encode and decode path is byte-identical to before: `UnionRepresentation.External` is the inert default, an empty chain is inert, and field customization adds nothing on the default path. The new representation chain reuses the existing per-representation capability probe (lifted from throw-or-pass into choose-first-passing) and the existing untagged capture-once-replay decode machinery, rather than adding a parallel engine.
- Decode lookup keys are modeled as a typed `WireKey` (a `ByName` opaque over `String` and a `ByFieldId` opaque over `Int`), keeping the field-name and numeric field-id namespaces disjoint so a self-describing codec (lookup by name) and Protobuf (lookup by numeric tag) cannot alias inside one map. The `whenDefault` omit baseline is materialized once at config time through the field schema, so it is correct for product, collection, and sum field types.
- The write path now threads the caller `Frame` so a serialization failure reports the call site; `Frame.internal` is used only at frameless construction sites. This is why the kyo-doctest `Schema[kyo.Path]` instance gains an explicit `(using Frame.internal)`.


fwbrasil:
The PR looks good but there's a key aspect that I think we should follow up on: unions at the type level have no order and can easily end up represented in different orders in the same codebase since a lot comes from inference. I think ordering the deserializers could be enough to have a reasonable behavior. Or we could even have a dynamic priority oder.

fwbrasil (approved):

Issue #1719, opened 2026-06-27
[schema] Field-number annotation for the Protobuf codec (strict-interop pinning)


DamianReeves:
## Summary

Add a discoverable, derivation-time field-number annotation for the `kyo-schema` Protobuf codec so users can pin stable proto field numbers directly on a case class field, e.g. `@protoField(3) zip: String`.

## Motivation

The Protobuf codec assigns field numbers from a stable `MurmurHash3` of the field name. These are valid proto3 wire numbers and give schema-evolution stability, but they do not match a conventional hand-written `.proto` (1, 2, 3, ...). Today the only way to pin specific numbers is a runtime field-id override map, which is not discoverable from the type and easy to forget.

The strict/canonical conformance work (#1716) adds two non-failing steers toward pinning numbers for external interop:
- `protoSchema` annotates each hash-derived field number with a provenance comment.
- `Protobuf.fieldNumberAudit[A]` reports which fields rely on hash defaults vs. an explicit number.

Both point the user at a pin mechanism. A field-level annotation is the natural, discoverable mechanism, but it depends on annotation-derivation support that is being built separately, so it is tracked here rather than in #1716.

## Proposed

- A `@protoField(n: Int)` (name TBD) field annotation read by schema derivation, taking precedence over the hash-based default, equivalent to the existing field-id override for that field.
- Surfaced by `fieldNumberAudit` as `pinned = true`.
- Documented as the recommended way to pin field numbers for strict external interop.

## Depends on

The in-progress annotation-derivation work in `kyo-schema`.

## Related

- #1716 (Protobuf repeated/map fix + strict/permissive proto3 conformance; adds the non-failing field-number steers this annotation would complement)

fwbrasil:
Implemented in #1722, as a scoped `@proto.fieldNumber(n)` rather than the flat `@protoField(n)` proposed here. It desugars onto `Schema.fieldId`, so a pinned number takes precedence over the hash-derived default. Covered by `SchemaAnnotationTest.scala` and `ProtobufDifferentialTest.scala`.

PR #1721, opened 2026-06-27
[schema] fix Protobuf repeated/map codec, add proto3 conformance

### LINKED ISSUE #1716: [BUG]: Protobuf codec corrupts/breaks repeated (`List`) and `Map` fields
### Version

1.0.0-RC4

### Scala Version

3.8.3

### Expected Behavior

All of these case classes round-trip through `Protobuf.encode` / `decode` unchanged:

```
  Flat         : OK
  ListInt      : OK
  Nested       : OK
  ListNested   : OK
  MapStr       : OK
```

`Map[Int, V]` (and other standard protobuf map key types) should either derive a `Schema` or the limitation should be documented.

### Actual Behavior

```
  Flat         : OK
  ListInt      : MISMATCH
      in : ListInt(1,List(1, 2, 3))
      out: ListInt(1,List(1, 6538652, 2, 6538652, 3))
  Nested       : MISMATCH
      in : Nested(true,List(4, 5))
      out: Nested(true,List(4, 6538652, 5))
  ListNested   : THREW kyo.MissingFieldException: Missing required field 'a'. Add this field to the input.
  MapStr       : MISMATCH
      in : MapStr(9,Map(a -> 1, b -> 2))
      out: MapStr(9,Map(0 -> 1, 1 -> 2))
```

- **`List[scalar]`**: a constant junk value (`6538652` here) is inserted between every real element on decode. Looks like a packed-vs-unpacked framing mismatch — the decoder appears to read a field tag / length delimiter as a data element. Deterministic, and reproduces both top-level and nested in a sub-record.
- **`List[caseclass]`** (repeated embedded message): decode fails entirely with `MissingFieldException: Missing required field 'a'` — the framing for repeated messages is not read back correctly.
- **`Map[String, V]`**: values survive but the keys are discarded and replaced by their positional index (`0, 1, …`).
- **`Map[Int, V]`** (or any non-`String` key): does not compile — `No given Schema[Map[Int, V]] ... Define ... as a case class or sealed trait, or provide a given Schema[...]`.
- Scalar / `Option` / nested-record fields round-trip correctly; the problem is specific to repeated and map fields.

**Impact:** any schema containing a `List`/`Seq` or `Map` field cannot be safely serialized with the protobuf codec — data is silently corrupted (lists, string-keyed maps) or fails outright (repeated messages).

### Steps to Reproduce

Run the following (e.g. with scala-cli):

```scala
//> using dep io.getkyo::kyo-schema:1.0.0-RC4

object Repro:
  case class Flat(a: Int, b: String, c: Boolean) derives kyo.Schema
  case class ListInt(a: Int, b: List[Int]) derives kyo.Schema
  case class Nested(a: Boolean, b: List[Int]) derives kyo.Schema
  case class ListNested(a: Int, d: List[Nested]) derives kyo.Schema
  case class MapStr(a: Int, f: Map[String, Int]) derives kyo.Schema
  // case class MapInt(a: Int, f: Map[Int, Int]) derives kyo.Schema  // does not compile

  def check[A](label: String, value: A)(using kyo.Schema[A]): Unit =
    try
      val encoded = kyo.Protobuf.encode(value)
      val decoded = kyo.Protobuf.decode[A](encoded).getOrThrow
      val ok      = decoded == value
      println(f"  $label%-12s : ${if ok then "OK" else "MISMATCH"}")
      if !ok then
        println(s"      in : $value")
        println(s"      out: $decoded")
    catch case e: Throwable => println(f"  $label%-12s : THREW ${e.getClass.getName}: ${e.getMessage}")

  def main(args: Array[String]): Unit =
    check("Flat", Flat(42, "hi", true))
    check("ListInt", ListInt(1, List(1, 2, 3)))
    check("Nested", Nested(true, List(4, 5)))
    check("ListNested", ListNested(7, List(Nested(false, List(1)), Nested(true, List(2, 3)))))
    check("MapStr", MapStr(9, Map("a" -> 1, "b" -> 2)))
```

### Current Workaround

_No response_

DamianReeves:
## Problem

Two correctness bugs corrupted every repeated and map field (fixes #1716).

**Repeated fields.** `hasNextElement()` delegated to `hasNextField()` without tracking the field number, so only the first element of any repeated `List`/`Seq`/`Vector`/`Set`/`Chunk`/`Span` field was consumed. Elements past the first trashed the surrounding message parse, producing `MissingFieldException` for repeated message fields and silently truncated collections for repeated scalars.

**Map fields.** The writer emitted each map entry as a struct field with a synthetic number rather than as a proto3 `MapEntry` message (key at field 1, value at field 2). On decode, string keys were replaced by positional index strings (`"0"`, `"1"`, ...). Non-String map keys did not compile; no `mapSchema[K, V]` given existed.

Two secondary bugs completed the picture.

**`protoSchema` wrong field numbers.** The emitter used `idx + 1` sequential numbers while the wire carries MurmurHash3-derived numbers, making the `.proto` output inconsistent with the actual bytes.

**Map decode depth bypass.** The string-keyed map decode path never incremented the depth counter, so a deeply nested `Map[String, Map[String, ...]]` produced a `StackOverflowError` instead of `Result.Failure(LimitExceededException)`.

## What changed

User-facing summary of the change set:

- **Repeated fields round-trip.** `List`, `Seq`, `Vector`, `Set`, `Chunk`, and `Span` of scalars or case classes encode and decode correctly, top-level and nested.
- **Maps round-trip with keys preserved.** `Map[String, V]` keeps its keys; non-String keys (`Map[Int, V]`, `Map[Boolean, V]`, opaque/value-class keys) now derive a `Schema` via the new `mapSchema[K, V]` given and round-trip.
- **Absent repeated/map fields decode to empty** (proto3 semantics), not a failure.
- **Strict proto3 conformance by default.** New `Protobuf.Conformance` enum on `Protobuf.Config`; `Strict` (the default) rejects map keys proto3 cannot express natively, `Permissive` keeps the round-trippable extensions. **Breaking** for a non-native-key map under the default given (see Notes).
- **Packed repeated scalars.** Scalar repeated fields encode in proto3's canonical packed form; the reader accepts both packed and unpacked.
- **`protoSchema` emits the real wire field numbers** (was `idx + 1`), with a provenance comment on hash-derived numbers and a warning for the proto3 reserved range `19000-19999`. Suppress the provenance comment with `Config(protoSchemaProvenance = false)`.
- **New `Protobuf.fieldNumberAudit[A]`** reports each field's wire number and whether it is pinned or in the reserved range, without encoding or decoding.
- **Field-number pinning is wire-functional.** `Schema[A].fieldId(_.field)(n)` now pins the number on encode, decode, `protoSchema`, and the audit.
- **Map decode honors `maxDepth`** (no more `StackOverflowError` on a deeply nested map).
- README and CONTRIBUTING updated. Field-number annotation affordance tracked in getkyo/kyo#1719.

### Repeated and map round-trips

```scala
import kyo.*

case class Order(id: Int, items: List[String], counts: Map[Int, Int]) derives Schema, CanEqual

val order   = Order(1, List("pen", "ink"), Map(10 -> 2, 20 -> 5))
val bytes   = Protobuf.encode(order)
val decoded = Protobuf.decode[Order](bytes)
// decoded == Result.Success(order)
```

### Conformance mode

The zero-argument `given Protobuf` is `Strict`. Strict rejects a map key proto3 cannot
express natively (a value-class or float key); opt into the entry-message extension with
`Permissive`:

```scala
case class Pid(value: Int) derives Schema, CanEqual
case class Doc(tags: Map[Pid, String]) derives Schema, CanEqual

// default (Strict): a value-class key is rejected at encode
Protobuf.encode(Doc(Map(Pid(1) -> "x"))) // throws SchemaNotSerializableException

// Permissive: encodes via the entry-message form and round-trips
given Protobuf = Protobuf(Protobuf.Config(conformance = Protobuf.Conformance.Permissive))
val d = Doc(Map(Pid(1) -> "x"))
Protobuf.decode[Doc](Protobuf.encode(d)) // Result.Success(d)
```

A key that reduces to a native scalar (an opaque type or value class whose `Schema` is a
native scalar) stays allowed under Strict.

### protoSchema

```scala
case class User(id: Int, name: String) derives Schema

Protobuf.protoSchema[User]
// message User {
//   sint32 id = 1243642;    // hash-derived field number; pin via Schema.fieldId for stable external interop
//   string name = 1790876;  // hash-derived field number; pin via Schema.fieldId for stable external interop
// }

// suppress the provenance comments (the reserved-range warning is always kept):
given Protobuf = Protobuf(Protobuf.Config(protoSchemaProvenance = false))
Protobuf.protoSchema[User]
// message User {
//   sint32 id = 1243642;
//   string name = 1790876;
// }
```

### Pinning a stable field number

```scala
case class User(id: Int, name: String) derives Schema, CanEqual

// pin `name` to 7 on the wire; encode/decode/protoSchema/audit all honor it
given Schema[User] = Schema.derived[User].fieldId(_.name)(7)

val u = User(1, "ada")
Protobuf.decode[User](Protobuf.encode(u)) // Result.Success(u), with `name` at field 7
```

### Field-number audit

```scala
case class User(id: Int, name: String) derives Schema

Protobuf.fieldNumberAudit[User]
// Chunk(
//   FieldNumberInfo(path = "id",   name = "id",   number = 1243642, pinned = false, inReservedRange = false),
//   FieldNumberInfo(path = "name", name = "name", number = 1790876, pinned = false, inReservedRange = false)
// )

// enforce pinning in your own test/CI, at your chosen severity:
val unpinned = Protobuf.fieldNumberAudit[User].filter(!_.pinned)
assert(unpinned.isEmpty, s"pin field numbers for external interop: ${unpinned.map(_.path)}")
```

### Decode limits

```scala
// config-backed limits (defaults from Json.DefaultMaxDepth / DefaultMaxCollectionSize)
given Protobuf = Protobuf(Protobuf.Config(maxDepth = 16, maxCollectionSize = 1024))
Protobuf.decode[Order](bytes) // Result.Failure(LimitExceededException) if a bound is exceeded

// or pass explicit limits for a one-off
Protobuf.decode[Order](bytes, maxDepth = 8, maxCollectionSize = 256)
```

## Notes

**Default is `Strict`, but this does not break any correctly-functioning prior usage.** The repeated and map fields produced corrupt output before this PR, so no code that ran correctly against a prior release depended on non-canonical map-key encoding. A `Float`-, `Double`-, or product-type-keyed map that now raises `SchemaNotSerializableException` under `Strict` was previously encoding to bytes that could not be decoded back. To allow the non-canonical extensions:

```scala
given Protobuf = Protobuf(Protobuf.Config(conformance = Protobuf.Conformance.Permissive))
```

**Derivation-time field-number annotation is deferred to getkyo/kyo#1719.** `Schema.fieldId` at the call site is the current path for pinning a stable wire number. A compile-time annotation on a case-class field that makes the derived `Schema` carry the pin automatically is tracked separately.

### DIFF
--- a/kyo-schema/shared/src/main/scala/kyo/internal/FocusMacro.scala
+++ b/kyo-schema/shared/src/main/scala/kyo/internal/FocusMacro.scala
@@ -550,9 +550,9 @@ import scala.quoted.*
       * Each per-field schema resolves via `summonInline[Schema[ft]]` at the generated-code typer phase;
       * the in-flight `derived$Schema` is visible by forward-reference, so recursive products close the
       * cycle. The read body seeds defaults / `Maybe.empty` / `None`, tracks a `Long` required-field
-      * bitmap, ORs in `droppedFieldsMask` before the required check, and throws `MissingFieldException`
-      * (carrying `reader.frame`) for a missing required field. Used by both `derivedImpl`
-      * (`derives Schema`) and the `metaApplyImpl` Focus path.
+      * bitmap, ORs in reader-supplied pre-satisfied field masks before the required check, and throws
+      * `MissingFieldException` (carrying `reader.frame`) for a missing required field. Used by both
+      * `derivedImpl` (`derives Schema`) and the `metaApplyImpl` Focus path.
       */
     private def emitProductSchemaStatic[A: Type](using
         Quotes
@@ -735,10 +735,21 @@ import scala.quoted.*
         // READ: static per-field name-matched loop.
         def readBody(r: Expr[Reader]): Expr[A] =
             val requiredMask: Long = fields.zipWithIndex.foldLeft(0L) { case (m, (_, idx)) =>
-                if !hasDefaultFlags(idx) && !isMaybeFlags(idx) && !isOptionFlags(idx) then m | (1L << idx)
+                if !hasDefaultFlags(idx) && !isMaybeFlags(idx) && !isOptionFlags(idx) then
+                    m | (1L << idx)
                 else m
             }
             val fieldNames: List[String] = fields.map(_.name)
+
+            val absentDefaultableMaskExpr: Expr[Long] =
+                fields.indices.foldLeft('{ 0L }) { (mask, idx) =>
+                    if hasDefaultFlags(idx) || isMaybeFlags(idx) || isOptionFlags(idx) then mask
+                    else
+                        effectiveSchemaTypes(idx).asType match
+                            case '[t] =>
+                                val s = fieldSchemaExprTyped[t](idx)
+                                '{ $mask | kyo.internal.absentDefaultMask($s, ${ Expr(1L << idx) }) }
+                }
 
             val seedExprs: List[Expr[Any]] = fields.zipWithIndex.map { (f, idx) =>
                 val ft = tpe.memberType(f)
@@ -754,8 +765,10 @@ import scala.quoted.*
                     ft.asType match
                         case '[t] => '{ Option.empty[Any].asInstanceOf[t] }
                 else
-                    ft.asType match
-                        case '[t] => '{ null.asInstanceOf[t] }
+                    effectiveSchemaTypes(idx).asType match
+                        case '[t] =>
+                            val s = fieldSchemaExprTyped[t](idx)
+                            '{ kyo.internal.absentDefaultSeed($s) }
                 end if
             }
 
@@ -836,7 +849,10 @@ import scala.quoted.*
                     val maskExpr  = Expr(requiredMask)
                     val nExpr     = Expr(n)
                     Some('{
-                        val _combined = $seenRef | $r.droppedFieldsMask($nExpr)
+                        val _combined =
+                            $seenRef |
+                                $r.droppedFieldsMask($nExpr) |
+                                $r.absentDefaultedFieldsMask($nExpr, $absentDefaultableMaskExpr)
                         if (_combined & $maskExpr) != $maskExpr then
                             val _missing = java.lang.Long.numberOfTrailingZeros((~_combined) & $maskExpr).toInt
                             val _names   = $namesExpr
--- a/kyo-schema/CONTRIBUTING.md
+++ b/kyo-schema/CONTRIBUTING.md
@@ -396,6 +396,53 @@ def fieldId(name: String): Int =
     (MurmurHash3.stringHash(name) & 0x1fffff) + 1
 ```
 
+### Conformance modes (Strict and Permissive)
+
+`Protobuf.Config.conformance` is a CODEC-level concern, not a Schema slot. The schema derives the same structure in both modes; the codec gate is at encode time.
+
+`Conformance.Strict` (the default): encode rejects any map whose key type is not a proto3-native scalar (an integral, `Boolean`, `String`, or an opaque/value-class whose `Schema` reduces to one of those). Rejection throws `SchemaNotSerializableException` with a message prefixed `"non-canonical proto3 map key: "`. No bytes are written before the throw.
+
+`Conformance.Permissive`: all map key types are permitted. Non-canonical keys use the entry-message encoding and round-trip through this codec; they are not standard proto3 `map<K,V>` externally.
+
+The `given Protobuf = Protobuf()` zero-argument instance is Strict. Permissive requires an explicit config:
+
+```scala
+given Protobuf = Protobuf(Protobuf.Config(conformance = Protobuf.Conformance.Permissive))
+```
+
+### fieldNumberAudit and FieldNumberInfo
+
+`Protobuf.fieldNumberAudit[A]` is pure and total: no encode, no decode, no throw. It returns `Chunk[FieldNumberInfo]`, one row per message field, in declaration order, depth-first for nested messages.
+
+`FieldNumberInfo` fields:
+- `path`: dotted path from the root message (e.g., `"inner.id"` for a nested field)
+- `name`: leaf field name
+- `number`: the wire field number this codec uses (MurmurHash3 formula or a leaf-name override)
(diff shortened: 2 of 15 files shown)


kyo-schema/shared/src/main/scala/kyo/internal/FocusMacro.scala
```diff
@@ -593,6 +593,52 @@ import scala.quoted.*
 
         val hasDefaultFlags = fields.zipWithIndex.map((_, idx) => MacroUtils.hasDefault(sym, idx))
 
+        // Typed empty seeds for collection / map fields. Readers may opt into treating absent
+        // fields with these seeds as satisfied when the format defines that absence rule, as
+        // Protobuf does for repeated and map fields. Self-describing codecs do not opt in, so
+        // these fields remain required there.
+        val listSym   = TypeRepr.of[List[Any]].typeSymbol
```
fwbrasil: We can't specialize macros to specific types. Won't this break with other collection types?

DamianReeves: You were right. I removed the concrete collection-type specialization from `FocusMacro`.

The fix is now schema-driven:

- `Schema` has an internal `absentDefaultValue` slot.
- Built-in collection, map, and `Span[Byte]` schemas provide their empty absent values there.
- `FocusMacro` no longer checks for `List`, `Vector`, `Seq`, `Set`, `Map`, `Chunk`, or `Span` symbols. It only emits calls through `SchemaBridge` to ask each field schema for the absent seed and defaultable mask bit.
- Protobuf remains the only codec that treats those absent defaults as satisfying required fields via `Reader.absentDefaultedFieldsMask`; self-describing codecs keep missing collection/map fields as missing.

I also added a regression with a custom `PB1716Bag` collection schema and a derived holder. That verifies a non-built-in collection works through the schema path, which is the case this comment was pointing at.

PR #1725, opened 2026-07-01
[build] reduce Maven Central publishing footprint

fwbrasil:
### Problem
Maven Central is introducing per-namespace monthly publishing limits (file count, release size, release count) enforced from August 2026. Each io.getkyo release publishes about 197 artifact coordinates at 24 files each, placing the namespace in the top publishing percentile on both file count and size. Part of that footprint never needed to be on Central: the test-only WebAssembly target, benchmarks, examples, internal test fixtures, and empty platform aggregators. Separately, kyo-ffi-plugin ships a ~33 MB jar because it bundles the whole Scala 3 toolchain as plugin resources.

### Solution
* WebAssembly becomes a test-only platform. WasmPlatform sets publish/skip on every wasm row and on the kyoWasm aggregator, and release.yml no longer stages Wasm. Wasm still builds and tests in the CI matrix.
* Stop publishing non-library artifacts: kyo-bench, kyo-examples, kyo-tasty-fixtures-internal (a Test-only dependency), and the empty kyoJS and kyoNative aggregators.
* kyo-ffi-plugin no longer bundles the Scala 3 toolchain. It resolves kyo-ffi-codegen and its transitive toolchain from the consumer's resolvers at task time, version matched through a baked version.txt, reusing the existing ffiCodegenClasspath seam. The published plugin jar drops from ~33 MB to ~94 KB.

### Notes
* Codegen resolution is automatic and cached in the consumer's coursier cache. The in-repo integration test still overrides ffiCodegenClasspath with the codegen project's own classpath, so there is no publishLocal round-trip.
* Javadoc jars remain a separate, larger size lever and are not addressed here.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,19 +80,6 @@ jobs:
           PGP_SECRET: ${{ secrets.PGP_SECRET }}
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
-      - name: Stage Wasm (retriable)
-        uses: nick-fields/retry@v4
-        with:
-          max_attempts: 3
-          timeout_minutes: 120
-          command: sbt -Dplatform=Wasm "ci-release"
-        env:
-          CI_SONATYPE_RELEASE: ""
-          CI_CLEAN: ""
-          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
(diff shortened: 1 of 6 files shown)


hearnadam (approved):

DamianReeves:
A possible viable alternative is to publish the WASM artifacts to NPM.

DamianReeves (approved):

fwbrasil:
> A possible viable alternative is to publish the WASM artifacts to NPM.

We actually don't need separate artifacts fot wasm. @sjrd demonstrated that the JS artifacts are enough. The platform setup remains because we need to run tests in wasm anyway.

PR #1726, opened 2026-07-01
Replace Murmur hashing with XXHash

hearnadam:
### Problem

Kyo had several internal Murmur-based hashes used for deterministic IDs, buckets, and cache placement. These paths need a shared non-cryptographic hash implementation with explicit cross-platform byte-order behavior and support for both low-level inputs and `Span[Byte]`.
 
### Solution
Port the official xxHash 0.8.x Java XXH32 and XXH64 implementations into `kyo.internal.XXHash`. The port keeps the official constants, round functions, tail processing, and avalanche steps, with explicit little-endian reads for stable JVM, JS, Native, and Wasm behavior.

The API remains internal. The core implementation lives in `kyo-config` and supports:

- `Array[Byte]` XXH32 and XXH64 hashing
- sliced `Array[Byte]` hashing with range validation
- `String` XXH32 hashing as UTF-8 bytes
- `Int` hashing as four little-endian bytes

`kyo-data` adds `XXHashExtensions` with `Span[Byte]` overloads on `XXHash.type`, delegating to the same core implementation.

This migrates:
  - rollout bucket hashing
  - scheduler admission string keys
  - cache slot hashing
  - browser image hashes
  - workflow schema structural hashes
  - schema default field IDs
  - `Tag` content hashes

### Notes

  - Adds official xxHash sanity-vector coverage for XXH32 and XXH64.
  - Adds `Span[Byte]` overload tests that compare directly against the array core.
  - Updates schema documentation and generated field-number expectations for XXH32-derived IDs.
  - Validation run:
    - `kyo-configJVM/testOnly kyo.XXHashTest`
    - `kyo-dataJVM/testOnly kyo.XXHashExtensionsTest`


DamianReeves:
Did you consider Blake3 or is raw performance what we need optimization for?
If we think collision risk is something we want to mitigate Blake3 might be worth considering.
In my kymora project I use Blake3.

hearnadam:
> Did you consider Blake3 or is raw performance what we need optimization for?

Both Murmer and xxHash are non-cryptographic hashes. Murmer is fine if you need a good distribution of keys, but it's quite slow. The collision risk of xxHash is quite low overall.

DamianReeves:
> > Did you consider Blake3 or is raw performance what we need optimization for?
> 
> Both Murmer and xxHash are non-cryptographic hashes. Murmer is fine if you need a good distribution of keys, but it's quite slow. The collision risk of xxHash is quite low overall.

Okay sounds good.

DamianReeves (approved):

PR #1730, opened 2026-07-02
[kyo-compat] run Ox acquireReleaseWith release on interruption

ghostdogpr:
## Problem

Ox cancellation and timeout are thread interruption, so a cancelled `use` exits via `InterruptedException`. The Ox binding's hand-rolled `acquireReleaseWith` evaluated `use` under `catch case t if NonFatal(t)`, and `scala.util.control.NonFatal` classifies `InterruptedException` as fatal. The interrupt therefore propagated past the release line, so a timed-out `use` (e.g. `timeoutOption(...)(use)`) skipped release entirely and leaked the acquired resource. The `zio`/`ce`/`kyo` bindings avoid this because they delegate to native bracket primitives.

Two defects: the interrupt skipped release entirely, and even once caught, release ran interruptibly, so a scope tearing the fork down could interrupt release mid-flight.

## Fix

Modeled on Ox's own `useInterruptible` / `use`, whose release is documented as uninterruptible:

- Broaden the `use` catch from `NonFatal(t)` to all `Throwable`, so the interrupt is captured, release runs, then it is rethrown via `useResult.get` (cancellation semantics preserved).
- Run release inside `ox.uninterruptible(...)` so a pending interrupt cannot abort it before it completes.

`ensure` delegates to `acquireReleaseWith`, so it is fixed too.

## Tests

Added two regression tests to the shared `BracketEnsureTest` (one per public entry point) that cut `use` short with a timeout and assert release still runs. They fail on the pre-fix binding for the right reason ("release did not run") and pass after. The design waits past the `use` duration, so it is also correct on the non-cancelling `Future` binding (orphaned `use` completes and release fires). Verified failing-then-passing on ox and passing on all six bindings that share the suite.


kyo-compat/test/shared/src/test/scala/kyo/compat/BracketEnsureTest.scala
```diff
@@ -209,4 +209,35 @@ class BracketEnsureTest extends CompatTest:
...
 
+    "acquireReleaseWith runs release when use is cut short by a timeout" in run {
+        // A timed-out `use` must still release its resource. Ox cancellation is thread interruption,
+        // so the timed-out `use` exits via InterruptedException; on bindings without cancellation the
+        // `use` runs to completion orphaned. Either way release must run.
+        val released = new AtomicBoolean(false)
+        val bracket: CIO[Int] =
+            CIO.acquireReleaseWith(CIO.unit)(_ =>
+                CIO.defer { val _ = released.set(true) }
+            )(_ => CIO.sleep(200.millis).map(_ => 0))
```
fwbrasil: Can we avoid these timed sleeps in the tests? The perf of Github free runners vary a lot, so this has a tendency to become flaky. It seems in this case it could simply block forever with a latch

kyo-compat/test/shared/src/test/scala/kyo/compat/BracketEnsureTest.scala
```diff
@@ -209,4 +209,35 @@ class BracketEnsureTest extends CompatTest:
...
+        CIO.timeout(50.millis)(bracket).flatMap { result =>
+            // Wait past the use duration so release has fired on every binding before asserting.
+            CIO.sleep(500.millis).map { _ =>
```
fwbrasil: maybe the release could fulfill a promise that is waited for here?

kyo-compat/test/shared/src/test/scala/kyo/compat/BracketEnsureTest.scala
ghostdogpr: Removed them 👍

kyo-compat/test/shared/src/test/scala/kyo/compat/BracketEnsureTest.scala
ghostdogpr: Done 👍

PR #1742, opened 2026-07-11
[kyo-schema] add Ion Binary, BSON, and Ion Schema generation

DamianReeves:
## Summary

This adds three serialization formats to `kyo-schema` and, in the process, fixes several pre-existing `kyo-schema` defects that the new cross-format test coverage surfaced (including a real `Schema.derived`/enum issue on Protobuf). Everything is derived from a type's `Schema`, works across JVM, JS, and Native, and keeps the existing Ion text, JSON, YAML, Protobuf, and MessagePack behavior intact.

The diff is large (39 files, ~7.5k lines) so this description is organized as: **new capabilities** (with examples), **bug fixes** (with the enum one called out), then a **reviewer's guide** to navigating the change.

---

### 1. Ion Binary

Spec-valid Ion 1.0 binary (with the `E0 01 00 EA` BVM), the same Ion data model as the existing Ion text codec. Two entry points: through `Ion.Config` (format routing) or the standalone `IonBinary` codec.

```scala
import kyo.*

case class User(name: String, age: Int) derives Schema

val alice = User("Alice", 30)

// Standalone binary codec (default `given IonBinary` in scope via `import kyo.*`):
val bytes: Span[Byte] = IonBinary.encode(alice)
val back: Result[DecodeException, User] = IonBinary.decode[User](bytes)

// Or route the Ion byte helpers through binary via Ion.Config:
val config = Ion.Config(format = Ion.Format.Binary)
val bytes2: Span[Byte] = Ion.encodeBytes(alice, config)
Ion.decode[User](bytes2, config)
```

Decode returns `Result[DecodeException, A]`; malformed wire input (bad BVM, descriptor, VarUInt/VarInt, symbol table, trailing bytes) surfaces as a typed `ParseException`, never a thrown exception. Ion text remains fully preserved and unchanged; `Ion.Config` now selects text vs binary and the byte helpers route accordingly.

### 2. BSON

The document-oriented binary format used by MongoDB. The top-level value must be object-shaped (a case class or a string-keyed map), with strict single-root-document semantics by default.

```scala
val bytes: Span[Byte] = Bson.encode(alice)
Bson.decode[User](bytes)

// Configurable safety limits, same shape as Json.decode:
val config = Bson.Config(maxDepth = 8, maxCollectionSize = 64)
Bson.decode[User](bytes, config)
```

Notable BSON behavior, all enforced up front rather than producing bad bytes:

- Case classes encode as a BSON document keyed by field name, collections as BSON arrays, `Span[Byte]` as BSON binary (subtype 0), and `java.time.Instant` as the BSON UTC datetime type.
- BSON datetime is millisecond-precision only; a sub-millisecond `Instant` raises `SchemaNotSerializableException` rather than silently truncating.
- `BigInt` encodes only within the signed 64-bit int64 range; out-of-range raises `SchemaNotSerializableException` (no truncation). `BigDecimal` encodes losslessly to 34 digits via Decimal128.
- Root arrays, scalars, and null are rejected before any bytes are written.

### 3. Ion Schema (ISL 2.0) generation

An Ion Schema Language 2.0 document describing a type's wire shape, derived from its `Schema`.

```scala
// ISL text for a type:
val isl: String = Ion.ionSchemaString[User]

// Or build the document model and render / stream it:
val doc: IonSchema = IonSchema.from[User]
IonSchema.encode(doc)                 // String
IonSchema.encodeTo(doc, outputStream) // streams UTF-8 ISL, flushes, does not close
```

`encodeTo` writes incrementally to the `OutputStream` as the document renders (it does not materialize the full document in memory first). Products, options, collections, string-keyed and non-string-keyed maps, and sealed traits map to ISL type expressions; supported `Schema.Constraint` metadata renders as ISL constraints.

### 4. Supporting model changes

- **`Structure` base-value expansion**: the shared representation gains `Bytes`, `Instant`, and `Duration` base values, so these round-trip through every format that can express them (Ion text/binary, BSON, MsgPack, Protobuf, JSON, YAML) rather than each codec special-casing them.
- **`Ion.Config.annotationEmissionMode`** (`Suppress` default / `Emit`): an opt-in `Codec.Writer` SPI (`annotations` / `canWriteAnnotations`, both no-op by default so every other codec's output is byte-for-byte unchanged) lets Ion encoding emit schema annotations as Ion type annotations when a caller opts in. The default preserves prior Ion output exactly.
- **Typed failures throughout**: decode paths return `Result[DecodeException, A]`; unrepresentable encode inputs fail with `SchemaNotSerializableException`.

---

## Bug fixes surfaced by the new coverage

The cross-format and generative test coverage added here exercised code paths the existing suites did not, and turned up several real, pre-existing defects. Each was fixed reproduce-first (a failing test added before the fix).

### `Schema.derived` + enums (Protobuf)

Two distinct enum/`Schema.derived` defects on Protobuf, both fixed:

1. **Canonical-shape validation was bypassed on the generic entry points.** A schema shape Protobuf cannot canonically represent (for example an enum used as a map key) was rejected only by `Protobuf.encode` directly; the generic `Schema[A]#encode[C]` / `encodeString[C]` path silently emitted non-canonical wire. Fixed by adding an internal `Codec.validate` hook that every `Schema.encode`/`encodeString` entry point now calls, so Protobuf's conformance check applies uniformly regardless of entry point.

2. **The field-number audit was blind to a variant-level pin.** A `@proto.fieldNumber` pin on a sealed-trait / enum variant's own field was invisible to the audit (wire bytes were already correct; the reporting was wrong). Fixed by reading the field's own captured annotation as a fallback.

A 25-case enum-derivation probe matrix (simple, parameterized, mixed, nested, enum-in-enum, large-arity enums; as product fields, `Option`/`List`/`Set` elements, and map keys; under transforms and each union representation; across all codecs) is now a permanent regression suite. It also confirmed the many enum shapes that were already correct (ordinal never leaks into the wire; only the declared name does).

### Protobuf field-id pins not applied through several paths

The generative transform coverage surfaced a family of related pin/transform bugs, each fixed reproduce-first with byte-level evidence:

- Field-id pins were not threaded through the generic `Schema.encode`/`decode` entry points (only through `Protobuf.encode`/`decode` directly), so the generic path wrote hash-derived field numbers that mismatched the pinned ids.
- Pins on **nested** schemas (a pinned field inside a nested product or container element) were never applied at all.
- A nested schema's pins **replaced the shared override state with no restore**, so an outer schema's own pin on a field written *after* a nested-pinned field emitted a default field number. This one passed self-round-trip (write and read were wrong together) but broke interop with any other Protobuf tooling; it is now fixed with save/restore scoping per nesting depth.

### Map-field decode corruption under schema transforms

Decoding a real `Map[String, V]` field corrupted on every format when the schema also carried a field transform (`.rename` / `.drop` / `.denyUnknownFields`), because the transform-aware reader reused a cached field name for map entry keys instead of consuming the wire key. Fixed; a cross-format transform-plus-Map regression matrix guards it.

### Yaml class-initialization deadlock

Running `kyo.YamlTest` in isolation (`testOnly kyo.YamlTest`) deadlocked: `Yaml`'s static initializer forced `ReaderConfig.Default` while `ReaderConfig`'s constructor defaults read `Yaml` members, so concurrent first-touch from opposite ends could wedge class initialization for the JVM lifetime. Full-suite runs happened to dodge it by initialization order. Fixed by making the shared limit constants `inline val` (so config defaults fold to literals, removing the cross-object read) and making the default pipeline `lazy`; a concurrent first-touch regression guard races class init from fresh class loaders under a bounded deadline. The same init-safety pattern was applied uniformly across the Ion/Bson/Json config constants, guarded by one `CodecInitTest`.

---

## Testing

- Full `kyo-schema` suite green on **JVM (2903), JS (2898), and Native (2898)** on the squashed commit.
- New permanent regression suites: `IonBinaryConformanceTest`, `BsonConformanceTest`, the enum-derivation matrix in `SchemaTest`, the transform/pin/Map matrices in `CodecTest`, and `CodecInitTest`.
- Conformance fixtures assert stable wire bytes and specific decode-failure types with expected messages (not just "some failure"), so a fixture cannot pass on a wrong-reason failure.
- README and `CONTRIBUTING.md` updated for the new formats and codec-authoring conventions; README examples compile under doctest.

---

## Reviewer's guide (for navigating a large diff)

New public surface (start here):
- `kyo-schema/shared/src/main/scala/kyo/IonBinary.scala` — standalone Ion Binary codec.
- `kyo-schema/shared/src/main/scala/kyo/Bson.scala` + `internal/bson/*` — BSON codec, writer, reader, Decimal128.
- `kyo-schema/shared/src/main/scala/kyo/IonSchema.scala` — ISL 2.0 generation.
- `kyo-schema/shared/src/main/scala/kyo/Ion.scala` — `Config` format routing, `AnnotationEmissionMode`, `ionSchema` helpers.
- `kyo-schema/shared/src/main/scala/kyo/Structure.scala` — base-value expansion.

Cross-cutting internals (the bug-fix surface):
- `kyo-schema/shared/src/main/scala/kyo/internal/SchemaSerializer.scala` — the transform/pin threading and the Map-decode fix.
- `kyo-schema/shared/src/main/scala/kyo/{Schema,Protobuf,Codec}.scala` — nested pin threading, the `validate` hook, the field-number audit.
- `kyo-schema/shared/src/main/scala/kyo/Yaml.scala` — init-deadlock fix.

Docs and tests fill the remainder; the conformance and matrix suites are the fastest way to see the intended behavior of each new format and each fix.


DamianReeves:
### CI status: the failing jobs are pre-existing failures on `main`, not introduced by this branch

The red jobs here (`build (Native)` linux-x64, `build (JVM)` linux-arm64 and linux-x64) fail with tests in `kyo-stm` and `kyo-browser`, never in `kyo-schema`, and they fail the same way on this PR's base commit on `main`.

Evidence:
- Base commit `baa4c1eb8` on `main` ([run 29054083495](https://github.com/getkyo/kyo/actions/runs/29054083495)): Native x64 fails `BrowserSettlementTest` / `BrowserVerifyReadTest`; JVM arm64 fails `STMStressTest`.
- This PR ([run 29164148209](https://github.com/getkyo/kyo/actions/runs/29164148209)), across all 4 rerun attempts: the failures stay in those same two modules (`STMStressTest` on JVM, the kyo-browser CDP suites on Native x64). Zero `kyo-schema` failures. No kernel OOM on any job.

Tracked in #1743.

Separately, the earlier native OOM on this PR is resolved. This branch shards the `kyo-schema` native test link into its own sbt driver (`NATIVE_HEAVY`), so its whole-program optimize runs against a fresh heap; Native x64 now links `kyo-schema` and runs 1421/1426 tests, failing only on the pre-existing kyo-browser tests above.

kyo-schema/jvm/src/test/scala/kyo/CodecInitTest.scala
```diff
@@ -0,0 +1,130 @@
...
+// against Ion.Config) to satisfy the guard-every-codec mandate, but neither pairing
+// carries a live class-init-lock dependency in either direction: the raced classes
+// cannot deadlock against each other today, and the case only proves the two classes
+// initialize concurrently without wedging. A genuine future Json.Config or
+// IonBinary.Config forced by its own codec object's clinit would need its own
+// dedicated race case against that new companion. Raw threads and bounded waits are
+// required: the subject is JVM class initialization itself, which cannot be
+// suspended through effect-based primitives, and the trial loop below iterates the
+// same raw mechanics for the same reason.
+class CodecInitTest extends kyo.test.Test[Any]:
```
fwbrasil, line 37: Is there a concrete issue this addresses? Ideally companions shouldn't hold global mutable state

kyo-schema/shared/src/main/scala/kyo/internal/bson/BsonFormat.scala
```diff
@@ -0,0 +1,35 @@
...
+    val Int64: Int      = 0x12
+    val Decimal128: Int = 0x13
+
+    val BinarySubtypeGeneric: Int = 0x00
+    val BinarySubtypeOld: Int     = 0x02
+    val MinDocumentLength: Int    = 5
+end BsonFormat
+
+private[kyo] enum BsonValue derives CanEqual:
+    case DocumentValue(fields: Vector[(String, BsonValue)])
```
fwbrasil, line 24: side note: it'd be nice to have an ordered map in `kyo-data` for this kind of use case

kyo-schema/shared/src/main/scala/kyo/internal/bson/BsonReader.scala
```diff
@@ -0,0 +1,586 @@
...
+            case NullValue          => "null"
+            case Int32Value(_)      => "int32"
+            case Int64Value(_)      => "int64"
+            case Decimal128Value(_) => "decimal128"
+    end actualName
+
+end BsonReader
+
+object BsonReader:
+    def apply(input: Array[Byte], config: kyo.Bson.Config)(using frame: kyo.Frame): BsonReader =
```
fwbrasil: `Span` would be ideal since `input` doesn't seem to mutate

kyo-schema/shared/src/main/scala/kyo/internal/SchemaSerializer.scala
```diff
@@ -56,12 +56,64 @@ private[kyo] object SchemaSerializer:
...
+      * change (this schema carried no overrides of its own, or the writer is not a `ProtobufWriter`),
+      * so there is nothing to restore and an ambient ancestor override, if any, is left untouched.
+      */
+    private[kyo] def threadFieldIdOverridesForWrite(schema: Schema[?], writer: Writer): Maybe[Map[String, Int]] =
+        writer match
+            case pw: ProtobufWriter =>
+                val overrides = schema.fieldIdNameOverrides
+                if overrides.nonEmpty then
+                    val prior = pw.fieldIdOverridesSnapshot
+                    val _     = pw.withFieldIdOverrides(overrides)
```
fwbrasil: Not sure I understand the purpose of this line. Does it have a side effect?

kyo-schema/shared/src/main/scala/kyo/Ion.scala
```diff
@@ -1,43 +1,102 @@
...
+      *   the Ion wire format to use
+      * @param maxDepth
+      *   maximum nesting depth for structs and lists when decoding through configured helpers
+      * @param maxCollectionSize
+      *   maximum number of entries in a single collection or struct when decoding through configured helpers
+      * @param annotationEmissionMode
+      *   controls whether Ion writers emit captured schema annotations as Ion type annotations
+      */
+    final case class Config(
+        format: Ion.Format = Ion.Format.Text,
```
fwbrasil, line 76: maybe binary is a better default?

kyo-schema/shared/src/main/scala/kyo/IonSchema.scala
```diff
@@ -0,0 +1,646 @@
...
+        out.toString
+    end encode
+
+    /** Encodes an Ion Schema document as UTF-8 ISL text to an output stream.
+      *
+      * The document is written incrementally as it renders, with no intermediate full-document
+      * string materialized. The write happens synchronously on the calling thread. The stream is
+      * flushed once rendering completes but is never closed; the caller owns its lifecycle.
+      */
+    def encodeTo(schema: IonSchema, out: OutputStream): Unit =
```
fwbrasil: would it be possible to avoid exposing methods that perform side effects? we typically require `AllowUnsafe` for that

kyo-schema/shared/src/main/scala/kyo/Json.scala
```diff
@@ -19,10 +19,10 @@ end Json
   */
 object Json:
     /** Default maximum nesting depth for objects/arrays in JSON decoding (DoS limit). */
-    val DefaultMaxDepth: Int = 512
+    inline val DefaultMaxDepth = 512
```
fwbrasil, line 22: I think we can keep the explicit return type. User-facing methods should always be fully typed

kyo-schema/CONTRIBUTING.md
```diff
@@ -652,7 +656,19 @@ A codec that carries configuration (Yaml's `writerConfig`) uses a default-argume
 final class Yaml(writerConfig: Yaml.WriterConfig = Yaml.WriterConfig.Default) extends Codec:
 ```
 
-If the new format is **self-describing** (can materialize an arbitrary wire value into `Structure.Value` without a schema), its Reader subclass should also extend `Codec.IntrospectingReader` and implement `readStructure(): Structure.Value`. Json and Yaml extend it; Protobuf does not [kyo-schema/shared/src/main/scala/kyo/Codec.scala:146-163].
+If the new format is **self-describing** (can materialize an arbitrary wire value into `Structure.Value` without a schema), its Reader subclass should also extend `Codec.IntrospectingReader` and implement `readStructure(): Structure.Value`. Json and Yaml extend it; Protobuf does not [kyo-schema/shared/src/main/scala/kyo/Codec.scala:157-175]. Ion Binary and BSON are both self-describing binary formats and both implement `readStructure` this way [kyo-schema/shared/src/main/scala/kyo/internal/ionbinary/IonBinaryReader.scala:252-262] [kyo-schema/shared/src/main/scala/kyo/internal/bson/BsonReader.scala:307-317].
```
fwbrasil, line 659: The module is getting larger and larger with new formats. We should consider having separate modules for each eventually

fwbrasil (approved):
LGTM besides the minor comments

kyo-schema/jvm/src/test/scala/kyo/CodecInitTest.scala
DamianReeves: There was a race with initialization of vals and constants.

DamianReeves:
### Responses to review comments

Thanks for the review. Addressing each in fix commit `1aa0571`:

- **`internal/bson/BsonReader.scala` (Span):** Done. `BsonReader` now takes a `Span[Byte]` and `Bson.newReader` passes it through without the `.toArray` copy (`Span` is an opaque `Array` alias here, so it stays zero-cost).
- **`IonSchema.scala` (side-effecting methods / `AllowUnsafe`):** Done. Both `encodeTo` overloads now require `AllowUnsafe` with a `// Unsafe:` note, since each writes synchronously to a caller-supplied `OutputStream`; the test call sites opt in.
- **`internal/SchemaSerializer.scala:85` (purpose / side effect):** It is an intentional side effect: `withFieldIdOverrides` mutates the writer's active field-id override map and returns the writer for chaining (discarded here). The prior map is captured on the line above and restored once this schema's write completes. Added a comment saying so.
- **`Json.scala` (explicit type on the constants):** These are `inline val` compile-time constants, so the type is already the literal singleton (e.g. `512`). A widened `: Int` is rejected by the compiler (`inline value must have a literal constant type`), so they stay inferred. Happy to convert them from `inline val` to plain `val`s with explicit `Int` types if you prefer that trade-off.
- **`jvm/.../CodecInitTest.scala` (concrete issue / global mutable state):** The config companions hold only an immutable `val Default`, no global mutable state. The concrete issue was a class-initialization deadlock: a codec object and its config companion each read the other's `val` constants during class init (the Yaml pairing reproduced it), so two threads first-touching from opposite ends could wedge both classes for the JVM lifetime. `CodecInitTest` is the regression guard, and the decode-limit constants were made `inline val`s to remove the runtime cross-reads. The only mutable state is per-`Codec.Reader` instance (decode limits, reset per decode), not on any companion.
- **`internal/bson/BsonFormat.scala` (ordered map in `kyo-data`):** Agreed, would fit this case well. Out of scope for this PR; noting it as a follow-up.
- **`CONTRIBUTING.md` (separate modules per format):** Agreed. Splitting the formats into separate modules is a sensible step as the surface grows; a follow-up rather than part of this PR.

Issue #1748, opened 2026-07-16
[kyo-schema] omitEmptyCollections breaks decode for an explicitly array-form String-key Dict/OrderedMap


DamianReeves:
## Summary

Under `omitEmptyCollections` (or a per-field `.omit(_.f).whenEmpty`), an empty `Dict[String, V]` or `OrderedMap[String, V]` field round-trips correctly with the default given, but **fails to decode** when a caller explicitly binds the general array-form given (`dictSchema`/`orderedMapSchema`) for a `String` key instead of the object-form default (`stringDictSchema`/`stringOrderedMapSchema`).

It fails loudly with a `TypeMismatchException`, not silently, and only on a non-default given binding. This is a follow-up to the omit-policy fix that made these types honor the policy at all.

## Reproduction

```scala
import kyo.*

case class Record(name: String, m: Dict[String, Int], count: Int)

// Explicitly bind the GENERAL given (array-of-pairs wire form) for a String key,
// overriding the more-specific object-form default `stringDictSchema`.
given Schema[Dict[String, Int]] = Schema.dictSchema[String, Int]
given Schema[Record]            = Schema.derived[Record].omitEmptyCollections

val encoded = Json.encode(Record("alice", Dict.empty, 7))
// encoded == {"name":"alice","count":7}   the empty field is correctly omitted

Json.decode[Record](encoded)
// Failure(TypeMismatchException: Type mismatch: expected Sequence but got Record(Chunk.Indexed()))
```

A non-empty value round-trips fine under the same binding (`{"name":"alice","m":[{"key":"x","value":1}],"count":7}`), so the defect is specific to the omitted-then-reinjected empty value.

## Root cause

When an empty collection field is omitted on encode, the decoder synthesizes the field back with an empty value (`SchemaSerializer.omitDefaultedNames` at the decode side, symmetric with the encode-side omit gate). For `Dict`/`OrderedMap` the synthetic empty value is produced by `emptyMappingWireValue`, which chooses the wire shape from the field's **declared key type**:

- `Structure.Type.Primitive(PrimitiveKind.String, _)` key -> empty `Structure.Value.Record` (object `{}`)
- any other key -> empty `Structure.Value.Sequence` (array `[]`)

This is correct for the default givens, because for a `String` key the default is `stringDictSchema`, which writes the object form. But the **general** `dictSchema` writes the array form for every key type, including `String`, and both givens declare a byte-identical structure:

```scala
[...]

Issue #1747, opened 2026-07-15
[kyo-schema] Protobuf codec mishandles nested and empty collections


DamianReeves:
### Context

Work on #1744 (adding `OrderedMap` to `kyo-data` and teaching `kyo-schema` to serialize it) surfaced a cluster of defects in the Protobuf codec. They are not independent bugs. They share one root cause:

> **proto3's wire format cannot natively express an empty `repeated` or a nested `repeated`. The writer emits bytes anyway without validating, and the reader then guesses.**

Everything below was measured against the real codecs at `c77de2e`, not inferred from reading.

---

## Symptom 1: an empty collection value silently corrupts the *next* entry

proto3 cannot represent an empty `repeated`, so an entry whose value is an empty collection ships **key-only**. The decoder discards the one signal that would reveal this, then reads a value anyway, consuming the *following* entry's value.

`Schema.scala:2853-2863`, `mapSchema`'s read loop:

```scala
if reader.hasNextElement() then
    reader.checkCollectionSize(count)
    discard(reader.objectStart())
    // hasNextField advances past the inter-field separator (a no-op for
    // Protobuf, the comma for a self-describing codec) before each field read.
    discard(reader.hasNextField())
    val _ = reader.field() // "key"
    val k = kSchema.serializeRead(reader)
    discard(reader.hasNextField())   // <-- return value discarded
    val _ = reader.field() // "value"
    val v = vSchema.serializeRead(reader)   // <-- reads a value that may not be there
```

`hasNextField()` returns exactly the information needed (is there a value field at all?), and `discard` throws it away.

Measured, `Map[Int, Chunk[String]]` over the Protobuf codec:

```
input:   1 -> [a],  2 -> [],  3 -> [c]
decoded: 1 -> [a],  2 -> [c], 3 -> [c]     <-- entry 2 stole entry 3's value
```

**An empty value in the last position throws. In any earlier position it is silent.** Position is the only discriminator:

| case | before fix |
|---|---|
| no empty value | PASS |
| empty in FIRST position | silent corruption |
| empty in MIDDLE position | silent corruption |
| empty in LAST position | throws |
| all values empty | corrupt/throw |

Non-empty collection values pass, which is why this survived normal use.

---

## Symptom 2: nested collections collapse to identical bytes

Three distinct values, one encoding. Measured with a `Chunk[Chunk[String]]` value:

```
1 -> []        = 8aa9ea04020802
1 -> [[]]      = 8aa9ea04020802
[...]

PR #1749, opened 2026-07-16
Add OrderedDict to kyo-data, and use it across kyo-schema

### LINKED ISSUE #1744: [kyo-data] Add an insertion-ordered map type
### Context

Follow-up from PR #1742 review ([comment](https://github.com/getkyo/kyo/pull/1742#discussion_r3575098279)). `kyo-schema`'s BSON support wants an insertion-ordered map for building documents; today it works around the lack of one.

### Proposal

Add an ordered (insertion-preserving) map type to `kyo-data`, so codecs and other modules that must preserve key order have a shared, allocation-conscious type instead of ad-hoc workarounds.

### Notes

- Surfaced by `kyo-schema/shared/src/main/scala/kyo/internal/bson/BsonFormat.scala`.
- Out of scope for the format work in #1742; filed as a standalone `kyo-data` enhancement.

DamianReeves:
# Add `OrderedDict` to kyo-data, and use it across kyo-schema

Closes #1744.

## Why

`kyo-schema`'s BSON document builder needs a map that preserves key insertion order and worked around the absence of one (surfaced in the [#1742 review](https://github.com/getkyo/kyo/pull/1742#discussion_r3575098279)). Codecs and other modules that must keep key order had no shared type to reach for. This adds that type and migrates the first consumer onto it.

While building it, several pre-existing `kyo-schema` codec defects surfaced. They are fixed here, each with a test that fails without the fix.

### `OrderedDict[K, V]` + `OrderedDictBuilder` (kyo-data)

An immutable map that iterates in insertion order at every size, with the same dual representation `Dict` uses: a flat `Span` for 8 or fewer entries, a `TreeSeqMap` ordered by insertion beyond that. `Dict` is unchanged.

```scala
opaque type OrderedDict[K, V] = Span[K | V] | TreeSeqMap[K, V]
```

Order is a contract, not an accident:

```scala
val stages = OrderedDict("draft" -> 1, "review" -> 2, "published" -> 3)

stages.toChunk.map(_._1)                    // Chunk("draft", "review", "published")
stages.update("draft", 9).toChunk.map(_._1) // unchanged order: update keeps the key's position
stages.remove("draft").update("draft", 1)   // "draft" now appended last: re-add is a fresh insertion
```

Like `Dict`, it is opaque and provides no `CanEqual`; use `map.is(other)` for structural equality, not `==`. `OrderedDictBuilder` builds one from repeated `add`s and, like `ChunkBuilder`/`DictBuilder`, pools its buffer in a thread-local.

### kyo-schema consumes it three ways

The BSON document store is backed by `OrderedDict` instead of the ad-hoc workaround, and two `Schema` givens let a case class carry an `OrderedDict` field:

```scala
given stringOrderedDictSchema[V](using vSchema0: => Schema[V]): Schema[OrderedDict[String, V]]
given orderedDictSchema[K, V](using kSchema0: => Schema[K], vSchema0: => Schema[V]): Schema[OrderedDict[K, V]]
```

Insertion order survives a kyo encode/decode round-trip on every codec:

```scala
case class Config(byLevel: OrderedDict[Int, String]) derives Schema

val c       = Config(OrderedDict(3 -> "gold", 1 -> "bronze", 2 -> "silver"))
val decoded = Json.decode[Config](Json.encode(c)).getOrThrow
decoded.byLevel.toChunk.map(_._1) // Chunk(3, 1, 2), not sorted
```

proto3 does not mandate entry order for `map` fields, so a foreign Protobuf reader may reorder entries; both givens document this. kyo's own encode/decode always preserves order.

## Codec fixes found along the way

1. **`Schema.dictSchema` non-String-key wire form (BREAKING).** A non-String-key `Dict` wrote each entry as a bare two-element array and failed to decode on six of the seven codecs, with Protobuf silently emitting corrupt bytes. It now writes a two-field `key`/`value` record, the same form `mapSchema` uses. See the breaking note below.
2. **Empty `Dict`/`OrderedDict` field on Protobuf.** proto3 writes nothing for an empty map field, so an empty one decoded to `MissingFieldException`. All four map-like givens now declare an absent default, matching the `Map` givens.
3. **`omitEmptyCollections` / `whenEmpty` ignored `Dict`/`OrderedDict`.** The policy silently no-op'd on these two types while working on `Map`, `Chunk`, `List`, `Vector`, `Set`, `Seq`. It now drops an empty one like every other collection. The discriminator reads the field's declared wire shape rather than a runtime class, so it never emits a backing-array identity string.

## Breaking change

`Schema.dictSchema` changes the wire form for a **non-String-key** `Dict`. MsgPack bytes serialized by an earlier version for such a `Dict` cannot be read by the new code. MsgPack was the only codec that decoded the old form (the other six failed, and Protobuf emitted corrupt bytes), so this bounds where affected data can exist: re-serialize any such data. `stringDictSchema` (String keys) is unaffected.

### Added
- `[kyo-data]` `OrderedDict[K, V]`: an immutable map that iterates in insertion order
- `[kyo-data]` `OrderedDictBuilder[K, V]`: build an `OrderedDict` from repeated adds
- `[kyo-schema]` `Schema.stringOrderedDictSchema` and `Schema.orderedDictSchema`: derive a `Schema` for a case class with an `OrderedDict` field

### Changed
- `[kyo-schema]` `Schema.dictSchema`: non-String-key `Dict` serializes each entry as a two-field `key`/`value` record instead of a bare two-element array. **BREAKING** for previously-serialized MsgPack bytes of a non-String-key `Dict` (see above).
- `[kyo-schema]` `Schema.dictSchema` / `Schema.stringDictSchema`: a case class field holding an empty `Dict` now decodes on Protobuf instead of failing with `MissingFieldException`, matching the `Map` givens.

## Follow-up issues

Two narrower codec limitations are documented and tracked rather than fixed here:

- **[#1747](https://github.com/getkyo/kyo/issues/1747)**: an empty-collection *value* inside a map entry can decode incorrectly on the Protobuf codec (a shared defect across the `mapSchema`, `dictSchema`, and `orderedDictSchema` givens). Disclosed in `kyo-schema/README.md`.
- **[#1748](https://github.com/getkyo/kyo/issues/1748)**: `omitEmptyCollections` does not yet support an explicitly array-form given bound for a String key; that combination fails loud on decode. Documented on the policy and pinned by a characterization test.

## Testing

Tests run on JVM, JS, Native, and Wasm. Cross-platform tests live under `shared/src`.

- kyo-data: `OrderedDictTest` (44), `OrderedDictBuilderTest` (10)
- kyo-schema: `SchemaTest` (304), `ProtobufTest` (147), plus the per-codec suites (Json/Bson/Ion/IonBinary/MsgPack/Yaml)

The order tests use fixtures whose insertion order differs from both the sorted order and the hash order of their keys, so an order-losing decode fails the assertion (an immutable `Map` preserves insertion order up to 4 entries, which a smaller fixture would hide). The Protobuf tests pin encoded byte sizes, not just round-trip.

Adds a `kyo-data` contributor guide and `OrderedDict` coverage in both module READMEs (doctest-validated).

## Commits

17 commits, kept as history rather than squashed. The `dictSchema` fix lands as a reproduce-then-fix pair (`43f48f057` reds the failure, `c03728fdd` fixes it) so the reproduction is on record for `git bisect`.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/OrderedDict.scala
+++ b/kyo-data/shared/src/main/scala/kyo/OrderedDict.scala
@@ -4,7 +4,7 @@ import scala.annotation.tailrec
 import scala.collection.immutable.TreeSeqMap
 import scala.reflect.ClassTag
 
-/** An immutable map that preserves the insertion order of its keys at every size. OrderedMap uses a dual representation optimized for both
+/** An immutable map that preserves the insertion order of its keys at every size. OrderedDict uses a dual representation optimized for both
   * small and large collections:
   *
   *   - For up to 8 entries, keys and values are stored in a flat [[Span]] with linear-scan lookup, avoiding hashing overhead and providing
@@ -23,9 +23,9 @@ import scala.reflect.ClassTag
   * #### Creation
   *
   * ```scala
-  * val empty = OrderedMap.empty[String, Int]
-  * val doc   = OrderedMap("a" -> 1, "b" -> 2, "c" -> 3)
-  * val fromMap = OrderedMap.from(Map("x" -> 10))
+  * val empty = OrderedDict.empty[String, Int]
+  * val doc   = OrderedDict("a" -> 1, "b" -> 2, "c" -> 3)
+  * val fromMap = OrderedDict.from(Map("x" -> 10))
   * ```
   *
   * #### Access
@@ -38,12 +38,12 @@ import scala.reflect.ClassTag
   *
   * #### Modification
   *
-  * All modifications return a new OrderedMap, leaving the original unchanged:
+  * All modifications return a new OrderedDict, leaving the original unchanged:
   *
   * ```scala
-  * doc.update("a", 10)  // OrderedMap("a" -> 10, "b" -> 2, "c" -> 3)
-  * doc.remove("b")      // OrderedMap("a" -> 1, "c" -> 3)
-  * doc ++ OrderedMap("d" -> 4)
+  * doc.update("a", 10)  // OrderedDict("a" -> 10, "b" -> 2, "c" -> 3)
+  * doc.remove("b")      // OrderedDict("a" -> 1, "c" -> 3)
+  * doc ++ OrderedDict("d" -> 4)
   * ```
   *
   * #### Iteration and Transformation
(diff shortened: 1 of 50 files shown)


kyo-data/shared/src/main/scala/kyo/OrderedMap.scala
```diff
@@ -0,0 +1,676 @@
...
+  * OrderedMap uses an unboxed representation through Scala 3's opaque types, so wrapping and unwrapping incurs no runtime overhead.
+  *
+  * @tparam K
+  *   the type of keys
+  * @tparam V
+  *   the type of values
+  * @see
+  *   [[OrderedMapBuilder]] for incremental construction
+  */
+opaque type OrderedMap[K, V] = Span[K | V] | TreeSeqMap[K, V]
```
fwbrasil: Can we name it `OrderedDict` for consistency?

kyo-data/shared/src/main/scala/kyo/OrderedMap.scala
```diff
@@ -0,0 +1,676 @@
...
+  * iteration order unspecified.
+  *
+  * Operations are designed to avoid tuple allocations for key-value pairs. Iteration, transformation, and lookup methods accept separate
+  * key and value parameters rather than `(K, V)` tuples, eliminating boxing overhead in hot paths.
+  *
+  * #### Creation
+  *
+  * ```scala
+  * val empty = OrderedMap.empty[String, Int]
+  * val doc   = OrderedMap("a" -> 1, "b" -> 2, "c" -> 3)
```
fwbrasil: A main premisse of `Dict` was not offering methods that imply tuple allocations. This is following the current design of `Dict` but we should follow up to avoid this constructor. I can be separate since it's pre-existing

kyo-data/shared/src/main/scala/kyo/OrderedDict.scala
```diff
@@ -0,0 +1,676 @@
...
+  * doc.update("a", 10)  // OrderedMap("a" -> 10, "b" -> 2, "c" -> 3)
+  * doc.remove("b")      // OrderedMap("a" -> 1, "c" -> 3)
+  * doc ++ OrderedMap("d" -> 4)
+  * ```
+  *
+  * #### Iteration and Transformation
+  *
+  * ```scala
+  * doc.foreach((k, v) => println(s"$k -> $v"))
+  * doc.map((k, v) => (k.toUpperCase, v * 2))
```
fwbrasil, line 53: hmm another method that implies tuples but it also follows `Dict`

kyo-data/shared/src/main/scala/kyo/OrderedMap.scala
```diff
@@ -0,0 +1,676 @@
...
+            )
+        end mapValues
+
+        /** Returns all keys as a [[Span]] in insertion order. */
+        def keys(using ClassTag[K]): Span[K] =
+            reduce(
+                span =>
+                    val n   = Span.size(span) / 2
+                    val arr = new Array[K](n)
+                    @tailrec def loop(i: Int): Unit =
```
fwbrasil: it seems this could use `System.arraycopy`

kyo-data/shared/src/main/scala/kyo/OrderedMap.scala
```diff
@@ -0,0 +1,676 @@
...
+                    Span.fromUnsafe(arr)
+            )
+
+        /** Returns all values as a [[Span]] in insertion order. */
+        def values(using ClassTag[V]): Span[V] =
+            reduce(
+                span =>
+                    val n   = Span.size(span) / 2
+                    val arr = new Array[V](n)
+                    @tailrec def loop(i: Int): Unit =
```
fwbrasil: ditto

kyo-data/shared/src/main/scala/kyo/OrderedMap.scala
DamianReeves: Done in 7f4bf1e: renamed `OrderedMap` -> `OrderedDict` (the type, its builder, the derived Schema givens, and every reference across kyo-data, kyo-schema, and kyo-config).

kyo-data/shared/src/main/scala/kyo/OrderedMap.scala
DamianReeves: Done in a18be34: the small-size backing span stores keys and values in two contiguous halves, so `keys` is now a single `System.arraycopy` of the first half.

kyo-data/shared/src/main/scala/kyo/OrderedMap.scala
DamianReeves: Done in a18be34: same treatment for `values`, copying the second (values) half in one `System.arraycopy`.

kyo-data/shared/src/main/scala/kyo/OrderedMap.scala
DamianReeves: Agreed. This mirrors `Dict`'s current constructor, so I kept it here to match `Dict` rather than diverge. Happy to do the tuple-free constructor as a separate follow-up covering both `Dict` and `OrderedDict`, as you suggested.

kyo-data/shared/src/main/scala/kyo/OrderedDict.scala
DamianReeves: Same here: follows `Dict`. I'll fold this into the same separate follow-up that drops the tuple-implying methods on both types.

DamianReeves:
Follow-up for the two tuple-allocation notes (the `apply` constructor and `map` on `Dict`/`OrderedDict`) filed as #1754.

PR #1752, opened 2026-07-17
[kyo-data] restore constant-time Tag hashing on the dispatch path

fwbrasil:
## Problem

`Tag.fastPathEqual` and `Tag.checkTypes` compared tags by streaming their encoded strings through XXH32 on every call instead of comparing the memoized `String.hashCode`. Both methods sit on the kernel's per-operation tag-comparison path, so every non-identical tag comparison re-streamed both encoded strings, up to four times per comparison. This slowed effect dispatch roughly 10x on CI (`STMTest` went from 3s to 30s on linux-arm64) and starved the STM stress tests' bounded-retry writers into `FailedTransaction`, keeping `main` red on linux-arm64 since July 2.

`XXHash.hash32(String)` had the same per-call cost baked into the primitive itself, streaming UTF-8 bytes through a dedicated state machine (`Utf8State32`) on every invocation rather than reusing anything cheap and already available on the `String`.

## Solution

`Tag.fastPathEqual` and `Tag.checkTypes` compare memoized `hashCode`s directly instead of routing through `XXHash.hash32`.

`XXHash.hash32(String)` now delegates to `hashInt(input.hashCode)`: XXH32 applied to the four little-endian bytes of the JLS `String.hashCode`. This is constant-time and allocation-free for a reused string, deterministic across platforms, and still avalanche-mixed by the XXH32 finalizer, with collision pairs identical to plain `String.hashCode`. The now-dead `Utf8State32` streaming state machine and the seeded `String` overload are removed.

Because `Tag.hash`, the Protobuf/MessagePack field IDs (`Schema.fieldId`, `Protobuf`), and the config/scheduler key-based bucketing all derive from `hash32(String)`, every one of those call sites becomes constant-time for reused strings, not just the two kernel methods above.

## Notes

String-derived constants are regenerated to match the new derivation: `Tag.hash` pins for `Tag[Int]`/`Tag[String]` in `TagTest`, the Protobuf field-number examples in `kyo-schema/README.md`, and the reserved-range test fixture in `ProtobufTest` (`PBReserved.r1641` renamed to `r1635` to land back in the proto3 reserved range under the new derivation). No release contains the old values, so this constant churn is compat-free.

Interop docs (`kyo-schema/CONTRIBUTING.md`, `kyo-schema/README.md`) now spell out the exact derivation, XXH32 over the JLS `hashCode` bytes, not over UTF-8 content, so an external implementation reproducing Protobuf field numbers can match it precisely.

`kyo-config/README.md` and `kyo-scheduler/README.md` drop "XXH32" in favor of "stable string hash" when describing the string-keyed bucketing and rejection helpers, since the underlying primitive is no longer a direct content hash.

Validated: kyo-config, kyo-data, kyo-schema, and kyo-flow JVM suites green; `STMTest` back to 3.3s locally; `STMStressTest` green across 5 consecutive runs.

### DIFF
--- a/kyo-config/shared/src/main/scala/kyo/internal/XXHash.scala
+++ b/kyo-config/shared/src/main/scala/kyo/internal/XXHash.scala
@@ -5,6 +5,10 @@ package kyo.internal
   * This is a Scala port of the official Java XXH32 and XXH64 implementations from xxHash 0.8.x. The port keeps the same constants, round
   * functions, tail processing, and avalanche steps, while replacing platform-specific memory access with explicit little-endian array reads.
   *
+  * The `String` overload is not a UTF-8 content hash: it applies XXH32 to the memoized JLS `String.hashCode` so that hashing a reused
+  * string is constant-time and allocation-free (see `hash32(String)`). To hash a composite key (for example a case class), prefer
+  * `hashInt(key.hashCode)` over rendering the key to a string.
+  *
   * Return values preserve the exact xxHash bit pattern in signed Scala `Int` and `Long` values. Callers that need unsigned formatting should
   * format the returned bits explicitly.
   */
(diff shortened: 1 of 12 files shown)


kyo-config/shared/src/main/scala/kyo/internal/XXHash.scala
```diff
@@ -67,48 +71,15 @@ private[kyo] object XXHash {
...
-            } else {
-                state.add(0xe0 | (c >>> 12))
-                state.add(0x80 | ((c >>> 6) & 0x3f))
-                state.add(0x80 | (c & 0x3f))
-            }
-            i += 1
-        }
-        state.digest()
-    }
+    def hash32(input: String): Int =
```
hearnadam, line 81: isn't this problematic for cross-jvm equality? Wouldn't it be better to just skip using XXHash for tag?

fwbrasil: I don't think so. `String.hashCode` is part of the JVM spec and stable across runs and distributions. I think we also don't store it so the hash is always calculated during execution

fwbrasil: I had initially fixed only tag but I worry we'll end up hitting this perf issue again if we don't fix it centrally

DamianReeves (approved):
LGTM

DamianReeves (approved):
LGTM

PR #1758, opened 2026-07-19
[kyo-net] add the cross-platform network transport module

fwbrasil:
## Problem

kyo-http and kyo-jsonrpc each hand-rolled their own low-level transport code (sockets, TLS, platform I/O drivers), duplicating connection lifecycle and TLS-engine logic across modules with no shared home and no reuse path for other protocols that need raw sockets.

## Solution

Add kyo-net, a cross-platform transport layer (TCP, Unix-domain sockets, stdio, TLS) over a completion-based I/O driver on every Kyo platform: Linux io_uring with an epoll fallback, macOS/BSD kqueue, the JVM NIO floor, and Node.js on JS. TLS is a single engine abstraction with a priority-ordered backend: BoringSSL, built from a pinned upstream commit and linked statically, as the primary on JVM and Native, falling back to the JDK SSLEngine on the JVM and to system OpenSSL on Native when a BoringSSL build is not staged.

kyo-http's transport moves onto kyo-net: kyo-http keeps the byte-stream and protocol layers (HTTP/1.1 framing, routing, WebSocket) and becomes a consumer of kyo-net's Connection/Transport instead of owning socket and TLS code directly. Alongside the move:

- Outbound header names and values are validated against RFC 9110 (name is a token, value carries no control character other than HTAB) before they reach the wire, closing a response-splitting/request-smuggling gap; connectRaw (used for CONNECT proxying and container exec, a peer-forwarding path) is brought under the same guard, along with the request path and derived Host header, and the WebSocket client's Sec-WebSocket-Protocol value gets the same token check the server side already applied to the echoed subprotocol.
- ConnectionPool's release path (releasingConn) now decides from the actual request outcome instead of a finalizer error argument, so a request that aborts with a typed HttpException (a rejected redirect, a non-2xx failOnError, a redirect loop) closes its connection instead of returning it to the pool; on the shared default client, which is never closed, that connection previously leaked its descriptor for the life of the process.
- HttpServer.init, initUnscoped, and their initWith variants now surface a bind failure as Abort.fail(HttpBindException) instead of throwing it as a defect, so a caller recovers it with Abort.run[HttpBindException] like any other typed failure. Callers that bind a server (kyo-caliban, kyo-flow, kyo-ai, kyo-browser, kyo-ui) are updated for the new effect row.
- This builds on the response-size cap already merged to origin/main (#1753): HttpPayloadTooLargeException's message and scaladoc are corrected to describe the response body it actually caps (it is thrown only by the client backend; the server's 413 path never uses it), and ChromeDownloader now streams its ~200 MiB archive to disk via getStreamBytes instead of buffering it through getBinary, which rejected the download once it exceeded the cap.

kyo-jsonrpc's wire transports move onto kyo-net Connections: unixDomain binds through Transport.listenUnix on every platform, replacing three separate per-platform UDS backends (a raw java.nio server on the JVM, throwing stubs on JS/Native) with one implementation that runs on JVM, JS, Native, and Wasm. contentLengthStdio becomes a cross-platform factory over Transport.stdio; the JVM-only stream-injection overload is removed since arbitrary byte-stream pairs go through the public JsonRpcWireTransport seam instead.

NetConnectionClosedException.operation is now a sealed Operation enum (Read, Send, Handshake, Upgrade, Start, Close) instead of a String, so a caller branches on a typed value rather than matching message text.

## Notes

An intermittent Scala Native crash can surface under concurrent load. It traces to an upstream data race in Scala Native's module-init publish path (a winning thread's context fields are read by a racing thread before they are fully written), not to a kyo defect. The fix is authored and open upstream as scala-native/scala-native#4992. kyo carries no workaround for it: the race lives in Scala Native's own runtime module initialization and its vulnerable set includes runtime-internal modules an application cannot reach, so no application-level mitigation closes it. It resolves when the upstream fix releases.


fwbrasil:
Kyo's throughput on mac is ~2x of the competitors and allocation rate is up to 30x smaller. I'll try to run on linux as well, uring might produce even better results! 

benchmark results: https://jmh.morethan.io/?source=https://gist.githubusercontent.com/fwbrasil/81586371c6671cc988a9095a00edd4b4/raw/16a410de61da588d6b18c222408fe26cd4295253/jmh-result.json

PR #1759, opened 2026-07-19
[kyo-system,kyo-eventlog] File/system capability and typed event-log stack

DamianReeves:
## Summary

This change introduces a **file/system capability layer** and a **typed event-log stack** as two new modules, and migrates every module that touched the filesystem off raw `java.nio`/`java.io` onto the new capability effects.

- **`kyo-system`** (new): the `Path`/`FileSystem` capability. `PathRead`/`PathWrite` effects, a `FileSystem` service SPI, and swappable backends (host, in-memory, root-confined, copy-on-write overlay, zip/archive). `kyo.Path` and `kyo.System` move here out of `kyo-core`.
- **`kyo-eventlog`** (new): a raw `Journal` capability plus a typed `EventLog` facade, durable file-backed and JVM-only H2 backends, cross-format `migrate`/`upcast`, and a decider-style `EventSourced` layer.
- **`kyo-data`**: a `FileSize` value type.
- **Callers migrated** to the capability rows: `kyo-ai`, `kyo-browser`, `kyo-doctest`, `kyo-jsonrpc`, `kyo-pod`, `kyo-tasty`, `kyo-website`, `kyo-http`.
- **`kyo-website`**: a tutorial route rail plus three eventlog tutorials.
- **`kyo-examples`**: a car-rental EventLog example.

All source and tests are cross-platform (JVM, JS, Native, Wasm) with shared tests. This branch is merged up to current `main` and reconciles the `Path` relocation with `main`'s new `kyo-stats-machine` (see *Integration with main*).

Net change: **229 files, +35,115 / -6,291**. `kyo-core` shrinks by ~4,100 lines as `Path`/`System` move out.

---

## The file/system capability (`kyo-system`)

`Path` operations are values in `PathRead` / `PathWrite` effect rows, discharged by `Path.run` (and `Scope.run` for scoped temp resources), instead of eager IO:

```scala
import kyo.*

val program: String < (PathRead & PathWrite & Sync & Scope) =
    for
        dir  <- Path.tempDir("demo")
        file  = dir / "greeting.txt"
        _    <- file.write("hello")
        text <- file.read
    yield text

Scope.run(Path.run(program))   // : String < Sync
```

A `FileSystem` service backs the capability, so the same program runs against different backends:

- **host** (real OS filesystem),
- **in-memory** (deterministic tests, no disk),
- **root-confined** (a chroot-like boundary),
- **overlay** (copy-on-write staging with a durable commit protocol, intent log, and crash recovery),
- **zip/archive** (read/write into an archive, all platforms).

The surface also carries a **positioned channel tier** (`Path.ReadHandle`/`WriteHandle`, `readChunk`/`position`), an **advisory file-lock** capability, and `transaction` / `sandbox` / virtual-path combinators over the overlay.

---

### Raw `Journal` capability

An append-only, stream-addressed log as an `ArrowEffect`, with durable backends:

```scala
for
    backend <- Journal.Backend.inMemory                  // or Journal.Backend.file(dir, configuration)
    result  <- Journal.run(backend) {
        for
            _      <- Journal.append(streamId, ExpectedOffset.NoStream, Chunk(event))
            events <- Journal.read(streamId, Event.StreamOffset.first, maxCount = 100)
        yield events
    }
yield result
```

Backends: in-memory, **FileJournal** (durable, segmented, JSONL or binary segment formats, group commit, SWMR lock-skipping reads, crash recovery), a **JVM-only H2** backend implementing the SPI directly, and a **zip** journal. `FileSize` (kyo-data) types the segment-rollover threshold; an `Fsync` enum types the durability knob.

### Typed `EventLog` facade with a fluent builder

Register each domain member once; `import log.given` then routes `append` by the runtime value with no per-call witness:

```scala
for
    log <- EventLog.builder[QuestEvent](journalId)
        .define[QuestStarted](stream = Event.StreamSelector.constant(questStream))
        .define[PartyJoined](stream = Event.StreamSelector.by(vehicleName)(e => Chunk(e.id)))
        .build                                           // .codecs(...) is optional; defaults apply
    backend <- Journal.Backend.inMemory
    events  <- Journal.run(backend) {
        import log.given
        for
            _  <- log.append(QuestStarted("q1"))         // routes to questStream
            rs <- log.read(questStream, Event.StreamOffset.first, 100)
        yield rs
    }
yield events
```

`Builder` proves route completeness at `build` through a compile-time `RouteCoverage` (a `Mirror.SumOf` iteration for enum/sealed, a union-decomposition macro for a bare union): a missing `.define` is a compile error, and two members sharing a structure name fail `build` loudly rather than silently shadowing. A value typed at the domain supertype (the natural type of a decider's folded `Chunk[A]`) dispatches by its runtime member; an unrouted append aborts a typed `PreparationFailure`, never a panic.

### Event vocabulary

A single top-level `Event` namespace: `Event.New` (submitted) and `Event.Recorded` (stored) lifecycle shapes; `Event.Id`/`Event.Type`/`Event.StreamId`/`Event.StreamOffset`/`Event.Metadata`; typed metadata attributes (`AttributeKey`/`Attribute`); and per-append override facets (`AppendDirective.expected` / `replaceStreamId` / `replaceEventId` / `withMetadata`).

### `EventSourced`: deciders

```scala
val decider = new EventSourced.Decider[Command, State, QuestEvent]:
    def initialState                       = State.empty
    def decide(s: State, c: Command)       = Chunk(...)     // < Abort[DecideFailure]
    def evolve(s: State, e: QuestEvent)    = ...            // < Abort[EvolveFailure]

for
    store <- EventSourced.EventStore.builder[QuestEvent](journalId).define[QuestStarted](...).build
    _     <- EventSourced.decide(store, streamId, command, decider)
yield ()
```

### Cross-format migration

```scala
EventLog.migrate(source, target, streams)                  // raw byte-preserving copy across media/formats
EventLog.migrateWith(reader, target, targetCodecs, streams, transform)  // rewrite once
reader.upcast[Y](f)                                        // zero-cost read-time view
```

---

## Changelog by area

**kyo-system (new, 37 files):** PathRead/PathWrite effects + Service SPI; `Path`/`System` relocated from kyo-core; host, in-memory, root-confined, overlay (durable commit, intent log, crash recovery, whiteouts, conflict resolution), and zip backends; positioned channel tier; advisory file locks; transaction/sandbox/virtual combinators; `FileSystem` rename (from `PathService`) and `CommitStrategy` (from `Disposition`).

**kyo-eventlog (new, 70 files):** raw `Journal` ArrowEffect; `EventLog.Codecs` codec authority; typed `EventLog` facade + fluent `Builder` with compile-time route coverage; `EventStore`/`EventSourced` deciders; `Event` vocabulary (New/Recorded, opaque ids, typed metadata attributes, append directives); FileJournal (segmented, JSONL/binary, group commit, SWMR reads, crash recovery) re-layered over the FileSystem capability; JVM-only H2 backend; zip journal; `Journal.Reader` + logical `JournalEntryRef` resolver; typed `migrate`/`migrateWith`/`upcast`.

**kyo-data:** `FileSize` value type with byte/decimal-unit constructors.

**Path-capability migration:** kyo-ai, kyo-browser, kyo-doctest, kyo-jsonrpc, kyo-pod, kyo-tasty, kyo-website, kyo-http discharge their filesystem IO through PathRead/PathWrite.

**kyo-website:** tutorial child-route rail (model, content loading, SSG emission, SPA routing) + three eventlog tutorials; site generation migrated to Path rows.

**kyo-examples:** car-rental EventLog example.

---

## Integration with `main`

`main` merged `kyo-stats-machine` (#1756), which reads host metrics through `Path.ReadHandle.readLong` and added `readLong` + its `AbsentLong` sentinel to kyo-core `Path`. This branch relocates `Path` to kyo-system. Reconciled by:

- grafting `readLong` + the `ReadHandle` companion (`AbsentLong`, the shared `parseLeadingLong` parser) onto kyo-system `Path`, and implementing `readLong` on the in-memory handle;
- pointing `kyo-stats-machine` at kyo-system for the relocated `Path`/`System`;
- adapting the `readLong` tests to kyo-system's effect rows and porting the shared parser tests.

The branch is a superset of current `main`; every affected module is green.

---

## Verification

Cross-platform green (JVM / JS / Native / Wasm): kyo-eventlog (+ doctest), kyo-system, kyo-stats-machine, kyo-examples. Shared tests only; no platform-specific test dodging.


kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -72,10 +72,22 @@ object Async extends AsyncPlatformSpecific:
...
+    private[kyo] def parseConcurrencyDefault(raw: Maybe[String], default: Int): Int =
+        raw.map { s =>
+            s.toIntOption.getOrElse(
+                throw new IllegalArgumentException(
+                    s"Invalid value for system property 'kyo.async.concurrency.default': '$s' (expected Int)"
+                )
+            )
+        }.getOrElse(default)
+
     val defaultConcurrency =
```
fwbrasil, line 84: Can you migrate this to `StaticFlag`? this is old code before it was introduced

kyo-core/shared/src/main/scala/kyo/KyoAppInterrupts.scala
```diff
@@ -13,7 +13,8 @@ private[kyo] trait KyoAppInterrupts:
                 promise
                     .completeDiscard(Result.panic(Interrupted(Frame.internal, s"Interrupt Signal: $signal")))
 
-        if System.live.unsafe.operatingSystem() != System.OS.Windows then
+        // Unsafe: kyo.System lives in kyo-system (which depends on kyo-core); detect Windows from java.lang.System os.name (empty on JS, where signal registration is a no-op anyway).
+        if !java.lang.System.getProperty("os.name", "").startsWith("Windows") then
```
fwbrasil, line 17: see `Platform.isWindows`

kyo-core/shared/src/main/scala/kyo/StreamCoreExtensions.scala
```diff
@@ -1124,69 +1124,6 @@ object StreamCoreExtensions:
 
     end extension
 
-    /** Shared write logic: opens a write handle via Scope, runs the body, and removes the partial file on failure. */
```
fwbrasil, line 1127: is removal intentional?

kyo-data/shared/src/main/scala/kyo/FileSize.scala
```diff
@@ -0,0 +1,282 @@
...
+  * Arithmetic saturates on overflow: addition and multiplication that would exceed `Long.MaxValue`
+  * clamp to `Long.MaxValue`; subtraction clamps to `FileSize.Zero`.
+  *
+  * The `show` method produces a human-readable string at the coarsest lossless binary unit:
+  * `"64.mib"`, `"1.gib"`, `"FileSize.Zero"`.
+  *
+  * @see
+  *   [[Duration]] for the analogous time-span type
+  */
+opaque type FileSize = Long
```
fwbrasil, line 21: Couldn't this be used for memory configs for example? Can we have a more generic name?

kyo-doctest/jvm/src/main/scala/kyo/doctest/internal/LinkValidator.scala
```diff
@@ -32,8 +32,10 @@ private[kyo] object LinkValidator:
       * touched. Cross-document anchor checks read each referenced README at most once (memoised across the call).
       */
     def validate(file: Path)(using Frame): Chunk[Doctest.Failure] < (Sync & Abort[Doctest.Error]) =
-        Abort.recover[FileReadException](e => Abort.fail(Doctest.Error.IoError(file, "read", e))) {
-            file.read.flatMap { raw =>
+        // Per-operation runner: the IoError message must carry the "read" op label; a single boundary
+        // runner at validate() entry would lose the distinction between "read" here and "exists" in validateLink.
+        Abort.recover[FileException](e => Abort.fail(Doctest.Error.IoError(file, "read", e))) {
+            Path.runReadOnly(file.read).flatMap { raw =>
```
fwbrasil, line 38: it's pre-existing but it'd be nice to avoid `flatMap` in Kyo computations here

fwbrasil:
I'll continue the review tomorrow but there's a key aspect I think we should double check. The current approach with a tracked effect has two main benefits I see:

1. Restricting computations to read only
2. If an `Isolate` isn't provided, it's not possible to fork fibers with the pending effect

The second seems only valid if the backends have concurrency restrictions. If not and we keep an `ArrowEffect`, we should provide an `Isolate` evidence.

An alternative design could be using a `Local` where the default is regular file reading and then a method to override it for a computation. If we were in pre-LLM times, I'd say we shouldn't track but it's not as clear cut to me now since LLMs don't have much trouble with this kind of idiom.

Sorry for starting this discussion again but it's an important design decision.

@hearnadam could you share your thoughts as well?

DamianReeves:
> I'll continue the review tomorrow but there's a key aspect I think we should double check. The current approach with a tracked effect has two main benefits I see:
> 
> 1. Restricting computations to read only
> 2. If an `Isolate` isn't provided, it's not possible to fork fibers with the pending effect
> 
> The second seems only valid if the backends have concurrency restrictions. If not and we keep an `ArrowEffect`, we should provide an `Isolate` evidence.
> 
> An alternative design could be using a `Local` where the default is regular file reading and then a method to override it for a computation. If we were in pre-LLM times, I'd say we shouldn't track but it's not as clear cut to me now since LLMs don't have much trouble with this kind of idiom.
> 
> Sorry for starting this discussion again but it's an important design decision.
> 
> @hearnadam could you share your thoughts as well?

Given the comments here, will work on splitting this PR to have an initial portion that is focused on kyo-system and the Path changes, and another PR with the eventlog changes.

kyo-system/shared/src/main/scala/kyo/internal/FileSystemCrc32.scala
```diff
@@ -0,0 +1,70 @@
+package kyo.internal
+
+/** Table-driven CRC32 with the reflected IEEE 802.3 polynomial (0xEDB88320). Produces bit-identical
+  * output to `java.util.zip.CRC32` for any input. Promoted from the inline copy that lived in
+  * `OverlayFileSystem.scala`'s `WriteOpLog` (staged-journal record framing); this shared copy now
```
hearnadam, line 5: We should add pass over comments to not reference historical state. Comments only need to identify the current behavior.

kyo-system/shared/src/main/scala/kyo/Path.scala
```diff
@@ -0,0 +1,1339 @@
...
+  * only after a runner discharges it. Discharge with [[Path.runReadOnly]] (read-only) or
+  * [[Path.run]] (read and write). `PathWrite <: PathRead`: a write-capable context also satisfies
+  * reads, and a read-only runner rejects write programs at the call site.
+  *
+  * @see
+  *   [[Path.run]], [[Path.runReadOnly]], [[Path.runWith]], [[Path.runReadOnlyWith]] for the runners
+  * @see
+  *   [[FileSystem]] for the pluggable backend the runners install
+  */
+sealed trait PathRead extends ArrowEffect[[A] =>> Path.Op[A], Id]
```
hearnadam, line 61: I'm not convinced a full effect is necessary. Why wouldn't a Local with explicit tracking be possible here?

We actually could do something clever and use an implicit type to represent the effect, whether it's transparent or opaque.
```
def read(path: Path)(using FileSystemMode): X < S
```

Similar to what you did with FileSystem...

DamianReeves:
Will break up this PR

PR #1761, opened 2026-07-20
[aeron] cross-platform kyo-aeron (JVM/JS/Native/Wasm) via FFI + MsgPack

fwbrasil:
## Problem

kyo-aeron was JVM-only, built directly on the `io.aeron` Java client. `Topic`, `AeronClient`, and the publish/subscribe transport had no JS, Native, or Wasm implementation.

## Solution

- **`AeronTransport`** abstracts the transport over two backends: `JvmAeronTransport` (the `io.aeron` Java client, JVM only, keeping the zero-copy `tryClaim` hot path) and `FfiAeronTransport` (the `kyo_aeron.c` shim over the Aeron C client and embedded media driver, shared by JS, Wasm, and Native). The shim is bound through kyo-ffi: koffi on JS and Wasm, which share one `js-wasm/` selector and the same koffi-on-Node backend, and `@extern` codegen on Native. All three statically link libaeron 1.50.2; the JVM uses no FFI.
- **`Topic`, `AeronClient`, and the `TopicException` hierarchy** move to `shared`, with the same public signatures, the same effect rows, and the same on-wire bytes on every platform. No `io.aeron` type appears on the public surface.
- **Serialization** moves from upickle (`Topic.AsMessage`, a upickle `ReadWriter`) to kyo-schema's MsgPack codec, so a message type now `derives Schema` and kyo's own cross-platform serialization drops the external dependency. The wire frame is an `Envelope(typeTag, Chunk[A])`, and the receiver checks `typeTag` so a subscriber rejects a mismatched on-wire type.
- **`Topic.publish`** checks the encoded size before the connectivity check, so an oversize message fails terminally with `TopicMessageTooLargeException` instead of exhausting the retry schedule and reporting a backpressure exhaustion whose advice could never help. `maxLen == 0`, the closed-publication sentinel, still falls through to the connectivity path.
- **`AeronClient.connectUnscoped`** connects a client whose lifetime the caller owns, for a long-lived session released on its own teardown path (mirrors `Fiber.initUnscoped` and `Command.spawnUnscoped`). `Topic.run`'s `Scope`-managed overloads are unchanged.
- **build.sbt**: kyo-aeron becomes a JVM/JS/Native/Wasm `crossProject` on the FFI plugin, and the JS, Native, and Wasm aggregates include it. `kyo-aeron/scripts/build-aeron.sh` stages the libaeron static archive per os-arch and the CI setup action runs it on every leg. The staged tree is gitignored, and only `aeron_driver_static` is linked since it embeds the client. The script is module-local, like `kyo-ffi/scripts`, because it stages only for this module.

kyo-compiler consumes kyo-aeron for its forked-worker IPC, so it moves with the surface. Its `Envelope`/`Request`/`Response` wire types and the `Compiler.AsMessage` alias derive a kyo-schema `Schema` instead of a upickle `ReadWriter` (the opaque `Uri` codec via `Schema.transform`, the hand-written `CompilerException` codec likewise), and its client plumbing goes from a raw `io.aeron.Aeron` to `AeronClient.connectUnscoped` / `AeronClient.connect` with `Topic.run(client)`. It keeps owning its `MediaDriver`.

## Notes

Also fixes a latent kyo-test bug the JS work surfaced: the power-assert macro eagerly stringifies every captured operand (`Recorder.render`, via `String.valueOf` then `toString`) for passing and failing asserts alike. On Scala.js a koffi `void*` pointer throws "Cannot convert object to primitive value" on string coercion, so any assert that captured an FFI handle crashed. `Recorder.render` is now total: an unrenderable value degrades to a `<unrenderable>` placeholder, bounded in length.


kyo-aeron/js-wasm/src/main/scala/kyo/internal/AeronPlatformTransport.scala
```diff
@@ -0,0 +1,71 @@
...
+private[kyo] object AeronPlatformTransport:
+
+    /** Starts an embedded media driver in `dir` and connects a client to it.
+      *
+      * `dir` must be unique per call (callers pass `Path.tempDir`), since Aeron otherwise routes
+      * every runtime through its single default directory. The `@Ffi.blocking` downcalls dispatch
+      * to a libuv worker, leaving the Node event loop free during the ~10s connect.
+      */
+    def embedded(dir: String)(using Frame): AeronRuntime < Async =
+        Sync.Unsafe.defer(Ffi.load[AeronBindings]).map { bindings =>
```
hearnadam, line 21: just noticed - should we make FFI all caps?

kyo-aeron/shared/src/main/scala/kyo/internal/FfiAeronTransport.scala
```diff
@@ -0,0 +1,175 @@
...
+        else Absent
+
+    override def injectError(errcode: Int, errmsg: String)(using AllowUnsafe): Unit =
+        bindings.testInjectError(client, errcode, errmsg)
+
+end FfiAeronTransport
+
+private[kyo] object FfiAeronTransport:
+
+    /** Matches the shim's `KYO_AERON_SLOT_INITIAL_CAP`. 64 KiB holds an unfragmented IPC message
```
hearnadam: nit: these should be the same constant then?

kyo-aeron/shared/src/main/scala/kyo/internal/AeronTransport.scala
```diff
@@ -0,0 +1,85 @@
...
+  * `freeAsync*`. `Topic.addPublicationDeadline` / `addSubscriptionDeadline` drive the loop with
+  * `Async.sleep(addBackoff)` bounded by a deadline, surfacing `TopicAddTimeoutException` on expiry.
+  *
+  * `asyncAdd*` returns `Maybe` so a closed client surfaces as `Absent` rather than an exception or
+  * a use-after-free: the FFI impl detects it via the C shim's live-bundle registry (closed client
+  * returns NULL), the JVM impl maps the `AeronException` thrown on a closed client. `Topic.publish`
+  * maps `Absent` to `TopicPublicationClosedException`; `Topic.stream` maps it to
+  * `TopicBackpressureExhaustedException` (transient, retried per the retry schedule).
+  */
+private[kyo] trait AeronTransport:
```
hearnadam:
```suggestion
private[kyo] abstract class AeronTransport:
```

kyo-aeron/js-wasm/src/main/scala/kyo/internal/AeronPlatformTransport.scala
```diff
@@ -0,0 +1,71 @@
...
+
+    /** Maps a connect failure, treating the absent-driver NULL as the only expected one and
+      * re-raising everything else as a panic so a genuine defect stays a defect.
+      */
+    private def mapConnectFailure(e: Any)(using Frame): Nothing < Abort[TopicTransportFailedException] =
+        e match
+            case n: FfiNullPointer =>
+                Abort.fail(TopicTransportFailedException(Maybe(n.getMessage).filter(_.nonEmpty).getOrElse(n.toString)))
+            case t: Throwable => Abort.panic(t)
+            case other        => Abort.panic(new RuntimeException(s"unexpected Aeron connect failure: $other"))
```
hearnadam: Should this retain the cause? is this case even possible?

kyo-aeron/jvm/src/main/scala/kyo/internal/AeronPlatformTransport.scala
```diff
@@ -0,0 +1,73 @@
...
+    def embedded(dir: String)(using Frame): AeronRuntime < Async =
+        Sync.defer {
+            val driver = MediaDriver.launchEmbedded(
+                new io.aeron.driver.MediaDriver.Context().aeronDirectoryName(dir).dirDeleteOnStart(true)
+            )
+            // The recording error handler captures fatal conductor conditions (driver timeout,
+            // conductor service timeout, buffer full) into errorSlot, which fatalError reads at the
+            // offer/poll boundary. The default handler instead spawns an exit daemon thread on
+            // DriverTimeoutException.
+            val errorSlot = new java.util.concurrent.atomic.AtomicReference[String](null)
```
hearnadam: AtomicRef.unsafe?

kyo-aeron/js-wasm/src/main/scala/kyo/internal/AeronPlatformTransport.scala
fwbrasil: yeah, I'll add a todo to follow up separately

kyo-aeron/js-wasm/src/main/scala/kyo/internal/AeronPlatformTransport.scala
fwbrasil: This is actually a rough edge in `kyo-ffi`. It declares the failure of fibers as `Any`, which doesn't make sense given that the point is tracking typed failures. It should be `Nothing` and use the panic channel. I'll follow up on this separately as well

fwbrasil: I got confused. The `Any` type param of `Fiber.Unsafe[A, Any]` is actually pending effects, not failure tracking, so there's only the panic channel already. The issue is `kyo-aeron` incorrectly using `Abort[Any]`. Fixing it

PR #1763, opened 2026-07-21
[kyo-test] add schema-based snapshot testing

DamianReeves:
# Schema-based snapshot testing

## Problem

`kyo-test/snapshot` snapshots a value only through `Render[A].asString`, a flat `toString`-shaped string with no field names and one format. Three costs follow from that:

1. **Non-deterministic fields**: a timestamp, instant, or process id makes a snapshot perpetually unstable, because nothing can normalize the value before storage.
2. **Opaque mismatches**: a failure surfaces only as a raw line diff, never as "which field changed".
3. **Silent schema drift**: a type that gains or loses a field drifts with no schema-evolution signal.

`kyo-schema` already ships the machinery to close all three gaps (`Schema[A]` derivation with pluggable `Codec`, `Modify[A]` for typed field-level normalization, `Changeset` for structural comparison), but nothing in kyo-test wired it into the snapshot path.

## Solution

A new `assertSchemaSnapshot[A]` assertion on `SnapshotTestBase[S]` (inherited by the discoverable `SnapshotTest[S]`), alongside the existing `Render`-based `assertSnapshot`, which is unchanged in behavior and stays available for types without a `Schema`.

```scala
protected inline def assertSchemaSnapshot[A](
    inline actual: A,
    inline name: String,
    inline config: SnapshotConfig[A] => SnapshotConfig[A] = identity[SnapshotConfig[A]]
)(using inline schema: Schema[A], inline frame: Frame, inline as: kyo.test.AssertScope): Unit < (S & Async & Abort[Throwable] & Scope)
```

Given a `Schema[A]` in scope, it normalizes `actual` through the config's `Modify[A]`, encodes it via the suite's `SnapshotCodec` (default `SnapshotCodec.Yaml`, a readable field-named text format), and stores the result under `${snapshotDir}/${suite}/${name}.${ext}`. Each codec has its own extension (`snap.yaml`, `snap.json`, `snap.pb`, ...), all distinct from `assertSnapshot`'s plain `.snap`, so the two assertions never collide on the same name.

### `SnapshotCodec`: any codec, text vs binary

`kyo.Codec` carries no text-versus-binary flag, so the snapshot module wraps it in a two-case enum that decides both the storage format and whether a mismatch can carry a textual diff:

```scala
enum SnapshotCodec(val codec: Codec, val ext: String):
    case Text(override val codec: Codec, override val ext: String)   extends SnapshotCodec(codec, ext)
    case Binary(override val codec: Codec, override val ext: String) extends SnapshotCodec(codec, ext)
```

Seven presets cover every codec kyo-schema ships; `Text`/`Binary` remain the open extension point for any custom `Codec`:

| Preset | Kind | Extension | Stored as |
|---|---|---|---|
| `SnapshotCodec.Yaml` (default) | Text | `snap.yaml` | UTF-8 string |
| `SnapshotCodec.Json` | Text | `snap.json` | UTF-8 string |
| `SnapshotCodec.Ion` | Text | `snap.ion` | UTF-8 string |
| `SnapshotCodec.Protobuf` | Binary | `snap.pb` | raw wire bytes |
| `SnapshotCodec.Bson` | Binary | `snap.bson` | raw wire bytes |
| `SnapshotCodec.MsgPack` | Binary | `snap.msgpack` | raw wire bytes |
| `SnapshotCodec.IonBinary` | Binary | `snap.ionb` | raw wire bytes |

Binary snapshots are the real wire artifact: stored byte-identical to codec output (no base64, no trailing-newline munging), so a `.pb` file stays inspectable with format tooling. `SnapshotStore` gains a raw-bytes sibling path to support this:

```scala
def readBytes(path: String): Maybe[Span[Byte]]
def writeBytes(path: String, content: Span[Byte]): Unit
```

with platform implementations for JVM/Native (NIO) and JS/Wasm (Node `fs`).

The format is selected per suite by overriding a hook, mirroring the existing `snapshotDir`:

```scala
protected def snapshotCodec: SnapshotCodec = SnapshotCodec.Yaml
```

### `SnapshotConfig`: the single per-call customization point

Per-call customization goes through an immutable builder passed as a lambda, so future capabilities (for example a custom comparison strategy) become additive methods on the config, never new parameters on `assertSchemaSnapshot`, and existing call sites keep compiling unchanged:

```scala
final class SnapshotConfig[A] private[snapshot] (private[snapshot] val modify: Modify[A]):
    /** Adds a field-normalization pass, scrubbing non-deterministic fields before encode and before compare. */
    def normalize(f: Modify[A] => Modify[A]): SnapshotConfig[A]

object SnapshotConfig:
    def apply[A]: SnapshotConfig[A] = new SnapshotConfig[A](Modify.apply[A])
```

Today it carries exactly one capability, `.normalize`, which composes a `Modify[A]` transform applied before both encode and compare. Multiple `.normalize` calls accumulate in order. Format selection deliberately stays a per-suite `snapshotCodec` concern, not a `SnapshotConfig` knob: the two are complementary customization surfaces at different scopes.

### Comparison semantics

On a run with an existing stored snapshot:

1. The stored bytes are decoded via `Schema[A]`.
2. **Decode succeeds**: the stored value and the normalized actual are compared structurally through `Changeset`. Structural equality passes, so re-serialization noise (a hand-reformatted YAML file that still decodes to an equal value) never fails the assertion: the compare is format-tolerant.
3. **Values differ**: the failure reports the changed field paths (nested paths dotted, e.g. `b.y`), plus a unified text diff for text codecs.
4. **Decode fails**: the type's shape has drifted from the stored snapshot (field added, removed, retyped). This fails with the distinct `SnapshotSchemaEvolution` exception, never conflated with an ordinary value mismatch. A `Result.Panic` on decode (a raw codec defect, not a schema mismatch) propagates unwrapped.

First-run and update-mode behavior matches `assertSnapshot`: the first run writes the proposed snapshot and fails with `SnapshotNotFound`; `KYO_TEST_SNAPSHOT=update` (or the `snapshotUpdateMode` override) accepts new or changed snapshots.

### Basic: readable YAML snapshot

```scala
import kyo.*
import kyo.test.*
import kyo.test.snapshot.*

case class Statement(id: Int, balanceCents: Long) derives Schema

class StatementSnapshotTest extends SnapshotTest[Any]:
    "monthly statement" in {
        assertSchemaSnapshot(Statement(1, balanceCents = 12_300L), "monthly")
    }
end StatementSnapshotTest
```

Stored at `test-snapshots/StatementSnapshotTest/monthly.snap.yaml` as field-named YAML.

### Normalizing non-deterministic fields

```scala
case class Statement(id: Int, balanceCents: Long, generatedAt: Long) derives Schema

class SchemaStatementSnapshotTest extends SnapshotTest[Any]:
    "monthly statement" in {
        val statement = Statement(1, balanceCents = 12_300L, generatedAt = java.lang.System.currentTimeMillis())
        assertSchemaSnapshot(statement, "monthly-schema", _.normalize(_.set(_.generatedAt)(0L)))
    }
end SchemaStatementSnapshotTest
```

The `generatedAt` field is scrubbed to `0L` before both encoding and comparison, so a repeated run with a fresh timestamp still matches the stored snapshot. `Modify[A]` supports `.set` and `.update` on nested fields, so `_.normalize(_.update(_.request.id)(_ => "<redacted>"))` works the same way.

### Per-suite format selection

```scala
class ProtoSnapshotTest extends SnapshotTest[Any]:
    override protected def snapshotCodec: SnapshotCodec = SnapshotCodec.Protobuf

    "wire format" in {
        assertSchemaSnapshot(Statement(1, 12_300L), "wire")
    }
end ProtoSnapshotTest
```

Stored at `test-snapshots/ProtoSnapshotTest/wire.snap.pb` as the exact Protobuf wire bytes.

### Failure output

A value mismatch names the changed fields and, for text codecs, includes a unified diff:

```
Snapshot mismatch
changed fields: balanceCents, account.owner

--- stored
+++ actual
 id: 1
-balanceCents: 12300
+balanceCents: 45600
```

A schema drift (stored snapshot no longer decodes) fails with `SnapshotSchemaEvolution` and the decode error, keeping "the type changed shape" distinct from "the value changed".

## Changelog

**Added**

- `assertSchemaSnapshot[A]` on `SnapshotTestBase[S]` / `SnapshotTest[S]`: schema-driven snapshot assertion with normalization, structural comparison, and schema-evolution detection.
- `SnapshotCodec`: `Text`/`Binary` enum wrapping any kyo-schema `Codec`, with seven presets (`Yaml` default, `Json`, `Ion`, `Protobuf`, `Bson`, `MsgPack`, `IonBinary`), each with a distinct snapshot extension.
- `SnapshotConfig[A]`: extensible per-call config builder, currently carrying `.normalize(Modify[A] => Modify[A])`.
- `SnapshotSchemaEvolution`: distinct failure for a stored snapshot that no longer decodes under the current schema.
- `snapshotCodec` per-suite override hook (default `SnapshotCodec.Yaml`), mirroring `snapshotDir`.
- `SnapshotStore.readBytes`/`writeBytes`: raw-bytes storage path for binary codecs, implemented for JVM/Native (NIO) and JS/Wasm (Node `fs`).
- Module `CONTRIBUTING.md` for `kyo-test/snapshot` and an `assertSchemaSnapshot` section in the `kyo-test` README.
- Tests: `SnapshotSchemaTest`, `SnapshotCodecTest`, `SnapshotConfigTest`, `SnapshotSchemaEvolutionTest`, `SnapshotStoreBytesTest`, plus a foreign-package compile proof (`SnapshotForeignPackageTest`), all cross-platform (JVM/JS/Native).

**Changed**

- `assertSnapshot`: behavior-preserving refactor only; its inline name validation now delegates to the shared `validateSnapshotName` helper used by both assertions.
- `build.sbt`: `kyo-test-snapshot` now depends on `kyo-schema`.
- `AssertScope.noAssertionDiagram`: mentions `assertSchemaSnapshot` alongside the existing assertions.

**Unchanged**

- The `Render`-based `assertSnapshot` path, its `.snap` extension, and its failure behavior.
- kyo-schema's public API (consumed as-is).

## Notes

- `SnapshotConfig.modify` is `private[snapshot]`; the inline `assertSchemaSnapshot` body resolves it through a non-inline in-package helper so a suite in a foreign package can still call it with a config lambda (covered by `SnapshotForeignPackageTest`).
- The config parameter defaults to `identity[SnapshotConfig[A]]` (an explicitly typed witness) rather than bare `identity`, avoiding a type-inference pitfall with defaulted function parameters on inline methods.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2914,6 +2914,7 @@ lazy val `kyo-test-snapshot` =
         .crossType(CrossType.Full)
         .dependsOn(`kyo-test-api`)
         .dependsOn(`kyo-data`)
+        .dependsOn(`kyo-schema`)
         .dependsOn(`kyo-test-runner` % Test)
         .in(file("kyo-test/snapshot"))
         .settings(
--- a/kyo-test/README.md
+++ b/kyo-test/README.md
@@ -458,7 +458,7 @@ val doubleShrinks: Chunk[Double] = Shrink.double(2.7) // 2.0 (integral neighbor)
 
 ## Write meaningful tests: no-assertion enforcement
 
-A leaf that runs to completion having evaluated zero assertions is itself a bug: it proves nothing. kyo-test enforces this at run time. After a leaf body joins, the runner counts how many assertions it evaluated; a leaf that would otherwise pass with a count of zero is flipped to `Failed` with the message `leaf passed without evaluating any assertion`. The whole assert family counts: `assert`, `fail`, `intercept`, `assertEventually`, `typeCheck`, `assertSnapshot`, and a property `forAll`. `cancel` and `assume` do not count, because they steer the outcome rather than make a claim.
+A leaf that runs to completion having evaluated zero assertions is itself a bug: it proves nothing. kyo-test enforces this at run time. After a leaf body joins, the runner counts how many assertions it evaluated; a leaf that would otherwise pass with a count of zero is flipped to `Failed` with the message `leaf passed without evaluating any assertion`. The whole assert family counts: `assert`, `fail`, `intercept`, `assertEventually`, `typeCheck`, `assertSnapshot`, `assertSchemaSnapshot`, and a property `forAll`. `cancel` and `assume` do not count, because they steer the outcome rather than make a claim.
 
 This is a run-level setting, `failOnNoAssertion`, on by default. ScalaTest forced this at compile time through its `Assertion` return type; kyo-test cannot, because a leaf body is an effectful `Unit < (S & Async & Abort[Any] & Scope)`, so the check runs after the body completes.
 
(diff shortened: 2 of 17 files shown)


DamianReeves:
Upstream PR with this change merged

PR #1764, opened 2026-07-22
[kyo-markdown] add a cross-platform Markdown-to-UI renderer

fwbrasil:
## Problem

kyo has no shared way to turn a Markdown string into renderable UI. A module that needs to show authored content (a docs page, a rendered message) has to reach for a third-party Markdown library, which pins it to one platform and pulls in a syntax highlighter that does not cross-compile.

## Solution

Add `kyo-markdown`, a cross-platform module whose single entry point, `Markdown.render(source)`, returns `Rendered(article, headings)`: a kyo-ui `UI` article tree ready to mount, plus a heading outline for a table of contents. Each `Heading` carries the same id the article's in-page anchor uses, so an outline link and its target stay in lockstep.

`render` is pure and total: a construct outside the supported grammar degrades to text rather than failing, so no call needs error handling. The grammar covers ATX headings, paragraphs, ordered and unordered lists, GFM pipe tables, fenced code, blockquotes, and an inline span set, parsed with kyo-parse combinators. There is no third-party Markdown dependency and no JVM-only highlighter, so the module builds on JVM, JS, Native, and Wasm, kyo-ui's full platform set.

## Notes

Raw HTML in the source passes through un-escaped, the trust model a CommonMark renderer applies by default; a caller rendering untrusted Markdown sanitizes the output itself.


DamianReeves (approved):
LGTM

PR #1772, opened 2026-07-23
[kyo-core] fix HalfOpen channel/queue close semantics

fwbrasil:
## Problem

Two related bugs on the `closeAwaitEmpty` (HalfOpen) path.

1. `Channel.flush` livelock. When `closeAwaitEmpty` leaves the backing queue `HalfOpen` with a parked put, a consumer that makes space drives the put-transfer branch into an infinite loop: `queue.offer` fails `Closed` for a `HalfOpen` queue, but the branch only checked `.contains(true)` and re-enqueued the parked put unchanged, so the tailrec `flush` re-evaluated the identical state forever. On single-threaded JS this freezes the event loop (dead timers, one pegged core, no allocation, no output); on the JVM it burns a carrier thread.

2. `Queue.close()` no-op on HalfOpen. `close()` only transitioned from `Open`, so a hard `close()` on a queue with a pending `closeAwaitEmpty` (for example `WritePump.teardown` closing a channel whose producer parked on a full outbound) did nothing: the await-empty promise never settled and the parked put hung forever.

## Solution

1. Match the `queue.offer` result three ways in both flush put-transfer arms. `Success(true)` completes the put, `Success(false)` re-enqueues (a genuine full queue), and a `Closed` failure fails the parked put, the same outcome the `FullyClosed` drain already delivers, at the right time. The take-transfer branch gets the same treatment: when a take is interrupted during a `HalfOpen` drain, its value is retried against the remaining takers rather than dropped through a placeholder put that the transfer arm would then fail.

2. `Queue.close()` now escalates from `HalfOpen` as well as `Open`: it CASes to `FullyClosed` and completes the pending await-empty promise with `false` (the queue did not drain to empty before this hard close), then drains the backlog, failing any parked puts and takes.

## Notes

`close()` on a queue with a pending `closeAwaitEmpty` is now a hard close, an externally visible change: `Scope`-managed channels and queues that sit in `HalfOpen` at scope exit (for example the stream extensions) now fully close and settle the await `false` instead of no-oping. Scope exit means the stream is done, so the new behavior is correct at each reviewed site.

Coverage: Channel-level regression tests for the flush livelock (single and batch put) and the hard-close escalation, a `QueueTest` for the `close()`-aborts-`closeAwaitEmpty` contract (backlog returned, await settles `false`, later ops fail `Closed`), and a looped close-vs-drain race test for the `handleHalfOpen`/escalate CAS.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Channel.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Channel.scala
@@ -224,7 +224,7 @@ object Channel:
           * closed.
           *
           * @return
-          *   true if the channel was successfully closed and emptied, false if it was already closed
+          *   true if the channel was successfully closed and emptied, false if it was already closed or a hard `close()` aborted the drain
           */
         def closeAwaitEmpty(using Frame): Boolean < Async = Sync.Unsafe.defer(self.closeAwaitEmpty().safe.get)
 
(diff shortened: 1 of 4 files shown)


kyo-core/shared/src/main/scala/kyo/Channel.scala
```diff
@@ -802,12 +802,26 @@ object Channel:
...
+                                            // Full but open: hold as a placeholder put for a later transfer.
+                                            val placeholder = Promise.Unsafe.init[Unit, Abort[Closed]]()
+                                            discard(puts.offer(Put.Value(value, placeholder)))
+                                        case _ =>
+                                            // HalfOpen/closed (a closeAwaitEmpty drain): the offer can never succeed, and a placeholder put would
+                                            // just be failed by the transfer arm, dropping the value. Retry delivery against the remaining takers.
+                                            // If none are waiting, the value's only consumer interrupted and the closing queue will not re-buffer
+                                            // it, so it is forfeited (the drain settles one element short).
+                                            @tailrec
+                                            def retryTransfer(): Unit =
```
hearnadam, line 819: Won't this retry forever if takers is never refilled?

fwbrasil: I don't think so, it stops when `takes.poll` returns `Absent`

PR #1774, opened 2026-07-23
[build] drop Scala.js incremental linker state in CI to fit the sbt heap

DamianReeves:
## Summary

Follow-up to #1769. That PR's CI showed both JS jobs (x64 + arm64) and the arm64 Wasm job dying with `java.lang.OutOfMemoryError: Java heap space` in the sbt process during the test phase: the schema split means CI now links a test binary per schema module in one 12G-heap sbt process, and the Scala.js incremental linker state retained per module overflowed it. #1769 was merged before this fix landed on its branch, so main is exposed to the same OOM.

The fix enables the Scala.js linker's batch mode — which drops incremental state after each link — only when `insideCI`, for both `js-settings` and `wasm-settings`. Local builds keep incremental linking; CI trades irrelevant relink speed for a bounded footprint.

## Test Plan

- [x] `sbt "show kyo-schemaJS/scalaJSLinkerConfig"` loads and shows `batchMode = false` locally (CI-only activation)
- [ ] This PR's JS/Wasm CI jobs complete without heap exhaustion

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2759,7 +2759,15 @@ lazy val `js-settings` = Seq(
     bspEnabled                                  := false,
     Test / parallelExecution                    := false,
     jsEnv                                       := new NodeJSEnv(NodeJSEnv.Config().withArgs(List("--max_old_space_size=5120"))),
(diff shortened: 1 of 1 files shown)


DamianReeves:
Abandoning

PR #1775, opened 2026-07-24
[kyo-i18n] add a cross-platform internationalization module

xsistens:
## Problem

kyo has no shared way to translate an application. A module that needs localized text has to reach for a third-party i18n runtime, which pins it to one platform (the common Fluent runtime is JavaScript-only) and threads a translator dependency through every call site that renders a string. Nothing ties the active locale to kyo-ui's reactivity, so a language switch means re-rendering by hand.

## Solution

Add `kyo-i18n`, a cross-platform module that parses and formats [Fluent](https://projectfluent.org/) (`.ftl`) bundles in pure Scala and depends only on `kyo-core`. It builds on JVM, JS, Native, and Wasm, with no third-party Fluent runtime.

The active locale is a `Signal`, and `I18n.t(key, args)` returns a bare `Signal[String]` that a UI layer lifts into a reactive text node: switching the locale re-renders exactly the affected leaves and nothing else. `I18n.now` and `I18n.at` are the `< Sync` point-in-time counterparts for values that must not re-translate (toasts, log lines, exception text). Interpolation covers `{ $name }` placeholders; an argument that is itself a `Signal[String]` (typically a nested `t`) keeps the composed message reactive to that argument too, and the `i18n"..."` interpolator composes literals and arguments into one reactive signal.

A handle is built once by `I18n.init` or `I18n.initInMemory`, then made ambient with `I18n.let`. `init` loads each locale's text through a caller-supplied function, so a platform-specific source (an HTTP fetch, a classpath resource) stays in the caller; passing a directory in place of that function reads `<locale>.ftl` from it instead, for the JVM, Native, and Node. Everything else reads the handle back through a `Local`, so a translated value reads like a plain call: no `using` parameter, no pending effect on the reactive path, no global mutable state. The `Local`'s default is an empty handle, so a translation read before `init` renders a visible miss marker rather than throwing.

A `Locale` is a normalized primary language subtag (`de-DE` and `DE` both become `de`), so a config value, a stored preference, and a `navigator.language` prefix all compare equal. `Locale.default` reads the host environment (`navigator.language` in a browser, otherwise `user.language` or `LC_ALL`/`LANG`) and `Locale.preferred` narrows that to the locales an application actually has bundles for, so a start locale can honor the reader without ever naming one that would render nothing. A host that names no language yields `Locale.Undetermined`, the BCP-47 `und` subtag. The module fixes no locale set: the consumer decides which locales exist by which bundles it loads.

## Notes

The directory overload is covered by a `jvm-native` test rather than the shared suite: reaching `Path` from the Scala.js test bundle would pull in the `node:fs` module import, and those tests link as `NoModule`.

A key that is missing, or present but value-less, renders as `‹key›` rather than falling back to another locale or an empty string, so a gap in a bundle is visible in the output. The parser accepts the flat, interpolation-only subset of Fluent (`id = pattern` messages, indented continuation lines, `#` comments); select expressions, message attributes, and term references are not parsed. A line outside the subset is ignored but reported through `kyo.Log` at warn level, so a bundle written against full Fluent shows up in the log rather than as mysteriously absent text; the two specified behaviors stay quiet, namely a value-less message being absent and a non-placeholder brace staying literal. A `Select` AST node is reserved so plural selection, when first needed, is an internal parser addition rather than a change to any message that already parses.

### DIFF
--- a/kyo-i18n/README.md
+++ b/kyo-i18n/README.md
@@ -1,0 +1,186 @@
+# kyo-i18n
+
+Translate an application from `.ftl` bundles, with the active locale held as a `Signal` so a language switch re-renders exactly the affected text and nothing else. Bundles are parsed and formatted in pure Scala, so the module works across JVM, JavaScript, Scala Native, and WebAssembly and depends only on `kyo-core`. Translations read like plain calls: the resolved handle is provided once at the application root and read back ambiently, so no dependency is threaded through call sites.
+
+<!-- doctest:setup
+```scala
+val demoBundles = Dict(
+    Locale("en") -> "greet = Hello, { $name }!",
+    Locale("de") -> "greet = Hallo, { $name }!"
+)
+```
+-->
+
+```scala
+I18n.initInMemory(demoBundles, Locale("en")).map { i18n =>
+    I18n.let(i18n) {
+        val greeting: Signal[String] = I18n.t("greet", Dict("name" -> "Sam"))
+        greeting.current
+    }
+}
+// "Hello, Sam!"; set the locale to Locale("de") and every t leaf re-renders to German
+```
+
+`I18n.t` is the center: it returns a bare `Signal[String]` that a UI layer lifts into a reactive text node, so switching the locale patches the affected leaves in place. `I18n.now` and `I18n.at` are the point-in-time counterparts for values that must not re-translate. A handle built by `I18n.init` is made ambient with `I18n.let`; everything else reads it back.
+
+## Installation
+
+Add the dependency to your `build.sbt`:
+
+```scala doctest:expect=skipped
+libraryDependencies += "io.getkyo" %% "kyo-i18n" % "<latest version>"
+```
+
+All public types live in the `kyo` package:
+
+```scala
+import kyo.*
+```
+
+**Note:** the effectful methods take an implicit `kyo.Frame` used for source-location reporting. The signatures shown here omit `(using Frame)` for readability; the compiler synthesizes it at each call site.
+
+## Locales
+
+A `Locale` is a normalized primary language subtag. `apply` and `parse` lowercase the input and drop any region or script, so a value from a config file, a stored preference, and a browser `navigator.language` all compare equal:
+
+```scala doctest:scope=nested
+Locale.parse("de-DE") // Present(Locale("de"))
+Locale.parse("EN")    // Present(Locale("en"))
+Locale.parse("")      // Absent
+
+Locale("de").code // "de"
+```
+
+`Locale.default` reads the host environment: `navigator.language` in a browser, otherwise the `user.language` property or `LC_ALL`/`LANG`. It reports whatever the host prefers, which need not be a language the application has a bundle for, so `Locale.preferred` picks the host locale only when it is one of the supported set and falls back otherwise:
+
+```scala doctest:scope=nested
+val supported = Seq(Locale("en"), Locale("de"))
+
+Locale.preferred(supported, Locale("en"))
+// the host locale if it is en or de, otherwise en
+```
+
+When the host names no language at all, `Locale.default` is `Locale.Undetermined`, the BCP-47 `und` subtag.
+
+The module fixes no locale set. The consumer decides which locales exist by which bundles it loads and which start locale it passes to `init`.
+
+## Providing translations
+
+`initInMemory` builds a handle from bundle text you already hold:
+
+```scala doctest:scope=nested
+val bundles = Dict(
+    Locale("en") -> "hello = Hello",
+    Locale("de") -> "hello = Hallo"
+)
+
+I18n.initInMemory(bundles, Locale("en")).map { i18n =>
+    I18n.let(i18n)(I18n.now("hello"))
+}
+// "Hello"
+```
+
+`init` is the usual choice for an application: it loads each locale's text through a function you supply, so the source of the bundles stays in your code. On the client that loader is typically an HTTP fetch, on the server a classpath resource; its effects flow into the result:
+
+```scala doctest:scope=nested
+val supported = Seq(Locale("en"), Locale("de"))
+
+def loadBundle(locale: Locale): String < Async =
+    Async.defer(s"hello = Hello from ${locale.code}")
+
+I18n.init(supported, Locale("en"))(loadBundle).map { i18n =>
+    I18n.let(i18n)(I18n.now("hello"))
+}
+// "Hello from en"
+```
+
+Pair it with `Locale.preferred` to start in the reader's own language whenever a bundle exists for it:
+
+```scala doctest:expect=skipped
+for
+    start <- Locale.preferred(supported, Locale("en"))
+    i18n  <- I18n.init(supported, start)(loadBundle)
+yield i18n
+end for
+```
+
+On the JVM, Scala Native, and Node the bundles usually sit on disk, so `init` also takes a directory in place of the loader and reads `<locale>.ftl` from it:
+
+```scala doctest:expect=skipped
+for
+    start <- Locale.preferred(supported, Locale("en"))
+    i18n  <- I18n.init(supported, start)(Path("resources") / "i18n")
+yield i18n
+end for
+// reads resources/i18n/en.ftl and resources/i18n/de.ftl
+```
+
+That overload carries `Abort[FileReadException]` for the reads. A browser build never links it, since nothing there calls it.
+
+`let` provides the handle as the ambient source for its body. A `t` leaf built anywhere resolves against whichever handle is ambient when it is sampled; a leaf sampled outside any `let` renders miss markers.
+
+## Translating
+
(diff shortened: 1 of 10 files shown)


kyo-i18n/shared/src/main/scala/kyo/internal/Ftl.scala
```diff
@@ -0,0 +1,98 @@
...
+        parts.foreach {
+            case Part.Literal(text) => sb.append(text)
+            case Part.VarRef(name)  => sb.append(args.getOrElse(name, s"{$$$name}"))
+        }
+        sb.toString
+    end render
+end Message
+
+/** A parsed translation bundle: message id to [[Message]]. One bundle holds one locale's translations. */
+final private[kyo] case class Bundle(messages: Map[String, Message]):
```
fwbrasil: please prefer `kyo.Dict` over `Map`

kyo-i18n/shared/src/main/scala/kyo/internal/Ftl.scala
```diff
@@ -0,0 +1,98 @@
...
+
+    final private case class St(key: Maybe[String], buf: String, done: Map[String, String])
+
+    def parse(ftl: String): Map[String, Message] =
+        // Value-less message (blank after `=`) is absent, not empty (Fluent semantics): lookup renders the miss marker.
+        def flush(st: St): Map[String, String] =
+            st.key match
+                case Present(k) if st.buf.trim.nonEmpty => st.done.updated(k, st.buf)
+                case _                                  => st.done
+        val end = ftl.linesIterator.foldLeft(St(Absent, "", Map.empty)) { (st, line) =>
```
fwbrasil: This seems to discard content with the wrong format. Can you add `kyo.Log.warn` so it isn't a silent failure?

kyo-i18n/shared/src/main/scala/kyo/I18n.scala
```diff
@@ -0,0 +1,184 @@
...
+        bundles.get(loc).fold(Absent: Maybe[kyo.internal.Message])(_.get(key)) match
+            case Present(message) => message.render(I18n.coerce(args))
+            case Absent           => s"${I18n.MissOpen}$key${I18n.MissClose}"
+
+    private[kyo] def leaf(key: String, args: I18n.Args)(using Frame): Signal[String] =
+        val sigArgs = args.iterator.collect { case (name, sig: Signal[String] @unchecked) => name -> sig }.toList
+        if sigArgs.isEmpty then locale.map(loc => lookup(loc, key, args))
+        else
+            val inputs: Seq[Signal[I18n.Sample]] =
+                locale.map(I18n.Sample.Loc(_)) +: sigArgs.map((_, sig) => sig.map(I18n.Sample.Str(_)))
```
fwbrasil: use `Chunk` for this please

kyo-i18n/shared/src/main/scala/kyo/I18n.scala
```diff
@@ -0,0 +1,184 @@
...
+
+    private[kyo] enum Sample derives CanEqual:
+        case Loc(locale: Locale)
+        case Str(value: String)
+
+    private[kyo] val MissOpen  = '‹'
+    private[kyo] val MissClose = '›'
+
+    private val i18nLocal: Local[I18n] =
+        Local.init(new I18n(Map.empty, Signal.initConst(Locale("und"))(using Frame.internal), _ => ()))
```
fwbrasil: what's `und`? could we have a more sensible default based on the environment?

kyo-i18n/shared/src/main/scala/kyo/I18n.scala
```diff
@@ -0,0 +1,184 @@
...
+    /** Provides `handle` as the ambient translation source for `v`. Every [[t]] leaf built inside (whenever it
+      * is sampled) resolves against it; leaves sampled outside any `let` render miss markers.
+      */
+    def let[A, S](handle: I18n)(v: A < S)(using Frame): A < S =
+        i18nLocal.let(handle)(v)
+
+    /** Builds a handle from in-memory bundle text, keyed by locale, starting at `start`. The fetch-free
+      * counterpart to [[init]], for tests and for callers that already hold the bundle sources.
+      */
+    def initInMemory(bundles: Map[Locale, String], start: Locale)(using Frame): I18n < Sync =
```
fwbrasil: `Dict`

kyo-i18n/shared/src/test/scala/kyo/I18nTest.scala
```diff
@@ -0,0 +1,136 @@
...
+                }
+            yield assert(result == ("Hello", "Hallo"))
+        }
+        "a forked next wakes on a locale switch" in {
+            for
+                h <- I18n.initInMemory(bundles, en)
+                result <- I18n.let(h) {
+                    for
+                        f <- Fiber.initUnscoped(I18n.t("hello").next)
+                        _ <- Async.sleep(100.millis)
```
fwbrasil: Please avoid sleeps in tests, CI hosts are quite slow and this can become flaky. Rely on `assertEventually` instead

kyo-i18n/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+I18n.initInMemory(bundles, Locale("en")).map { i18n =>
+    I18n.let(i18n)(I18n.now("hello"))
+}
+// "Hello"
+```
+
+`let` provides the handle as the ambient source for its body. A `t` leaf built anywhere resolves against whichever handle is ambient when it is sampled; a leaf sampled outside any `let` renders miss markers.
+
+## Translating
```
fwbrasil, line 122: Can we have examples with the init methods that includes the load function? I imagine that'll be the most common use. Also, should we have a convenience init that uses `Path`?

fwbrasil:
Thank you for this! I'll actually need it as well :)

kyo-i18n/shared/src/main/scala/kyo/internal/Ftl.scala
xsistens: Done. `Dict.get` returns `Maybe` directly, so `Bundle.get`'s `fold(Absent)(Present(_))` collapsed to a one-liner, and the same wrapping in `I18n.lookup` went with it.

kyo-i18n/shared/src/main/scala/kyo/internal/Ftl.scala
xsistens: Done. The parser now collects the lines it can't handle and emits one `Log.warn` per bundle listing them with line numbers, rather than one warning per line.

Two cases stay quiet because they are specified rather than accidental: a value-less message is absent per Fluent semantics, and a brace that is not `{ $var }` stays literal by design. Happy to report those too if you disagree.

kyo-i18n/shared/src/main/scala/kyo/I18n.scala
xsistens: Done, it is a `Chunk` now.

kyo-i18n/shared/src/main/scala/kyo/I18n.scala
xsistens: `und` is the registered "undetermined" subtag from the IANA Language Subtag Registry that BCP 47 draws from, with `Scope: special`. So it is a valid language tag rather than a placeholder, which is what let it stand where a `Locale` was structurally required but none had been established. It is now named `Locale.Undetermined` and documented as such.

This particular occurrence is gone, though: it guarded a `samples.head` case that could not happen, so I dropped the `Sample` enum and combine the locale with the argument signals directly.

And yes to the environment default. `Locale.default` reads `navigator.language`, else the `user.language` property, else `LC_ALL`/`LANG`. Since the host may prefer a language you have no bundle for, `Locale.preferred(supported, fallback)` picks it only when one exists.

kyo-i18n/shared/src/main/scala/kyo/I18n.scala
xsistens: Done. I took `Args` with it, so it reads `I18n.t("greet", Dict("name" -> "Sam"))`. Leaving one `Map` in the public API seemed worse than changing both, but I can revert that half if you meant only the bundles.

kyo-i18n/shared/src/test/scala/kyo/I18nTest.scala
xsistens: Done, `assertEventually` on the locale ref's waiters. The test builds its handle over a ref it owns so it can observe them, rather than exposing the ref on `I18n` just for the test.

kyo-i18n/README.md
xsistens: Both added. The README shows `init` with a loader function, and passing a directory in place of that function reads `<locale>.ftl` from it: `I18n.init(locales, start)(Path("resources") / "i18n")`. Same call shape either way.

The directory form is covered by a jvm-native test rather than the shared suite: reaching `Path` from the Scala.js test bundle pulls in the `node:fs` import, and those tests link as `NoModule`.

PR #1776, opened 2026-07-24
[kyo-prelude] add Abort.tap / Abort.tapError

xsistens:
## Problem

`Abort` has a consuming recovery family (`recover`, `recoverError`, `recoverOrThrow`, `fold`) and a success-side `tap` (`Stream.tap`, `Pipe.tap`, `KyoCombinators.tap`), but no way to observe a failure and let it keep propagating. Code that wants to log or record a failure on its way out has to `recover` and re-raise it by hand, which reads as handling the error even when it is only being observed.

## Solution

Add `Abort.tap` and `Abort.tapError` on the `Abort` companion, next to `recover`/`recoverError`.

`tap` runs a side effect when the computation fails with an `E`, then re-raises the same failure; success values and panics pass through without evaluating the observer. `tapError` additionally observes panics, receiving the full `Result.Error` (`Failure(e)` or `Panic(t)`). Both leave the error in the effect row, so an enclosing handler still decides its fate. The callback returns `Any < S`, matching the existing success-side `tap`, and its result is discarded.

## Notes

Each combinator delegates to its consuming sibling with a re-raise (`recover(e => onFail(e).andThen(Abort.fail(e)))` / `recoverError(err => onError(err).andThen(Abort.error(err)))`), so it shares the exact handler semantics of the pair: `E` is matched via `ConcreteTag`, and for `tap` the panic channel is preserved untouched. The tests thread the observed value through `Var` so the side effect stays inside the effect row.

### DIFF
--- a/kyo-prelude/README.md
+++ b/kyo-prelude/README.md
@@ -114,6 +114,24 @@ assert(Abort.run(partial).eval == Result.succeed(Result.succeed(1)))
 
 When you want recovery only for declared domain errors and want unexpected panics to escape, use `recover` (leaves `Abort[Nothing]`) or `recoverOrThrow` (throws). When you want to consume the whole result including unexpected exceptions, use `recover(onFail, onPanic)` or `fold(onSuccess, onFail, onPanic)`. The single-handler variants always leave panic in some form; the multi-handler variants consume it. `Abort.recoverError` recovers from any `Error[E]` (both `Failure` and `Panic`) with a single handler. `Abort.foldError` folds over an `Error[E]` alongside a success branch. `Abort.foldOrThrow` folds success and typed failures while rethrowing panics as exceptions.
 
+### Observing a failure without consuming it
+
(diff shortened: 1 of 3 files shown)


DamianReeves:
You should add README changes for the containing module so that our docs remain up to date.

xsistens:
I will do that later today. Thanks for the reminder. I have to check how documentation is done for available modules first.

kyo-prelude/shared/src/test/scala/kyo/AbortTapTest.scala
```diff
@@ -0,0 +1,71 @@
+package kyo
+
+import kyo.*
+
+/** Tests for the non-consuming `Abort.tap`/`Abort.tapError` pair: the side effect fires on the matching error, the error is
+  * re-raised unchanged (still observable by an enclosing handler), and the success path never evaluates the observer. The
+  * observed value is threaded through a `Var` so the side effect stays inside the effect row.
+  */
+class AbortTapTest extends kyo.test.Test[Any]:
```
fwbrasil: Can you move this to `AbortTest`?

xsistens: Sure, thanks for taking a look at it!

DamianReeves:
Can you rebase thsi?

DamianReeves (approved):
LGTM

xsistens:
Rebase done.

PR #1777, opened 2026-07-24
Add cross-platform UUID capability

DamianReeves:
# Add cross-platform UUID capability

## Summary

This PR introduces a first-class, cross-platform UUID capability across `kyo-data`, `kyo-schema`, and `kyo-core`.

It adds:

- An opaque `kyo.UUID` value type with parsing, formatting, inspection, ordering, and byte conversion.
- Deterministic RFC version 5 and Kyo-profile version 8 construction.
- Secure effectful RFC version 4 generation.
- Monotonic RFC version 7 generation.
- Dynamically scoped generation through `UUID.let`.
- Canonical text generation through `UUID.v4String`.
- JSON and MsgPack schema integration.
- System property and environment parsing.
- Secure entropy implementations for JVM, JavaScript, Native, and Wasm.
- Scoped UUID generation for HTTP multipart boundaries and container staging paths.
- Updated module documentation for the UUID value, schema support, secure generation,
  HTTP integration, and container staging behavior.

This PR intentionally preserves `Random.uuid`, `Random.Unsafe.uuid()`, their behavior, and all existing call sites. Any migration away from that compatibility surface is deferred to a separate change if requested by maintainers and the community.

## Motivation

Kyo previously represented UUIDs primarily through `java.util.UUID` or UUID-formatted strings. UUID generation also appeared in APIs whose capability did not communicate the required security or time-ordering guarantees.

The new design separates UUID responsibilities by layer:

| Module | Responsibility |
| --- | --- |
| `kyo-data` | Pure UUID value, parsing, formatting, inspection, and deterministic algorithms |
| `kyo-schema` | Schema and codec integration |
| `kyo-core` | Secure effectful generation and dynamic scoping |

This preserves module dependency direction while providing a Kyo-native public API.

### Pure UUID values

```scala
import kyo.*

val parsed: Result[UUID.InvalidUUID, UUID] =
    UUID.parse("550e8400-e29b-41d4-a716-446655440000")

val urn: Result[UUID.InvalidUUID, UUID] =
    UUID.parseUrn("urn:uuid:550e8400-e29b-41d4-a716-446655440000")

val nil: UUID = UUID.nil
val max: UUID = UUID.max
```

Canonical parsing accepts uppercase or lowercase hexadecimal digits, but requires the exact `8-4-4-4-12` layout. The canonical representation is always lowercase:

```scala
val id = UUID
    .parse("550E8400-E29B-41D4-A716-446655440000")
    .getOrThrow

assert(id.show == "550e8400-e29b-41d4-a716-446655440000")
```

Malformed input produces a typed failure:

```scala
UUID.parse("not-a-uuid") match
    case Result.Failure(error: UUID.InvalidUUID) =>
        println(error.problem.show)
    case _ =>
        ()
```

### Byte conversion

UUID bytes use network byte order:

```scala
val id    = UUID.parse("550e8400-e29b-41d4-a716-446655440000").getOrThrow
val bytes = id.bytes

assert(bytes.size == 16)
assert(UUID.fromBytes(bytes).getOrThrow == id)
```

`UUID.fromBytes` requires exactly 16 bytes and returns `UUID.InvalidUUID` for any other length.

### Inspection and ordering

```scala
val id = UUID.parse("01890f3e-7c00-7000-8000-000000000001").getOrThrow

assert(id.version == 7)
assert(id.variant == UUID.Variant.RFC)
assert(id.unixTimestampMillis == Present(1689955200000L))
```

UUID ordering uses unsigned bytewise comparison across all 128 bits:

```scala
val first  = UUID.parse("00000000-0000-0000-0000-000000000001").getOrThrow
val second = UUID.parse("80000000-0000-0000-0000-000000000000").getOrThrow

assert(first.compare(second) < 0)
assert(summon[Ordering[UUID]].compare(first, second) < 0)
```

### Deterministic UUIDs

RFC version 5 UUIDs use SHA-1 as required by the standard:

```scala
import java.nio.charset.StandardCharsets

val dnsNamespace =
    UUID.parse("6ba7b810-9dad-11d1-80b4-00c04fd430c8").getOrThrow

val name =
    Span.from("www.example.com".getBytes(StandardCharsets.UTF_8))

val id = UUID.v5(dnsNamespace, name)

assert(id.show == "2ed6657d-e927-568b-95e1-2665a8aea6a2")
assert(id.version == 5)
assert(id.variant == UUID.Variant.RFC)
```

The Kyo version 8 profile uses a versioned domain marker and SHA-256:

```scala
val namespace =
    UUID.parse("6ba7b810-9dad-11d1-80b4-00c04fd430c8").getOrThrow

val name =
    Span.from("customer-123".getBytes(StandardCharsets.UTF_8))

val id = UUID.v8Sha256(namespace, name)

assert(id.version == 8)
assert(id.variant == UUID.Variant.RFC)
```

Equal namespace and name inputs always produce the same UUID.

### Version 4

```scala
import kyo.*

val randomId: UUID < Sync =
    UUID.v4
```

When canonical text is needed directly:

```scala
val randomIdText: String < Sync =
    UUID.v4String
```

`UUID.v4String` delegates to the dynamically scoped generator and renders the resulting UUID using its canonical lowercase representation.

### Version 7

```scala
val timeOrderedId: UUID < Sync =
    UUID.v7
```

Each `UUIDGenerator` instance maintains its own atomic version 7 state.

When the clock repeats or moves backward, the generator:

1. Retains the last effective millisecond.
2. Increments the 74-bit payload.
3. Preserves strict UUID ordering.

When the payload is exhausted, it advances the logical millisecond and samples a new secure payload. It never wraps to an older UUID.

### Dynamic scoping

Applications can install a custom generator without threading it through every call:

```scala
def withGenerator[A, S](
    generator: UUIDGenerator
)(
    value: A < S
)(using Frame): A < (S & Sync) =
    UUID.let(generator)(value)
```

For example:

```scala
val program: (UUID, UUID) < Sync =
    for
        first  <- UUID.v4
        second <- UUID.v7
    yield (first, second)

val scoped: (UUID, UUID) < Sync =
    UUID.let(customGenerator)(program)
```

Nested scopes restore the enclosing generator, and concurrent scopes remain isolated.

The capability-first entry points remain available:

```scala
val v4: UUID < Sync = UUIDGenerator.v4
val v7: UUID < Sync = UUIDGenerator.v7

val scoped =
    UUIDGenerator.let(customGenerator)(UUIDGenerator.v4)
```

## Platform entropy

`UUIDGenerator.live` uses platform cryptographic entropy only.

| Platform | Entropy source |
| --- | --- |
| JVM | Lazily initialized `java.security.SecureRandom` |
| JavaScript | `globalThis.crypto.getRandomValues` |
| Wasm | `globalThis.crypto.getRandomValues` |
| Native POSIX | `/dev/urandom`, with complete reads and deterministic cleanup |
| Native Windows | `BCryptGenRandom` through a small native bridge |

Entropy failures remain `Sync` panics. The generator never falls back to:

- `Random`
- timestamp-only UUIDs
- process counters
- weak pseudo-random generators

The Windows bridge reports the exact `NTSTATUS` on failure. POSIX reads close their source exactly once and preserve both primary and cleanup failures.

## SHA implementation

The deterministic algorithms require SHA-1 for RFC version 5 and SHA-256 for the Kyo version 8 profile.

Scala.js and Scala Native do not provide a common portable digest API suitable for the shared `kyo-data` layer. This PR therefore adds private, pure SHA-1 and SHA-256 implementations in `kyo-data`.

These implementations:

- Are internal and not exposed as general-purpose cryptography APIs.
- Are deterministic and cross-platform.
- Support incremental chunk processing.
- Are tested against published vectors and JVM `MessageDigest`.
- Do not participate in random entropy generation.
- Use SHA-1 only where the UUID version 5 standard requires it.

The previous HTTP-local SHA-1 implementation is removed in favor of the shared internal implementation.

### Schema instances

```scala
val kyoSchema: Schema[UUID] =
    Schema.uuidSchema

val javaSchema: Schema[java.util.UUID] =
    Schema.javaUuidSchema
```

Implicit resolution works for both types:

```scala
val kyoSchema  = summon[Schema[UUID]]
val javaSchema = summon[Schema[java.util.UUID]]
```

`Schema[UUID]` is represented as a string primitive and uses canonical lowercase text.

### JSON

```scala
val id =
    UUID.parse("550e8400-e29b-41d4-a716-446655440000").getOrThrow

val encoded: String =
    Json.encode(id)

assert(encoded == "\"550e8400-e29b-41d4-a716-446655440000\"")

val decoded: Result[Throwable, UUID] =
    Json.decode[UUID](encoded)
```

JSON decoding preserves typed UUID parse failures for malformed hexadecimal characters, braces, URN input, missing separators, and noncanonical unhyphenated input.

### MsgPack

```scala
val id =
    UUID.parse("fedcba98-7654-4321-8fed-cba987654321").getOrThrow

val bytes   = MsgPack.encode(id)
val decoded = MsgPack.decode[UUID](bytes).getOrThrow

assert(decoded == id)
```

Tests verify that all 128 bits, including high-bit values, survive round trips. The existing `java.util.UUID` codec behavior remains supported.

## System parsing

`kyo.UUID` can be read directly from environment variables and system properties:

```scala
val serviceId: Maybe[UUID] < (Sync & Abort[UUID.InvalidUUID]) =
    System.property[UUID]("service.id")

val deploymentId: Maybe[UUID] < (Sync & Abort[UUID.InvalidUUID]) =
    System.env[UUID]("DEPLOYMENT_ID")
```

Malformed values raise the typed `UUID.InvalidUUID` error.

## HTTP integration

New HTTP multipart boundaries are generated through the scoped secure UUID capability.

This applies consistently to:

- buffered requests
- streaming requests
- buffered responses
- streaming responses
- direct client execution
- pooled client execution
- redirected requests

Existing valid multipart boundaries are preserved. Missing or invalid generated boundaries are replaced consistently in both the header and body.

Boundary parsing now handles:

- case-insensitive parameter names
- reordered MIME parameters
- quoted boundary values
- quoted semicolons
- RFC 2045 token validation for unquoted values
- RFC 2046 boundary character and length constraints
- the 70-character boundary limit
- invalid trailing spaces

This prevents header and body boundary disagreement while keeping request and response decoding symmetric.

## Container staging integration

Temporary container-copy staging paths now use:

```scala
UUID.v4String
```

The staging directory remains scope-managed and is removed safely after use.

Regression coverage verifies that a generated staging directory cannot overwrite or remove the exact legacy staging path.

### `Random.uuid` remains unchanged

This PR does not remove, deprecate, or reroute:

```scala
Random.uuid
random.uuid
Random.Unsafe.uuid()
```

Existing call sites continue to use the existing dynamically scoped `Random` capability.

`Random` remains explicitly documented as non-cryptographic. Callers that require secure UUID generation can opt into:

```scala
UUID.v4
UUID.v4String
UUID.v7
```

A possible migration away from `Random.uuid` is intentionally left for a separate PR following maintainer and community feedback.

### Java UUID compatibility

The implicit schema and codec path for `java.util.UUID` remains available.

One explicit schema-name change should be noted:

```scala
// kyo.UUID
Schema.uuidSchema

// java.util.UUID
Schema.javaUuidSchema
```

Code relying only on `summon[Schema[java.util.UUID]]` is unaffected. Code explicitly referencing the previous Java UUID given by name should use `Schema.javaUuidSchema`.

## Error model

| Operation | Failure model |
| --- | --- |
| `UUID.parse` | `Result[UUID.InvalidUUID, UUID]` |
| `UUID.parseUrn` | `Result[UUID.InvalidUUID, UUID]` |
| `UUID.fromBytes` | `Result[UUID.InvalidUUID, UUID]` |
| JSON UUID decoding | Typed codec failure retaining the UUID parse error |
| Secure generation | Untracked panic in `Sync` if entropy or clock access fails |
| Version 7 exhaustion | Untracked panic rather than timestamp or payload wraparound |

Generation does not silently recover through weaker randomness.

## Testing

The implementation followed red, green, refactor TDD cycles. Test names describe behavior rather than requirement identifiers.

### JVM

- Full `kyo-coreJVM/test`
- Full `kyo-flowJVM/test`
- Full `kyo-httpJVM/test`
- Full `kyo-podJVM/test`
- `kyo-dataJVM/test`
- `kyo-schema-jsonJVM/test`
- `kyo-schema-msgpackJVM/test`
- `kyo-doctestJVM/test`


kyo-core/shared/src/main/scala/kyo/UUIDGenerator.scala
```diff
@@ -0,0 +1,210 @@
...
+
+    /** Generates a monotonic RFC version 7 UUID. */
+    def v7(using Frame): UUID < Sync
+end UUIDGenerator
+
+object UUIDGenerator:
+
+    /** The default generator backed by the platform secure entropy source and the dynamically scoped Kyo clock. */
+    val live: UUIDGenerator =
+        make(Clock.now.map(_.toJava.toEpochMilli), UUIDEntropyPlatform.live)
```
hearnadam: I think `nowWith` exists?

kyo-core/shared/src/main/scala/kyo/UUIDGenerator.scala
```diff
@@ -0,0 +1,210 @@
...
+
+    private def make(clockMillis: Frame ?=> Long < Sync, entropy: UUIDEntropyPlatform): TestControl =
+        // Unsafe: construction creates the generator's private atomic state before any effectful operation can observe it.
+        import AllowUnsafe.embrace.danger
+        new Default(clockMillis, entropy, AtomicRef.Unsafe.init[V7State](V7State.Empty).safe)
+    end make
+
+    private def makeFinite(clockMillis: Chunk[Long], entropy: Chunk[Byte]): TestControl =
+        val clockIndex   = new AtomicInteger(0)
+        val entropyIndex = new AtomicInteger(0)
```
hearnadam: Should we use the Kyo version unsafely here?

kyo-data/shared/src/main/scala/kyo/UUID.scala
```diff
@@ -0,0 +1,322 @@
...
+/** A universally unique identifier as specified by RFC 9562 (which obsoletes RFC 4122).
+  *
+  * A UUID is a 128-bit value, stored here as a pair of `Long`s (the most and least significant 64 bits, matching the layout of
+  * `java.util.UUID`). The canonical textual form is the lowercase, hyphenated `8-4-4-4-12` grouping of hex digits, for example
+  * `550e8400-e29b-41d4-a716-446655440000`.
+  *
+  * This type carries only pure value operations: parsing, formatting, byte conversion, and the deterministic name-based constructors
+  * [[UUID.v5]] and [[UUID.v8Sha256]]. Random and time-based generation (v4, v7) is an effectful capability that lives in `kyo-core`.
+  */
+opaque type UUID = (Long, Long)
```
hearnadam: A class would be better here as it would be specialized

kyo-core/shared/src/main/scala/kyo/UUIDGenerator.scala
DamianReeves: Fixed.

kyo-core/shared/src/main/scala/kyo/UUIDGenerator.scala
DamianReeves: Thanks, updated as suggested.

kyo-data/shared/src/main/scala/kyo/UUID.scala
DamianReeves: Updated as well.

fwbrasil (approved):

PR #1780, opened 2026-07-24
[kyo-ui] add UI.mounted with node-scope supervision

xsistens:
## Problem

kyo-ui renders a reactive tree of pure UI values, but a component that must run an effect to
produce its UI (load data, open a subscription, read ambient config) has no first-class place in
that tree. Today it is expressed as a signal holding a cell that a separately-run fiber fills,
wired by hand at each site, with no standard handling for the pending window, for a failure during
or after the effect, or for re-running the component when its inputs change.

## Solution

Add `UI.mounted`, which lifts an effectful component (`UI < (Async & Scope)`) into the reactive
tree as an ordinary node.

- `UI.mounted(effect).keyed(k).placeholder(ui).onError(f).whilePending(KeepLatest | Placeholder)`:
  a per-region keyed instance table runs each component on a runner fiber; a re-key evicts the old
  instance before painting the new one (break before make), and `KeepLatest` holds the previous
  painted UI while the re-keyed instance is pending. A duplicate key renders an inline error card
  in the second slot rather than silently sharing one instance; an unstable key (repeated churn on
  one path) emits a churn hint.
- Failure routing: `.onError` receives a `UI.MountError` carrying the throwable, a rendering-ready
  message, the node's frame, path and key, and a `source` (`Effect`, `Panic`, `Fork`) saying which
  path the failure came from. Without `.onError` the node falls back to a default inline error
  node. The enclosing region stays alive either way, and a later key change re-mounts cleanly.
- Node-scope supervision via `UI.fork`: a mount effect allocates and returns, so anything
  long-lived it opens keeps running on the node's `Scope` afterwards. Forked with `UI.fork`, such a
  fiber dying with a non-interrupt terminal failure flips the already-published node into the same
  error routing a failed mount takes (`MountError.source == Fork`). At most one flip per instance,
  teardown interrupts exempt.
- Effect-typed event handlers: every `on*` setter is polymorphic in `S` behind the same
  `Isolate[S, Sync, S]` gate as `UI.mounted`, so a handler can read the ambient context effects the
  mount can. Pure `Any < Async` handlers still infer `S = Any` and stay source-compatible.
- Dispatch descends through static intermediates before taking a deep-reactive jump, so a click on
  a reactive label inside a button still runs the button's own handler.

## Changes since the first review

- **`kyo-core` is untouched (diff is 0).** The first version added a `Fiber.Supervisor` propagated
  through a `Local` read on every `Fiber.init`, in the hot path. `Local.init` is inheritable and
  `Fiber.onComplete` is public, so the primitive does not need `kyo-core` at all: `UI.fork` now
  lives in kyo-ui, mirrors `Fiber.init`'s signature, and attaches an `onComplete` reporter only
  when a mount effect installed one. `Fiber.init` itself is back to exactly what it was.
- **Supervision is opt-in.** A plain `Fiber.init` inside a mount effect is still scoped to the
  node, but its failure stays invisible to the node. That is the documented escape hatch for a
  fiber whose failure the effect observes itself (`Fiber.initUnscoped` remains fire and forget).
  Outside a mount effect, `UI.fork` *is* `Fiber.init`.
- **`UI.boundary` is not part of this PR anymore.** Pending aggregation and typed error selection
  for a subtree are a separate concern with a separate design question (whether a boundary should
  gate reveal at all), and they had no production consumer here. Removed and parked.

## Why not a plain `Signal[UI]`

A mount without a key is close to that: run an effect, publish into a cell, render the cell. The
machinery is what a *keyed* mount buys, and it is not expressible with a signal alone: an instance
that survives re-renders of its enclosing region, its `Scope` and resources kept alive across them,
torn down deterministically when the key changes, with the previous paint held during the swap. A
router that mounts a page per route, or a panel keyed by the entity it shows, needs exactly that
identity; a `Signal[UI]` re-runs the effect on every re-render of the region above it.

## Performance

`UI.mounted` is additive: it adds new AST and engine paths rather than changing existing ones. An
SSR-render micro-benchmark (`HtmlRenderer.render` over a ~481-node static tree, 30k iterations after
warmup) shows the render hot path for ordinary non-mounted trees is unchanged: byte-identical
output, timing within run-to-run noise. With `kyo-core` untouched, `Fiber.init` has no added cost
in this PR at all.

On the reactive side the same principle is enforced structurally: a region's mount registry is
created on first use, not per region. Only a region that actually contains a `UI.mounted` node ever
puts anything in one, so a region without mounts allocates nothing for them and skips the
per-emission pass that would otherwise collect mount keys for it. `evictExcept` is `inline`, so an
unforced registry does not even evaluate its key set.

## Changes in this revision

- Rebased onto current `main`.
- A region's mount registry is created on first use rather than per region (see Performance). Every
  reactive region owned one, but only a region containing a mounted node ever puts anything in it,
  so a thousand-row list paid a thousand registries and two passes over all rows per emission to be
  told there was nothing to evict. `MountedTest` gains a case for a mount that first appears in a
  later emission of its region, which is the path deferred creation introduces.
- One supervision test waited on a fixed 150ms sleep for a paint that takes longer than that under
  load. It now waits on the paint itself and keeps a sleep only as the settle window that proves no
  second paint follows.

### DIFF
--- a/kyo-ui/README.md
+++ b/kyo-ui/README.md
@@ -344,6 +344,54 @@ If you need both a signal-driven attribute AND a chainable element method, set t
 
 When you pass either a constant `String` or a `SignalRef[String]` to `.value`, the framework wraps it in `UI.Bound[String]`: `Const(v)` or `Ref(ref)`. You normally never name this type; it surfaces in AST pattern matching (see [Pattern-matching on UI](#pattern-matching-on-ui-ast-access)).
 
+## Effectful component mounts
+
+The reactive boundaries above re-render pure `UI` values as signals change. A component that must run an *effect* to produce its UI (load data, open a subscription, read an ambient service) needs more: a node that runs `UI < (Async & Scope)`, shows something while it is pending, and gives a failure somewhere to land. `UI.mounted` is that node.
+
+`UI.mounted(effect)` lifts an effectful component into the tree as an ordinary child. The effect runs when the node attaches, inside a `Scope` tied to the node's lifetime (torn down on detach), so resources it acquires are released when the node goes away; an `Abort[Throwable]` failure is caught into the node's error state instead of propagating. The builder tunes the surrounding behavior:
+
+| Modifier | Effect |
+| --- | --- |
+| `.keyed(k)` | Continuity anchor: the live instance (its `Scope`, resources, published content) is kept across re-renders of the enclosing reactive region while an equal key is re-emitted, and torn down when the key changes or disappears. Any `CanEqual` value works; the idiomatic key is the component's singleton object, tupled with the non-`Signal` inputs the effect closed over. |
+| `.placeholder(ui)` | Content painted while the effect is pending on first mount. Defaults to an empty node. |
+| `.onError(f)` | Renders a failure of the node, receiving a `UI.MountError` (the `error` itself, a rendering-ready `message`, the node's `frame`, `path` and `key`, and a `source` saying which path it came from); the failure is also logged. The enclosing region stays alive, and a later key change re-mounts cleanly. |
+| `.whilePending(KeepLatest \| Placeholder)` | What a keyed node shows while it re-mounts: `KeepLatest` (the default) keeps the last rendered content visible as an inert frozen snapshot (its nested regions and handlers are already torn down) until the new instance publishes; `Placeholder` drops back to the placeholder, the honest choice for interactive content where a frozen-but-visible form would swallow input. On first mount the placeholder always wins. |
+
+```scala
+import UI.*
+import kyo.*
+
+def profileCard(userId: Int): UI < (Async & Scope) =
+    for name <- Signal.initRef("...")
+    yield div(h2(name), p(s"user #$userId"))
+
+val card: UI =
+    UI.mounted(profileCard(42))
+        .keyed(42)
+        .placeholder(div("loading..."))
+        .onError(e => div(s"failed: ${e.message}"))
+        .whilePending(PendingPolicy.KeepLatest)
+```
+
+A value input that should UPDATE a live instance belongs in a `Signal` parameter, not the key; a non-`Signal` value the effect captures belongs IN the key, or it is silently frozen at mount time. A duplicate key under one region renders an inline error node in the second slot rather than silently sharing one instance.
+
+### Failure after the first paint
+
+A mount can also fail *after* it has already published its UI: a background fiber it forked (a data feed, a subscription drain, a poll loop) terminates with an error. Fork those with `UI.fork` instead of `Fiber.init` and node-scope supervision covers that window: a non-interrupt failure flips the already-published node into the same error routing as an up-front failure: its `.onError` first, then a default error node.
+
+```scala
+import UI.*
+import kyo.*
+
+def liveTicker(quotes: Stream[String, Async]): UI < (Async & Scope) =
+    for
+        price <- Signal.initRef("...")
+        _     <- UI.fork(quotes.foreach(p => price.set(p)))
+    yield div(span(price))
+```
+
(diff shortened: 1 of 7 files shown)


kyo-core/shared/src/main/scala/kyo/Fiber.scala
```diff
@@ -135,7 +141,56 @@ object Fiber:
         reduce: Reducible[Abort[E]],
         frame: Frame
     ): Fiber[A, reduce.SReduced & S2] < (Sync & S & Scope) =
-        Scope.acquireRelease(initUnscoped[E, A, S, S2](v))(_.interrupt)
+        Supervisor.local.use {
```
fwbrasil: This is a critical and hot code path, I'd prefer to avoid introducing `Supervisor` as designed here. Would it be possible to have an isolated impl like this and then wire it where we need instead of propagating via the `Local`?

```scala
class Supervisor(watch: (Fiber[?, ?], Result[?, ?]) => Any < Sync):
     def apply[A, E](v: A < (Async & Abort[E])): A < (Abort[E] & Async) = 
            Fiber.init(v).map(fiber => fiber.onComplete(result => watch(fiber, result).andThen(fiber.get))
```

kyo-ui/README.md
```diff
@@ -341,6 +341,43 @@ If you need both a signal-driven attribute AND a chainable element method, set t
...
+```scala
+import UI.*
+import kyo.*
+
+def profileCard(userId: Int): UI < (Async & Scope) =
+    for name <- Signal.initRef("...")
+    yield div(h2(name), p(s"user #$userId"))
+
+val card: UI =
+    UI.mounted(profileCard(42))
```
fwbrasil, line 369: I haven't reviewed the rest of the code but it includes a lot of machinery. I wonder if there's a simpler path via a `Signal[UI]`? It seems possible to define a method that takes a computation `UI < (Async & Scope)` and transforms it into `Signal[UI]` via `initRaw` wiring the computation in the functions

kyo-core/shared/src/main/scala/kyo/Fiber.scala
xsistens: Done, `kyo-core` is untouched now (diff 0) and `Fiber.init` is back to exactly what it was. You were right about the cost too: `Fiber.init` is a `Scope.acquireRelease`, which already suspends the context once, and the supervisor `Local` piled a second one on top unconditionally, even with nothing installed.

I couldn't use the sketch as-is though: it wraps one computation and only sees that computation's own fiber, which `Abort.run` catches anyway. The one I need to see outlives the effect, since a mount allocates and returns while the feed it forked keeps running on the node's `Scope`. That does need something ambient, but not in `kyo-core`: `Local.init` is inheritable and `Fiber.onComplete` is public, so `UI.fork` lives in kyo-ui now and wires the reporter itself. Side effect: supervision is opt-in, a plain `Fiber.init` doesn't participate. More under "Changes since the first review".

kyo-ui/README.md
xsistens: Fair, and for the simple case you're right: a mount without a key is basically that already, run the computation, publish into a cell, render the cell.

The machinery is there for identity. A `Signal[UI]` built inside the render function is a fresh signal on every re-render of the region above it, so the computation re-runs and the `Scope` gets rebuilt; hoist it out and you're managing instance lifetime by hand at every call site, which is the thing this replaces. A keyed mount keeps the live instance across those re-renders with its `Scope` and resources intact, tears it down when the key changes, and holds the last paint during the swap. A router mounting a page per route wants exactly that.

Also, it shrank since you looked: `UI.boundary` is out and `kyo-core` is untouched. Longer version under "Why not a plain `Signal[UI]`".

xsistens:
Just saw that there are a few places in this PR where I'm using Seq and Map again. I will switch that over to the respective kyo data structures.

PS: I saw that the current version of kyo-ui uses Seq already for storing the path, so I keep it like that in this scenarios. If that should be adjusted we should do it in a separate PR that covers all occurences.

kyo-ui/shared/src/main/scala/kyo/internal/ReactiveUI.scala
```diff
@@ -90,18 +148,43 @@ private[kyo] object ReactiveUI:
...
+                        case Absent => (true: Boolean) // not (or no longer) wired: bubble
+                    }
+                ReactiveUI(
+                    path,
+                    Signal.initConst(spec.placeholderUI),
+                    isConst = false,
+                    Seq.empty,
+                    handle,
+                    svgContext = svg,
+                    mountedSpec = Present(spec)
```
ahoy-jon, line 176: all args by name?

PR #1781, opened 2026-07-24
[kyo-ui] add declarative focus seeding and return (focusAuto / focusRestore)

xsistens:
## Problem

kyo-ui has attributes for keyboard focus *order* and *containment* (`tabIndex`, `focusTrap`, `focusGroup`), but nothing that moves focus in response to the tree changing. Two common patterns have no declarative form: focusing an element the moment it appears (a panel or field that opens), and returning focus to where it was when that element goes away (a modal closing). Doing it by hand means reaching for the DOM from an effect and tracking the previously focused element yourself, and it is easy to steal focus on an unrelated re-render.

## Solution

Add two attributes in the `data-kyo-focus-*` family, wired in both transports (`HtmlRenderer.clientJs` for server-push, `DomBackend` for the SPA):

- `focusAuto(true)`: when a re-render inserts the element NEWLY into the DOM (its `data-kyo-path` was not present in the replaced subtree before), the client calls `.focus()` on it once. A re-render of an already-visible element (same path) does not re-seed, so the user's focus is never stolen by an echo re-render. Seeding runs after the client's existing focus/caret restore, so opening a panel wins over restore-to-trigger. On initial page load a focus-auto element already present is seeded, consistent with native `autofocus`. The element is not made focusable automatically; pair with `tabIndex(-1)` for elements that are not natively focusable.
- `focusRestore(true)`: when the seeded element leaves the document, the client returns focus to the element focused just before seeding (located by `data-kyo-path`, skipped if it is gone) — the modal-returns-focus-to-its-trigger pattern.

Both are pure attribute setters returning `Self`; the behavior lives entirely in the two client transports, keyed off the emitted `data-kyo-focus-auto` / `data-kyo-focus-restore` markers, and composes with the existing `focusTrap` / `focusGroup` family.

## Notes

Seeding is edge-triggered on NEW insertion — compared against the set of focus-auto paths present before the patch — so it never fights the user for focus. Tested as `focusAuto`/`focusRestore` cases in `FocusableTest` (real-Chrome integration test, server-push transport).

### DIFF
--- a/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
+++ b/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
@@ -8,6 +8,12 @@ import scala.scalajs.js
 /** Scala.js UI backend. Mounts a UI into the browser DOM. */
 private[kyo] object DomBackend:
 
+    /** Focus seeding/restore stack for `data-kyo-focus-auto`/`data-kyo-focus-restore` (seeded path, prior-focus path,
+      * restore flag). Mirrors `__frs` in HtmlRenderer.clientJs. Module-level mutable state is safe: all mutation runs
+      * inside `Sync.defer` on the single-threaded JS runtime.
+      */
+    private var focusReturnStack: List[(String, Maybe[String], Boolean)] = Nil
+
     /** Mount a UI into the page body. */
     def mount(ui: UI)(using Frame): Unit < (Async & Scope) =
         mountInto(ui, document.body)
@@ -41,7 +47,9 @@ private[kyo] object DomBackend:
             html <- HtmlRenderer.render(ui, Seq.empty)
             _    <- Sync.defer(container.innerHTML = html)
             _    <- applyJsProps(container)
-            _    <- Sync.defer(beginAnimationsSync(container))
+            // Initial mount: everything is new, so pass an empty old-path set (mirrors clientJs startup faSeed).
+            _ <- Sync.defer(seedFocusAuto(container, Set.empty))
+            _ <- Sync.defer(beginAnimationsSync(container))
             exchange = LocalExchange(root)
             dispatch <- ReactiveUI.subscribe(root, exchange)
             // Single-consumer drain owned by the ambient page Scope: every JS event effect is run by a
@@ -110,6 +118,9 @@ private[kyo] object DomBackend:
                                 else pathAttr
                             else null
                         val (selStart, selEnd) = if insideRegion then readSelection(ae) else (Absent, Absent)
+                        // Focus-auto paths present BEFORE the replace, so seeding fires only for newly-appeared
+                        // elements (a re-render of an already-open overlay must not steal focus).
+                        val oldFocusAuto = focusAutoPaths(el)
                         el.outerHTML = finalHtml
                         val updated = document.querySelector(s"""[data-kyo-path="$pathAttr"]""")
                         if updated != null then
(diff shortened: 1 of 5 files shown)


kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
```diff
@@ -158,6 +173,64 @@ private[kyo] object DomBackend:
...
+            val p = el.getAttribute("data-kyo-path")
+            p != null && !oldSet.contains(p)
+        }.foreach { el =>
+            val ae = document.activeElement
+            val ret =
+                if ae != null && (ae ne document.body) then Maybe(ae.getAttribute("data-kyo-path"))
+                else Absent
+            focusReturnStack =
+                (el.getAttribute("data-kyo-path"), ret, el.hasAttribute("data-kyo-focus-restore")) :: focusReturnStack
+            val _ = el.asInstanceOf[scalajs.js.Dynamic].focus()
```
fwbrasil: please `kyo.discard` instead of `val _` in PRs

kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
```diff
@@ -158,6 +173,64 @@ private[kyo] object DomBackend:
...
+      * `data-kyo-focus-restore`, focus the recorded previous element (skip silently if it is gone too).
+      * Mirrors `frSweep` in HtmlRenderer.clientJs.
+      */
+    private def sweepFocusAuto(): Unit =
+        var continue = true
+        while continue do
+            focusReturnStack match
+                case (faPath, ret, restore) :: rest =>
+                    if document.querySelector(s"""[data-kyo-path="$faPath"][data-kyo-focus-auto]""") != null then
+                        continue = false
```
fwbrasil: why short circuit here?

kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
xsistens: Done — both call sites this PR adds now use `discard(...)` (the seed in `seedFocusAuto`, and the return-focus call from the sweep, which moved into a small `focusIfPresent` helper).

I left the three pre-existing `val _` in this file alone (two in `restoreFocus`, one reflow force in `beginAnimationsSync`) since they're already on `main` and unrelated to this change — happy to fold them in if you'd rather have the file consistent.

kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
xsistens: It's the LIFO invariant rather than an optimization — but the question made me find a real bug underneath it, so: good catch.

The stack is newest-seed-first. Stopping at the first entry whose element is still in the document is what keeps the sweep from restoring focus *out of* a seed that is still on screen. If it instead scanned the whole stack for dead entries, two independent panels A (older) and B (newer) would break: closing A while B is open would fire A's restore and yank focus out of B.

What the code got wrong is what happened to A's entry afterwards — it stayed on the stack, still armed. So when B later closed, the sweep popped B and restored focus to B's trigger (correct), then kept going, popped A, and restored to A's trigger, overwriting it. The user ends up on the trigger of a panel they closed a while ago.

Popping only one entry per sweep is not the fix either: for a genuinely nested close (modal plus inner popover removed in the same patch) the popover's return target lives inside the modal and is gone too, so the unwind has to continue until it finds one that is still there — otherwise focus lands on `<body>`.

The rule is now two-part, and both halves are in the doc comment: stop at the first live seed; below that keep unwinding but let at most one restore land, dropping the deeper stale entries silently so they cannot fire on a later sweep either. I also rewrote the loop as `@tailrec` — the `var continue` was what made the exit read like a bail-out instead of a rule.

Covered by a new `FocusableTest` case with two sibling panels closed out of order; it fails on the previous logic (focus ends on A's trigger instead of B's).

PR #1782, opened 2026-07-24
[kyo-ui] opt-in per-level event consumption (stopPropagation)

xsistens:
> **Merge before its stacked follow-ups.** #1783 (`onContextMenu`) and #1786 (`onFileSelect`) are based on this branch and opt their new events into the `keepBubbling` introduced here. Merge this PR first; their diffs narrow to just their own feature once this lands.

## Problem

kyo-ui dispatches a bubbling event to every element on the path from the innermost target outward, firing each ancestor that declared a handler for that event's type. There is no way for an element to say "I handled this; do not also run the handlers above me." Nested overlays are the common case: an Escape keydown inside an inner panel also reaches the outer panel's handler, so both close at once.

## Solution

Add `stopPropagation(v: Boolean)`. When the dispatch walk reaches an element that carries the flag AND declared a handler for the arriving event's type, the event stops there — the ancestors' handlers do not fire. It only consumes the types that element actually handles; other event types pass through unaffected. The default (unset or `false`) keeps the existing bubble-through semantics.

The stop lives server-side in `ReactiveUI.dispatchToElement` (`keepBubbling`), the single source of truth for both transports; `data-kyo-stop="1"` is emitted as an inspection marker only, and the clients never branch on it. It applies to the bubbling events (click, keydown, keyup, hover, unhover, scroll); focus and blur never bubble in kyo-ui.

## Notes

Tested as `stopPropagation` cases in `EventHandlerTest` (real-Chrome integration test, server-push transport).

### DIFF
--- a/kyo-ui/README.md
+++ b/kyo-ui/README.md
@@ -101,6 +101,8 @@ div.onClickSelf(Console.printLine("background"))(
 )
 ```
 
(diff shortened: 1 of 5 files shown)


DamianReeves (approved):
LGTM

PR #1783, opened 2026-07-24
[kyo-ui] onContextMenu event (right-click), both transports (follow-up: #1780, #1782)

xsistens:
> **⚠ Depends on #1780 and #1782 — please merge those first.** `onContextMenu` uses the effect-typed handler signature (`[S](…)(using Isolate[S, Sync, S])`) added by #1780 and opts into #1782's `keepBubbling` (a `stopPropagation(true)` ancestor consumes contextmenu events like any other), so this builds on both. The diff shows their commits until they land.

## Problem

kyo-ui has click, key, focus/blur, hover, and scroll events, but no right-click. An app that wants a custom context menu has to reach around the framework: attach a raw DOM listener, call `preventDefault` itself, and translate the event back into the reactive tree by hand.

## Solution

Add `onContextMenu`, the `contextmenu` (right-click) event, wired in both transports (the server-push clientJs and the SPA `DomBackend`). Like `onClick` it takes a by-name effectful action or a `MouseEvent`-typed handler in #1780's effect-typed form (`[S](…)(using Isolate[S, Sync, S])`), and dispatches through the reactive tree with the same bubble-through semantics: every ancestor on the path that declared a context-menu handler fires, innermost first; via #1782's `keepBubbling`, an ancestor that sets `stopPropagation(true)` consumes it.

When the right-click lands on an element (or a descendant) that has a context-menu handler somewhere in its ancestor chain, the client calls `preventDefault` to suppress the browser's native menu; a right-click with no handler in its chain keeps the native menu, so the suppression is opt-in per subtree.

## Notes

Tested as `onContextMenu` cases in `EventHandlerTest` (real-Chrome integration test, server-push transport).

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -121,6 +121,12 @@ object Fiber:
 
     /** Runs an asynchronous computation in a new Fiber guaranteeing eventual interruption via the [[Scope]] effect.
       *
+      * If a [[Fiber.Supervisor]] is installed in the current context (see [[Supervisor.let]]), the spawned fiber is additionally
+      * SUPERVISED: any non-interrupt terminal failure (an `Abort` failure or a panic that is not an [[Interrupted]]) is reported to the
+      * supervisor exactly once. Interruptions — including the scope-close interrupt this very method registers — are never reported.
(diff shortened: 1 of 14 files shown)


fwbrasil:
Please avoid opening PRs that have dependencies on multiple other PRs, even if draft. To validate CI, it's possible to setup in a fork

PR #1784, opened 2026-07-24
[kyo-ui] SPA event delegation forwards ancestor-declared bubbling events

xsistens:
## Problem

kyo-ui's reactive dispatch bubbles an event to every ancestor that declared a handler for its type (a keydown inside a panel reaches the panel's own `onKeyDown`). The SPA transport (`DomBackend`, the client-only mount) honors this only for a handler declared on the event's own target: its forwarding gate reads just the target element's `data-kyo-ev`, so an event whose handler sits on an ancestor is dropped at the JS boundary before the bubble dispatch can run. The server-push transport already ancestor-walks and is unaffected.

## Solution

Walk the ancestor chain in the SPA forwarding gate. A small `ChainTypes` wrapper replaces the target-only `data-kyo-ev` set; its `contains(type)` checks each ancestor up to `document.body` via `declaredInChain`, so the gate matches the dispatch's bubble semantics (the same ancestor-walk the server-push client already performs). Every use site of the old set was `contains(...)`, so the wrapper is a drop-in.

## Notes

Verified: `kyo-uiJS` compile; the event-delegation suite (`EventHandlerTest`, `TypedEventTest`, `UIEventWiringTest`, `UITestSpa*`) stays green: 103 tests, no regression. Full `kyo-uiJS` suite green with the new test infrastructure (2052 tests).

The `withUI` browser harness runs the server-push transport (through `UIServer`), whose ancestor-walk already handles this correctly, so it cannot reach this SPA code path. The forwarding gate is unit-tested in isolation instead: `declaredInChain` is hoisted to a `private[kyo]` method, and `DomBackendDelegationTest` exercises it against a real jsdom document. Since the Node.js test environment has no DOM globals (and the jsdom jsEnv cannot load the CommonJS test bundle), a small `DomTestEnv` helper requires jsdom lazily and installs the window's globals onto `globalThis`; suites that never touch the DOM never load it. The ancestor-declared case fails under the pre-fix target-only semantics, so the test is non-vacuous. jsdom is installed for the JS CI target in the setup action (version pinned there: the repo gitignores `*.json`, so there is no `package.json`).

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -19,6 +19,14 @@ runs:
       with:
         node-version: '24'
 
+    # jsdom for the DOM-backed Scala.js unit tests (kyo-ui DomTestEnv); loaded lazily via require,
+    # so only suites that install the DOM need it. JS only: the helper lives in the js-only test tree.
+    # The repo ignores *.json, so the version is pinned here rather than in a package.json.
+    - name: Install Node.js test dependencies
+      if: ${{ inputs.target == 'JS' }}
+      shell: bash
+      run: npm install --no-save --no-fund --no-audit jsdom@^30
+
     - uses: actions/cache@v5
       env:
         FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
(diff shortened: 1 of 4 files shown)


fwbrasil:
could this be tested via scalajs-specific tests? even if only the specific logic in isolation

xsistens:
Done in 757e0b6. `declaredInChain` is now a proper method and `DomBackendDelegationTest` runs it against a real jsdom document, including the ancestor case (which fails without the fix). The jsdom jsEnv itself can't load our test bundle (CommonJS, because of the kyo-http node builtins), so I added a small `DomTestEnv` helper that requires jsdom lazily and copies the window globals onto `globalThis`, same mechanism jest uses. Suites that don't touch the DOM never load it, and CI installs jsdom for the JS target in the setup action. If we want DOM-level unit tests beyond kyo-ui at some point, `DomTestEnv` could move into kyo-test, but for now I kept it in kyo-ui's test tree.

fwbrasil (approved):

PR #1785, opened 2026-07-25
[kyo-ui] SPA local mount fired form onSubmit twice on a submit-button click

xsistens:
## Problem

A click on a submit button makes the browser fire a native `submit` event right after the `click`. The engine already emulates `onSubmit` from a click that bubbles to a `<form>` (matching browser behavior), so forwarding both the click and the native submit runs the form's `onSubmit` twice. The server-push transport (`HtmlRenderer.clientJs`) already guards this with a `_kyoClickSubmit` flag; the client-only SPA transport (`DomBackend`) does not, so a locally-mounted app double-fires.

## Solution

Mirror the server-push guard in `DomBackend`'s event delegation: set `clickSubmitGuard` on every dispatched Click and clear it on a 0-timeout; the synchronously-following native `submit` is dropped while the flag is set (its `preventDefault` still applies, so the browser never performs its own form navigation). A standalone submit (Enter in an input, a programmatic submit) has no preceding click, so it is unaffected.

## Notes

Adds a counting regression test to `FormValidationScenarioItTest` asserting a single button click runs `onSubmit` exactly once. As with the SPA transport itself, the browser test harness (`withUI`) drives the server-push transport — which already guards this — so the test pins the expected "once" behavior that this change aligns the SPA local-mount transport to, rather than failing without it. The client-only `DomBackend` path has no browser harness; it is kept in sync by construction with the clientJs guard it mirrors.

### DIFF
--- a/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
+++ b/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
@@ -257,6 +257,11 @@ private[kyo] object DomBackend:
         final class ChainTypes(target: dom.Element):
             def contains(t: String): Boolean = declaredInChain(target, t)
 
+        // A submit button's click makes the browser fire a native `submit` right after, but the Click dispatch
+        // already emulates onSubmit; this flag (set on Click, cleared on a 0-timeout) suppresses that one
+        // following native submit so the form handler runs once. Mirrors clientJs's `_kyoClickSubmit` guard.
+        var clickSubmitGuard = false
+
         val handler: scalajs.js.Function1[dom.Event, Unit] = (e: dom.Event) =>
             findPathElement(e.target.asInstanceOf[dom.Element]).foreach { target =>
                 val path    = parsePath(target.getAttribute("data-kyo-path"))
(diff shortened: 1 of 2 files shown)


kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
```diff
@@ -248,6 +253,8 @@ private[kyo] object DomBackend:
                         // is handled by UILocation's interceptor. Prevent-defaulting every anchor here
                         // would also kill those.
                         if target.tagName.toLowerCase == "a" && evTypes.contains("click") then e.preventDefault()
+                        clickSubmitGuard = true
+                        discard(dom.window.setTimeout(() => clickSubmitGuard = false, 0))
```
fwbrasil, line 286: I don't have much experience with JS. Is there a scheduling guarantee that makes this safe and will execute after `submit` is processed?

xsistens: Yes. When a submit button is clicked, the browser fires the native `submit` synchronously as part of the click's default action — both events happen within the same event-loop task. A `setTimeout` callback is a new task and can only run after the current one completes, so the clear can never run between the click and its native submit; the guard is guaranteed to still be set when the submit arrives.

The 0-timeout's only job is to make sure a *later, unrelated* submit isn't swallowed — any later submit arrives in a new task, by which point the flag is already cleared.

Edge cases: Enter-in-an-input fires a synthetic click on the form's submit button first, so it goes through the same click→submit pair and is handled identically. A programmatic `form.requestSubmit()` has no preceding click, so the guard is false and it passes through.

This is the same reasoning the server-push transport's `_kyoClickSubmit` guard in `clientJs` already relies on — this change just mirrors it.

fwbrasil (approved):

fwbrasil:
There's a merge conflict

PR #1786, opened 2026-07-25
[kyo-ui] onFileSelect: file-metadata selection event (follow-up: #1780, #1782)

xsistens:
> **⚠ Depends on #1780 and #1782 — please merge those first.** `onFileSelect` uses the effect-typed handler signature (`[S](…)(using Isolate[S, Sync, S])`) added by #1780 and opts into #1782's `keepBubbling` (a `stopPropagation(true)` element consumes file-select events like any other), so this builds on both. The diff shows their commits until they land.

## Problem

kyo-ui's file input surfaces a selected file only through `onChange`, which carries a single file's text content. There is no way to receive file metadata (name, size, MIME type) or to handle multiple selected files.

## Solution

Add `onFileSelect`, wired in both transports. Declaring it emits a `fileselect` token in `data-kyo-ev`; the client then reads every file selected on a descendant file input — a `FileReader` per file, aggregated once all reads complete — and posts a single `FileSelect` event carrying a `Seq[UI.FilePayload]` (name, size, MIME type, and text content per file). The event bubbles like `Change` and, via #1782's `keepBubbling`, is consumed by an ancestor that sets `stopPropagation(true)`. A file input with only `onChange` keeps the existing single-file, content-only path, so current callers are unaffected. `FileInput.multiple` renders the HTML `multiple` attribute.

## Notes

Test: `FileSelectScenarioItTest` (real-Chrome integration test, server-push transport).

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -121,6 +121,12 @@ object Fiber:
 
     /** Runs an asynchronous computation in a new Fiber guaranteeing eventual interruption via the [[Scope]] effect.
       *
+      * If a [[Fiber.Supervisor]] is installed in the current context (see [[Supervisor.let]]), the spawned fiber is additionally
+      * SUPERVISED: any non-interrupt terminal failure (an `Abort` failure or a panic that is not an [[Interrupted]]) is reported to the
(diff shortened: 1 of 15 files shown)


fwbrasil:
Please avoid opening PRs that have dependencies on multiple other PRs, even if draft. To validate CI, it's possible to setup in a fork

PR #1787, opened 2026-07-25
[kyo-ui] declarative input filtering and masking (inputFilter / inputMask)

xsistens:
## Problem

kyo-ui text inputs accept any typed or pasted character; there is no declarative way to constrain input to digits, a decimal number, or a fixed format, so an app has to post-validate and correct the field after the fact (the rejected character is briefly visible, and the caret jumps).

## Solution

Add `inputFilter` and `inputMask`, enforced client-side by a document-level `beforeinput` capture listener wired in both transports (the server-push clientJs and the SPA `DomBackend`), keyed off `data-kyo-filter` / `data-kyo-mask` markers.

- `inputFilter(pattern)`: `"int"`, `"decimal"`, or a literal allowed-character set. A disallowed character is rejected before it enters the field, and filtered out of pasted or dropped text — so the value never shows a character it would reject.
- `inputMask(mask)`: Prime-style tokens (`9` = digit, `a` = letter, `*` = alphanumeric, any other character is a literal, e.g. `"(999) 999-9999"`). The field formats progressively — literals auto-insert, each position restricts its character class, backspace steps over literals, and the caret stays correct. The server sees the masked value through the normal `input`/`change` events.

## Notes

Test: `InputMaskScenarioItTest` (real-Chrome integration test, server-push transport).

### DIFF
--- a/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
+++ b/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
@@ -61,6 +61,7 @@ private[kyo] object DomBackend:
             _    <- Sync.defer(seedEnter(container, Set.empty))
             _    <- Sync.defer(seedFocusAuto(container, Set.empty))
             _    <- Sync.defer(beginAnimationsSync(container))
+            _    <- setupInputMasking()
             exchange = LocalExchange(root)
             dispatch <- ReactiveUI.subscribe(root, exchange)
             // Single-consumer drain owned by the ambient page Scope: every JS event effect is run by a
@@ -168,22 +169,8 @@ private[kyo] object DomBackend:
                 else located
             val _ = focusTarget.asInstanceOf[scalajs.js.Dynamic].focus()
             (selStart, selEnd) match
-                case (Present(s), Present(e)) =>
-                    val dyn = focusTarget.asInstanceOf[scalajs.js.Dynamic]
-                    if scalajs.js.typeOf(dyn.setSelectionRange) == "function" then
-                        try
-                            val _ = dyn.setSelectionRange(s, e)
-                        catch
-                            // setSelectionRange throws InvalidStateError on input types that do not
-                            // support text selection (e.g. email, number). Mirrors HtmlRenderer.clientJs:583:
-                            // `catch(e){if(e.name!=='InvalidStateError')throw e;}`. Re-throw any other
-                            // JS exception so genuine failures are not silently dropped.
-                            case ex: scalajs.js.JavaScriptException
-                                if ex.exception.asInstanceOf[scalajs.js.Dynamic].name.asInstanceOf[String] == "InvalidStateError" =>
-                                ()
-                    end if
-                case _ => ()
-            end match
+                case (Present(s), Present(e)) => setSelection(focusTarget, s, e)
+                case _                        => ()
         end if
     end restoreFocus
 
@@ -256,6 +243,25 @@ private[kyo] object DomBackend:
             true
         end if
     end focusIfPresent
+
+    /** Moves the caret on `el`, tolerating the two documented ways that is a no-op.
+      *
+      * Elements outside input and textarea (select, contenteditable) have no `setSelectionRange` at all, and on
+      * input types without a text selection (email, number, both of which kyo-ui offers as text inputs) it throws
+      * `InvalidStateError`. In either case the value is already set and only the caret stays put. Any other
+      * JavaScript exception is a real failure and propagates rather than being swallowed. Mirrored by
+      * `kyoSetCaret` in `HtmlRenderer.clientJs`.
+      */
+    private def setSelection(el: dom.Element, start: Int, end: Int): Unit =
+        val dyn = el.asInstanceOf[scalajs.js.Dynamic]
+        if scalajs.js.typeOf(dyn.setSelectionRange) == "function" then
+            try discard(dyn.setSelectionRange(start, end))
+            catch
+                case ex: scalajs.js.JavaScriptException
+                    if ex.exception.asInstanceOf[scalajs.js.Dynamic].name.asInstanceOf[String] == "InvalidStateError" =>
+                    ()
+        end if
+    end setSelection
 
     // Bridge a Kyo Async computation from a JS callback boundary by offering it to the page-scoped drain
     // channel. The single AllowUnsafe site narrows to the offer crossing (the JS callback has no Kyo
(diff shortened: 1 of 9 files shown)


kyo-ui/shared/src/main/scala/kyo/UI.scala
```diff
@@ -883,6 +883,21 @@ object UI:
 
             /** Binds the input to a `SignalRef` two-way: edits write the new text back into the ref, and ref changes update the input. */
             def value(v: SignalRef[String]): Self
+
+            /** Declarative client-local input FILTER: rejects disallowed characters as they are typed or pasted, via a
+              * document-level `beforeinput` capture listener. `pattern` is `"int"` (digits only), `"decimal"` (digits plus one
+              * separator), or a literal set of allowed characters. The server only ever sees the filtered value through the
```
fwbrasil: A major goal of kyo-ui is making APIs properly typed, could we have a proper type hierarchy or a type union for this?

fwbrasil: another option is method overloads

kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
```diff
@@ -374,4 +375,173 @@ private[kyo] object DomBackend:
         if p == null || p.isEmpty then Seq.empty
         else p.split("\\.").toSeq
 
+    // ---- input filter/mask mirror (SPA transport) ----
+
+    private def filterStr(pat: String, str: String, curVal: String): String =
```
fwbrasil: We have a lot of behaviors in these methods that aren't tested. Could they be tested via unit tests? Browser tests would be fine as well

kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
```diff
@@ -374,4 +375,173 @@ private[kyo] object DomBackend:
...
+        val ev   = scalajs.js.Dynamic.newInstance(ctor)("input", scalajs.js.Dynamic.literal(bubbles = true))
+        discard(t.asInstanceOf[scalajs.js.Dynamic].dispatchEvent(ev))
+    end dispatchInput
+
+    private def setMasked(t: dom.html.Input, v: String): Unit =
+        t.value = v
+        val dyn = t.asInstanceOf[scalajs.js.Dynamic]
+        if scalajs.js.typeOf(dyn.setSelectionRange) == "function" then
+            try discard(dyn.setSelectionRange(v.length, v.length))
+            catch case _: Throwable => ()
```
fwbrasil: why swallow the exception?

fwbrasil:
There's a merge conflict

kyo-ui/shared/src/main/scala/kyo/UI.scala
xsistens: Done. `inputFilter` now takes a typed `InputFilter` enum, following the `FileAccept` / `Href` / `Target` pattern already in this file:

```scala
enum InputFilter derives CanEqual:
    case Digits                 // "digits"
    case Decimal                // "decimal"
    case Allowed(chars: String) // "chars:" ++ chars
```

On the type-union option specifically: a union of string literal types plus `String` collapses, because `String` subsumes its own singleton types, so `"digits" | "decimal" | String` is just `String`. A sealed ADT is the only union here that keeps the cases distinguishable.

Typing it also fixed a latent bug rather than just tidying the signature. The old wire format branched on `pat == "int"` and treated everything else as a character set, so `inputFilter("int")` meaning "allow i, n and t" was inexpressible, and a typo like `"integer"` silently degraded to "allow i, n, t, e, g, r". The enum encodes the set behind an explicit `chars:` prefix, mirroring `Href.External`'s `scheme:value`, so a set can never be read as a keyword. `InputMaskingTest` pins both directions, including `Allowed("int")` and `Allowed("chars:")`.

While typing it I also moved the two fields off the universal `Attrs` onto `TextInputAttrs`, where `value` and `placeholder` already live. They were previously carried by every `div`, `span` and per-datum chart `Attrs()`.

xsistens: I went with the enum instead, for a reason specific to this API rather than a general preference. In kyo-ui overloads carry the constant-versus-reactive distinction (`value(String)` vs `value(SignalRef[String])`), where the two arms take genuinely different types. Here both arms would take a `String`, so an `inputFilter(String)` overload would reintroduce exactly the ambiguity the enum removes: `inputFilter("int")` would be back to meaning two different things depending on a lookup table.

The enum keeps the escape hatch explicit and the API to one method. `inputFilter(InputFilter.Allowed("int"))` is unambiguous, and the compiler lists the alternatives at the call site.

One related change came out of this. The two setters moved from `TextInput` to a new `sealed trait ConstrainedInput extends TextInput`, mixed into the seven free-text inputs but not into `NumberInput`. `numberInput.inputMask("(999) 999-9999")` used to compile and was broken at runtime: a `number` field normalizes `.value` to `""` on invalid content, and the mask's own literals are never valid there. It is now a compile error, with `typeCheckFailure` tests pinning it.

kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
xsistens: You are right, and it was worse than untested. `UITest.withUI` starts `UI.runHandlers` plus an `HttpServer`, so the browser harness only ever exercises the server-push transport. Nothing in the repo loads the Scala.js backend, so this file had zero coverage.

Three changes:

1. The character-level logic moved to `kyo-ui/shared/src/main/scala/kyo/internal/InputMasking.scala`. `DomBackend` keeps only the DOM wiring. In `shared` the logic is testable on JVM, JS, Native and Wasm without a browser and without new infrastructure. `InputMaskingTest` now has 62 tests covering branches that had none, including `maskRaw`'s literal-mismatch arm, the `decimal` second-separator rules, the ASCII-only classes, and the invariant `maskRaw(m, maskFormat(m, raw)) == raw` that the whole design rests on.

2. The server-push transport carries a hand-written JavaScript copy, so unit-testing the Scala half would only cover half the feature. That copy is now a named fragment, `HtmlRenderer.inputMaskJs`, and `InputMaskingJsParityTest` loads it into a real page and drives it through the same case table as the Scala suite. One table, two implementations, so a new row covers both. It earned its keep immediately: with the mask parser's escape branch deliberately disabled it fails with `js=+4\1 23 expected=+49 123`.

3. `InputMaskingScenarioItTest` grew from 5 to 13 cases, covering what only a browser shows: paste via `insertFromPaste` with a real `DataTransfer`, typing over a selection, caret placement, mask capacity, and `type=email`.

What remains uncovered is stated in the suite's scaladoc rather than left implicit: `setupInputMasking`, `setValue`, `setFilteredAt`, `dispatchInput` and the `compositionend` listener are thin DOM shims over now-tested logic, and a genuine Scala.js browser harness is separate scope.

kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
xsistens: No good reason, and the right idiom was already 300 lines up in the same file. `restoreFocus` catches only `JavaScriptException` with `name == "InvalidStateError"` and rethrows everything else; `clientJs` did the same. The new code fell behind that.

Rather than just narrowing the catch, I merged all three call sites onto one helper per language, `DomBackend.setSelection` and `kyoSetCaret`. It tolerates exactly two documented no-ops, an element with no selection API at all (`select`, contenteditable) and an input type that throws `InvalidStateError` (`email`, `number`), and propagates everything else. `restoreFocus` lost its inline try/catch, so the diff removes lines here.

The dead `typeof setSelectionRange === "function"` guard turned out to be hiding something real. On an `HTMLInputElement` that property always exists, so the guard never fired, but `Interactive.data` lets any element carry `data-kyo-filter`, and both handlers then cast to `dom.html.Input` unchecked. A throw out of a `beforeinput` capture listener on `document.body` breaks typing for the whole page. Both handlers now gate on `tagName == "INPUT" || tagName == "TEXTAREA"` up front, the same test `clientJs` already used in the morph path.

The `type=email` case is covered end to end by an IT: the second keystroke is what proves the client script survived the first.

fwbrasil:
merge conflict again

PR #1788, opened 2026-07-25
[kyo-ui] enter/leave CSS transition lifecycle (enterTransition / leaveTransition)

xsistens:
## Problem

kyo-ui renders reactive updates by replacing an element's `outerHTML`. That swaps the DOM node for a fresh one, so a CSS transition or `@keyframes` animation on the old node never runs — there is no declarative way to animate an element as it appears or disappears.

## Solution

Add `enterTransition` and `leaveTransition`, client-local animation that survives the `outerHTML` replace, wired in both transports (the server-push clientJs and the SPA `DomBackend`), keyed off `data-kyo-enter` / `data-kyo-leave` markers.

- `enterTransition(classes)`: when a patch inserts an element newly into the DOM (its `data-kyo-path` was not present in the replaced subtree before, and at initial page load), the client adds the space-separated classes right after insertion, forces a reflow, then removes them on the next animation frame — so a CSS transition runs from the enter-from state the classes describe. An echo re-render of an already-present element does not replay.
- `leaveTransition(classes)`: when a Replace or Remove removes the element, the client captures its `getBoundingClientRect` (nothing crosses the wire), applies the patch, then appends a visual GHOST to `<body>` — a deep clone, `position:fixed` at the captured rect, `pointer-events:none`, with `data-kyo-*`/`id` stripped so it is inert — and adds the leave classes on the next frame. The ghost is removed on `transitionend`/`animationend` or a 1-second safety timeout. Only the outermost removed leave-element per patch spawns a ghost.

## Notes

Test: `EnterLeaveTransitionScenarioItTest` (real-Chrome integration test, server-push transport).

### DIFF
--- a/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
+++ b/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
@@ -41,6 +41,7 @@ private[kyo] object DomBackend:
             html <- HtmlRenderer.render(ui, Seq.empty)
             _    <- Sync.defer(container.innerHTML = html)
             _    <- applyJsProps(container)
+            _    <- Sync.defer(seedEnter(container, Set.empty))
             _    <- Sync.defer(beginAnimationsSync(container))
             exchange = LocalExchange(root)
             dispatch <- ReactiveUI.subscribe(root, exchange)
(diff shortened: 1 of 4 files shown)


fwbrasil:
general question about your PRs: are you validating these changes manually with a browser as well? It's quite difficult for me to reason about how browser JS works so I'm mostly validating the overall logic but I can't really validate all the mutability and side effectful methods called

xsistens:
Yes. Since the in-repo tests can't reach the SPA/`DomBackend` paths, that side is validated manually in a real browser. Concretely: there's an upcoming kyo-ui-components PR with a demo page that exercises these features; I run AI-driven browser testing against it, and afterwards go through it personally to catch visual regressions that automated checks tend to miss, such as flicker on re-render, layout shifts, or janky transitions.

fwbrasil (approved):

fwbrasil:
Thanks for the clarification. Can you fix the merge conflicts and push? it'll trigger CI automatically

PR #1789, opened 2026-07-25
[kyo-ui] add an opt-in session handler-error policy (onHandlerError) (follow-up: #1780)

xsistens:
> **⚠ Depends on #1780 (`UI.mounted`) — please merge that first.** The policy threads through the mount/normalize/`MountDispatch`/`safeDispatch` dispatch machinery introduced by #1780, so it builds on it. The diff shows #1780's commit until it lands.

## Problem

When an event handler fails or panics and nothing below it consumed the error, the engine logs it out-of-band and keeps the event bubbling — but an app has no way to react to those escaped errors in one place. For example: a dozen buttons whose `onClick` handlers each call a backend, and the app wants any one of them failing to pop the same error toast. Today that means wrapping every single handler; there is no one place to say "any unhandled handler error becomes a toast."

## Solution

Add an opt-in `onHandlerError` policy: `UI.runMount` and `UI.runHandlers` gain overloads taking `Throwable => Unit < Async`, the app's single central decision for any event-handler failure or panic that nothing below consumed.

It is threaded structurally from the transport entry point through `normalize` → `MountDispatch` → `safeDispatch`, deliberately **not** a `Local`: root-only is the truth, and it works identically on the server-push transport (where the browser drives server-side session fibers over a WebSocket), on which a `Local` installed at the mount site does not reach those per-session fibers.

`safeDispatch` still logs out-of-band and never stops the bubble chain; the policy is simply offered the error after the log. Handlers keep their local options:

- a handler that wants a different fate **consumes** the error itself (`Abort.recover`) before it reaches this seam, so the policy never sees it;
- a handler that only wants to **observe** re-raises after its side effect and lets the error bubble on to the policy.

With no policy installed the behavior is unchanged (log-only).

## Notes

Tested in `UIEventWiringTest` (direct-engine dispatch, no browser), alongside the existing `safeDispatch` recover test: the policy sees an Abort and a panic, a successful handler never invokes it, a locally-consumed failure is kept from it while an observed-and-re-raised one reaches it, and with no policy the log-only status quo holds.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -121,6 +121,12 @@ object Fiber:
 
     /** Runs an asynchronous computation in a new Fiber guaranteeing eventual interruption via the [[Scope]] effect.
       *
+      * If a [[Fiber.Supervisor]] is installed in the current context (see [[Supervisor.let]]), the spawned fiber is additionally
+      * SUPERVISED: any non-interrupt terminal failure (an `Abort` failure or a panic that is not an [[Interrupted]]) is reported to the
+      * supervisor exactly once. Interruptions — including the scope-close interrupt this very method registers — are never reported.
+      * With no supervisor installed (the default), behavior is unchanged. [[initUnscoped]] and [[use]] never supervise — they are the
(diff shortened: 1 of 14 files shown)


fwbrasil:
https://github.com/getkyo/kyo/pull/1780 has some major open questions so I'm closing this for now

PR #1790, opened 2026-07-25
[kyo-ui] focus restore skips reactive wrapper spans (follow-up: #1781)

xsistens:
> **⚠ Depends on #1781 (declarative focus seeding / return) — please merge that first.** This refines the focus-restore path introduced by #1781 (`data-kyo-focus-auto` / `-restore`, `restoreFocus`, the focus-return sweep), so it builds on it. The diff shows #1781's commit until it lands.

## Problem

A reactive child — an element whose content or attributes are driven by a signal, e.g. a trigger whose label updates live, `state.map(s => UI.button(s.title))` — renders inside a `<span data-kyo-reactive>` wrapper. That wrapper is styled `display: contents` (layout-transparent) and shares its `data-kyo-path` with the first element it wraps. So when focus restore looks an element up by path with a plain first-match query, it can hit that transparent, unfocusable wrapper instead of the real element — and `.focus()` on the wrapper goes nowhere.

This bites focus restore to a trigger that happens to be a reactive child: opening a `focusRestore` panel records the trigger's path as the return target, and on close the sweep queries that path, lands on the wrapper span, and the trigger never regains focus.

## Solution

Both transports now prefer a `:not([data-kyo-reactive])` match when locating an element by path for focus, falling back to the plain query when there is no non-wrapper match:

- browser mount (`DomBackend`): a `locateByPath` helper, used by the caret/focus restore after a reactive region replace and by the focus-return sweep;
- server-push client (`HtmlRenderer.clientJs`): the same as an `lbp(p)` helper, used in the replace handler's focus/caret restore and in `frSweep`.

## Notes

`FocusableTest` gains a scenario where the trigger is a reactive-wrapped button: open the `focusRestore` panel, close it, and assert focus returns to the button. Without the wrapper skip the restore lands on the `data-kyo-reactive` span and the button never regains focus (verified: the test fails on the base branch and passes here).

### DIFF
--- a/kyo-ui/README.md
+++ b/kyo-ui/README.md
@@ -61,6 +61,7 @@ Every `Element` carries an `Attrs` record. The chainable setters are:
 - `.hidden(v: Boolean)`: hide the element. The `Signal[Boolean]` overload exists but lives under "Reactivity" below because it changes the return type.
 - `.style(v: Style)` / `.style(f: Style.type => Style)`: attach styling, see [Styles](#styles).
 - `.tabIndex(v)`, `.focusTrap(v)`, `.focusGroup(id)`: focus order and grouping for keyboard navigation.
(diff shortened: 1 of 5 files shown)


xsistens:
Superseded by the comment-marker region model now in #1794: reactive regions no longer paint a wrapper element at all (including the transparent span that used to appear around signal-bound fields on their first patch), so there is nothing for focus restore to skip. The replacement behavior — an element path resolves directly; a region path falls back to the region registry and focuses the first focus-capable element in the range — ships as part of #1794.

Issue #1638, opened 2026-05-25
[BUG]: Signal.streamChanges emits continuously when no changes have occurred


markehammons:
### Version

1.0.0-RC2

### Scala Version

3.8.3

### Expected Behavior

The documentation says streamChanges should emit only when changes have occurred, so Signal.initConst(5) should return either Chunk(5) or Chunk.empty[Int] on `.streamChanges.run`.

### Actual Behavior

```
        val x = Signal.initConst(5)

        x.streamChanges
          .take(5)
          .run // returns Chunk(5,5,5,5,5)
```

### Steps to Reproduce

See above code.

fwbrasil:
@markehammons is this still valid? you had mentioned a solution in discord I think

markehammons:
@fwbrasil this should still be valid. I had other issues with streamChanges that I haven't diagnosed properly yet. The fact remains though that streamChanges on a `Signal.initConst` emits continuously, which doesn't match the API's described behavior.

fwbrasil:
fixed by https://github.com/getkyo/kyo/pull/1791

PR #1791, opened 2026-07-25
[kyo-core] Signal: a constant must never complete `next`

xsistens:
## Problem

`nextWith` is documented as waiting for a value change: "It will wait until the signal's value changes before completing." `Signal.initConst(v).next` did not honor that, completing immediately with the constant.

Everything built on that race then misbehaves. `Signal.awaitAny` returns instantly, `combineLatest(ref, const).next` fires with no change to report, and an `observe` over such a signal busy-loops, re-arming a discarded `Async.sleep(repairInterval)` timer thousands of times a second and leaking the heap until the process is OOM-killed.

The kyo-website SSG has exactly this shape: the search dropdown is `queryRef.combineLatest(searchIndex)` with a constant empty index at build time, and `DocsApp` pairs `route` with a constant `contentLoading` flag. Pairing any constant with a mutable signal triggers it.

## Solution

`initConst`'s `nextWith` becomes `Async.never`: a value that never changes has no next change to report. `current`/`currentWith` are unchanged, and no combinator needs special casing, because a constant now simply never wins their race.

Two degenerate cases follow the same rule and no longer complete either: `zipAll`/`combineLatestAll` over an empty sequence are constants, and `awaitAny(Seq.empty)` watches nothing that could ever change.

`zip` waits for ALL its inputs to change, so pairing one with a constant now yields a `next` that can never fire. That follows from the documented contract and no caller in the repo does it, but it is the sharpest edge of the change.

## Notes

The `initConst`/`initConstWith` tests asserted the old immediate `next`; they now cover `current` and assert that `next` does not fire, as do the empty-sequence tests. Regression test in `SignalTest`: `combineLatest(ref, const).next` does not complete until `ref` actually changes (asserted via `ref.waiters`). Reverting either one-line change fails exactly the tests that pin it and no others.


fwbrasil:
I think this is actually a bug in `Signal.initConst`. The `nextWith` contract implies firing only on changes.

xsistens:
You're right, that's the better place for it. My first draft deliberately left `initConst.next` alone: I read the immediate completion as an established contract that direct callers might depend on, so I routed the combinators around it through a separate internal channel rather than risk a behavior change. Your point that `nextWith` already means "fires on change" makes that caution unnecessary, and fixing it at the source is much smaller. Done in 1d6d317: `initConst`'s `nextWith` is now `Async.never`, and the internal change channel is gone entirely. No combinator needs special casing anymore, since a constant simply never wins their race.

Two degenerate cases follow the same rule and no longer complete either: `zipAll`/`combineLatestAll` over an empty sequence are constants, and `awaitAny(Seq.empty)` watches nothing that can change. The `initConst`/`initConstWith` tests that pinned the old immediate `next` are updated to match.

One consequence worth naming: since `zip` waits for all its inputs to change, `zip(ref, const).next` can now never fire. That follows from the contract and nothing in the repo does it, but it is the sharpest edge of the change, so it has a test of its own.

fwbrasil:
it seems this fixes https://github.com/getkyo/kyo/issues/1638

PR #1792, opened 2026-07-25
[kyo-ui] leave-ghost pins its size against inherited %-based min/max (follow-up: #1788)

xsistens:
> **⚠ Depends on #1788 (enter/leave CSS transition lifecycle) — please merge that first.** This refines the leave-ghost that #1788 introduced, so it builds on it. The diff shows #1788's commit until it lands.

## Problem

A leave transition clones the departing element into a `position: fixed` ghost re-parented to `<body>` and pins it to the element's last `getBoundingClientRect`, so it animates out from where it was. But the clone keeps the source's classes — and a percentage min/max size on those classes (e.g. an overlay panel's `min-width: 100%`) now resolves against the ghost's **new** parent, `<body>` (the viewport), overriding the pinned pixel width. The ghost balloons full-width for the duration of the transition.

## Solution

Right after pinning the rect, neutralize the four percentage-sensitive constraints on the ghost — `min-width`/`max-width`/`min-height`/`max-height` — in both transports (browser mount `DomBackend` and server-push `HtmlRenderer.clientJs`), so the pixel pin holds regardless of what the cloned classes declare.

## Notes

Covered by the existing enter/leave ghost scenario test (the ghost is still spawned and torn down correctly — no regression). The change is a layout-constraint neutralization on the already-tested ghost element.

### DIFF
--- a/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
+++ b/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
@@ -41,6 +41,7 @@ private[kyo] object DomBackend:
             html <- HtmlRenderer.render(ui, Seq.empty)
             _    <- Sync.defer(container.innerHTML = html)
             _    <- applyJsProps(container)
+            _    <- Sync.defer(seedEnter(container, Set.empty))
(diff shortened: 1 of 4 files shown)


fwbrasil:
merge conflicts

PR #1793, opened 2026-07-25
[kyo-ui] add preventScrollKeys: suppress native page scroll for nav keys

xsistens:
## Problem

Consider a custom listbox: the user holds `ArrowDown` to move the highlight through the options. The element's `onKeyDown` moves the highlight — but the browser *also* scrolls the whole page, because `ArrowDown` is a page-scroll key and nothing declined that default. The list scrolls out from under the user.

A kyo-ui handler cannot fix this itself. Handlers here run asynchronously — and on the server-push transport they run on the server, a network round trip away — so by the time a handler could call `preventDefault()`, the browser has already scrolled. Declining a browser default has to happen synchronously, in the browser, at the moment of the keydown.

## Solution

Add `preventScrollKeys`, which marks an element and its subtree as a keyboard-navigation region. It emits a `data-kyo-scroll-keys` attribute, and both transports (the browser-mount `DomBackend` and the server-push client script) read it in the browser: while focus is on that element or a descendant, they call `preventDefault()` synchronously for the page-scrolling navigation keys, then still forward the keydown so the element's own `onKeyDown` runs. The key is suppressed for scrolling, not swallowed.

Vertical keys (`ArrowUp`/`ArrowDown`/`PageUp`/`PageDown`) are always suppressed. Horizontal and edge keys (`ArrowLeft`/`ArrowRight`/`Home`/`End`) are suppressed only when the focused target is not a text-editable field (`<input>`/`<textarea>`/`<select>`/`contenteditable`), so caret movement inside a filter input keeps working.

## Notes

`FocusableTest` gains a scenario: a focused `preventScrollKeys` button sits on a deliberately tall (3000px) page. Pressing `PageDown` leaves `window.scrollY` at 0 while the button's `onKeyDown` still fires (verified: without the fix the page scrolls and the assertion fails).

### DIFF
--- a/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
+++ b/kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
@@ -362,6 +362,14 @@ private[kyo] object DomBackend:
         var clickSubmitGuard = false
 
         val handler: scalajs.js.Function1[dom.Event, Unit] = (e: dom.Event) =>
+            // Runs before path resolution so it fires even when the focused descendant is not a path element;
+            // the keydown is still forwarded below, so the region's own onKeyDown runs (see `scrollKeyPrevented`).
+            if e.`type` == "keydown" then
+                val ke  = e.asInstanceOf[dom.KeyboardEvent]
+                val tgt = e.target.asInstanceOf[dom.Element]
+                if tgt != null && scrollKeyPrevented(ke.key, tgt) && tgt.closest("[data-kyo-scroll-keys]") != null then
+                    e.preventDefault()
+            end if
             findPathElement(e.target.asInstanceOf[dom.Element]).foreach { target =>
                 val path    = parsePath(target.getAttribute("data-kyo-path"))
                 val evTypes = ChainTypes(target)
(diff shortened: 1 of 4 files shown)


kyo-ui/js-wasm/src/main/scala/kyo/internal/DomBackend.scala
```diff
@@ -374,4 +382,21 @@ private[kyo] object DomBackend:
...
+      * input keeps caret movement.
+      */
+    private def scrollKeyPrevented(key: String, target: dom.Element): Boolean =
+        def editable =
+            val tag = target.tagName
+            tag == "INPUT" || tag == "TEXTAREA" || tag == "SELECT" ||
+            target.asInstanceOf[scalajs.js.Dynamic].isContentEditable.asInstanceOf[Boolean]
+        end editable
+        key match
+            case "ArrowUp" | "ArrowDown" | "PageUp" | "PageDown" => true
```
fwbrasil: Isn't `"ArrowUp" | "ArrowDown"` needed in case of an editable multi-line text box?

xsistens: Good point. In a textarea (or contenteditable) the default for ArrowUp/Down and PageUp/Down is caret movement inside the field, not a page scroll, so preventing it there just breaks line navigation without gaining anything. Same goes for a native `<select>`, which uses the vertical arrows to change the selected option.

Pushed a fix: vertical keys now pass through when the focused target consumes them itself (textarea, select, contenteditable), in both transports, plus a test that caret movement in a textarea inside a region keeps working (fails without the fix).

Single-line inputs stay suppressed on purpose. The case behind that: a searchable dropdown, where focus sits in the filter input while ArrowDown is supposed to move the highlighted entry in the option list below (via the element's `onKeyDown`). A single-line input has no own use for vertical arrows, so nothing is lost — but without the suppression the page scrolls while you navigate the options.

fwbrasil (approved):

fwbrasil:
@xsistens there's a conflict. Please focus on clearing the backlog of PRs. It's becoming difficult to manage.

fwbrasil:
conflict

PR #1794, opened 2026-07-26
[kyo-ui] comment-marker region delimiters + in-place range morphing (both transports) (follow-up: #1780)

xsistens:
> **⚠ Depends on #1780 (UI.mounted): please merge that first.** The diff shows #1780's commit until it lands.

## Problem

Reactive regions (`Reactive`/`Foreach`/`Mounted`/`Boundary`) paint as a wrapper element, `<span data-kyo-path=".." data-kyo-reactive>` (`<g>` in SVG), and updates replace that element's subtree. That representation is structurally broken in every parser-sensitive context, on `main` today:

1. **Tables, initial paint.** Both transports produce the page via an HTML string parse (`innerHTML` / full document). A `<span>` between `<table>` and its rows is invalid, so the parser *foster-parents* the wrapper out of the table: `table(rows.foreach(r => tr(td(r))))` paints its anchor before the table, subscribe-side paths and painted DOM disagree, and every subsequent patch targets a node outside the table. A body-wide text assert can't see this: the existing TableTest "foreach rows add remove" passed while the rows' updates landed outside the table.

2. **Tables, patch path.** Region updates re-parse from an HTML string via `<template>`. The fragment parser runs in "in body" mode, where a `<tr>` start tag is a parse error that is *silently dropped*: a row-region payload loses all its rows even when the anchor is found.

3. **Select.** The "in select" insertion mode drops unknown tags wholesale, so a reactive region as a direct `<select>` child never paints its wrapper and its updates are dead.

There is no wrapper tag that fixes this: in `<tr>` context a `<td>` anchor start tag *closes* the open cell ("in cell" mode), so no element can hold cell content through the parser.

## Solution

Regions render as **comment-node markers** (`<!--kyo:PATH-->content<!--/kyo:PATH-->`), the representation lit-html/Vue/Solid use for the same reason: comments are valid in every parse context (table, tr, select, svg: the `<g>` special case disappears too), and the region injects no element into user markup (`[data-kyo-reactive] { display: contents; }` leaves the base stylesheet).

- **Registry instead of querySelector.** Comments are not selectable; a `path -> (start, end)` registry is built by a `TreeWalker` scan after the initial paint, refreshed by range-scoped rescans after each patch, and connectivity-validated on lookup (stale → one full rescan → retry → silent no-op, preserving the pre-reveal Boundary contract).
- **Range morphing.** `morphChildren` generalizes to a sibling-range morph bounded by the markers; a nested region's marker span reconciles as ONE logical child keyed by its region path (never positionally: mispairing would corrupt marker text). Out-of-range siblings are never touched.
- **Context-aware payload parse.** Region payloads are bare content fragments parsed in the context the live parent dictates (`<table><tbody>`, `<table><tbody><tr>`, `<select>`, `<svg>` wraps), fixing (2) for good.
- **Mount flags ride the markers.** `data-kyo-mount`/`data-kyo-mount-slot` become ` m`/` s` flags on the open marker; `HtmlOp.Replace` carries `mount` explicitly (the payload no longer has a wrapper root to carry it).
- **Element regions stay wrapper-free.** A signal-bound field (`input.value(ref)`) is its own region root and is morphed 1:1 in place: the transparent wrapper `<span>` that used to appear around such fields on their first patch is gone entirely.
- **Parser-implied `<tbody>` repair.** The whole-page parser keeps an open marker that precedes a table's first row as a `<table>` child while the rows land in the implied `<tbody>`; scan-time pairing adopts the marker into the row group so the pair shares a parent again.
- **Marker escaping.** `%`→`%25`, `-`→`%2D`, space→`%20` per path: Foreach keys are arbitrary user strings and must not be able to terminate the comment (`-->`).

Both transports change in lockstep (SPA `DomBackend` + the server-push inline client), same as the base morphing work in this PR.

## Tests

- `TableRegionTest` (new, real Chrome, structural asserts: `table tr` / `select option`, not body-wide text): rows as direct table child paint inside the table and patch; reactive cell inside `tr`; keyed row add/remove/reorder; static sibling rows untouched by range patches; mounted rows in a table; nested (`$r`) region in a table; reactive options in `<select>`.
- `RegionMarkerTest` (new): the marker byte-format contract incl. hostile-key escaping.
- `MountedTest` gains two regressions for reactive-family children of `fragment(...)` splices (subscribe-side paths normalized at the child path, matching the painted markers).
- Full kyo-ui JVM suite green on this branch: 2091 tests, 0 failures (28 min, real Chrome).

### Path selectors escape user-provided keys

Keyed-list keys are user data and become `data-kyo-path` segments; both transports query elements with `[data-kyo-path="..."]` attribute selectors. A key containing a double quote or backslash broke (or redirected) the query. Both selector builders now escape those metacharacters: `pathSelector` on the SPA side and `__kyoPathSel` in the embedded client, kept in lockstep (the client helper is built via `fromCharCode` because the embedded script lives in an interpolated Scala literal whose escape processing must not touch the emitted JS).

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -121,6 +121,12 @@ object Fiber:
 
     /** Runs an asynchronous computation in a new Fiber guaranteeing eventual interruption via the [[Scope]] effect.
       *
+      * If a [[Fiber.Supervisor]] is installed in the current context (see [[Supervisor.let]]), the spawned fiber is additionally
+      * SUPERVISED: any non-interrupt terminal failure (an `Abort` failure or a panic that is not an [[Interrupted]]) is reported to the
+      * supervisor exactly once. Interruptions (including the scope-close interrupt this very method registers) are never reported.
+      * With no supervisor installed (the default), behavior is unchanged. [[initUnscoped]] and [[use]] never supervise: they are the
+      * escape hatches for fire-and-forget fibers and for fibers whose failure the caller observes itself.
+      *
       * @param v
       *   The computation to run
       * @return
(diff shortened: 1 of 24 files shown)


fwbrasil:
merge conflicts

fwbrasil:
https://github.com/getkyo/kyo/pull/1780 has some major open questions so I'm closing this for now

PR #1795, opened 2026-07-26
[kyo-ui] pointer/measure/command wire-protocol extension (UI.Commands) (follow-up: #1782, #1794)

xsistens:
> **⚠ Depends on #1780, #1782, and #1794: please merge those first.** The pointer setters use the effect-typed handler form and the per-session `Env` from #1780, and pointer dispatch opts into #1782's `keepBubbling`; the branch is stacked on #1794 to keep the series linear (this feature does not itself depend on morphing). The diff shows their commits until they land.

## Problem

Some component interactions cannot be expressed as a render diff. A drag needs a stream of pointer positions that keeps flowing after the pointer leaves the element. An overlay that flips above vs. below its trigger needs the trigger's measured geometry *before* it renders. An OTP field needs to move focus to the next box, and a chat log needs to scroll to the bottom: imperatively, from a handler, without re-rendering the whole subtree. Today render diffs are the only server→client channel, so none of these are possible.

## Solution

A cohesive extension of the client↔server protocol, tied together by a new per-session `UI.Commands` channel, so the parts land together:

- **Pointer events with drag sessions.** `onPointerDown`/`onPointerMove`/`onPointerUp`, carrying a `UI.PointerEvent` (position relative to the target, its bounding rect, the button mask; derives `Schema` as the wire payload). Pointer-down calls `setPointerCapture`, so the move stream keeps flowing even when the pointer leaves the element until pointer-up releases it; moves are rAF-coalesced to one per frame. Both transports. Unlocks drag components: a color picker, a slider drag, a draggable dialog.
- **`UI.Commands`**: the first imperative server→client channel a component can drive. Provided per session via `Env[Commands]`, obtained with `UI.commands` in any event handler or `UI.mounted` effect. Built over a per-transport `emit: HtmlOp => Unit < Async`: the server-push transport wraps the session body in `Env.run(commands)` on the session fiber; the browser mount applies ops to the local DOM directly.
- **Measure round-trip.** `requestMeasure(path)(sink)` emits a `RequestMeasure`; the client replies with a `Measure` event that the transport routes to `deliverMeasure`, outside the element dispatch tree, since it resolves a pending request, not an `onXxx` handler. The pattern is: request → callback updates a `Signal` → re-render with correct geometry.
- **Command op.** `Command(path, verb)`, whitelisting `focus` and `scrollIntoView` (unknown verbs are ignored, so the whitelist can grow compatibly). Unlocks OTP auto-advance and auto-scroll.
- **Self-addressing (id-addressed twins).** `commandById` / `requestMeasureById`, plus `freshId`. A reusable component mints an id with `UI.Commands.freshId`, stamps it on its own element (`id` is a settable prop), and then drives commands *at itself* by id, without ever learning its own structural render path. The client resolves the target with `getElementById`; the verb whitelist and the measure round-trip are shared with the path-addressed forms.

## Notes

Tested with real-Chrome scenario tests on the server-push transport: `PointerScenarioItTest` (drag session start → move stream → release), `MeasureScenarioItTest` (request → reply → geometry), `CommandScenarioItTest` (focus / scrollIntoView verbs), `SelfAddressScenarioItTest` (freshId → self-stamp → focusById / requestMeasureById). Commands used inside a `UI.mounted` effect work because the subscription that forks the mount fibers runs inside `Env.run(commands)`. The morph, mount, event-wiring and render suites pass unchanged.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -121,6 +121,12 @@ object Fiber:
 
     /** Runs an asynchronous computation in a new Fiber guaranteeing eventual interruption via the [[Scope]] effect.
       *
+      * If a [[Fiber.Supervisor]] is installed in the current context (see [[Supervisor.let]]), the spawned fiber is additionally
+      * SUPERVISED: any non-interrupt terminal failure (an `Abort` failure or a panic that is not an [[Interrupted]]) is reported to the
+      * supervisor exactly once. Interruptions (including the scope-close interrupt this very method registers) are never reported.
+      * With no supervisor installed (the default), behavior is unchanged. [[initUnscoped]] and [[use]] never supervise: they are the
+      * escape hatches for fire-and-forget fibers and for fibers whose failure the caller observes itself.
+      *
       * @param v
       *   The computation to run
       * @return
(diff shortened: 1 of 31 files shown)


fwbrasil:
Please avoid creating PRs, even if draft, if they depend on multiple other PRs

PR #1796, opened 2026-07-26
[kyo-ui] in-place reactive attribute patching + viewport observation (follow-up: #1795)

xsistens:
> **⚠ Depends on #1780, #1782, #1794, and #1795: please merge those first.** This builds on the `UI.Commands` channel (#1795, id-addressed measure and freshId) and the morph (#1794, whose `morphAttrs` learns the ownership shield here); the branch is stacked accordingly. The diff shows their commits until they land.

## Problem

A reactive binding replaces its element via `outerHTML`, which resets any in-flight CSS transition, so you cannot fade a panel in by toggling a class, or animate a bound style, through the normal reactive channel. And a component that needs to react to *its own* viewport geometry (an overlay that repositions on scroll/resize) has no way to observe it continuously.

## Solution

Two capabilities on the id-addressed imperative channel:

- **In-place patching**: `bindClassById` / `bindStyleById` fork a mount-scoped fiber over a `Signal` and emit `SetClassById` / `SetStyleById`, which the client applies to `getElementById(id)` **without replacing the element**, so CSS transitions on the class/style fire. Style merges per-declaration (no `style=""` clobber).
- **Viewport observation**: `observeViewportById` registers a persistent sink and emits `ObserveViewportById`; the client measures now and attaches capturing scroll + resize listeners that stream `MeasureById(id)` replies. A `Scope` finalizer emits `UnobserveViewportById` and drops the sink.

Because these ops write class/style *out of the render pass*, the in-place morph must not reconcile them away on a later re-render (server HTML never carries the client-set value). Ownership is tracked **on the element** as a `__kyoOwn` expando dict of owned attribute names: `SetClassById`/`SetStyleById` mark their element, and `morphAttrs` skips any attribute whose name is owned. The flag is reclaimed with the node: no session-lived set that only ever grows (`freshId` mints a unique id per component instance). Mirrored across both transports.

## Notes

Test (`DomBackendTest`, server-push): a class bound via `bindClassById` survives a real re-render morph of its element (a sibling signal bumps the enclosing reactive region, so `morphAttrs` runs against server HTML that lacks the class). Without the ownership shield the morph reconciles the class away, verified (the test fails when the shield is removed). The per-attribute (rather than class/style-only) shape of the shield anticipates the id-addressed attribute channel (`bindAttrById`) in a follow-up.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -121,6 +121,12 @@ object Fiber:
 
     /** Runs an asynchronous computation in a new Fiber guaranteeing eventual interruption via the [[Scope]] effect.
       *
+      * If a [[Fiber.Supervisor]] is installed in the current context (see [[Supervisor.let]]), the spawned fiber is additionally
+      * SUPERVISED: any non-interrupt terminal failure (an `Abort` failure or a panic that is not an [[Interrupted]]) is reported to the
+      * supervisor exactly once. Interruptions (including the scope-close interrupt this very method registers) are never reported.
+      * With no supervisor installed (the default), behavior is unchanged. [[initUnscoped]] and [[use]] never supervise: they are the
+      * escape hatches for fire-and-forget fibers and for fibers whose failure the caller observes itself.
+      *
       * @param v
       *   The computation to run
       * @return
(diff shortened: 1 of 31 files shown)


fwbrasil:
Please avoid opening PRs that depend on multiple other PRs

PR #1797, opened 2026-07-26
[kyo-ui] declarative reactive attribute/class/boolean binding channels (follow-up: #1796)

xsistens:
> **⚠ Depends on #1780, #1782, #1794, #1795, and #1796: please merge those first.** This is the ergonomic API over the in-place channel (#1796) and the id-addressed measure/self-address infrastructure (#1795), and it extends the morph's ownership shield (#1794/#1796); the branch is stacked accordingly. The diff shows their commits until they land.

## Problem

A reactive attribute or class today rides a full region re-render: the element is repainted, which resets any in-flight CSS transition. So you cannot fade a badge between severities by toggling a class, track `aria-expanded` on a toggle button, or reactively disable a field, without a re-render that loses the animation.

## Solution

The declarative API over the id-addressed in-place channel: a reactive `Signal` drives an attribute, class, or boolean attribute of an element **in place** (no re-render, no re-mount), so CSS transitions fire and structural selectors keep matching.

- **`bindAttrById` / `SetAttrById`**: the self-addressing (by DOM id) twin of the class/style ops, for a component that stamped its own id.
- **Declarative reactive channels**, bound on the element and observed by a mount-scoped fiber the region's subscription forks:
  - a reactive string attribute: `placeholder(Signal[String])`, `aria`, `title` → `SetAttrByPath`;
  - a reactive class: `cssClass(name, Signal[Boolean])` → `SetClassByPath` (`classList.toggle`, so transitions fire);
  - a reactive boolean attribute: `disabled(Signal)`, `readOnly(Signal)` → `SetBoolAttrByPath` (presence toggled in place).

  The element renders the signal's **current** value as an ordinary attribute, so SSR / first paint is byte-identical; the observer patches it on later emissions. The setters accept a `Boolean | Signal[Boolean]` union so the constant and reactive forms share one call site, and the three channel setters share three inline helpers.

Because these writes land out of the render pass, the morph must not reconcile them away on a later re-render. Ownership lives **on the element** as a `__kyoOwn` expando dict of owned attribute names (reclaimed with the node, no session-lived set that grows); every in-place write marks its attribute, and `morphAttrs` skips any owned name. Both transports.

## Notes

Tests (`DomBackendTest`, server-push): an imperatively-bound attribute, and a declaratively-bound reactive string attribute, class, and boolean attribute, each patch in place and **survive a re-render morph** of their element (a sibling signal bumps the enclosing region so `morphAttrs` runs against server HTML that lacks the value), plus a reactive attribute bound inside a `UI.mounted` block. SSR golden HTML stays byte-identical (a reactive binding renders its current value as a plain attribute).

### Fragment row roots with binding channels

The element-children walk promoted an element carrying a binding channel (bound input ref or reactive attr/class/boolean channel) to its own reactive node, but the fragment-children walk (keyed `Foreach` rows) did not: a row root like `tr.cssClass("danger", selected.map(_ == row.id))` accepted the setter and its observers silently never started. Both walks now share a single promotion predicate (`needsOwnNode`), the two duplicated channel-observer fork blocks are folded into one helper (`forkChannelObservers`), and a browser test pins the row-root class patch across select, keyed reorder, and post-reorder re-subscription.

### Skip of redundant first emissions after subscribe

Every `observe` fires once with the current value on subscription. After a region paints (and after every keyed-list re-render re-subscribes all children), each child region and channel observer therefore immediately repainted exactly what the enclosing render had just painted: on a 1k-row keyed list this compounds into thousands of redundant patches trailing every paint. Normalize now captures render-time snapshots (normalize happens-before the HTML render on every paint path, so skipping a first emission that still equals its snapshot is safe by construction); the skip path still claims mounts and subscribes the already-walked children. A change landing between render and subscribe differs from the snapshot and still fires. Direct-engine tests pin both directions (no redundant op after subscribe; a racing change is delivered), and a browser test pins that a keyed row whose root is itself a reactive region patches at the row path.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -121,6 +121,12 @@ object Fiber:
 
     /** Runs an asynchronous computation in a new Fiber guaranteeing eventual interruption via the [[Scope]] effect.
       *
+      * If a [[Fiber.Supervisor]] is installed in the current context (see [[Supervisor.let]]), the spawned fiber is additionally
+      * SUPERVISED: any non-interrupt terminal failure (an `Abort` failure or a panic that is not an [[Interrupted]]) is reported to the
+      * supervisor exactly once. Interruptions (including the scope-close interrupt this very method registers) are never reported.
+      * With no supervisor installed (the default), behavior is unchanged. [[initUnscoped]] and [[use]] never supervise: they are the
+      * escape hatches for fire-and-forget fibers and for fibers whose failure the caller observes itself.
+      *
       * @param v
       *   The computation to run
       * @return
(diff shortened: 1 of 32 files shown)


fwbrasil:
Please avoid creating PRs that depend on multiple other PRs

PR #1798, opened 2026-07-26
[kyo-ui] onScrollPosition: native scroll-position event (follow-up: #1780, #1782, #1784)

xsistens:
> **⚠ Depends on #1780, #1782, and #1784 — please merge those first.** The `onScrollPosition` setter uses the effect-typed handler form from #1780, dispatch opts into #1782's `keepBubbling`, and the SPA capture-phase listener uses #1784's ancestor-chain forwarding. The diff shows their commits until they land.

## Problem

kyo-ui has `onScroll` — a per-notch mouse-wheel delta — but no way to observe an element's actual scroll **position**. Server-authoritative virtual scrolling needs it: the browser scrolls a full-height spacer natively and the server repositions the rendered window from the reported `scrollTop`. A wheel delta cannot give that — it misses scrollbar drags, touch scrolling, and keyboard scrolling, and it is a delta rather than the resulting offset.

## Solution

Add `onScrollPosition`, carrying a `UI.ScrollPositionEvent` (`scrollTop` / `scrollLeft` / `targetId`), wired in both transports. Because the browser owns the scroll and reports the resulting position, it fires for every scroll input — wheel, scrollbar drag, touch, keyboard.

Scroll does not bubble, so a capture-phase body listener catches descendant viewport scrolls; a scroll burst is rAF-coalesced to one event per frame. A page-level scroll whose target has no `data-kyo-path` resolves to nothing and is ignored. Declaring it emits the `scroll` token in `data-kyo-ev`, and dispatch bubbles through the reactive tree with the same `stopPropagation` semantics as the other events.

## Notes

Test (`EventHandlerTest`, server-push): an element with `onScrollPosition` over a short `overflow:auto` viewport reports its `scrollTop` to the handler after the element is scrolled.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -121,6 +121,12 @@ object Fiber:
 
     /** Runs an asynchronous computation in a new Fiber guaranteeing eventual interruption via the [[Scope]] effect.
       *
+      * If a [[Fiber.Supervisor]] is installed in the current context (see [[Supervisor.let]]), the spawned fiber is additionally
+      * SUPERVISED: any non-interrupt terminal failure (an `Abort` failure or a panic that is not an [[Interrupted]]) is reported to the
+      * supervisor exactly once. Interruptions — including the scope-close interrupt this very method registers — are never reported.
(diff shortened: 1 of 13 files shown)


fwbrasil:
Please avoid opening PRs that depend on multiple other PRs

PR #1799, opened 2026-07-26
[kyo-system] add the overlay filesystem and stage writes over any backend

### LINKED ISSUE #1815: Process: stdin feeds spawn daemon threads outside Scope, so they outlive and ignore cancellation
Part of #1813.

`feedInputStream` and `feedStream` in `kyo-core/jvm-native/src/main/scala/kyo/internal/ProcessPlatformSpecific.scala` each start a raw daemon thread to write into the child's stdin:

```scala
private def feedInputStream(is: InputStream, processStdin: OutputStream): Unit =
    val t = new Thread(() =>
        try
            val buf = new Array[Byte](8192)
            var n   = is.read(buf)
            while n >= 0 do
                processStdin.write(buf, 0, n)
                n = is.read(buf)
            processStdin.close()
        catch case _: IOException => ()
        finally
            try is.close()
            catch case _: IOException => ()
    )
    t.setDaemon(true)
    t.start()
```

`feedStream` does the same with a pre-evaluated byte array.

### Problem

The thread is started and the handle discarded. There is no `Scope` registration, no reference retained, and no cancellation path, so:

- Interrupting the fiber that spawned the process does not stop the feed. The thread keeps reading its source and writing to the child's stdin.
- Nothing joins it, so a caller cannot know when the feed has finished or whether it failed. `IOException` is swallowed on both paths.
- The lifetime is the JVM's, not the computation's. Being a daemon thread means it does not block shutdown, but that is the only bound on it.

This is the same class of ownership problem as #1814, from the other direction: there the work is on a scheduler thread that should not be blocked, here it is on a thread the effect system does not know about at all.

### Suggested direction

Tie the feed to the `Scope` that owns the process, so it is cancelled when the process handle is released and its failure is observable rather than swallowed.

### Notes

Longstanding and present on `main`. #1799 moves the file to `kyo-system` unmodified; the `feedInputStream` and `feedStream` region is byte-identical.

### LINKED ISSUE #1816: Process: waitFor(timeout) measures its deadline with a platform timer instead of the ambient Clock
Part of #1813.

`Process.waitFor(timeout)` delegates to the platform's own timer: `CompletableFuture.orTimeout` on JVM and Native, `setTimeout` on JS.

```scala
// kyo-core/shared/src/main/scala/kyo/Process.scala
def waitFor(timeout: Duration)(using Frame): Maybe[ExitCode] < Async =
    Sync.Unsafe.defer(self.unsafe.waitFor(timeout).safe.get)
```

```scala
// kyo-core/jvm-native/src/main/scala/kyo/internal/ProcessPlatformSpecific.scala
discard(jp.onExit()
    .orTimeout(timeout.toMillis, TimeUnit.MILLISECONDS)
    ...
```

### Problem

Both timers are correct in wall time, but neither consults the ambient `Clock`, so the deadline does not move under `Clock.withTimeControl`.

Two consequences:

- The behaviour cannot be asserted except by waiting the deadline out in real time. A test for "returns `Absent` on expiry" costs the full timeout.
- It is load-sensitive. On a busy machine the process can be carried past a deadline the caller believed it controlled, and there is no way to make that deterministic from a test.

The signature already declares `Async`, so no widening is needed to express the deadline against `Clock`; the row was simply not being used.

### Note on testing this

Asserting only that the wait returns `Absent` under `Clock.withTimeControl` does not distinguish the two implementations. A platform timer reaches the same answer, just by burning the full timeout in real time. The distinguishing assertion is on elapsed wall time: with a `Clock`-driven deadline almost none passes.

### Status

Fixed in #1799 by routing through `Async.timeout`, which builds the deadline from the ambient `Clock` and arms no timer for a non-finite duration. The `Unsafe` tier keeps its platform timer, since it sits outside the effect system and cannot consult the ambient `Clock`.

Longstanding and present on `main`; #1799 moves the file unmodified.

### LINKED ISSUE #1817: Process: waitFor() never returns for an already-signalled process on JS
Part of #1813.

On JS, `Process.waitFor()` checks whether the child has already exited before registering an `exit` listener. That pre-check reads only `child.exitCode`:

```scala
// kyo-core/js-wasm/src/main/scala/kyo/internal/ProcessPlatformSpecific.scala
val ec = child.exitCode
if ec != null && !js.isUndefined(ec) then
    resolved = true
    p.completeDiscard(Result.succeed(Process.ExitCode(ec.asInstanceOf[Int])))
end if
if !resolved then
    discard(child.on("exit", ...))
```

Node leaves `exitCode` as `null` when a child dies from a signal, recording the signal in `signalCode` instead. So for a signalled child the pre-check does not fire, and the code falls through to registering an `exit` listener. If the `exit` event has already been emitted by then, that listener never runs, nothing completes the promise, and the wait never returns.

### Reproduction

```scala
Scope.run {
    for
        proc   <- Command("sleep", "60").spawn
        _      <- proc.destroyForcibly
        result <- proc.waitFor          // never completes on JS
    yield result
}
```

The same program returns `ExitCode.Signaled(9)` on JVM.

### Why it was not caught

The timed overload `waitFor(timeout)` used a separate platform path that did not go through this pre-check, so a test using a timeout would return `Absent` on expiry rather than hanging. Only the untimed `waitFor` reaches it.

### Status

Fixed in #1799: the pre-check now consults `signalCode` as well and builds the result through the existing `exitCodeFrom` helper, so an already-signalled child resolves immediately.

Longstanding and present on `main`; #1799 moves the file with only the fixes described here.

DamianReeves:
### What this adds

The overlay filesystem, the top of the path-capability stack. `OverlayFileSystem` stages writes in an upper layer over any lower `FileSystem` and applies them on commit. The upper is a `PathTrie` keyed by paths resolved through the lower, so a subtree staged beneath a symlinked parent commits under the name the lower resolves.

`Path.stageWrites`, `Path.commitWritesOnSuccess` and `Path.discardWrites` run a `PathWrite` program against an overlay over the installed service, so staging is reachable without naming a backend. Commit derives its plan from the staged state, re-checks every observation the work was built on against the live lower, and aborts with `CommitConflict` when one no longer holds; `commitWith` resolves conflicts instead. A commit survives a crash: content stages first, an intent log records the plan, application follows, and a marker declares completion. `FileSystem.overlayRecovering` replays a leftover staging directory.

The conformance, snapshot and watch suites gain the overlay beside the host, in-memory and zip backends, and `demo/ConfigWorkspaceDemo` runs the surface end to end.

Staged keys follow the lower backend’s case policy while preserving entry spelling for listings and replay. Alias moves preserve content and identity, and empty channel creation records provenance. Directory validation and acquisition cleanup follow the updated lower-stack contracts. Synchronous overlay factories expose `FileSystem.Watch`; generic factories preserve arbitrary lower effects in reads and writes.

Host commits preserve destination access controls on staged replacement files. Recovery logs containing file contents use private staging directories. Directory privacy and permission-source metadata survive nested overlays and intent-log replay. New files inherit their final destination parent's access controls, including through nested overlays and recovery. Directory resolution retains privacy requirements for child writes, and durable apply markers prevent recovery from clearing completed children. Directory sync plans survive restarts, including newly created ancestors and retries after deletion. The unreleased intent-log format is version 6.

Linux private staging directories restore owner traversal after inherited default ACLs remove it from `mkdir(0700)`. Restoration uses the exclusively created, identity-checked descriptor and keeps group, other, and named ACL access masked.

Host commits require directory synchronization. Windows host commits fail at the staging-parent barrier before publishing staged targets. Staged reads and discarded writes remain available; in-memory commits work on Windows. Recovery of an existing intent can partially apply its plan before a directory barrier fails and retains the intent for retry. Shared tests verify the exact Windows error, retained data, private staging, partial replay, and cleanup.

### Stack

The lower PRs through #1867 are in `main`. The remaining stack is #1868 → #1869 → #1870 → #1871 → #1799, rebased onto `origin/main` (`3b8679117`). Merge bottom up.

### Validation

Full `kyo-system` tests on JVM, JS, Native, and Wasm, plus JVM doctests. Regression coverage includes the updated lower-stack contracts. Local runtime checks ran on macOS.

Follow-up validation passed: 1,337 JVM, 1,303 JS, 1,281 Native, and 1,258 Wasm tests, plus 41 doctests. The [Linux ACL probe](https://github.com/getkyo/kyo/actions/runs/35159099174) reproduced the original failure and verified private creation, replay, collisions, and unchanged parent permissions. The [Windows ARM64 run](https://github.com/getkyo/kyo/actions/runs/35161614377) passed the full JVM and Node system suites, including the explicit commit and recovery failure tests.


kyo-aeron/shared/src/test/scala/kyo/AeronTransportTest.scala
```diff
@@ -589,7 +589,7 @@ class AeronTransportTest extends Test:
                                 subResult.isEmpty,
                                 s"iteration $i: asyncAddSubscription on closed client must return Absent; got $subResult"
                             )
-                        }.andThen(dir.removeAll)
```
fwbrasil: is this change intentional?

kyo-browser/shared/src/main/scala/kyo/internal/ChromeDownloader.scala
```diff
@@ -77,7 +77,7 @@ private[kyo] object ChromeDownloader:
             root     <- cacheRoot
             versionDir = root / s"${artifactName(build)}-$v-$platform"
             exec       = executablePath(versionDir, platform, build)
-            cached <- exec.exists
+            cached <- Abort.recover[FileSystemException](_ => false)(Path.runReadOnly(exec.exists))
```
fwbrasil, line 80: Isn't it safer to map the exception type instead of falling back to `false`?

kyo-browser/shared/src/test/scala/kyo/internal/BrowserLauncherTest.scala
```diff
@@ -202,9 +201,8 @@ class BrowserLauncherTest extends BaseBrowserTest:
     "pollDevToolsActivePort returns the URL when DevToolsActivePort exists and is well-formed" in {
         Scope.run {
             for
-                tmp <- Path.tempDir("kyo-browser-pollDevTools-happy-")
-                _   <- Scope.ensure(Abort.run[FileFsException](tmp.removeAll).unit)
```
fwbrasil: is this removal intentional?

kyo-browser/shared/src/test/scala/kyo/internal/ChromeDownloaderTest.scala
```diff
@@ -52,12 +52,8 @@ class ChromeDownloaderTest extends BaseBrowserTest:
     // ---- version override path is reflected in the cached directory ----
 
     /** Creates a temp directory that auto-deletes when the enclosing scope closes. */
-    private def tempDirScoped(prefix: String)(using Frame): Path < (Scope & Sync & Abort[FileFsException]) =
-        Path.tempDir(prefix).map { p =>
-            Scope.acquireRelease(p) { dir =>
-                Abort.run[FileFsException](dir.removeAll).unit
```
fwbrasil: I guess `tempDir` already registers the finalizer?

kyo-browser/shared/src/test/scala/kyo/BrowserDownloadTest.scala
```diff
@@ -122,7 +122,7 @@ class BrowserDownloadTest extends BrowserTest:
                 // Inverse poll: loop exits cleanly iff file never landed; Abort.fail = file landed.
                 _ <- assertNeverLands(tempPath / fileName, s"denyDownloads-no-landing-at-$fileName", 10, 50.millis)
                 // Confirm the file is absent after the full poll window (the deny contract was upheld).
-                absent <- (tempPath / fileName).exists
+                absent <- Abort.recover[FileSystemException](_ => false)(Path.runReadOnly((tempPath / fileName).exists))
```
fwbrasil, line 125: It seems we could let just let the failure propagate

kyo-core/shared/src/main/scala/kyo/Async.scala
```diff
@@ -72,11 +76,7 @@ object Async extends AsyncPlatformSpecific:
       *
       * Note: This only affects collection processing methods. Operations like race and gather run with unbounded concurrency.
       */
-    val defaultConcurrency =
-        import AllowUnsafe.embrace.danger
-        given Frame = Frame.internal
-        Sync.Unsafe.evalOrThrow(System.property[Int]("kyo.async.concurrency.default", Runtime.getRuntime().availableProcessors() * 2))
-    end defaultConcurrency
+    val defaultConcurrency: Int = async.concurrency.default()
```
fwbrasil, line 90: thanks for the cleanup!

kyo-system/shared/src/main/scala/kyo/StreamFileExtensions.scala
```diff
@@ -0,0 +1,69 @@
...
+  *
+  * These sinks acquire a write handle via a capability op in a `Scope`, stream the content, and remove the
+  * partially-written file on failure (the handle's `finish()` is never reached on failure, triggering
+  * delete-on-close). They live in kyo-system alongside [[kyo.Path]] because they couple to the file
+  * capability; kyo-core keeps the platform-neutral stream combinators.
+  *
+  * @see
+  *   [[kyo.Path]] for the file-path capability the sinks target
+  */
+object StreamFileExtensions:
```
fwbrasil: The pattern we've set is an extension file per module, this should be `StreamSystemExtensions`

DamianReeves:

|---|---|

```scala
}
```

### What this branch needs

Beyond the rename set, `OverlayFileSystem` needs a `temp` member, since `FileSystem.Write` now declares one:

```scala
def temp(prefix: String, suffix: String)(using Frame): Path.TempFileHandle < (S & Abort[FileStructureException])
```

The design call is whether the temp file is staged in the overlay or written straight to the base. Staging it means the overlay's staged-changes contract now covers entries nobody intended to commit; going to the base means a temp file escapes the overlay. Whichever way it goes, it should be stated and should match `tempDir`'s existing choice here.

kyo-aeron/shared/src/test/scala/kyo/AeronTransportTest.scala
DamianReeves: Yes. The current test uses `withEmbeddedRuntime`, whose `Scope` finalizer closes the runtime and removes its directory, including when the body fails. The trailing `andThen(dir.removeAll)` is no longer needed.

kyo-browser/shared/src/test/scala/kyo/internal/BrowserLauncherTest.scala
DamianReeves: Yes. `Path.tempDir` now registers recursive removal with the enclosing `Scope`, through the filesystem service that created it. This test keeps that scope, so the explicit second finalizer is redundant.

kyo-browser/shared/src/test/scala/kyo/internal/ChromeDownloaderTest.scala
DamianReeves: Yes. `Path.tempDir` owns the scoped cleanup now. The helper is now `hostTempDir`, which delegates through `Path.run` and retains the enclosing `Scope`.

kyo-core/shared/src/main/scala/kyo/Async.scala
DamianReeves: Kept this cleanup in the rebased stack: `defaultConcurrency` delegates to the typed configuration accessor.

kyo-system/shared/src/main/scala/kyo/StreamFileExtensions.scala
DamianReeves: This is now `StreamSystemExtensions`. The refreshed stack retains the module-based name that landed in `main`.

kyo-browser/shared/src/main/scala/kyo/internal/ChromeDownloader.scala
DamianReeves: Fixed in [4b5bbd128](https://github.com/getkyo/kyo/commit/4b5bbd128ee1cd1330ec3073f6a4d555478b412d) in the lower stack. A failed cache inspection now returns `BrowserSetupFailedException` with the original filesystem cause and never invokes the downloader. The confinement regression reproduced the unwanted download before the fix and now verifies the exact cause, path, operation, and zero download calls. The browser suites pass 43 tests on each of JVM, JS, Native, and Wasm.

kyo-browser/shared/src/test/scala/kyo/BrowserDownloadTest.scala
DamianReeves: Fixed in [4b5bbd128](https://github.com/getkyo/kyo/commit/4b5bbd128ee1cd1330ec3073f6a4d555478b412d) in the lower stack. Both the polling helper and the final download-denial assertion propagate filesystem errors, so a failed existence check cannot pass as an absent file. The download suite passes all 12 tests on JVM, JS, Native, and Wasm.

PR #1800, opened 2026-07-26
[kyo-eventlog] add event log support

DamianReeves:
This PR contains the eventlog changes split from #1759.

It is stacked on #1799 so the review diff contains only eventlog-related changes. Once #1799 merges, this PR can be retargeted to `main`.

Validation:
- `kyo-eventlogJVM/test`
- `kyo-eventlogJS/test`
- `kyo-eventlogNative/test`
- `kyo-eventlogWasm/test`


DamianReeves:

|---|---|

```scala
}
```

### What this branch needs

The rename set above, nothing more. No `FileSystem.Write` implementation of its own, so the new `temp` member arrives already satisfied by the backends below, and there are no `Path.temp` callsites in `kyo-eventlog` to reclassify.

Being top of the stack, this branch inherits whatever naming decisions the branches below settle on, in particular the `watch` collision raised on #1867.

PR #1803, opened 2026-07-28
Fix schema JSON encoding without private JDK access

### LINKED ISSUE #1801: [BUG]: Kyo-schema fails to run without -add-opens=java.base/java.lang=ALL-UNNAMED
### Version

1.0.0-RC5

### Scala Version

3.8.4

### Expected Behavior

Application compiles well with JVM defaults.

### Actual Behavior

A hello world Kyo app using `kyo-schema` fails on JVM defaults. To make it work it's required to set ` --add-opens=java.base/java.lang=ALL-UNNAMED`.

### Steps to Reproduce

Use the following ScalaCLI reproducer:
```scala
//> using scala 3.8.4
//> using jvm 25
//> using dep io.getkyo::kyo-schema:1.0.0-RC5

import kyo.*

final case class Person(name: String, age: Int) derives Schema

@main def run(): Unit =
  val json = Schema[Person].encodeString[Json](Person("Alice", 30))
  println(s"encoded ok: $json")
```

Run as follows to observe it failing:
```sh
env -u JAVA_TOOL_OPTIONS scala-cli run --server=false .
```
You should observe:
```sh
Exception in thread "main" java.lang.ExceptionInInitializerError
        at kyo.internal.JsonWriter.resultString(JsonWriter.scala:193)
        at kyo.Schema.encodeString(Schema.scala:219)
        at repro$package$.run(repro.scala:30)
        at run.main(repro.scala:27)
Caused by: java.lang.IllegalAccessException: module java.base does not open java.lang to unnamed module @5702b3b1
        at java.base/java.lang.invoke.MethodHandles.privateLookupIn(MethodHandles.java:268)
        at kyo.internal.AsciiStringFactory$.<clinit>(AsciiStringFactory.scala:13)
        ... 4 more
```

Use following to run app successfully:
```sh
 env -u JAVA_TOOL_OPTIONS scala-cli run --server=false \
  --java-opt --add-opens=java.base/java.lang=ALL-UNNAMED .
```
You should get a result:
```sh
encoded ok: {"name":"Alice","age":30}
```

### Current Workaround

We need to set `--add-opens=java.base/java.lang=ALL-UNNAMED` everywhere possible. I set it in Mill build for tests (as running tests fails without flag). Need to add it to `nativeImageOptions` for GraalVm Native Image to  be able to build app successfully. Unfortunately setting it in sbt/mill build is not enough, and to run tests in IDE (IntelliJ) I need to ensure this flag is properly propagated to make IntelliJ's own MUnit runner seeing it (and this part is not that straightforward). If the dev is patient enough, will be able to compile and run kyo app and its tests. Less experienced devs might struggle with it or at least see it as poor DX.

### Pre-analysis

I see for Loom scheduler this flag is optional.  Missing → graceful fallback to normal pool + one  WARNING  log line. Harmless, by design. Just a perf hint.

Regarding `kyo-schema`  `AsciiStringFactory`  - flag is mandatory, no fallback. Missing → hard  `ExceptionInInitializerError`  on first JSON encode. 
This means, unlike Loom, schema has zero fallback - it crashes.

I think it's  brittle: relies on JDK internals; a future JDK could break it.

DamianReeves:
Fixes #1801

### Problem

Kyo Schema JSON encoding initializes a JVM-specific `AsciiStringFactory` that uses `MethodHandles.privateLookupIn` to invoke a private `String(byte[], byte)` constructor. On a standard JVM where `java.lang` is not opened, the first JSON string result fails with `ExceptionInInitializerError` and `IllegalAccessException`.

The repository-wide test JVM option opened `java.lang`, which masked this failure in the module test suite.

### Solution

Construct JSON result strings from the written byte range with the public UTF-8 `String` constructor in shared `JsonWriter` code.

This change also:

- removes the JVM, Scala.js/Wasm, and Native `AsciiStringFactory` implementations
- removes the inherited `java.lang` opening from `kyo-schema-jsonJVM` tests
- adds a JVM regression asserting `java.lang` remains closed while ASCII and UTF-8 output succeeds
- moves direct `JsonWriter` depth and comma-tracking tests into a shared writer suite
- updates maintained schema documentation to describe the shared implementation

### Notes

- Reproduced the reported failure with Kyo 1.0.0-RC5, Scala 3.8.4, and JDK 25 without `--add-opens`.
- Verified the same reproducer against this branch without `--add-opens`: `encoded ok: {"name":"Alice","age":30}`.
- Passed the complete `kyo-schema-json` test suites on JVM, Scala.js, Scala Native, and Wasm.
- Passed `scalafmtCheckAll`.

#### Benchmark

Compared the original copied-buffer plus private-constructor path with the shared public `String(byte[], offset, length, UTF_8)` path on JDK 25.0.1. JMH used 3 forks, 3 one-second warmup iterations, 5 one-second measurement iterations, and the GC profiler. Inputs were exact 32-byte and 4096-byte JSON-shaped ASCII and multibyte UTF-8 payloads. Setup asserted both implementations returned identical strings.

| Payload | Original throughput | Public UTF-8 throughput | Original allocation | Public UTF-8 allocation |
| --- | ---: | ---: | ---: | ---: |
| ASCII, 32 bytes | 129.68M ± 12.54M ops/s | 147.12M ± 0.96M ops/s | 72 B/op | 72 B/op |
| ASCII, 4096 bytes | 2.319M ± 0.227M ops/s | 7.316M ± 0.759M ops/s | 4136.003 B/op | 4136.001 B/op |
| UTF-8, 32 bytes | 31.22M ± 5.12M ops/s | 38.53M ± 0.49M ops/s | 216 B/op | 216 B/op |
| UTF-8, 4096 bytes | 389.5K ± 88.1K ops/s | 439.8K ± 17.5K ops/s | 16024.019 B/op | 16024.016 B/op |

Values show 99.9% confidence intervals. The public implementation had a higher point estimate in every workload with effectively identical allocation. The UTF-8 4096-byte intervals overlap, so the supported conclusion is no observed regression, not a proven speedup for that case. These results support removing the private optimization instead of retaining fallback complexity.


fwbrasil:
the code was a perf optimization. Can we keep it and fallback if it add-opens isn't set? I believe the scheduler has a mechanism like that

DamianReeves:
> the code was a perf optimization. Can we keep it and fallback if it add-opens isn't set? I believe the scheduler has a mechanism like that

I based the approach on what jsoniter-scala and zio-blocks is doing and ran some benchmarks that seemed to show this solution is as or more performant.

baldram:
> > the code was a perf optimization. Can we keep it and fallback if it add-opens isn't set? I believe the scheduler has a mechanism like that
> 
> I based the approach on what jsoniter-scala and zio-blocks is doing and ran some benchmarks that seemed to show this solution is as or more performant.

Andrij obsessively optimizes, and benchmarks are his regular tests to avoid regressions. He's the person behind both solutions. I think it's a great reference indeed.

Issue #1802, opened 2026-07-28
kyo-ai: Anthropic backend cannot express a prompt-cache breakpoint


mynamiki:
Three related gaps in `AnthropicCompletion`, all rooted in `system` being a
single `String` instead of Anthropic's content-block array.

### 1. No cache breakpoint can be expressed, so no prompt caching happens

`Prompt`'s scaladoc states the intent explicitly:

> Instructions are added to the context as SEPARATE system messages (not one
> concatenated block) so providers can cache individual instruction blocks.

On Anthropic that intent cannot be realized today. The `/messages` API caches a
prefix only when a block carries an explicit `cache_control: {"type":
"ephemeral"}` marker — there is no implicit prefix caching the way OpenAI does
it. The request DTO has no place to put one:

```scala
case class Request(
    model: String,
    messages: List[Message],
    system: Maybe[String],   // <- a plain string; no block, no cache_control
    ...
) derives Schema
```
`kyo-ai/shared/src/main/scala/kyo/ai/completion/AnthropicCompletion.scala:131-140`

So a workload with a large static instruction prefix reused across many calls
pays full input price on every call.

### 2. System messages after the first are demoted to `user`

Anthropic carries the system prompt as a top-level field, so the backend lifts
the first message out:

```scala
val (system, body) =
    ctx.messages.headMaybe match
        case Present(SystemMessage(c)) => (Present(c), ctx.messages.drop(1))
        case _                         => (Absent, ctx.messages)
```
`AnthropicCompletion.scala:168-171`

Every remaining `SystemMessage` then falls through to:

```scala
case SystemMessage(content) =>
    Message(Role.User.name, List(Content("text", text = Present(s"[INTERNAL SYSTEM INSTRUCTION] $content"))))
```
`AnthropicCompletion.scala:201-202`

Since `Prompt` emits one system message per instruction by design, any prompt
built from more than one instruction loses system-role authority for
instructions 2..N on this backend — they arrive as user turns carrying a
text marker. Guardrail-style instructions are exactly the ones that suffer, and
a rule that arrives as if the user had typed it is weaker than one in `system`.

### 3. Cache usage is not reported

```scala
case class Usage(input_tokens: Int, output_tokens: Int) derives Schema
```
`AnthropicCompletion.scala:141`

Anthropic returns `cache_creation_input_tokens` and `cache_read_input_tokens`.
Without them a user cannot verify a cache hit or measure the saving, which
makes the feature unfalsifiable in practice.

[...]

PR #1806, opened 2026-07-29
kyo-ai: model Anthropic's system field as a content-block array

mynamiki:
Anthropic's Messages API accepts `system` as `string | Array<TextBlockParam>`. kyo-ai
sends the string form, which works, but a plain string has nowhere to carry a
`cache_control` breakpoint — so prompt caching cannot be expressed on this wire at
all. Context: #1802.

This models `system` as a block array and adds an optional `cache_control` to
`Content`.

**Semantics are unchanged.** `Completion.fitSystemMessages` still merges the
contiguous leading run into one message; this changes only how that merged content
travels — one text block instead of a bare string. A single-instruction conversation
produces a one-element array, which the API treats identically.

**Nothing is hardcoded and no behaviour is introduced.** `cache_control` is `Absent`
everywhere; the change only gives the marker a home on the wire. It is also the field
an automatic layer would populate, so this doesn't constrain the transparent design
described in #1802 — a single marker, as you suggested, fits here without further
change.

**One judgement call worth flagging.** The array form is emitted unconditionally. It
is the documented shape for every current Anthropic model and the string form is
sugar over it, so there is no model variation to encode. The only scenario where it
could matter is a third-party endpoint pointed at via `apiUrl` that accepts the
string form only. I left it unconditional rather than add a config for a case that
may not exist — happy to gate it behind a provider config if you'd rather not carry
that risk.

**Scope deliberately excluded:**

- Any threshold or automatic-marking logic — that is the transparent design, and yours.
- Any change to non-leading system messages. The `<system-reminder>` path is correct
  as it stands.
- The cache usage counters from item (3) of #1802: on re-reading, `Usage` already
  carries both and `usageStats` already folds them into `AIStats`, with the same
  subset semantics as the OpenAI backend (`cachedInputTokens` as a subset of
  `inputTokens`). There was nothing to add.

**Tests:** a new case asserts the wire shape (`system` as an array of `text` blocks)
while checking that the leading-run merge still delivers every instruction. Full
`kyo-aiJVM/test` suite green locally.

### DIFF
--- a/kyo-ai/shared/src/main/scala/kyo/ai/completion/AnthropicCompletion.scala
+++ b/kyo-ai/shared/src/main/scala/kyo/ai/completion/AnthropicCompletion.scala
@@ -12,7 +12,7 @@ import kyo.ai.Context.{Message as ContextMessage, *}
 /** The Anthropic completion backend (the `/messages` API).
   *
   * Serializes the conversation to the Anthropic request shape via typed `Content` DTOs that `Json.encode`
-  * emits directly: the first context message is lifted into the top-level `system` field, the rest map to
+  * emits directly: the first context message is lifted into the top-level `system` block array, the rest map to
   * user/assistant `Message`s, a tool result becomes a USER-role message with a `tool_result` block, and a
   * user image serializes as a multi-part text-then-image content list. A reply's `tool_use.input` is an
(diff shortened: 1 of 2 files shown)


mynamiki:
Closing this: it does not meet the project's contribution standard. Resubmitting properly.

PR #1808, opened 2026-07-29
kyo-ai: declare the Anthropic system-slot encoding as a config

mynamiki:
### Problem

The Anthropic Messages API accepts `system` as `string | Array<TextBlockParam>`.
kyo-ai sends the string form, which works, but a bare string has nowhere to carry a
`cache_control` breakpoint, so prompt caching cannot be expressed on this wire.
Context: #1802.

### Solution

`Config.SystemEncoding` declares which form an endpoint accepts, defaulted per
provider (Anthropic declares `Blocks`) and overridable per model entry, with a
`systemEncoding` setter for a config re-aimed at another endpoint via `apiUrl`. The
Anthropic backend reads the declared fact; it does not decide the shape itself.
`Content` gains an optional `cache_control`, absent everywhere, so the marker has a
place to live.

`Request.system` becomes `Maybe[Structure.Value]` because the two encodings are
different JSON types in one field, the same approach `Content.input` already uses.

### Notes

- Merging is unchanged: `Completion.fitSystemMessages` still merges the leading run,
  and this only changes how the merged content is encoded.
- No caching behaviour is introduced. Placement of a marker, thresholds, and TTL
  remain open for the transparent design in #1802, which would populate this same
  field.
- Both encodings are covered by tests. Full `kyo-aiJVM/test` is green.
- One observation while validating: `LLMStreamIntegrationTest` failed once in a
  full-suite run and did not reproduce. It passes on the base commit, in isolation,
  and in a repeated full-suite run. That test drives an external CLI and this change
  does not touch the ClaudeCode path.

### DIFF
--- a/kyo-ai/shared/src/main/scala/kyo/ai/completion/AnthropicCompletion.scala
+++ b/kyo-ai/shared/src/main/scala/kyo/ai/completion/AnthropicCompletion.scala
@@ -12,7 +12,7 @@ import kyo.ai.Context.{Message as ContextMessage, *}
 /** The Anthropic completion backend (the `/messages` API).
   *
   * Serializes the conversation to the Anthropic request shape via typed `Content` DTOs that `Json.encode`
-  * emits directly: the first context message is lifted into the top-level `system` field, the rest map to
+  * emits directly: the first context message is lifted into the top-level `system` block array, the rest map to
   * user/assistant `Message`s, a tool result becomes a USER-role message with a `tool_result` block, and a
   * user image serializes as a multi-part text-then-image content list. A reply's `tool_use.input` is an
   * arbitrary object, carried as a `Structure.Value` and re-encoded to the raw argument JSON the `Call`
@@ -202,8 +202,13 @@ private[completion] object AnthropicCompletion extends Completion:
             // Extended-thinking reply blocks ({"type":"thinking","thinking":...,"signature":...}); decoded so
             // the reply parses, then ignored by `read` (the reasoning is not surfaced as a transcript message).
             thinking: Maybe[String] = Absent,
-            signature: Maybe[String] = Absent
+            signature: Maybe[String] = Absent,
+            // Prompt-cache breakpoint. Anthropic caches a prefix only against an explicit marker on a
+            // block, so this is the only place it can be expressed. Absent everywhere today; the field
+            // exists so the marker HAS a home on the wire.
+            cache_control: Maybe[CacheControl] = Absent
         ) derives Schema
+        case class CacheControl(`type`: String) derives Schema
         case class Source(`type`: String, media_type: String, data: String) derives Schema
         case class Message(role: String, content: List[Content]) derives Schema
         // strict = true enables Anthropic structured outputs: token-level constrained decoding against the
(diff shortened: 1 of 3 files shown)


kyo-ai/shared/src/main/scala/kyo/ai/completion/AnthropicCompletion.scala
```diff
@@ -221,7 +226,10 @@ private[completion] object AnthropicCompletion extends Completion:
         case class Request(
             model: String,
             messages: List[Message],
-            system: Maybe[String],
+            // `string | Array<TextBlockParam>` on the wire, so the field is untyped here and the entry's
+            // declared Config.SystemEncoding decides which of the two is emitted. Only the block form can
+            // carry a per-block `cache_control` breakpoint; the string form cannot express it at all.
+            system: Maybe[Structure.Value],
```
fwbrasil, line 232: could this be `system: Maybe[String | List[Content]]` and use a plain string if possible? we'd save tokens in the json structure

kyo-ai/shared/src/main/scala/kyo/ai/completion/AnthropicCompletion.scala
```diff
@@ -289,9 +297,17 @@ private[completion] object AnthropicCompletion extends Completion:
                         ctx.messages,
                         content => UserMessage(systemReminder(content), Absent)
                     )
+                // The encoding is the entry's declared fact, never a provider name read here: an endpoint
+                // that accepts the string form only says so on its config (Config.SystemEncoding).
+                def encodeSystem(content: String): Structure.Value =
+                    config.systemEncoding match
```
fwbrasil, line 303: I can't see the much value for this config. I was expecting: 
- a config for how the system messages are encoded depending if the model accepts multiple system messages or not
- logic to bundle multiple initial system messages into a single entry with multiple text contents

btw, maybe a good optimization is folding sequential messages in general. If we have multiple sequential messages of the same role, they could be in a single API-level message with multiple content blocks. I think it'd reduce token use.

kyo-ai/shared/src/main/scala/kyo/ai/completion/AnthropicCompletion.scala
```diff
@@ -202,8 +202,13 @@ private[completion] object AnthropicCompletion extends Completion:
...
             // the reply parses, then ignored by `read` (the reasoning is not surfaced as a transcript message).
             thinking: Maybe[String] = Absent,
-            signature: Maybe[String] = Absent
+            signature: Maybe[String] = Absent,
+            // Prompt-cache breakpoint. Anthropic caches a prefix only against an explicit marker on a
+            // block, so this is the only place it can be expressed. Absent everywhere today; the field
+            // exists so the marker HAS a home on the wire.
+            cache_control: Maybe[CacheControl] = Absent
         ) derives Schema
+        case class CacheControl(`type`: String) derives Schema
```
fwbrasil, line 211: can we expand this to the supported json schema?
```json
{
  "$schema": "http://json-schema.org",
  "$id": "https://anthropic.com",
  "title": "Anthropic CacheControl Block",
  "description": "Configuration object used to inject explicit prompt caching breakpoints at structural points within Anthropic API requests.",
  "type": "object",
  "required": [
    "type"
  ],
  "additionalProperties": false,
  "properties": {
    "type": {
      "description": "The caching model strategy. Defines the persistence lifestyle of the prompt segment. Anthropic currently relies solely on short-term ephemeral blocks.",
      "type": "string",
      "enum": [
        "ephemeral"
      ]
    },
    "ttl": {
      "description": "Time-To-Live duration string. Sets the rolling expiration clock for the cache segment. Every cache hit completely resets this timer back to zero. Omitting this field prompts the gateway to resolve to the default tier.",
      "type": "string",
      "enum": [
        "5m",
        "1h"
      ],
      "default": "5m"
    }
  }
}
```

PR #1809, opened 2026-07-29
[build] prepare the codebase for Scala 3.9

fwbrasil:
## Problem

Scala 3.9 tightens several language rules and deprecates APIs kyo uses. None of it shows up on 3.8.4, so it accumulates unseen and then fails everywhere at once on the day of the upgrade. Two latent defects in the build are in the same bucket: wrong today, fatal on 3.9.

- **`$` is reserved for compiler use.** The FFI emitters generate 11 identifiers containing it across 52 sites, and `-Werror` promotes each warning to a failure.
- **Type ascriptions on patterns** other than variable and number-literal patterns are no longer allowed. Six declarations use them.
- **`-1.seconds` and `-25.toByte` no longer parse.** The leading `-1.` reads as an incomplete float literal. Three occurrences are ordinary code; the other 47,279 are in a kyo-tasty fixture that spends 103,146 source lines encoding 100KB of bytes, one `Byte` literal per line.
- **`SourceFile.getJPath` and `jpath` are deprecated**, with no Option-returning successor; `path` is what remains.
- **`UpcallBridge` binds a `MethodHandle` to a wildcard val**, which crashes the 3.9 compiler outright: a `CyclicReference` raised while its classfile parser reads the generic signature of the JDK's `java.lang.invoke.ClassSpecializer`. Reported upstream as https://github.com/scala/scala3/issues/26646.
- **kyo-net targets a release its own APIs do not exist in.** It calls `java.lang.foreign`, which is JDK 22+, but applies only `kyo-settings` and so compiles under the global `-release 17`. The seven other modules using that API all apply `foreignRelease`. The published artifact therefore claims Java 17 while calling methods absent there, and a consumer on 17 through 21 links successfully and fails at runtime. This is a live defect, not a 3.9 concern.
- **The doctest scripted fixtures pin their own Scala version.** They hardcode 3.8.3 while the runner classpath is injected from the build, which is on 3.8.4. The two have already drifted; it stays harmless only until the standard library changes underneath, at which point the forked driver dies with `NoSuchMethodError`.

## Solution

Adopt the new rules and correct the two defects, without moving the Scala version:

- Emitters drop the sigil (`__kyoScratch`, `__kyoMark`, and nine others; `{Trait}$externs` becomes `{Trait}_externs`). The golden files and emitter tests that assert on emitted text move with them, so the module is never left red.
- Pattern ascriptions move to named vals, or are dropped where the right-hand side already fixes the type.
- The three ordinary negative literals are parenthesized. The kyo-tasty fixture is re-encoded as hex payloads decoded at runtime, which removes the construct rather than escaping 47,279 instances of it, and takes the file from 103,146 lines to 1,985.
- `OpenApiMacro` reads `SourceFile.path`. It may be a virtual path when the unit is not on disk, which the existing `Files.exists` check already rejects the same way a missing directory is.
- `UpcallBridge` uses `discard`, which says the same thing, is what the neighbouring files already use for a dropped `java.util.Map.put`, and compiles on both versions, so it need not wait for the compiler fix.
- `kyo-net` applies `foreignRelease`, making the artifact honest about the JDK it already requires.
- The doctest fixtures read the Scala version the plugin injects, alongside `plugin.version` and the runner classpath, instead of restating it.
- `FrameTest` keeps `with$Dollar` and silences the warning with `@nowarn`. The `$` is what the test is for: it asserts `calleeName` round-trips an identifier containing one, so renaming would have deleted the coverage rather than fixed it.

## Notes

Every change here compiles and tests on 3.8.4, so this lands independently of any upgrade. Verified per touched module: kyo-ffi-codegen, kyo-ffi, kyo-data, kyo-tasty and kyo-schema-json all green, plus a full JVM compile of main and test sources with `-Werror` on.

The re-encoded fixture was checked byte for byte against the literals it replaces: all 97 payloads and all 102,735 bytes decode identically. The net effect on the tree is about 101,000 fewer lines, and the compiler no longer parses 100,000 separate tokens to build those arrays.

**kyo-net's class file version goes from 61 to 69**, so tooling that reads it will now report Java 25. That is the visible consequence of the release fix and the one thing here worth a second opinion.

Not included, and deliberately: the `scala3Version` bump itself, and kyo-tasty's reader. The latter accepts TASTy at major 28 minor 8, while 3.9 emits minor 9, so it rejects every file 3.9 produces. That needs porting against whatever the minor bump changed and belongs with the migration, not with preparing for it.

A second 3.9 compiler crash turned up during validation, an assertion failure resolving `jdk.internal.foreign.LayoutPath` out of the synthesized rt.jar that sbt supplies under `-release 17` on a JDK 25 host. It is avoided here as a side effect of the kyo-net release fix. Both crashes are regressions: clean on 3.8.4, present since 3.9.0-RC1 and still on nightly.


kyo-ffi/jvm/src/main/scala/kyo/ffi/internal/UpcallBridge.scala
```diff
@@ -354,7 +355,9 @@ object UpcallBridge:
             val allArgs: Array[Class[?]] = (fnCls +: argClses).toArray
             val mt                       = MethodType.methodType(retCls, allArgs).nn
             val raw                      = MethodHandles.lookup().nn.findVirtual(moduleClass, s"invokeShape_$shape", mt).nn
-            val _                        = m.put(shape, raw.bindTo(UpcallBridge).nn)
+            // `val _ = <MethodHandle>` crashes the Scala 3.9.0 compiler (CyclicReference on the JDK's
+            // java.lang.invoke.ClassSpecializer, present since 3.9.0-RC1). discard is equivalent here.
```
hearnadam, line 359: nit: I don't think we need this? Isn't kyo.discard the preference?

hearnadam (approved):

Issue #1812, opened 2026-07-29
[schema] Derivation support for types built through a validating smart constructor


DamianReeves:
## Motivation

kyo-schema has no way to derive a `Schema` for a type whose construction is guarded by a validating smart constructor. The fields are readable, so encoding is derivable — but decoding must go through the constructor, both to reach the type at all and to honour its invariant.

The clearest instance is the `sealed abstract case class` idiom, where `abstract` deliberately suppresses `apply`/`copy` synthesis:

```scala
sealed abstract case class Refined private (value: Int)
object Refined:
  def make(v: Int): Result[String, Refined] =
    if v > 0 then Result.succeed(new Refined(v) {}) else Result.fail("must be positive")
```

Verified behaviour of that shape:

- `Refined(5)` → `object Refined does not take parameters` (no synthesised `apply`)
- `r.copy(value = 9)` → `value copy is not a member of Refined`
- from another file → `Cannot extend sealed class Refined in a different source file`

So derived code cannot construct one by any route. Today `Schema.derived` reports:

```
Cannot derive Schema for sealed trait Refined: no case class or object variants found.
```

which describes neither the type nor the obstacle.

This is not specific to that idiom — it applies to any type with a validating constructor: refined types, newtypes with invariants, domain primitives that reject bad input. Deriving a `Schema` that bypasses the invariant would be worse than failing, so the gap is real.

## Proposal

An explicit derivation entry point that takes the constructor:

```scala
object Refined:
  given Schema[Refined] = Schema.derivedVia(make)
```

`derivedVia` would derive the product shape from the constructor parameters for encoding, and route decoding through the supplied function, folding its failure into the decode error so a rejected value surfaces as a decode failure rather than an exception or a bypassed invariant.

Accepting the common result shapes — `B => A`, `B => Result[E, A]`, `B => Option[A]`, `B => Either[E, A]` — would cover most cases.

Why explicit rather than inferred:

- No annotation to learn, and no guessing at companion method names (`apply`, `make`, `of`, `from`, …), which would be fragile and surprising when it picked the wrong one.
- The author states the intent at the point where the invariant lives.
- It generalises beyond `sealed abstract case class` to any validated type, including ones with a public constructor that should not be used for decoding.

## Minimum useful outcome

Even without `derivedVia`, the error message should name the real obstacle rather than misclassifying the type — something like:

```
Cannot derive Schema for Refined: no accessible constructor
(abstract case class). Provide a Schema explicitly, or derive via its
[...]

PR #1818, opened 2026-07-30
[kyo-schema] derive sealed case classes as products, and add derivedVia for validating constructors

### LINKED ISSUE #1811: [BUG]: Schema.derived fails on any sealed case class, reporting it as a sealed trait with no variants
## Summary

`Schema.derived` fails on any `sealed case class`, reporting the class as a sealed trait:

```
Cannot derive Schema for sealed trait Sealed: no case class or object variants found.
```

`sealed` restricts where subclasses may be declared; it does not change a type's shape. A `sealed case class` is a product with a public constructor and should derive exactly as the same class without `sealed` does.

## Reproduction

```scala
//> using scala 3.8.4
//> using dep io.getkyo::kyo-schema::1.0.0-RC5

import kyo.Schema

sealed case class Sealed(x: Int)
given Schema[Sealed] = Schema.derived
```

```
[error] ./sealed-repro.scala:7:24
[error] Cannot derive Schema for sealed trait Sealed: no case class or object variants found.
[error] given Schema[Sealed] = Schema.derived
[error]                        ^^^^^^^^^^^^^^
```

Non-recursive, no type parameters, no self-type, no private constructor, top-level. `sealed` is the only unusual thing about it.

## Controls

Removing `sealed` — compiles:

```scala
final case class Plain(x: Int) derives Schema
case class Bare(x: Int)        derives Schema
```

A genuine sealed hierarchy — compiles:

```scala
sealed trait Shape derives Schema
object Shape:
  case class Circle(r: Int) extends Shape
  case class Square(s: Int) extends Shape
```

Sealed-*trait* derivation works. The trigger is a class carrying both `sealed` and `case`.

## Why product derivation is unambiguously correct here

There is no sum-type reading to prefer, because Scala permits no derivable variants of a `sealed case class`:

- A case-class variant is rejected: `case class Sub(y: Int) extends Base(y)` → *"case class Sub has case ancestor class Base, but case-to-case inheritance is prohibited."*
- A non-case variant compiles, but the parent's synthesised members misreport it — `Base(1) == SubPlain(1)` is `true` in both directions, `match { case Base(x) => }` matches the subclass, and `copy` on the subclass returns a `Base`.

So the message's premise is true but vacuous: there could never have been case-class or object variants to find.

## Hypothesis

Guessing from the error text, not from reading the macro: derivation appears to test `Flags.Sealed` before `Flags.Case`. A class with both takes the sealed-trait path, which then requires a non-empty `children` list that a case class does not have. Checking `Case` first would route it to product derivation.

The message describing a class as a "sealed trait" is consistent with this.

## Versions

Identical error text on both:

- `1.0.0-RC5` (Maven Central)
- `1.0.0-RC5+30-c491bfca-SNAPSHOT` (Sonatype snapshots)

Scala 3.8.4, JDK 17 (also seen on 25), macOS 15 aarch64. Reproduced standalone via scala-cli and originally via Mill 1.2.0-RC1.

## Impact

Found deriving schemas for [finos/morphir-scala](https://github.com/finos/morphir-scala), where every naming type (`Name`, `Path`, `FQName`, `PackageName`, `ModuleName`, …) is declared `sealed case class`. Since the whole IR is built from those, derivation is blocked for the entire model.

Workaround is `final case class` — strictly stronger than `sealed`, and source/binary compatible since `sealed` already prevented out-of-file subclassing. But it is a source change forced on downstream code, and unavailable where `sealed` is load-bearing in a published API.

---

`sealed abstract case class` fails with the same message but is a genuinely different situation — it has no accessible constructor, so it needs its own handling. Filed separately.

DamianReeves:
Fixes #1811, #1812

### Problem

Two separate gaps in `Schema` derivation, both surfacing as the same wrong message.

**#1811 (bug).** `Schema.derived` failed on any `sealed case class`, reporting it as a sealed trait:

```
Cannot derive Schema for sealed trait Sealed: no case class or object variants found.
```

`sealed` restricts where subclasses may be declared; it does not change a type's shape. A `sealed case class` is a product with a public constructor and should derive exactly as the same class without `sealed`. There is no sum reading to prefer either, because Scala prohibits case-to-case inheritance, so there could never have been variants to find. The message's premise was true but vacuous.

`derivedImpl` tested `Flags.Sealed` before `Flags.Case`, so a class carrying both took the sum path and aborted on an empty `children` list. Corroborating evidence sits in the same file: `metaApplyImpl` already tested Case first, so `Schema[Sealed]` worked while `Schema.derived[Sealed]` failed on the same type.

Found deriving schemas for [finos/morphir-scala](https://github.com/finos/morphir-scala), where every naming type (`Name`, `Path`, `FQName`, ...) is declared `sealed case class`, which blocked derivation for the whole IR.

**#1812 (missing capability).** A type whose construction is guarded by a validating smart constructor had no derivation route at all. The fields are readable, so encoding is derivable, but decoding has to go through the constructor, both to reach the type and to honour its invariant:

```scala
sealed abstract case class Refined private (value: Int)
object Refined:
    def make(v: Int): Result[String, Refined] =
        if v > 0 then Result.succeed(new Refined(v) {}) else Result.fail("must be positive")
```

`abstract` deliberately suppresses `apply` and `copy`, so derived code could not construct one by any route. This is not specific to that idiom: it applies to any refined type, newtype with an invariant, or domain primitive that rejects bad input. Deriving a Schema that bypasses the invariant would be worse than failing.

### Solution

**#1811.** The product test now comes first in `derivedImpl`, and it excludes abstract case classes via a new `isConstructibleCaseClass`. An abstract case class gets `rejectAbstractCaseClass`, which names the missing constructor and points at `derivedVia` instead of misclassifying the type. `metaApplyImpl` picks up the same guard, so both entry points agree.

**`Schema.derivedVia(make)`.** The encode side is the plain product emission, so the wire shape stays byte-identical to the same fields on a case class. The decode side reads those fields exactly as a product does and then applies `make` in place of the primary-constructor call, through a new `constructVia` hook in `emitProductSchemaStatic`. Everything before that point (field decoding, defaults, the required-field bitmap) is shared verbatim rather than reimplemented.

Arity 1 through 8, one argument per case field in declaration order. Arity and per-field types are checked in the macro, because the overloads infer their parameter types from the supplied function rather than from `A`; a mismatch is a compile error naming the offending field.

`Schema.Constructed[R, A]` reads the value out of whatever shape the constructor reports its outcome in: the type itself, `Result`, `Maybe`, `Option`, `Either`, or `Try`. Any other outcome type works with its own given. The total instance (`Constructed[A, A]`) sits in a lower-priority parent, so a `Result`-returning constructor is not ambiguous when no expected type pins the constructed type.

A refusal becomes `ConstructorRejectedException`, a `DecodeException`, so it arrives as the `Result.Failure` every other malformed input arrives as instead of a raw throw escaping `decode` or an invariant quietly bypassed. A `Try` failure keeps its exception as the cause.

Naming: `derivedVia`, not `derivedWith`. Per CONTRIBUTING, `fooWith(f)` in kyo means "do foo, then apply `f` to the result" (`initWith`, `runWith`, `initUnscopedWith`). Here `make` is an input to derivation, not a continuation over its output, so `With` would read backwards.

### Notes

- **A third bug fell out of the same dispatch logic.** The variant emitter tested `Sealed && Trait` to spot an intermediate sealed parent, so a `sealed abstract class` mid-hierarchy failed that test and was emitted as a zero-field product, which emits `New` on an abstract class and does not compile. It now routes any sealed intermediate back through the sum emitter, trait or abstract class. Covered by its own test.
- **Reproductions were written first.** Both issues' repros are in `SchemaTest.scala` and were confirmed failing with the issue's exact message before any implementation change.
- 24 new tests: 8 for the sealed-case-class shapes (product structure, multi-field, nesting, wire-shape parity with the unsealed twin, sum hierarchy unaffected, `sealed case class` as a sum variant, intermediate abstract class, the abstract-case-class message), 16 for `derivedVia` (all six outcome shapes, rejection paths, cross-codec invariant enforcement, `Option`-field wire treatment, nesting, inference with no expected type, and both compile-error paths).
- Full schema family green, 0 failures across all 8 modules. README doctest: 104 blocks, 0 failures, including the new "Validating smart constructors" section. JS and Native `Test/compile` pass; downstream consumers (`kyo-jsonrpc`, `kyo-tasty`, `kyo-examples`, `kyo-test-snapshot`, `kyo-doctest`) compile.
- `kyo-aeronJVM/Test/compile` could not be run locally: it fails in `ffiCompile` ("C compilation failed"). Confirmed to reproduce with this diff stashed, so it is a local C-toolchain issue and not from this change.
- `Schema[A]` and `Schema.derived[A]` still re-derive a plain schema for a *concrete* case class that has a `derivedVia` given, bypassing the invariant. That is `Schema[A]`'s documented behaviour (it never consults a user given) and is unchanged by this PR; nested field resolution goes through `summonInline`, so a `derivedVia` type used as a field does decode through its constructor. There is a test for that.
- `derivedVia` on a generic case class is covered: the case-field types resolve against the applied type, so a `Chunk[A]` field checks against the constructor's `Chunk[A]` parameter and the read body decodes through the summoned element schema.
- The `probe` / scaladoc breakage this PR originally tripped over (`MissingType: ... java.lang.foreign.type.MemorySegment` for `kyo-aeron`, `kyo-stats-machine`, `kyo-net`) was a `main` problem, not this branch's. I diagnosed and filed it as #1819; it is fixed on `main` by `3179315e4`, this branch is merged up to `06e1795c3`, and I verified both modules document cleanly here before closing the issue.

### DIFF
--- a/kyo-schema/shared/src/main/scala/kyo/Schema.scala
+++ b/kyo-schema/shared/src/main/scala/kyo/Schema.scala
@@ -4,6 +4,7 @@ import kyo.Codec.Reader
 import kyo.Codec.Writer
 import kyo.internal.StructureValueReader
 import kyo.internal.StructureValueWriter
+import scala.annotation.implicitNotFound
 import scala.annotation.nowarn
 import scala.annotation.publicInBinary
 import scala.annotation.tailrec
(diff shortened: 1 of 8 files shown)


kyo-schema/shared/src/main/scala/kyo/Schema.scala
```diff
@@ -1975,8 +1976,150 @@ object Schema:
...
+      *   the constructor's return type
+      * @tparam A
+      *   the constructed type
+      */
+    @implicitNotFound(
+        "Schema.derivedVia cannot read a constructed ${A} out of '${R}'. Supported outcomes are ${A} itself, " +
+            "Result[E, ${A}], Maybe[${A}], Option[${A}], Either[E, ${A}], and Try[${A}]. " +
+            "For any other shape, provide a given Schema.Constructed[${R}, ${A}]."
+    )
+    trait Constructed[R, A]:
```
hearnadam:
```suggestion
    abstract class Constructed[R, A]:
```

kyo-schema/shared/src/main/scala/kyo/Schema.scala
```diff
@@ -1975,8 +1976,150 @@ object Schema:
...
+                outcome match
+                    case Present(value) => Result.succeed(value)
+                    case _              => Result.fail("the constructor returned Absent")
+        end fromMaybe
+
+        given fromOption[A]: Constructed[Option[A], A] with
+            def asResult(outcome: Option[A]): Result[Any, A] =
+                outcome match
+                    case Some(value) => Result.succeed(value)
+                    case None        => Result.fail("the constructor returned None")
```
hearnadam: Doesn't Result have constructors for these?

kyo-schema/shared/src/main/scala/kyo/Schema.scala
DamianReeves: Applied: `Constructed` is now an `abstract class`. CONTRIBUTING is explicit that base types take class dispatch over interface dispatch, and `asResult` runs once per decode, so it is on a real path.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

kyo-schema/shared/src/main/scala/kyo/Schema.scala
DamianReeves: It does for two of them: `fromEither` and `fromTry` now go through `Result.fromEither` / `Result.fromTry`. `Result` has no `Option`/`Maybe` constructor, and those two need to supply the rejection description themselves since absence carries no error value, so they keep a body but express it via `Maybe#toResult(ifEmpty)`, with `Option` converting through `Maybe.fromOption` first per the Types convention. All 333 tests still pass, including the per-shape rejection assertions.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

fwbrasil (approved):

PR #1824, opened 2026-07-30
[kyo-ui] keyed Foreach row reuse: per-row subscription registry (follow-up: #1797)

xsistens:
> **⚠ Depends on #1780, #1782, #1794, #1795, #1796, and #1797: please merge those first.** Built on the reactive binding channels (#1797): retained rows keep their channel observers alive across list emissions. The diff shows their commits until they land.

## Problem

A keyed `Foreach` region treats every list emission as a whole-region change: the region re-renders all rows, re-walks the full fragment, and the per-value `Scope` cascade tears down and re-forks every child observer. Appending one row to a 1k-row keyed list costs ~1000 row renders, ~1000 normalize walks, and ~2000 fiber operations, none of which concern the row that actually changed. The redundant work dominates structural list operations (create/append/clear) and adds a re-subscription tail after every emission during which observers are still starting up.

## Solution

Per-row subscription reuse for `foreachKeyed`, following the pattern keyed lists use in fine-grained frameworks (SolidJS `For`, Airstream `split`): a row is an owned instance, not a per-emission artifact.

- Rows live in a `RowRegistry` (the `MountRegistry` ownership pattern): each `RowInstance` holds the cached render output, the walked reactive descendants, and an UNSCOPED runner fiber holding their subscriptions open. Row lifetime deliberately escapes the region's per-value cascade; the registry's owner scope ends it.
- Per emission the items are partitioned against the registry: a row whose key and item value persist is retained untouched (not re-rendered, not re-walked, not re-subscribed); a row whose item changed is rebuilt in place; added rows are built; removed and changed rows are closed AND awaited before the new value paints (the region-level break-before-make guarantee extended to rows).
- The paint is the same full-fragment publish as before, assembled from cached row UIs, so the wire format and the DOM reconciliation are unchanged. Retained rows need no observer retargeting at all: a keyed row's path (`regionPath :+ key`) is invariant under reorder and removal, and the keyed morph reuses the live node under that path.
- Event dispatch resolves through the same registry (shared via the normalize-time `ForeachSpec`): the path segment after the region path is the row key, so an event on a live row routes to that row's cached dispatch handler instead of re-rendering and re-walking the whole list per event (previously every click on a 1k-row list cost ~1000 row renders plus a full-fragment walk). The cached handler is exactly as fresh as a re-derived one: the registry retains a row only while its item value is unchanged, which is the same purity assumption the paint reuse already makes, so no handler ever closes over a stale item. Seeded rows (adopted from the initial walk, which produces no per-row handler) walk just their cached row UI on demand, and anything the registry cannot resolve falls back to the previous full-walk dispatch.
- `MountRegistry.evictExcept` gains a `preClaimed` set so mounts inside retained rows (which never re-claim) still participate in duplicate-key detection.

Gating keeps the semantics honest: reuse applies only to `foreachKeyed` (unkeyed rows are position-addressed, so paths shift on removal), and not to the `*Indexed` variants (the index is baked into row content). Duplicate keys within one emission disable reuse for that emission with a loud warning, keeping the painted output identical to the non-reusable path. Item comparison is value equality; items with identity equality degrade to rebuild-per-emission (correct, just unoptimized).

`Foreach` gains an `indexed: Boolean = false` field (source compatible; binary compatibility is not yet guaranteed pre-1.0).

## Tests

`ForeachReuseTest` (direct engine, recording exchange): appending calls the row render exactly once (for the added row); reordering renders nothing and the retained rows' channel observers still patch afterwards; removing a row closes its observers before the paint (no `on=true` patch can arrive for the removed row's key) while surviving rows keep patching; changing one item's value rebuilds exactly that row; dispatching click events on live rows (both a registry-built row with a cached handler and a seeded row without one) invokes the row's handler with zero list re-renders. All existing Foreach/Reactive/Mounted/TableRegion/DomBackend suites pass unchanged, and SSR golden HTML stays byte-identical (the paint payload is the same fragment shape).

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Fiber.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Fiber.scala
@@ -121,6 +121,12 @@ object Fiber:
 
     /** Runs an asynchronous computation in a new Fiber guaranteeing eventual interruption via the [[Scope]] effect.
       *
+      * If a [[Fiber.Supervisor]] is installed in the current context (see [[Supervisor.let]]), the spawned fiber is additionally
+      * SUPERVISED: any non-interrupt terminal failure (an `Abort` failure or a panic that is not an [[Interrupted]]) is reported to the
+      * supervisor exactly once. Interruptions (including the scope-close interrupt this very method registers) are never reported.
+      * With no supervisor installed (the default), behavior is unchanged. [[initUnscoped]] and [[use]] never supervise: they are the
+      * escape hatches for fire-and-forget fibers and for fibers whose failure the caller observes itself.
+      *
       * @param v
       *   The computation to run
       * @return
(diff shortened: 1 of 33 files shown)


fwbrasil:
Please hold to submit PRs, even if draft, when they have multiple dependencies on other PRs

PR #1828, opened 2026-07-31
[kyo-core] Signal: exact lossless observe on SignalRef via version-validated register/validate/await

xsistens:
> **⚠ Depends on #1827: please merge it first.** The kyo-browser native suite is only reliably green with the exception-free discriminator classification from #1827 (see the Scala Native paragraph below); the branch is stacked accordingly and the diff shows its commit until it lands.

## Problem

`Signal.observe` can miss a change for up to `repairInterval` (1s by default). The observe loop reads `current`, runs the observer body `f`, and only then registers on the next-change promise. A write that lands while `f` runs completes the previous promise with nobody parked on it: the observer then parks on the fresh promise and the change is stranded until either the next write or the periodic repair tick re-reads `current`.

The window is not narrow: it spans the entire execution of `f`. With many observers re-subscribing after a burst (a UI re-rendering a large list, for example), a measurable share of follow-up signal writes land inside it, and every one of them costs up to a second of latency. The repair timer that heals this also has a standing cost: every parked observer wakes once per interval, and each tick's cancelled race arm leaves a ghost waiter on the masked promise until the next write.

## Solution

Make `SignalRef` observation exact with a version-validated register/validate/await protocol, so no wakeup can be missed and no repair timer needs to be armed at all.

`SignalRef.Unsafe` gains a monotonic version counter, incremented once per distinct-value update, before the promise swap. Writer order is: value write, version increment, promise swap+complete. The observer reads the version before the value, runs `f`, and re-arms by capturing the next-change promise and re-checking the version:

- If the validate step sees a newer version, it returns immediately and the re-read observes the value (the value write happens before the increment).
- Otherwise the capture happened before that write's promise swap, and the swap completes exactly the captured promise.

Either way no change is stranded. An idle parked observer holds exactly one waiter and wakes only on a real change. Coalescing stays allowed: observation is level-based, not a queue.

The Scala Native "miscompile" warning at the previous comment block turned out to be a misdiagnosis, and this PR rewrites it with the actual finding. The historical kyo-browser native suite SIGSEGVs never involved `Signal`: the class is unreachable from that test binary (zero `Signal` symbols in the linked artifact), and core dumps show the real fault is a Scala Native 0.5.12 libunwind null read (`CFI_Parser::decodeFDE` inside `Throwable.fillInStackTrace`) on certain deep continuation stacks, layout-sensitive enough that unrelated source changes flipped it on and off. The kyo-side trigger was an exception thrown per non-numeric token in `SchemaSerializer.DiscriminatorReader.matchField`; with that classification made exception-free (using the adjacent `wireFieldId` helper), the full kyo-browser native suite is green on this change. Only a `Long` crosses `f` in this override; the promise handle is captured inside the race arm after `f` returned and dies with the re-arm call.

Delivery now comes in two documented tiers:

| Signal | Delivery |
|---|---|
| `SignalRef`, `map` chains rooted in one | exact, no repair timer armed |
| combinator-derived (`zip`, `combineLatest`, `switchMap`, `zipAll`, `combineLatestAll`, custom `initRaw`) | repairing loop, reconciled within `repairInterval` |

## API additions

- `observe(baseline: Maybe[A], repairInterval)(f)`: the (overridable) seeded primitive. `f` is skipped while `current` still equals `baseline`; existing `observe` overloads delegate with `Absent` and keep their exact semantics, including the initial emission.
- `observeChanges([repairInterval])(f)`: observes every change after subscription, skipping the value current at subscription time. This is the common "I already processed the current value" subscription, previously hand-rolled with a mutable first-emission flag.
- `streamChanges` is rebuilt on `observeChanges` and now matches its name and scaladoc: the subscription-time value is not emitted (previously it was), and a change can no longer strand until the next write (it had no repair path at all). `streamChanges(repairInterval)` overload added. No production call sites existed; tests are updated.

## Tests

New: a write landing while `f` runs is delivered with `repairInterval = Duration.Infinity` (hangs forever on the old loop), for the leaf and through `map`; the back-to-back loss harness re-run with `Infinity` and a tight await budget; `waiters == 1` across several would-be repair intervals; version increments only on distinct writes; `observeChanges` semantics (no initial emission, in-order changes, subscribe-race safety, interrupt); seeded-baseline semantics; `streamChanges` changes-only and no-repair delivery. All existing per-value `Scope` invariants (single live scope, interrupt cascade, no idle teardown) are unchanged and pass.

### DIFF
--- a/kyo-core/shared/src/main/scala/kyo/Signal.scala
+++ b/kyo-core/shared/src/main/scala/kyo/Signal.scala
@@ -89,13 +89,16 @@ sealed abstract class Signal[A](using CanEqual[A, A]) extends Serializable:
       * closed before the next `f` runs, at most one value's children are alive at a time and no waiter or fiber accumulates across changes.
       *
       * It is designed never to permanently miss the latest value, even under a write that races the observation, and never to tear a
-      * still-current value's `Scope` down on an idle timer. Every signal uses the same repairing loop: it reads `current`, runs `f`, then
-      * re-arms a `nextWith`/`Async.sleep(repairInterval)` race that holds the value's `Scope` open until the next change. A write that lands
-      * in the narrow window between reading `current` and registering `nextWith` is missed by the immediate wakeup and reconciled when the
-      * repair timer next fires and re-reads `current` (the hold re-waits on a still-current value, so a repair timer never closes its `Scope`).
-      * So the final value is always delivered: immediately in the common case, and within `repairInterval` in the worst case when a write
-      * races that window. Correctness never depends on `repairInterval` ; only the worst-case reconciliation latency does. This variant uses
-      * [[Signal.defaultRepairInterval]].
+      * still-current value's `Scope` down on an idle timer. Delivery comes in two tiers. A [[SignalRef]] (and a `map` chain rooted in one)
+      * observes exactly: a version-validated register/validate/await protocol makes every change wake the observer immediately, with no
+      * repair timer armed at all (see the `SignalRef.observe` override). Combinator-derived signals (`zip`, `combineLatest`, `switchMap`,
+      * `zipAll`, `combineLatestAll`, custom `initRaw`) use the repairing loop: it reads `current`, runs `f`, then re-arms a
+      * `nextWith`/`Async.sleep(repairInterval)` race that holds the value's `Scope` open until the next change. A write that lands in the
+      * window between reading `current` and registering `nextWith` is missed by the immediate wakeup and reconciled when the repair timer
+      * next fires and re-reads `current` (the hold re-waits on a still-current value, so a repair timer never closes its `Scope`). So the
+      * final value is always delivered: immediately in the common case, and within `repairInterval` in the worst case when a write races
+      * that window on a derived signal. Correctness never depends on `repairInterval` ; only the worst-case reconciliation latency does.
+      * This variant uses [[Signal.defaultRepairInterval]].
       *
       * @param f
       *   The per-value setup, run inside a fresh `Scope`; it may fork scoped children (`Fiber.init`) and should return once setup is done,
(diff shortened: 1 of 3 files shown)


fwbrasil:
Could you try to hold on PR creation? It'd be ideal to ensure you're able to clean the backlog of PRs before creating new ones. Also, our CI is quite busy and I need it to work on fixing flakiness. If you're posting PRs as a way to run CI, could you try to setup CI in your fork? It should be possible since we use vanilla free github runners. I'd appreciate if you could prepare a PR to `CONTRIBUTING.md` regarding these points if the fork CI works.

kyo-core/shared/src/main/scala/kyo/Signal.scala
```diff
@@ -530,16 +575,67 @@ object Signal:
...
+        //   - Otherwise the capture happened before that write's promise swap, and the swap completes exactly the
+        //     captured promise. Either way no change is stranded, with no repair timer armed at all (so an idle
+        //     parked observer costs nothing and holds exactly one waiter).
+        //   - Version-before-value is load-bearing: reading the value first could pair a fresh value with a stale
+        //     version and validate cleanly against a completion nobody held.
+        // Coalescing stays allowed (observation is level-based): a change-and-revert during `f` wakes the
+        // observer, re-reads an unchanged value, and re-arms.
+        //
+        // Scala Native note: earlier revisions of this override were blamed for kyo-browser native suite
+        // SIGSEGVs ("miscompiles", "heap corruption"). Root-caused 2026-07-31: the crashes never involved
```
fwbrasil, line 594: please ensure you open only reviewed PRs without artifacts related to the development process. I'm closing some PRs where the quality is too low, even if draft. Sorry for the disturbance but we need to keep the PR backlog clean

kyo-schema/shared/src/main/scala/kyo/internal/SchemaSerializer.scala
```diff
@@ -1608,9 +1608,14 @@ private[kyo] object SchemaSerializer:
...
-                    try parsed.toInt == CodecMacro.fieldId(expected)
-                    catch case _: NumberFormatException => false
+                    // the macro-generated case-class read body matches fields downstream. wireFieldId
+                    // classifies without throwing: the common token here is a plain field NAME, and paying
+                    // a NumberFormatException per non-id token is not merely slow on Scala Native, it is
+                    // fatal on some stacks: collecting the exception's stack trace walks the frames via
+                    // libunwind, which can segfault on deep continuation stacks (kyo-browser native suite,
+                    // SN 0.5.12: null read in CFI_Parser::decodeFDE during fillInStackTrace). fieldId is
+                    // always positive, so wireFieldId's -1 non-id sentinel never matches.
+                    wireFieldId(parsed) == CodecMacro.fieldId(expected)
```
fwbrasil, line 1618: this seems to leak from another pr

PR #1830, opened 2026-08-01
kyo-ai: declare whether an endpoint accepts a mid-conversation system message

### LINKED ISSUE #1810: [feature]: opt-in system-role messages for models that accept them mid-conversation
### Problem Statement

A system instruction that arrives after the conversation has started cannot ride as a
system message on this wire, so `Completion.fitSystemMessages` converts it into a user
turn wrapped in `<system-reminder>`. That is the right default and works on every model.

Four models do accept `{"role": "system", ...}` inside `messages`: Claude Fable 5,
Mythos 5, Opus 4.8 and Opus 5. Per the docs, such a message "has the same authority as
the top-level `system` field", and because it is appended to the end of the history "it
does not invalidate any cached prefix that came before it". It cannot be the first entry
in `messages`, and it is not available on Sonnet 5, so it cannot become unconditional.

### Proposed Solution

A declared config saying whether an entry accepts mid-conversation system messages,
defaulted per provider and overridable per model entry, following the pattern already
used for other per-endpoint facts (`SystemMessages`, `ReasoningOff`,
`ForcedToolChoice`). The `<system-reminder>` fallback stays for every entry that does
not declare it.

```scala
enum MidConversationSystem derives CanEqual:
    case Unsupported // converted to a user turn, today's behaviour
    case Accepted    // rides as a system-role message in `messages`
```

Declared per entry rather than derived from the model name, for a concrete reason: the
capability is not a property of the model alone. Amazon Bedrock documents it as
available on Opus 4.8 only, while the first-party API lists four models. The same model
name carries a different contract per endpoint, so anything deciding by name is wrong on
one of them.

### Alternative Solutions

*Keep today's behaviour unconditionally.* Correct everywhere and free to maintain. The
trade is that a caller who already chose a model with the capability does not get it.

*A third case on `SystemMessages`.* Considered and set aside: `SystemMessages` governs
how many system messages are delivered at the start, while this governs whether a later
one may ride with system authority. They are orthogonal, and folding them into one enum
makes invalid combinations representable.

### Current Workaround

None needed. The `<system-reminder>` path works. This is about not leaving a stronger
primitive unused where it exists.

### Additional Context

Split out of #1802 at @fwbrasil's request. Sources: [Using the Messages API, "System
role in messages"](https://platform.claude.com/docs/en/build-with-claude/working-with-messages)
and [Mid-conversation system messages](https://platform.claude.com/docs/en/build-with-claude/mid-conversation-system-messages);
the Bedrock contrast is in [its own guide](https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-mid-conversation-system.html).

The implementation follows in a PR against this ticket, so the shape is reviewed as
code. The placement rule is the part that constrains it most: a system message cannot
open `messages`, so the emission path has to fall back for a leading instruction
regardless of the entry's declaration.

mynamiki:
### Problem

A system instruction that arrives after the conversation has started is converted by
`Completion.fitSystemMessages` into a user turn wrapped in `<system-reminder>`. That is
correct on a wire with one system slot, and it stays the right default.

Some endpoints accept a system-role message inside the history: it carries the same
authority as the out-of-band slot, and because it is appended after the prefix it does
not invalidate a cached one. Where a caller already pinned such an entry, the stronger
primitive goes unused, and the wrapper arrives through the same channel as the untrusted
input it is meant to govern. Fixes #1810.

### Solution

`Config.MidConversationSystem` declares whether an entry accepts one, defaulted per
provider and overridable per entry, with a setter for a config re-aimed via `apiUrl`.

The variant carries the placement it allows (`AcceptedAfterOpeningTurn`), so the
completion impl applies the declared fact rather than deciding a rule of its own. An
endpoint whose contract differs adds its own case instead of the impl growing a
positional branch. The shared transform decides whether to convert; the impl decides
only how a surviving message is serialized. Neither reads a provider name.

`claude-opus-4-8` and `claude-fable-5` declare the capability, with the documented
behaviour as provenance, in the same style as the other declared facts on those entries.

### Notes

- **This changes wire behaviour on two catalog entries**, one of which is the provider
  default. On `claude-opus-4-8` and `claude-fable-5`, an instruction that arrives after
  the conversation has started now rides as a system-role message rather than a
  reminder-wrapped user turn. Every other entry is byte-identical to before.
- Only entries whose capability is documented declare it. `claude-sonnet-4-6`,
  `claude-haiku-4-5` and `claude-sonnet-5` stay `Unsupported`. The catalog has no entry
  for the other two models the documentation lists.
- A system message can never open the history, and that holds for every combination of
  declared facts, including `SystemMessages.AllDelivered` on a declaring entry. Written
  as a reproducing test first, since that combination produced an invalid request before
  the fix.
- `"a non-head system message serializes as a system-reminder user message"` moved to
  `claude-sonnet-5`. Its premise, that this wire has no system tail, is still true for
  every entry that does not declare the capability, and that is where it now guards.
- The residual risk is the one this codebase already carries for every declared fact: a
  gateway serving the same model name under a different contract. The per-entry override
  is how a caller states it.
- Full `kyo-aiJVM/test` is green.

### DIFF
--- a/kyo-ai/shared/src/main/scala/kyo/ai/Config.scala
+++ b/kyo-ai/shared/src/main/scala/kyo/ai/Config.scala
@@ -80,9 +80,19 @@ final case class Config private (
     // default applies, so an untouched config states nothing and never warns; a warning fires only on a
     // STATED amount the encoding cannot express. Held rather than dropped where it cannot ride, since
     // one config re-aims across providers, and held while reasoning is off.
-    reasoningAmount: Maybe[Config.Amount] = Absent
+    reasoningAmount: Maybe[Config.Amount] = Absent,
+    // Whether a later instruction may ride with system authority (see
+    // [[Config.MidConversationSystem]]). Defaulted so a positional construction elsewhere is unaffected.
+    midConversationSystem: Config.MidConversationSystem = Config.MidConversationSystem.Unsupported
 ):
-    def apiUrl(url: String): Config              = copy(apiUrl = url)
+    def apiUrl(url: String): Config = copy(apiUrl = url)
+
+    /** Declares whether this endpoint accepts a system instruction after the conversation has started
+      * (see [[Config.MidConversationSystem]]). Set it when `apiUrl` points at an endpoint whose contract
+      * differs from the provider's own.
+      */
+    def midConversationSystem(mode: Config.MidConversationSystem): Config =
+        copy(midConversationSystem = mode)
     def apiKey(key: String): Config              = copy(apiKey = Present(key))
     def apiOrg(org: String): Config              = copy(apiOrg = Present(org))
     def temperature(temperature: Double): Config = copy(temperature = Present(temperature.max(0).min(2)))
@@ -301,7 +311,8 @@ final case class Config private (
         reasoningOff: Maybe[Config.ReasoningOff] = Absent,
         forcedToolChoice: Maybe[Config.ForcedToolChoice] = Absent,
         systemInstructions: Maybe[Config.SystemMessages] = Absent,
-        invalidToolCalls: Maybe[Config.InvalidToolCalls] = Absent
+        invalidToolCalls: Maybe[Config.InvalidToolCalls] = Absent,
+        midConversationSystem: Maybe[Config.MidConversationSystem] = Absent
     ): Config =
         copy(
             provider = provider,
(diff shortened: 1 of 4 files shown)


DamianReeves:
@mynamiki , Sorry for the delay on this, can you rebase the PR and we'll allow it to run

PR #1832, opened 2026-08-02
[kyo-config] FlagPlatform: browser-safe process probe

xsistens:
## Problem

`Flag` crashes on Scala.js in browsers. `FlagPlatform.env`/`envNames` bind the `process` global to a `val` before their `typeof` guard runs:

```scala
val proc = js.Dynamic.global.process
if (js.typeOf(proc) == "undefined" || ...) ...
```

Scala.js compiles the first line to a bare `process` identifier read. In Node that yields the process object, but in a browser (no bundler shim) it throws `ReferenceError: process is not defined` before the guard on the next line ever executes. Any browser app that resolves a flag during startup dies with that error.

This is the same defect #1823 fixed in `kyo-core`'s `SystemPlatformSpecific`. That PR carried both modules; the merge took the `kyo-core` half, so `kyo-config` still has the crashing pattern. This is the remaining half.

## Solution

Keep the `typeof` guard inline on the global selection, which Scala.js compiles to a plain `typeof process` expression (safe on undeclared identifiers by JS semantics), and only bind the `val` after the guard passes — the same idiom `osName()`/`osArch()` and the merged `System.env` already use. Extracted as a `hasProcess` probe since both readers need it.

## Tests

`FlagPlatformTest` gains the missing-global topology, reproduced by deleting `process` from the global object for the duration of the read: `process` then becomes an *undeclared* identifier, which is exactly the state that makes a bare read throw while `typeof process` still answers `"undefined"`. Restoring it in a `finally` keeps the test runner (which talks over `process.stdout`) alive even when an assertion fails.

Both new cases fail with `ReferenceError: process is not defined` against the current code and pass with the fix. `envNames` had no coverage at all before, so its Node-path case comes along.

`SystemPlatformSpecificJsWasmTest` is new and adds the same coverage for the guard #1823 already merged — `env`, `osName` and `osArch` all read `process` and none of them had a test for the browser topology.

## Notes

No behavior change on Node: the same values are returned for present, absent and null variables.

The two modules are deliberately separate concerns here: `kyo-config` carries the fix, `kyo-core` only gains test coverage for code already on `main`.

### DIFF
--- a/kyo-config/js-wasm/src/main/scala/kyo/FlagPlatform.scala
+++ b/kyo-config/js-wasm/src/main/scala/kyo/FlagPlatform.scala
@@ -11,24 +11,37 @@ private[kyo] object FlagPlatform {
     def properties: Iterable[String] =
         java.lang.System.getProperties.propertyNames().asScala.map(_.toString).toList
 
+    /** Must stay INLINE on the global selection: only then does Scala.js emit a plain `typeof process`, which
+      * JS semantics make safe on an undeclared identifier. Binding the selection to a val first emits a bare
+      * `process` read, which throws `ReferenceError` before any guard can run.
+      */
+    private def hasProcess: Boolean =
+        js.typeOf(js.Dynamic.global.process) != "undefined"
+
     // A host with no `process` global (a browser, or a Wasm host with no Node shim) or a missing variable
     // falls back to the stdlib read: `java.lang.System.getenv` always returns null under Scala.js-Node, but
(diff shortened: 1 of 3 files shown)


fwbrasil:
thank you! I had a design exploration parked that flagged this issue. I won't be able to get to it soon so, in case you want to pick it up (or parts of it), I've posed an issue for it: https://github.com/getkyo/kyo/issues/1836

PR #1779, opened 2026-07-24
Remove Typelevel-related modules and dependencies

fwbrasil:
Fixes #1778

Only the benchamarks remain. I'll keep this open for a week as a draft so people can voice their concerns. There's [some interest](https://github.com/getkyo/kyo/issues/1778#issuecomment-5073108724) in maintaining the integrations in a separate repo already.

Issue #1778, opened 2026-07-24
Remove all integrations with Typelevel


fwbrasil:
I I've been banned from Typelevel repositories due to issues I had to report a [straightforward issue](https://github.com/typelevel/cats-effect/issues/4608). I know this might be an issue for some but I won't maintain integrations with projects I can't interact with even on basic issues.

**All modules integrating with Typelevel projects will be removed from Kyo's repository.**

Please speak up in case you'd feel strongly about keeping the integrations and I welcome people who might want to maintain the code outside of Kyo's repository.

calvinlfer:
hey @fwbrasil, i think the work that is done on this interop is very valuable. It's unfortunate what transpired :( 
I propose we should move this out to a separate repo

fwbrasil:
This is the diff in case people want to maintain the code in a separate repo: https://github.com/getkyo/kyo/pull/1779/changes

I'll work this week on supporting external bindings in kyo-compat. It seems the other modules can stand alone in a new repo without adaptations in kyo itself

fwbrasil:
I've prepared a [doc with guidance](https://gist.github.com/fwbrasil/c3a3c9316e404f12a58a3669550d39d1) to maintain the integrations externally and posted https://github.com/getkyo/kyo/pull/1840 to support external kyo-compat backends

PR #1840, opened 2026-08-04
[kyo-compat] let backends be maintained outside the repo

fwbrasil:
## Problem

#1779 removed kyo's cats-effect integrations, and #1778 invited the community to keep them alive in their own repositories. kyo-compat could not actually support that. It was built around a closed set of backends: every generated project hardcoded a dependency on `io.getkyo:kyo-compat-<name>`, and the cross-binding conformance suite that defines what it means to be a correct backend existed only as in-repo test wiring. An out-of-repo backend could not be declared, could not resolve to its own artifact, and could not be validated against the contract the built-in backends meet. So "maintain it elsewhere" was an invitation with no mechanism behind it; without this, the removed integrations have no path back.

## Solution

Decouple a backend's identity from who ships it, turning kyo-compat from a fixed set of five backends into an open surface. The built-in backends still live here; the plugin simply no longer assumes a backend is one of them, or that it comes from `io.getkyo`.

Three capabilities, each framed by what it unlocks:

- **External coordinates** are the load-bearing change. A backend can resolve to any `organization:artifact:version`. External and built-in coordinates go through a single resolution path, so this opens the door without changing how the five built-ins behave.

- **Vendoring** covers the consumer who would rather keep the binding as a project in their own build than depend on a published artifact. It mirrors the published path (the declaration has the same shape either way) and is guarded so a vendored backend that is never wired up fails loudly instead of silently resolving nothing.

- **The conformance suite becomes portable.** The plugin now carries the cross-binding suite and hands it to any binding on request, so an external binding is validated against the exact contract the built-ins pass rather than a private reimplementation. This is what makes a third-party binding trustworthy.

## Notes

- Built-ins are unchanged (coordinate defaults and a shared resolution path); a name-collision guard stops an external backend from shadowing a built-in.
- Covered by scripted tests (coordinate resolution, conformance extraction and wiring, vendoring and the unbound-guard error) and the README doctest.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3262,7 +3262,38 @@ lazy val `kyo-compat-plugin` = (project in file("kyo-compat/plugin"))
                 Def.task(streams.value.log.info("scripted skipped on Windows (sbt#6777 boot-server named-pipe flake)"))
             else
                 Def.task((scripted.toTask("")).value)
-        }).value
+        }).value,
+        // Bundle the cross-binding conformance suite (kyo-compat/test + test-streams)
+        // into the plugin jar as resources, plus an INDEX, so an external binding can
+        // pull it in via `.compatConformance`. Copied verbatim from the canonical suite
+        // the in-repo bindings compile against, so the bundle stays byte-identical.
+        Compile / resourceGenerators += Def.task {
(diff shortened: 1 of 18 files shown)


fwbrasil:
cc @ghostdogpr

Issue #1844, opened 2026-08-04
kyo-doctest: expect=runs and expect=crashes are documented but never execute the block


DamianReeves:
## Summary

`kyo-doctest` documents two runtime expectations, `expect=runs` and `expect=crashes`, but neither is implemented. Both are folded into the compile-only branch of the orchestrator, so no block body is ever executed. A block marked `expect=runs` whose body throws still passes, and a block marked `expect=crashes` that never throws also passes. The parsing side of the DSL is complete and tested; only the execution stage is missing.

This surfaced while adding a `## Storage sizes` section to `kyo-data/README.md`, where the examples carry `assert` calls whose whole purpose is to pin documented numeric values. Those asserts are currently never evaluated.

## Where it is

`kyo-doctest/jvm/src/main/scala/kyo/doctest/internal/Orchestrator.scala:179`

```scala
case Block.Expectation.Compiles | Block.Expectation.Runs | Block.Expectation.Crashes =>
    result match
        case Driver.Outcome.Ok(warnings) =>
            BlockOutcome.Success(block, fromCache, warnings.size)
        case Driver.Outcome.Failed(errors, _) =>
            ...
```

`toOutcomeFromResult` maps a `Driver.Outcome` (a compile result) to a `BlockOutcome`. `Skipped`, `FailsCompile` and `Warns` each get their own branch. `Runs` and `Crashes` share the `Compiles` branch, and there is no execution step anywhere else in the run: `Orchestrator.run` is parse, cache lookup, compile, translate diagnostics, build report.

## What the docs promise

`kyo-doctest/README.md`:

- line 104: "The default is 'compiles,' not 'runs.' A block whose body throws at runtime still passes validation unless you opt in with `expect=runs` or `expect=crashes`."
- line 157: "`expect=runs`: the block must type-check AND execute without throwing."
- line 174: "`expect=crashes`: the block must throw at runtime. Symmetrical to `fails-compile`: the example IS the failure being demonstrated."
- line 333: "A per-block compile failure (a type error inside one fenced block, a missed expectation, a runtime exception in an `expect=runs` block) lands in `Report.failures` as a `Doctest.Failure`."

## Reproduction

Add a block to any JVM module's `README.md` with a runtime assertion that is false:

````markdown
```scala doctest:expect=runs
import kyo.*

val size: ByteSize = 64L.mib
assert(size.show == "77.mib")
```
````

Then:

```sh
sbt 'kyo-dataJVM/doctest'
```

Observed:

[...]

PR #1845, opened 2026-08-04
[kyo-data] add ByteSize, an opaque type for quantities of storage

DamianReeves:
Adds `ByteSize` to `kyo-data`: an opaque type over `Long` (bytes) for talking about quantities of storage. It is a sibling of `Duration` in the same module and follows its structure closely.

## What it provides

- Unit constructors on `Long` and `Int`, binary (`64L.mib`, `1L.gib`, `1L.tib`) and decimal (`100.kb`, `10L.mb`, `2L.gb`).
- Factories: `ByteSize.fromBytes`, `ByteSize.fromUnits`, `ByteSize.Zero`.
- Text parsing: `ByteSize.parse("64MiB")`, accepting `"64 MiB"`, `"1.5GiB"`, `"512B"`, bare `"1024"` (bytes) and `"10MB"`, case-insensitively.
- `show`, rendering at the coarsest lossless binary unit (`"64.mib"`, `"1.gib"`, `"ByteSize.Zero"`), mirroring `Duration.show`.
- A `Flag.Reader.Scalar[ByteSize]` instance, so it works as a config flag type like `Duration` already does.

## Design points

**Non-negative by construction.** Negative and zero inputs clamp to `ByteSize.Zero`. There is no `Infinity` sentinel; the maximum representable value is `Long.MaxValue` bytes.

**Saturating arithmetic.** Addition and multiplication that would exceed `Long.MaxValue` clamp to it; subtraction clamps to `Zero`. The `fromUnits` overflow guard is checked before the multiply, not after, so it cannot rely on wrapped arithmetic.

**Exact integer math where it matters.** `fromUnits` and integral parse inputs compute `value * unit.bytes` without floating point, so a large decimal value such as `"8589934591GB"` is exact rather than rounded through a `Double`. Only fractional parse inputs (`"1.5GiB"`) go through `BigDecimal`, rounded `HALF_UP`.

**One numeric conversion path.** `parse` and the `Flag.Reader` instance both route through `private[kyo] def bytesFromParsed`, so the two cannot drift. The two regex patterns are duplicated deliberately: `Flag.Reader.apply` cannot take a `Frame` and `parse` returns `Result` rather than `Either`. `Duration` handles the same constraint the same way.

**`Units derives CanEqual`.** With `-language:strictEquality` on project-wide, comparing two `Units` values would otherwise be a compile error at every call site, including the tests here.

Following `Duration`'s current surface, this lands with `CanEqual` and the `Flag.Reader` only. No `Ordering`, `Render` or `Numeric`; those are reasonable follow-ups rather than something to preempt here.

## No in-repo consumer, on purpose

Nothing in the tree calls `ByteSize` yet. The first consumer arrives with the `kyo-system` `Path` work, which carries a `size: Maybe[ByteSize]` field. That branch is large and touches many modules; `ByteSize` is the one piece of it that is a self-contained foundation-layer value type with no dependency on the rest, so it lands first to shrink that review surface. Read "unused" as deliberate rather than accidental. The tests and the README section are what exercise the type here.

(The `structByteSize` / `verifyByteSize` / `newByteSize` identifiers in `kyo-ffi`, `kyo-net` and `kyo-offheap` are unrelated `Long` arithmetic that predate this and do not reference `kyo.ByteSize`.)

## Contributor guidance

`CONTRIBUTING.md` gains a `ByteSize` row in the kyo-data primitives table plus a short paragraph on where it belongs: storage and file sizes, buffer and byte-array capacities, read and write chunk sizes, packet and frame sizes, transfer limits and quotas. A raw `Long` or `Int` stays correct for an index, an offset, or a count of elements, since those are positions rather than sizes.

This is guidance for new and changed code. Nothing is migrated here, and existing APIs that thread byte counts as `Long` are not required to migrate.

## A doctest gap found along the way

The README examples carry `assert` calls so the documented numbers are checked rather than merely stated, and are marked `doctest:expect=runs`.

That expectation turns out not to be enforced. `expect=runs` and `expect=crashes` are documented in `kyo-doctest/README.md` but are folded into the compile-only branch at `Orchestrator.scala:179`, so no block body is ever executed. A README block with a deliberately false assert passes with `failures=0`. Filed as #1844 with a reproduction and follow-up tasks.

The markers stay, since they declare the intended semantics and will begin to be enforced once that lands. In the meantime every README assertion is mirrored into a `"README examples"` group in `ByteSizeTest`, where it does execute.

## Tests

113 leaves in `ByteSizeTest`, each asserting concrete values. Beyond the happy path:

- `fromUnits` saturation at the exact boundary, one step either side, for `TiB` and `GiB`.
- `parse` overflow at `Long.MaxValue` and one past it, rather than only at absurd magnitudes.
- `HALF_UP` rounding exercised at a tie (`"1.5B"`, `"2.5B"`, `"0.5B"`) and just below one, so a silent change to `FLOOR` or `HALF_DOWN` fails.
- `Math.round` in `*` exercised at a tie for the same reason.
- `Units` byte counts, symbols, and `all` in declaration order; the symbols are what `parse` matches against.
- `show` at `Long.MaxValue`, above `TiB`, and one byte below a `KiB`.

Both rounding and boundary groups were spot-checked by mutation: flipping the `fromUnits` guard from `>` to `>=` and `HALF_UP` to `FLOOR` each fail the corresponding leaves.

## Verification

| Command | Result |
|---|---|
| `sbt 'kyo-dataJVM/testOnly kyo.ByteSizeTest'` | 113 passed, 0 failed |
| `sbt 'kyo-dataJVM/test'` | green |
| `sbt 'kyo-dataJS/test'` | 3589 passed, 0 failed |
| `sbt 'kyo-dataNative/test'` | 696 passed, 0 failed |
| `sbt 'kyo-dataJVM/doctest'` | 45 blocks, 0 failures |

JS and Native matter here specifically: `BigDecimal` rounding and `Long` overflow semantics are where backends diverge, and this type's contract is built on both.

No `build.sbt` change is needed; `kyo-data` already cross-builds all four targets and everything `ByteSize` uses is on its classpath.

### DIFF
--- a/kyo-data/shared/src/main/scala/kyo/ByteSize.scala
+++ b/kyo-data/shared/src/main/scala/kyo/ByteSize.scala
@@ -1,0 +1,283 @@
+package kyo
+
+/** Represents a quantity of storage in bytes.
+  *
+  * `ByteSize` is an opaque type over `Long` (representing bytes). Values are non-negative:
+  * construction methods clamp negative inputs to `ByteSize.Zero`.
+  *
+  * Construct through the unit extensions (`64L.mib`, `1L.gib`), the factory methods
+  * (`ByteSize.fromBytes`, `ByteSize.fromUnits`), or by parsing from text
+  * (`ByteSize.parse("64MiB")`).
+  *
+  * Arithmetic saturates on overflow: addition and multiplication that would exceed `Long.MaxValue`
+  * clamp to `Long.MaxValue`; subtraction clamps to `ByteSize.Zero`.
+  *
+  * The `show` method produces a human-readable string at the coarsest lossless binary unit:
+  * `"64.mib"`, `"1.gib"`, `"ByteSize.Zero"`.
+  *
+  * @see
+  *   [[Duration]] for the analogous time-span type
+  */
+opaque type ByteSize = Long
+
+/** Companion object for ByteSize type. */
+object ByteSize:
+
+    inline given CanEqual[ByteSize, ByteSize] = CanEqual.derived
+
+    /** Exception thrown for invalid byte size parsing. */
+    class InvalidByteSize(message: String)(using Frame) extends KyoException(message)
+
+    /** Represents zero bytes. */
+    val Zero: ByteSize = 0L
+
+    /** Creates a `ByteSize` from a raw byte count.
+      *
+      * Negative values are clamped to `ByteSize.Zero`.
+      */
+    def fromBytes(value: Long): ByteSize =
+        if value <= 0 then ByteSize.Zero else value
+
+    /** Creates a `ByteSize` from a value and unit.
+      *
+      * The byte count is the exact integer product `value * unit.bytes`; no floating-point
+      * arithmetic is used, so results carry full precision at every magnitude. Negative and zero
+      * values are clamped to `ByteSize.Zero`. Values whose product would exceed `Long.MaxValue`
+      * clamp to `Long.MaxValue`.
+      */
+    def fromUnits(value: Long, unit: Units): ByteSize =
+        if value <= 0 then ByteSize.Zero
+        else if value > Long.MaxValue / unit.bytes then Long.MaxValue
+        else value * unit.bytes
+
+    /** Parses a human-readable byte size string.
+      *
+      * Accepted formats: `"64MiB"`, `"64 MiB"`, `"1.5GiB"`, `"512B"`, `"1024"` (bare digits equal
+      * bytes), `"10MB"`. Unit matching is case-insensitive. An integral value with a unit (for
+      * example `"10MB"`) yields the exact byte count `value * unit.bytes`; a fractional value (for
+      * example `"1.5GiB"`) is scaled with `BigDecimal` and rounded to whole bytes using `HALF_UP`.
+      * Negative values and values that overflow `Long.MaxValue` are rejected with an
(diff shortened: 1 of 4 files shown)


fwbrasil (approved):

kyo-data/shared/src/main/scala/kyo/ByteSize.scala
```diff
@@ -0,0 +1,283 @@
...
+  * Arithmetic saturates on overflow: addition and multiplication that would exceed `Long.MaxValue`
+  * clamp to `Long.MaxValue`; subtraction clamps to `ByteSize.Zero`.
+  *
+  * The `show` method produces a human-readable string at the coarsest lossless binary unit:
+  * `"64.mib"`, `"1.gib"`, `"ByteSize.Zero"`.
+  *
+  * @see
+  *   [[Duration]] for the analogous time-span type
+  */
+opaque type ByteSize = Long
```
hearnadam, line 21: What about calling this `Storage`?

DamianReeves: This used to be called `StorageSize` in a previous PR, and it was suggested by @fwbrasil to call it something more general since it could be used outside of storage scenarios (for example: memory, buffers, and transient non-storage scenarios).

PR #1852, opened 2026-08-05
[kyo-schema-json] JSONL/NDJSON framing, streaming, watch mode, and writing

DamianReeves:
This pull request is stacked on #1857 using GitHub's native base-branch mechanism. Review this layer against `kyo-system-extraction`.

## Problem

JSONL, also called NDJSON, is the interchange format for append-only record logs, AI agent transcripts, batch API files, and structured application logs. Kyo could encode and decode one JSON document, but it had no integrated API for reading, writing, or following streams of JSON records.

The original version split the pure codec and effectful streaming APIs across `kyo-schema-json` and a new `kyo-json` artifact. That split made the pull request harder to merge and introduced an extra public module before the broader module strategy is settled.

## Solution

All JSON and JSONL functionality now ships in `kyo-schema-json`.

### Pure JSONL framing

`Json.Lines` provides byte-level framing and whole-input operations:

- `decodeAll`, `decodeAllBytes`, and `decodeAllBytesResults`
- `decodeRecord`
- `encodeAll`, `encodeAllBytes`, and `encodeLine`
- immutable `Framer` state for chunk-independent streaming integration

Framing operates on bytes so UTF-8 characters remain intact across arbitrary chunk boundaries.

### Effectful JSONL streaming

The public API remains `kyo.Jsonl`. It provides:

- `pipe` and `pipeResults`
- `read` and `readResults`
- `follow` and `followResults`
- `encode`, `write`, and `append`

Strict variants emit every successfully decoded prefix record, then abort on the first failure. The `Results` variants emit per-record failures and continue whenever a record boundary makes recovery possible.

Every effectful entry point delegates framing and parsing to `Json.Lines`, so the pure and streaming surfaces share one set of JSONL rules.

### Supporting APIs

- `kyo-system` gains `Path.Origin`, `Path.tailBytes`, and the shared file-following loop used by text and JSONL drivers.
- `kyo-core` exposes `Stream.fromInputStream`.
- `Json.encode` and `Json.encodeBytes` honor a caller-scoped `Json` instance.
- The JSONL framer stores pending spans and concatenates once per record, avoiding quadratic residual copying.
- Scala.js append writes use the file handle cursor so `O_APPEND` behaves correctly across supported hosts.

## Artifact impact

There is no `kyo-json` project or artifact. Consumers add only:

```scala
libraryDependencies += "io.getkyo" %% "kyo-schema-json" % "<version>"
```

This artifact now brings in `kyo-schema` and `kyo-system` transitively.

## Verification

- Full `kyo-schema-json` tests pass on JVM, JavaScript, Native, and Wasm.
- All 64 `JsonlTest` cases pass on every platform.
- README doctests pass for `kyo-schema-json`, `kyo-core`, and `kyo-system`.
- `scalafmtCheckAll` passes.
- Direct downstream consumers compile, including schema tests, SQL, Tasty, JSON-RPC, examples, benchmarks, doctest support, and snapshot tests.
- Independent review found no critical or important issues.


README.md
```diff
@@ -317,7 +317,8 @@ The vertical an application developer assembles: HTTP services and clients, SQL
 | [kyo-sql-postgres](kyo-sql-postgres/README.md) | ✅  | ✅  | ✅     | ✅   | PostgreSQL driver speaking the wire protocol on kyo-net, plus `COPY` and `LISTEN`/`NOTIFY`                 |
 | [kyo-sql-mysql](kyo-sql-mysql/README.md)     | ✅  | ✅  | ✅     | ✅   | MySQL driver speaking the wire protocol on kyo-net, plus `LOAD DATA LOCAL INFILE`                          |
 | [kyo-schema](kyo-schema/README.md)           | ✅  | ✅  | ✅     | ✅   | One `derives Schema` powers validation, lenses, diffs, builders, and structural conversion; codecs plug in |
-| [kyo-schema-json](kyo-schema-json/README.md) | ✅  | ✅  | ✅     | ✅   | JSON codec for kyo-schema: `Json.encode`/`decode` text and bytes, safety limits, JSON Schema generation    |
+| [kyo-schema-json](kyo-schema-json/README.md) | ✅  | ✅  | ✅     | ✅   | JSON codec for kyo-schema: `Json.encode`/`decode`, `Json.Lines` JSONL framing, JSON Schema generation      |
+| [kyo-json](kyo-json/README.md)               | ✅  | ✅  | ✅     | ✅   | Streaming JSONL/NDJSON over files and byte streams: `Jsonl.read`, `Jsonl.follow`, `Jsonl.write`            |
```
fwbrasil: I think it can be confusing to users to know which module they should use. Would `kyo-schema-jsonl` work?

DamianReeves: Renaming doesn't scale here. I want to add jq-like functionality next, and that's `kyo-schema-jq`, then the next one, and so on. The problem is the extra module, not its name.

Proposal: drop the second module entirely.

1. **`kyo-schema-json` gains `kyo-prelude`, not `kyo-core`** - `Stream`/`Pipe`/`Abort`, no scheduler, no threads. That covers the whole streaming surface; I checked `Jsonl.pipe`/`pipeResults` and nothing needs `Sync` or `Async`.
2. **`kyo-system` (#1799) carries file operations for every format**, generic over the `Codec` SPI, so `path.readAs[A]` / `path.writeAs[A]` work for json, yaml, msgpack, bson, ion, and protobuf alike.

That gives one answer for all six formats: `kyo-schema-<format>` for codec and streaming, plus `kyo-system` if you want files. No `kyo-json`, and jq lands as a capability rather than a module.

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://gist.githubusercontent.com/DamianReeves/cc6bce1f59f635399b84bcdb8a9f1549/raw/kyo-system-layering-dark.svg">
  kyo-system layering
</picture>

One caveat: line framing can't be codec-generic. JSON escapes control characters so `\n` is a safe delimiter, but msgpack/bson/protobuf payloads can contain `0x0A`. So `kyo-system` would offer a generic whole-value tier plus a framer-parameterised streaming tier, with the framer coming from the format module.

kyo-schema-json/shared/src/main/scala/kyo/Jsonl.scala
```diff
@@ -0,0 +1,550 @@
...
+      * @param maxDepth
+      *   maximum nesting depth for objects/arrays (default `Json.DefaultMaxDepth`)
+      * @param maxCollectionSize
+      *   maximum number of entries in maps, sets, or arrays (default `Json.DefaultMaxCollectionSize`)
+      * @param maxLineBytes
+      *   the largest record the framer will accept, in bytes (default `Json.Lines.DefaultMaxLineBytes`)
+      * @return
+      *   a stream of per-record results that runs until interrupted, or until the pending bytes outgrow `maxLineBytes`
+      */
+    def followResults[A](
```
fwbrasil: My initial intuition was that `follow` meant following symbolic links. How about `watchResults`? it seems more consistent with the common terminology for file watching. Btw, this will be very useful for me! 🙏

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
```diff
@@ -0,0 +1,635 @@
...
+      * `bytes` by reference and two records holding identical content never match. Compare [[text]] or `bytes.is(other.bytes)` instead.
+      *
+      * @param bytes
+      *   the record's UTF-8 bytes, with any line terminator removed
+      * @param index
+      *   zero-based position among non-blank lines
+      * @param byteOffset
+      *   offset of the record's first byte within the overall input
+      */
+    final case class Record(bytes: Span[Byte], index: Long, byteOffset: Long):
```
fwbrasil: let's rename to avoid potential conflicts with `kyo.Record`

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
```diff
@@ -0,0 +1,635 @@
...
+        /** The record decoded as UTF-8 text. */
+        def text: String = new String(bytes.toArray, StandardCharsets.UTF_8)
```
fwbrasil, line 78: shouldn't this consider `index` and `byteOffset`?

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
```diff
@@ -0,0 +1,635 @@
...
+    end Record
+
+    /** One line a framer resolved out of the input.
+      *
+      * A line is either framed as a record or rejected for exceeding `maxLineBytes`. Both travel in one ordered sequence, so a rejected
+      * line stays exactly where it sat among the records around it, which is what lets a lenient consumer report one failure for it and
+      * carry on with the next record.
+      */
+    enum Line:
```
fwbrasil: could we avoid this enum and use `Result` instead?

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
```diff
@@ -0,0 +1,635 @@
...
+        def feed(chunk: Span[Byte])(using Frame): Framed =
+            if atStart then feedAtStart(chunk)
+            else scanChunk(pending, pendingSize, residualOffset, chunk)
+
+        /** Returns the unterminated trailing record, if the input ended without a newline.
+          *
+          * @return
+          *   the final record when the residual holds anything other than whitespace, `Absent` otherwise
+          */
+        def finish: Maybe[Record] =
```
fwbrasil: it seems this method finalizes a single line not all pending data? how about `finishLine`?

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
```diff
@@ -0,0 +1,635 @@
...
+          *
+          * Reached only while `atStart`, so `pending` holds at most two bytes and every one of them already matched the mark: that is the
+          * only way a framer stays in its start state. Comparing resumes at `pendingSize` rather than at zero for exactly that reason,
+          * which keeps the cross-boundary read here bounded to the three bytes of the mark.
+          *
+          * `total` is a `Long` for the same reason [[settle]] takes one: this is the other place the pending total grows, and a chunk
+          * large enough to wrap it would leave a framer holding a negative total and reading `Bom` at a negative index. The total is
+          * narrowed only on the branch that proves it below the mark's three bytes.
+          */
+        private def feedAtStart(chunk: Span[Byte])(using Frame): Framed =
```
fwbrasil: s/chunk/span

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
```diff
@@ -0,0 +1,635 @@
...
+    val DefaultMaxLineBytes: Int = 16 * 1024 * 1024
+
+    private val LimitName: String = "JSONL record length"
+
+    private val Newline: Byte = '\n'.toByte
+    private val Return: Byte  = '\r'.toByte
+    private val Tab: Byte     = '\t'.toByte
+    private val Space: Byte   = ' '.toByte
+
+    private val Bom: Span[Byte] =
```
fwbrasil, line 50: can you document the purpose and origin of these bytes?

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
```diff
@@ -0,0 +1,635 @@
...
+
+    /** Appends `piece` to the pending pieces, dropping it when it carries nothing.
+      *
+      * Keeping empty pieces out is what lets [[lastByte]] read the last held byte off the last piece.
+      *
+      * `Chunk.append` builds an append chain, on which `append` and `last` are both O(1) but `apply` is O(k). That is why [[join]] and
+      * [[isBlankPrefix]] each take `toIndexed` once and walk the result: indexing the chain directly would make walking the pieces
+      * quadratic in their count, which is the same shape of defect in the piece count that this representation removes in bytes.
+      */
+    private def hold(pieces: Chunk[Span[Byte]], piece: Span[Byte]): Chunk[Span[Byte]] =
```
fwbrasil: this method seems unsafe since it silently drops information. Can you review if the callers take that into consideration and if we could have a safer approach?

kyo-schema/shared/src/main/scala/kyo/SchemaException.scala
```diff
@@ -145,6 +145,41 @@ case class TrailingInputException(format: Codec, detail: String)(using Frame)
...
+  *   offset of the record's first byte within the framed input
+  * @param record
+  *   the raw record text, truncated in the rendered message but complete on the field
+  * @param cause
+  *   the underlying decode failure
+  */
+case class RecordDecodeException(
+    recordIndex: Long,
+    byteOffset: Long,
+    record: String,
```
fwbrasil: it's best to avoid payloads leaking in exceptions and logs since they can be sensitive information

kyo-system/shared/src/main/scala/kyo/Path.scala
```diff
@@ -586,20 +624,148 @@ object Path extends PathPlatformSpecific:
         runtime: Path
     ) derives CanEqual
 
+    /** What one poll of a followed file observed.
+      *
+      * Naming the outcomes keeps the loop that drives them a total match: each state's handling is a visible decision, and a state added
+      * later is a compile error at the driver rather than a silent fallthrough.
+      */
+    private enum Poll derives CanEqual:
```
fwbrasil: let's rename to avoid conflict with `kyo.Poll`

kyo-system/shared/src/main/scala/kyo/Path.scala
```diff
@@ -586,20 +624,148 @@ object Path extends PathPlatformSpecific:
...
+        /** The open file can no longer be measured. */
+        case Unreadable(error: Result.Error[FileReadException])
+    end Poll
+
+    /** What one [[Path.follow]] step produced from a block of bytes.
+      *
+      * `Stop` carries no state, which is what keeps a step that has finished from handing back state nothing will ever read. A step that
+      * knows its own consumer is done says so here rather than leaving the loop polling a file whose bytes can no longer be used.
+      */
+    private[kyo] enum Step[+V, +St]:
```
fwbrasil, line 651: can we use `Loop.Outcome2` instead?

kyo-schema-json/shared/src/main/scala/kyo/Jsonl.scala
DamianReeves: Renamed. `Jsonl.follow`/`followResults` are now `Jsonl.watch`/`watchResults`, and the shared `private[kyo]` loop they drive is `Path.watch`. I left `Path.tail`/`tailBytes` alone since those are deliberately the `tail -f` view.

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
DamianReeves: Renamed to `Json.Lines.Line`. The old `Line` enum is gone (your next comment), so the name was free.

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
DamianReeves: `bytes` is already the line's own slice with the terminator stripped, so `text` has nothing positional to add.

I think what you were looking at is the caution a few lines above: "do not compare two records with `==`". That was there only because `Span`'s derived equality is by reference. Fixed it properly instead of documenting it: `Line` now has structural `equals`/`hashCode` over `bytes`, `index`, and `byteOffset`, so two lines framed from the same input at different span sizes compare equal. The caution is gone.

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
DamianReeves: Done. A framed line is now `Result[LimitExceededException, Line]`: success carries the line, failure carries the size breach sitting where the line sat.

I kept `Framed` as an enum. Its two cases are the two things a breach can mean, and only `Continued` carries a framer, which is what keeps the unusable framer out of reach after a halt.

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
DamianReeves: Right, it resolves at most the one line the residual holds. Renamed to `finishLine`.

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
DamianReeves: Done, parameter names and the surrounding prose.

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
DamianReeves: Added: what the bytes are (the UTF-8 byte order mark, written by several Windows editors and export tools), and why the framer consumes them before framing starts (they belong to no record, and left in place they would prefix the first record, which then fails to parse since the mark is not JSON whitespace).

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
DamianReeves: It only dropped empty spans, so no bytes were lost, but the shape was fragile and you were right to flag it: the invariant it maintained (no empty piece, so the last held byte is on the last piece) lived in a comment, while the running total was threaded separately through five call sites.

Folded both into a `Pending` value that owns `add`, `endsWithReturn`, `join`, and `isBlankPrefix`. Three changes fall out of it:

- `lastByte` is gone, replaced by `endsWithReturn`, so no caller has to check non-emptiness first.
- The total is a `Long`, so a span landing against a ceiling near `Int.MaxValue` cannot wrap it negative. That widening used to be repeated at each of the two sites where the total grows.
- New test pins the one way an empty piece is reachable from outside, an empty feed, including that a CRLF split across it still strips the `\r`.

kyo-schema/shared/src/main/scala/kyo/SchemaException.scala
DamianReeves: Agreed. Dropped the `record` field and the truncated copy of it in the message. `recordIndex` and `byteOffset` stay, and they are what this type adds over its `cause`: they locate the record for anyone still holding the input without copying it anywhere the input is not already.

One thing to flag: `ParseException.input` does the same thing and predates this branch, so a `RecordDecodeException` wrapping one still surfaces the input through its cause's message. Happy to take that on, but it touches every codec's tests, so I left it out of this PR unless you want it here.

kyo-system/shared/src/main/scala/kyo/Path.scala
DamianReeves: Renamed to `Path.Observed`. `Jsonl` uses `kyo.Poll` a few lines away, so the collision was real.

kyo-system/shared/src/main/scala/kyo/Path.scala
DamianReeves: I don't think it can. `Loop.Outcome2` is opaque, so outside `Loop`'s scope the `O | Continue2[A, B]` union isn't visible and there's no way to match one; only `Loop.apply` destructures it. `Path.Step` is different: it's returned by a pure step callback and read by the watch loop itself, so it has to be a type the loop can match on.

Left `Step` as is and recorded the reason in `kyo-system/CONTRIBUTING.md`. If you'd rather have this, `Loop` could expose a fold over an `Outcome` and then `Path.watch` could use it. Say the word and I'll do that instead.

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
DamianReeves: Links to the code, for review:

- [`Pending`](https://github.com/getkyo/kyo/blob/45728e06dd76eb9fcdf445fe2c7297917cbca097/kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala#L97-L197), the value that now owns the pieces and the total together
- [`add`](https://github.com/getkyo/kyo/blob/45728e06dd76eb9fcdf445fe2c7297917cbca097/kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala#L120-L127), which is where the old `hold` went
- [`endsWithReturn`](https://github.com/getkyo/kyo/blob/45728e06dd76eb9fcdf445fe2c7297917cbca097/kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala#L129-L137), which replaced `lastByte`
- [the test for the empty feed](https://github.com/getkyo/kyo/blob/45728e06dd76eb9fcdf445fe2c7297917cbca097/kyo-schema-json/shared/src/test/scala/kyo/JsonLinesTest.scala#L314-L327)

kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala
DamianReeves: Yes, in the `Bom` scaladoc: [JsonLines.scala#L43-L49](https://github.com/getkyo/kyo/blob/5bc5b17dd/kyo-schema-json/shared/src/main/scala/kyo/JsonLines.scala#L43-L49).

The observable half was already in the docs (kyo-schema-json/README.md notes that byte offsets count a leading mark, kyo-schema/README.md lists mark-stripping among the framing rules), but the module guide only mentioned the mark in passing as a boundary-drift risk. Added a section to [kyo-schema-json/CONTRIBUTING.md](https://github.com/getkyo/kyo/blob/5bc5b17dd/kyo-schema-json/CONTRIBUTING.md#the-byte-order-mark-is-consumed-not-framed) covering the rule, that consuming the mark does not hide it from the reported offsets, and the part a refactor can break: the mark is recognized across span boundaries, so a framer holds a proper prefix of it in its start state rather than guessing, which is why `Framer.atStart` exists.

fwbrasil (approved):

PR #1854, opened 2026-08-06
[kyo-schema-json] use ByteSize for JSONL limits and read buffers

DamianReeves:
## Summary

Read buffer capacities and the JSONL record ceiling are caller-supplied amounts of storage, so they are now `ByteSize` values rather than bare `Int`s. Quantities that name a place or a count stay numeric: byte offsets, cursor positions, record indices, array lengths, and read counts are unchanged.

- `kyo-system`: `Path.readStream(charset, bufferSize)`, `Path.readBytesStream(bufferSize)`, `Path.tail(pollDelay, bufferSize)`, and `Path.tailBytes(bufferSize)` take a `ByteSize`. `Path.Origin.Offset` still takes a `Long`, because it names a position in a file.
- `kyo-core`: `Stream.fromInputStream(is, bufferSize)` takes a `ByteSize`.
- `kyo-schema-json`: `Json.Lines.DefaultMaxLineBytes` becomes `Json.Lines.DefaultMaxLineSize`, a `ByteSize` of 16 MiB, and the `maxLineBytes` parameter of `Json.Lines.Framer.init`, the `Json.Lines` decode entry points, and every `Jsonl` entry point becomes `maxLineSize: ByteSize`.

## Narrowing without an exception

Both places that narrow a `ByteSize` to something an array can hold are total functions, so no read grows an argument-validation failure in its effect row.

A buffer size is clamped to the range an array can address. `ByteSize.Zero` becomes one byte, because a zero-size buffer reads nothing on every pass and turns a read loop into a spin; anything above `Int.MaxValue` bytes becomes `Int.MaxValue`, which is what the `Int` parameter it replaced already did at its own ceiling.

A record ceiling is clamped by `Framer.init` to the new `Json.Lines.MaxLineSize` (`Int.MaxValue` bytes). A record is handed out as an array-backed `Span`, so a larger ceiling names records that cannot be built. Clamping at construction keeps every comparison inside the framer exact and keeps the size a `LimitExceededException` reports truthful.

While the ceiling was being retyped, the one remaining unsaturated narrowing in the framer went with it: `resolveSpanning` reported a breached line's size with a plain `.toInt` while `settle` saturated. Both now go through the same `saturateToInt`, so neither can report a negative length.

## Notes for review

- The `Path.Unsafe.ReadHandle.size()` return type is deliberately left as `Long`. It is a measurement, not a caller-supplied amount, and `Path.size` and `PathStat.sizeBytes` both report a `Long`; retyping only the handle would have split the three apart.
- This is the first use of `ByteSize` outside `kyo-data`.

## Verification

`sbt kyo-schema-jsonJVM/test kyo-systemJVM/test kyo-coreJVM/test` passes. Doctests pass for the four READMEs this touches.

New tests cover each clamp: `readBufferCapacity` at both ends in `kyo-core` and `kyo-system`, an end-to-end zero-buffer read through `Stream.fromInputStream`, `Path.readBytesStream`, and `Path.readStream`, and `Framer.init` clamping a ceiling above `MaxLineSize`.


fwbrasil:
This PR has some unrelated changes, maybe some issue with stacking or merging? I can review if it's intended

kyo-ai/shared/src/main/scala/kyo/ai/completion/ClaudeCodeCompletion.scala
```diff
@@ -284,6 +291,26 @@ private[completion] object ClaudeCodeCompletion extends HarnessCompletion("Claud
     private[completion] def userToolInfos(tools: Chunk[Tool.internal.Info[?, ?, LLM]]): Chunk[Tool.internal.Info[?, ?, LLM]] =
         tools.filterNot(_.name == Completion.resultToolName)
 
+    /** Writes the MCP configuration to a scoped file for the Claude CLI.
+      *
+      * Passing inline JSON is not portable on Windows: executable shims can consume the JSON's quotes while reconstructing the native
+      * command line, after which Claude treats the malformed value as a file path. A file path contains no JSON quoting and also avoids
+      * platform command-line length limits.
+      */
+    private[completion] def mcpConfigFile(config: String)(using Frame): Path < (Sync & Scope & Abort[AIGenException]) =
```
fwbrasil: stacking issues?

DamianReeves:
Good catch, and thanks for holding off. There were two things going on.

The real one: three commits on this branch had nothing to do with `ByteSize`. They were Windows fixes I picked up while verifying cross-platform (MSVC object placement in kyo-ffi, and the Claude CLI integration in kyo-ai). They are now their own PR, #1888, off main.

The other one was noise: the rest of what you saw was the stale merge base. This branch forked from #1852 before #1852 and #1857 landed, so GitHub was diffing against a base from before both. That is gone now.

The branch is rebased onto main and the diff is 15 files, all `ByteSize`. Two things changed in the rebuild that are worth your eye:

- The earlier version threw `IllegalArgumentException` when narrowing a `ByteSize` to an array capacity, which would have escaped a `Stream` as a panic. Both narrowing points are now total. A buffer size clamps to `[1, Int.MaxValue]`, and `Framer.init` clamps a record ceiling to `Json.Lines.MaxLineSize`, since a record is an array-backed `Span` and cannot exceed that anyway.
- The earlier version also retyped `Path.Unsafe.ReadHandle.size()` to `ByteSize`. I dropped that. It is a measurement rather than a caller-supplied amount, and it would have left the handle disagreeing with `Path.size` and `PathStat.sizeBytes`, which both stay `Long`. The line this PR draws is: caller-supplied capacities and ceilings become `ByteSize`, measured quantities and positions stay numeric.

DamianReeves:
CI notes for this branch. Every failure so far has been in a module this change does not touch, and the arm64 one reproduces on main.

**First attempt, three failures:**

- `linux-x64 / JVM`: `kyo.SignalTest > combineLatestAll > every individual signal can wake the combinator` hit the 2 minute leaf timeout. It syncs on waiter counts with ghost-waiter assumptions (`r1.waiters >= 2`, `r2.waiters >= 3`), and `assertEventually` retries on a fixed 10ms schedule with no attempt cap, so an assumption that never holds shows up as a hang rather than a failed assertion. Same test group as 715dffd76, 11f3721f1, 0059a3652, and 565275591. It passes locally, five runs in a row.
- `linux-arm64 / JVM`: `kyo-sql-testsJVM` failed the end-of-run leak check on one ESTABLISHED socket to the MySQL testcontainer, held by a pending IoUring read.
- `linux-x64 / Native`: the `kyo-browserNative` test binary crashed, exit 139 and then 134 on the retry.

Those jobs ran for 1h16m, 2h21m, and 3h57m, so the runners were heavily loaded.

**After rerun:** both linux-x64 jobs went green. The arm64 JVM job failed again on a different test, `kyo.internal.SqlConnectionCancelTest > Async.timeout on a statement in flight fires the wire cancel`, a 2 minute timeout in `kyo-sqlJVM`, after running 2h22m.

**That arm64 job is failing on main.** Run 32337833729, `build (linux-arm64) / build (JVM)`, fails the `kyo-sql-testsJVM` leak check with one ESTABLISHED socket held by a pending IoUring read: `Read(fd=44,id=188978561895,client)`. Same handle id this branch saw on its first attempt, so it reproduces on the base commit with none of this change present.

Nothing here touches `kyo-sql`, `kyo-sql-tests`, or `kyo-browser`. Locally, `kyo-schema-jsonJVM/test`, `kyo-systemJVM/test`, and `kyo-coreJVM/test` all pass, and the doctests pass for every README this touches.

kyo-ai/shared/src/main/scala/kyo/ai/completion/ClaudeCodeCompletion.scala
DamianReeves: Yes, stacking. This file had no business here. It has moved to #1888 along with the other two Windows commits, and this branch is now rebased on main, so the kyo-ai and kyo-ffi changes are gone from this diff.

fwbrasil (approved):

PR #1856, opened 2026-08-07
[kyo-data] add Glob

DamianReeves:
## Summary

- add `Glob`, a compiled and platform-independent matcher for slash-separated paths
- support `*`, `?`, complete-segment `**`, character classes, alternatives, and escaping
- make `Glob.ParseError` a `KyoException` with caller `Frame` context
- preserve the original pattern through `Glob.show` and `Render[Glob]`
- compile `glob"..."` literals into a versioned embedded automaton without runtime parsing or compilation

## Stack

This is the first PR in the native GitHub stack:

1. #1856, Glob
2. #1857, kyo-system extraction
3. #1852, JSONL in kyo-schema-json
4. #1854, ByteSize follow-up

## Verification

The 27 shared Glob tests pass on JVM, JavaScript, Native, and Wasm, including an externally consumed 4,096-character literal. The JVM bytecode regression test confirms generated literals call `GlobLiteral.fromEncodedV1` and do not call `Glob.parse`. The complete `kyo-dataJVM/test` suite, the `kyo-data` README doctest, and scalafmt checks also pass.


kyo-tasty/jvm/src/test/scala/kyo/StandardClasspathFidelityTest.scala
```diff
@@ -19,12 +19,12 @@ class StandardClasspathFidelityTest extends kyo.test.Test[Any]:
     // jrt:/ cold loads can still be slow on a contended runner; keep a generous per-leaf budget.
     override def timeout = Duration.fromJava(java.time.Duration.ofMinutes(3))
 
-    "allSymbols.count(isGiven) ~= 570 on standard classpath" in {
+    "allSymbols.count(isGiven) ~= 576 on standard classpath" in {
```
fwbrasil, line 22: we should change kyo-tasty tests to work with a fixed classpath

kyo-data/shared/src/main/scala/kyo/Glob.scala
```diff
@@ -0,0 +1,616 @@
...
+  * `*` and `?` stay within one segment. `**` is recursive only when it is the
+  * complete segment. Character classes, ranges, negation, alternatives, and
+  * backslash escaping are supported. Matching has polynomial bounds and does not
+  * use regular expressions or filesystem APIs.
+  *
+  * @see [[Glob.parse]]
+  * @see [[Glob.all]]
+  * @see [[Glob.CaseSensitivity]]
+  */
+opaque type Glob = Glob.Compiled
```
fwbrasil: Could we make `Glob.Compile` the `Glob` itself?

kyo-data/shared/src/main/scala/kyo/Glob.scala
```diff
@@ -0,0 +1,616 @@
...
+        if parts.size != 1 then
+            quotes.reflect.report.errorAndAbort("glob literals do not accept interpolated values")
+        val value = parts.head
+        // The macro implementation has no runtime caller frame. Parse failures are
+        // re-emitted by the compiler at the glob literal's source position.
+        Glob.parse(value)(using Frame.internal) match
+            case Result.Failure(error) =>
+                quotes.reflect.report.errorAndAbort(s"invalid glob at offset ${error.offset}: ${error.reason}")
+            case Result.Panic(error)  => throw error
+            case Result.Success(glob) => Glob.expression(glob)
```
fwbrasil, line 607: nice optimization!

kyo-data/shared/src/main/scala/kyo/Glob.scala
```diff
@@ -0,0 +1,616 @@
...
+      *
+      * The offset is a zero-based character offset into the original pattern.
+      * The reason is stable, human-readable text suitable for diagnostics.
+      *
+      * @param offset
+      *   location of the invalid syntax
+      * @param reason
+      *   explanation of the invalid syntax
+      */
+    final case class ParseError(offset: Int, reason: String)(using Frame)
```
fwbrasil: when printing to the console, this will show up as just `ParseError`. Can you rename to `GlobParseException`? "error" can be confused with Java's `Error` as well

fwbrasil (approved):

kyo-data/shared/src/main/scala/kyo/Glob.scala
DamianReeves: Done in 8a121cc80. Glob is now the final case class directly, with show and both matches overloads as members; the opaque alias and the Compiled wrapper are gone. Construction stays private[kyo], so the public surface is unchanged.

kyo-data/shared/src/main/scala/kyo/Glob.scala
DamianReeves: Renamed to GlobParseException in 8a121cc80, hoisted to a top-level type beside Glob so the qualified form does not stutter. kyo-dataJVM and JS suites plus the README doctest are green on the branch.

kyo-tasty/jvm/src/test/scala/kyo/StandardClasspathFidelityTest.scala
DamianReeves: Agreed, that range assertion is going to keep drifting as the classpath changes. Pinning the tests to a fixed classpath means putting together a stable jar set and rewiring TestClasspaths, which feels like its own PR rather than more off-topic weight on this one. I can do it as a follow-up right after this lands, or in this PR if you prefer.

PR #1857, opened 2026-08-07
[kyo-system] extract Path, System, Process, and Command from kyo-core

DamianReeves:
## Summary

- create `kyo-system` and move `Path`, `System`, `Process`, `Command`, and their exceptions out of `kyo-core`
- preserve the `kyo` package and existing consumer call sites
- move file-oriented stream extensions with `Path` to avoid a module cycle
- add the required downstream module dependencies and cross-platform configuration support

## Stack

This is the second PR in the native GitHub stack:

1. #1856, Glob
2. #1857, kyo-system extraction
3. #1852, JSONL in kyo-schema-json
4. #1854, ByteSize follow-up

Review this layer against the `glob` branch.

## Verification

The affected `kyo-system` and `kyo-core` suites pass across supported platforms. JVM `PathTest` passes 213 tests, with five Windows-only cases cancelled on Linux. Module README doctests pass.


kyo-system/shared/src/main/scala/kyo/StreamFileExtensions.scala
```diff
@@ -0,0 +1,78 @@
...
+
+/** Stream sinks that write byte and text streams to the file system.
+  *
+  * These live in kyo-system rather than kyo-core because they are typed in terms of [[kyo.Path]] and [[kyo.FileException]]. Keeping them in
+  * kyo-core would make kyo-core depend on kyo-system, which already depends on kyo-core.
+  *
+  * @see
+  *   [[kyo.Path]] for the path type these sinks target
+  */
+object StreamFileExtensions:
```
fwbrasil: Can you rename to `StreamSystemExceptions` with the module name for discoverability? It's the pattern we use in the codebase (see `StreamCoreExtensions`)

kyo-system/shared/src/main/scala/kyo/StreamFileExtensions.scala
```diff
@@ -0,0 +1,78 @@
...
+
+    /** Shared write logic: opens a write handle via Scope, runs the body, and removes the partial file on failure. */
+    private def writeWith[S](path: Path)(
+        body: Path.WriteHandle => Unit < (Sync & Abort[FileWriteException] & S)
+    )(using Frame): Unit < (Scope & Sync & Abort[FileException] & S) =
+        Scope
+            .acquireRelease(
+                Sync.Unsafe.defer(Abort.get(path.unsafe.openWrite(append = false, createFolders = true)))
```
fwbrasil: it'd be nice if the user can define `append` and `createFolders` via params

build.sbt
```diff
@@ -1038,6 +1058,7 @@ lazy val `kyo-ffi` =
         .crossType(CrossType.Full)
         .in(file("kyo-ffi"))
         .dependsOn(`kyo-core`)
+        .dependsOn(`kyo-system` % Test)
```
fwbrasil, line 1061: not a blocker for this PR but I'm wondering if we should rename `kyo-core` to `kyo-async` and then add a new `kyo-core` that includes `kyo-system` and `kyo-async`. It could reduce friction for users since `kyo-system` has features that are commonly used

fwbrasil (approved):
I've left minor comments, LGTM

kyo-system/shared/src/main/scala/kyo/StreamFileExtensions.scala
DamianReeves: Renamed to `StreamSystemExtensions`, file included, so it lines up with `StreamCoreExtensions`. The sink tests moved out of `PathTest` into a new `StreamSystemExtensionsTest` at the same time.

kyo-system/shared/src/main/scala/kyo/StreamFileExtensions.scala
DamianReeves: Both are parameters now, with the same names and defaults as the `Path` write methods: `append = false`, `createFolders = true`.

Two things worth flagging:

- The failure cleanup only deletes the file when `append = false`, since the sink truncated or created it. With `append = true` the file stays, because deleting it would throw away content the sink never wrote.
- Scala allows default arguments on only one alternative of an overloaded method. The `Stream[String, S]` sink already needed defaults for `charset`, so it keeps them, and the `Stream[Byte, S]` sink offers the two options as overloads instead.

New tests cover appending, the truncating default, folder creation, `createFolders = false` against a missing parent, and a failure part way through an append.

build.sbt
DamianReeves: Agreed, that split makes sense: `kyo-system` carries things people reach for early, so having a `kyo-core` that pulls in both would save a step. Keeping it out of this PR and doing it as a follow-up once this merges.

PR #1861, opened 2026-08-08
[kyo-system] fix the process lifecycle: deadlines, waits, stdin feeds, and child output

### LINKED ISSUE #1815: Process: stdin feeds spawn daemon threads outside Scope, so they outlive and ignore cancellation

### LINKED ISSUE #1816: Process: waitFor(timeout) measures its deadline with a platform timer instead of the ambient Clock

### LINKED ISSUE #1817: Process: waitFor() never returns for an already-signalled process on JS

DamianReeves:
### What this fixes

Three process lifecycle defects that predate the path-capability work:

- `Process.waitFor(timeout)` built its deadline from wall arithmetic; it now derives from `Clock`, and fixing that exposed a JavaScript wait bug that is fixed with it.
- Stdin feeds were not tied to the process that owned them; a feed now belongs to its process and ends with it.
- Child output larger than the pipe buffer could deadlock `text()` on the JVM; reads now stream on a dedicated thread.

The first commit extracts prerequisites (JavaScript signal reporting via `signalCode` and `exitCodeFrom`, the infinite-timeout guard, the `isAlive` fix, and the `text()` streaming read) that the two fix commits build on. They were never isolated commits upstream; this extraction carries them with their tests. One behavior change rides with it: a spawn failure now reports `Present(ExitCode(1))` instead of `Absent`.

A fourth commit renames `ExitCodeTest.scala` to `ProcessExitCodeTest.scala` so the test file shares a prefix with the source it covers.

The last commit answers the review: the two stdin feeds and the `text()` drain spawn through `Fiber.Unsafe.init` instead of `new Thread`, so the scheduler carries the blocking reads. `BlockingMonitor` drains a parked carrier's queue to the other workers and dispatches `Thread.interrupt()` to a parked carrier whose fiber was interrupted, which is what releases a feed parked on a quiet source. A process holds a stop action per feed rather than a `Thread`; `stopInputFeeds` and its contract are unchanged.

Fixes #1815
Fixes #1816
Fixes #1817

### Stack

### DIFF
--- a/kyo-system/jvm-native/src/main/scala/kyo/internal/ProcessPlatformSpecific.scala
+++ b/kyo-system/jvm-native/src/main/scala/kyo/internal/ProcessPlatformSpecific.scala
@@ -13,6 +13,23 @@ import scala.jdk.CollectionConverters.*
 
 final private[kyo] class JvmProcessUnsafe(private[internal] val jp: JProcess) extends Process.Unsafe:
 
+    // Threads feeding this process's stdin. Retained so the feed can be stopped with the process
+    // rather than running on after it, which is what an unheld thread does.
+    private val inputFeeds = new java.util.concurrent.ConcurrentLinkedQueue[Thread]()
+
+    private[internal] def registerInputFeed(t: Thread): Unit = discard(inputFeeds.add(t))
+
+    /** Stops every stdin feed. Idempotent, and safe to call after the process has already exited:
+      * a feed can still be parked reading its source even when the child is gone.
+      */
+    override def stopInputFeeds()(using AllowUnsafe): Unit =
+        var t = inputFeeds.poll()
+        while t != null do
+            if t.isAlive() then t.interrupt()
+            t = inputFeeds.poll()
+        end while
+    end stopInputFeeds
+
     def waitFor()(using AllowUnsafe, Frame): Fiber.Unsafe[ExitCode, Any] =
         val p = Promise.Unsafe.init[ExitCode, Any]()
         discard(jp.onExit().whenComplete { (exitedProcess, error) =>
@@ -259,32 +276,41 @@ final private[kyo] class JvmCommandUnsafe(
         end match
     end toJProcessBuilder
 
-    /** Feeds an InputStream into a process's stdin on a daemon thread, closing when done. */
-    private def feedInputStream(is: InputStream, processStdin: OutputStream): Unit =
+    /** Feeds an InputStream into a process's stdin on a daemon thread, closing when done.
+      *
+      * Returns the thread so the caller can retain it and stop the feed. A feed that nothing holds
+      * cannot be cancelled: interrupting the fiber that spawned the process would leave it reading
+      * its source and writing to the child for as long as the source produced bytes.
+      */
+    private def feedInputStream(is: InputStream, processStdin: OutputStream): Thread =
         val t = new Thread(() =>
             try
                 val buf = new Array[Byte](8192)
                 var n   = is.read(buf)
-                while n >= 0 do
+                while n >= 0 && !Thread.currentThread().isInterrupted() do
                     processStdin.write(buf, 0, n)
                     n = is.read(buf)
                 processStdin.close()
             catch
-                case _: IOException => ()
+                // An interrupt closes the streams underneath the read, so the resulting IOException
+                // is the ordinary way this thread stops and is not an error to report.
+                case _: IOException          => ()
+                case _: InterruptedException => ()
             finally
                 try is.close()
                 catch case _: IOException => ()
         )
         t.setDaemon(true)
         t.start()
+        t
     end feedInputStream
 
     /** Drains a Stream[Byte, Sync] into a process's stdin on a daemon thread, closing when done.
       *
       * The stream is evaluated eagerly (it is Sync-only), then the bytes are written in a background thread so that the caller (spawn) is
-      * not blocked by the write.
+      * not blocked by the write. Returns the thread for the same reason as [[feedInputStream]].
       */
-    private def feedStream(stream: Stream[Byte, Sync], processStdin: OutputStream)(using AllowUnsafe, Frame): Unit =
+    private def feedStream(stream: Stream[Byte, Sync], processStdin: OutputStream)(using AllowUnsafe, Frame): Thread =
         val chunk = Abort.run[Nothing](Sync.Unsafe.run(stream.run)).eval.getOrThrow
         val bytes = chunk.toArray
         val t = new Thread(() =>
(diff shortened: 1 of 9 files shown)


kyo-system/jvm-native/src/main/scala/kyo/internal/ProcessPlatformSpecific.scala
```diff
@@ -442,15 +472,31 @@ final private[kyo] class JvmCommandUnsafe(
...
+                // a single stream.
+                //
+                // End of stream is the right completion point for this operation: it yields the
+                // output text and does not report exit status, and the child closes stdout when it
+                // exits. The onExit handler remains only to surface a spawn failure, and completing
+                // an already-completed promise is a no-op, so whichever arrives first wins.
+                val out = proc.jp.getInputStream
+                // Unsafe: bridges a blocking pipe read off the scheduler. The thread ends at end of
+                // stream, which the child reaches by exiting.
+                val reader = new Thread(
```
fwbrasil: The scheduler is adaptive and handles blocking by default. There are exceptions but, in general,`new Thread` should be something we don't use in Kyo's codebase. Maybe use `Fiber.Unsafe.init`?

fwbrasil (approved):
LGTM other than `new Thread`

kyo-system/jvm-native/src/main/scala/kyo/internal/ProcessPlatformSpecific.scala
DamianReeves: Good call, and it turned out to be better than a like-for-like swap. All three pumps here (the two stdin feeds and the `text()` drain) now go through `Fiber.Unsafe.init`, the same bridge kyo-net's stdio pumps use.

The scheduler does more than tolerate the blocking read. `BlockingMonitor` marks the parked carrier blocked and drains its queue to the other workers, so the pump stops starving the pool, and it dispatches `Thread.interrupt()` to a parked carrier whose fiber was interrupted. That last part is what made the stdin feed work: the reason the feed held a `Thread` was to be able to interrupt a read parked on a quiet source, and interrupting the fiber now gets that for free. The read loop runs one read per `Loop` step so there is a safepoint between reads too.

So a process now holds a stop action per feed instead of a `Thread`. It interrupts the fiber and closes both ends, the close covering a source that ignores the interrupt. `CommandStdinFeedJvmTest` still pins the property it was written for and passes unchanged.

In 2f51e61.

PR #1862, opened 2026-08-08
[kyo-system] give the filesystem API a typed vocabulary

DamianReeves:
### What this adds

The typed vocabulary the filesystem work builds on:

- `FileSystemException` replaces `FileException` with a sealed hierarchy whose marker traits let a caller abort on reads, writes, or structure changes independently, and whose concrete cases record the failing operation.
- Write, move, and copy take named option values (`WriteOptions`, `MoveOptions`, `CopyOptions`, `Replace`, `Atomicity`) in place of positional booleans.
- Listing and walking take a compiled `Glob` rather than a host pattern string.
- `Path.tempDir` is `Scope` managed, and the unsafe tier returns typed results instead of raw booleans.

Roughly forty downstream files adapt mechanically to the renamed exceptions and the option values. `Path` keeps its direct signatures; no capability or runner appears in this pull request.

### Stack


kyo-system/jvm-native/src/main/scala/kyo/internal/PathPlatformSpecific.scala
```diff
@@ -319,85 +340,105 @@ final private[kyo] class NioPathUnsafe(val jpath: java.nio.file.Path) extends Pa
...
+            case e: AccessDeniedException                                 => Result.fail(FileAccessDeniedException(safe))
+            case e: java.nio.file.FileAlreadyExistsException              => Result.fail(FileAlreadyExistsException(safe))
+            case e: DirectoryNotEmptyException                            => Result.fail(FileDirectoryNotEmptyException(safe))
+            case e: IOException if NioExceptionBoundary.isNotDirectory(e) => Result.fail(FileNotADirectoryException(safe))
+            case e: IOException                                           => Result.fail(FileIOException(safe, operation, e))
+            case e: Throwable                                             => Result.panic(e)
 
 end NioPathUnsafe
 
+private[kyo] object NioExceptionBoundary:
```
fwbrasil, line 396: let's avoid leaking the underlying impl (`nio`) to the user-facing API so we have more freedom to swap internals

fwbrasil: nevermind, this is platform-specific code

kyo-system/shared/src/main/scala/kyo/internal/PathDirectories.scala
```diff
@@ -69,14 +69,22 @@ private[kyo] trait PathDirectories:
         )
     end platformProjectPaths
 
+    /** Creates a temporary directory in the platform temporary location and registers its recursive
+      * removal with the enclosing `Scope`. There is no unscoped public temp-directory primitive.
+      */
+    def tempDir(prefix: String = "kyo")(using Frame): Path < (Sync & Scope & Abort[FileStructureException]) =
```
fwbrasil, line 76: can we have an unscoped overload?

fwbrasil (approved):

kyo-system/shared/src/main/scala/kyo/internal/PathDirectories.scala
DamianReeves: Do you want me to make the default `tempDir` remain unscoped, and add a `tempDirScoped` instead?

DamianReeves: Actually scratch that `tempDirUnscoped` is more consistent with how we proceed.

PR #1863, opened 2026-08-08
[kyo-system] add a filesystem abstraction behind the Path API

DamianReeves:
### What this adds

`FileSystem.Read` and `FileSystem.Write` describe filesystem access as a service, and `HostFileSystem` implements them over the platform backends already in the module. The read and write conformance suites pin the contract against the host backend.

Nothing calls the new service yet, deliberately: no public `Path` signature changes and no module outside `kyo-system` changes in this pull request. That is what lets the capability flip arrive as its own reviewable change directly above.

One note for reviewers: `p.list(glob)` uses `Sensitive` matching while `FileSystem.host.list(p, glob)` probes the volume, so the two can disagree on a case-insensitive volume. The divergence lasts exactly one pull request and closes when the flip routes `Path` through the service.

### Stack

### DIFF
--- a/kyo-system/shared/src/main/scala/kyo/FileSystem.scala
+++ b/kyo-system/shared/src/main/scala/kyo/FileSystem.scala
@@ -1,0 +1,130 @@
+package kyo
+
+import java.nio.charset.Charset
+
+/** Filesystem backend tiers, effect-polymorphic in the backend effect `S`. [[Read]] exposes
+  * inspection and content reads. [[Write]] extends it with mutation and structure operations.
+  * Each public operation records the applicable typed failure category in its effect row and
+  * accepts a call-site [[Frame]].
+  *
+  * @see [[FileSystem.host]] for the backend over the platform implementations in this module
+  */
+object FileSystem:
+
+    trait Read[S]:
+        def defaultCaseSensitivity(using Frame): Glob.CaseSensitivity < S
+
+        // inspection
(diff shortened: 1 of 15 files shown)


kyo-system/shared/src/main/scala/kyo/FileSystem.scala
```diff
@@ -0,0 +1,130 @@
...
+/** Filesystem backend tiers, effect-polymorphic in the backend effect `S`. [[Read]] exposes
+  * inspection and content reads. [[Write]] extends it with mutation and structure operations.
+  * Each public operation records the applicable typed failure category in its effect row and
+  * accepts a call-site [[Frame]].
+  *
+  * @see [[FileSystem.host]] for the backend over the platform implementations in this module
+  */
+object FileSystem:
+
+    trait Read[S]:
```
fwbrasil: Let's use `abstract class` if possible. It's the default choice in Kyo given that it's more restricted (no multiple inheritance) and avoid interface dispatches, which are more expensive

fwbrasil (approved):

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
DamianReeves: Changed both tiers to abstract classes in 7539fc7db. Nothing needed the mixin: every implementor extends exactly one of them and Write extends Read in a single chain.

PR #1864, opened 2026-08-08
[kyo-system] track filesystem access in the type system

DamianReeves:
### What this adds

The capability flip. Reads carry `PathRead` and writes carry `PathWrite`, discharged by `Path.run` and `Path.runReadOnly` against the host backend, or by `Path.runWith` and `Path.runReadOnlyWith` against any other `FileSystem`. A function's effect row now states whether it touches the disk and whether it can modify it, and a caller can substitute the backend without changing the code under test.

`PathWrite <: PathRead`, so a write-capable context also satisfies reads, and `Path.runReadOnly` rejects a program containing a write at the call site rather than at runtime.

Every module that used the direct `Path` API moves to the capability form, which is why this is the widest pull request in the stack. `Path.tail`, `Path.tailBytes` and `Path.Origin` migrate with it and open through the installed read service.

### What the new callsites look like

An operation no longer names its own failure. It names the authority it needs:

```scala
// before
val text: String < (Sync & Abort[FileReadException])       = config.read
val done: Unit   < (Sync & Abort[FileWriteException])      = out.write(text)
val kids: Chunk[Path] < (Sync & Abort[FileStructureException]) = dir.list

// after
val text: String      < PathRead  = config.read
val done: Unit        < PathWrite = out.write(text)
val kids: Chunk[Path] < PathRead  = dir.list
```

A runner discharges the capability and folds the three per-operation failure types into the `FileSystemException` umbrella:

```scala
val loaded: String < (Sync & Abort[FileSystemException]) =
    Path.runReadOnly(config.read)

val saved: Unit < (Sync & Abort[FileSystemException]) =
    Path.run(out.write(text))
```

An existing `Abort.run` or `Abort.recover` at a callsite keeps its shape. It moves to the umbrella type and gains a runner inside it, because the failure is raised by the runner rather than by the operation:

```scala
// before
Abort.run[FileWriteException](path.write(html)).map {
    case Result.Success(_) => ...
    case Result.Failure(e) => ...
}

// after
Abort.run[FileSystemException](Path.run(path.write(html))).map {
    case Result.Success(_) => ...
    case Result.Failure(e) => ...
}
```

A public API that reads on the caller's behalf stops choosing the filesystem for them. It declares `PathRead` and lets the caller pick the runner:

```scala
// before
def init(locales: Seq[Locale], start: Locale)(dir: Path): I18n < (Sync & Abort[FileReadException])

// after
def init(locales: Seq[Locale], start: Locale)(dir: Path): I18n < (PathRead & Sync)
```

Which is what makes the backend substitutable. The same call runs against the host or against a supplied `FileSystem`, with no change to the code under test:

```scala
// production: the Local-selected backend, which defaults to the host
Path.runReadOnly(I18n.init(locales, Locale.en)(bundleDir))

// a test: any FileSystem.Read, with the same program
Path.runReadOnlyWith(fixtureFileSystem)(I18n.init(locales, Locale.en)(bundleDir))
```

Temporary paths go through the service too, so a substituted backend owns them the same way it owns everything else:

```scala
val tmpFile: Path < (PathWrite & Sync & Scope) = Path.temp("kyo-build-", ".json")
val tmpDir:  Path < (PathWrite & Sync & Scope) = Path.tempDir("kyo-build-")
```

Streaming reads, follow streams and the stream sinks carry the capability the same way:

```scala
val lines: Stream[String, PathRead & Scope & Sync]  = config.readLinesStream
val live:  Stream[String, PathRead & Async & Scope] = log.tail(pollDelay = 100.millis)
val sink:  Unit < (Scope & PathWrite & Sync)        = Stream.init(rows).writeLinesTo(out)

val records: Stream[Event, PathRead & Scope & Async & Abort[DecodeException]] =
    Jsonl.watch[Event](path, from = Path.Origin.End)
```

One consequence is worth knowing before reading the diff. A backend failure is raised by the runner, not by the frame that suspended the operation, so an `Abort.run` nested inside a runner cannot observe it. Code that needs to inspect a filesystem failure runs its own runner at the point it wants to catch:

```scala
}
```

### What changed, module by module

**kyo-system** carries the change. `Path.Op` gains the operation ADT for both capabilities plus `Op.Raise`, so a failure the read loop detects itself lands on the runner's `Abort` rather than adding an `Abort` row to every follower. Every extension method on `Path` suspends instead of calling the unsafe tier directly. The four runners (`run`, `runReadOnly`, `runWith`, `runReadOnlyWith`) and the `Isolate` instances that let a capability cross a fiber boundary are new. `Path.temp` and `Path.tempDir` both move from `PathDirectories` onto `Path` and carry `PathWrite`. Each suspends its own op (`Op.Temp`, `Op.TempDir`), the service vends a removal handle (`TempFileHandle`, `TempDirHandle`), and the `Scope` finalizer removes through the service that created the entry, so a temp path made by one backend is never deleted by another. The `tempUnscoped` and `tempDirUnscoped` variants stay host primitives on purpose: an entry whose lifetime the caller owns has no scope for a service-vended handle to hang from. `StreamSystemExtensions` opens its handle through `Path.Op.OpenWrite` and writes through `Path.Op.WriteChunk` and `Path.Op.WriteString`; because the failure now arrives at the runner, the append-preserving `finish()` moves into a single `Scope` finalizer that reads the computation's outcome, which also makes it independent of the finalizer parallelism `Scope.run` was given. `PathCapabilityTest` is new and pins each method's row, including the negative law that a write program does not compile under `runReadOnly`. `PathSnapshotTest` becomes `FileSystemSnapshotTest` and drives the backend directly. README and CONTRIBUTING are rewritten around the runner table.

**kyo-schema-json** moves the JSONL file surface onto the capability: `Jsonl.read`, `readResults`, `watch`, `watchResults`, `write` and `append` carry `PathRead` or `PathWrite` instead of `Abort[FileReadException]` or `Abort[FileWriteException]`, and the decode failure stays on its own `Abort[DecodeException]`. `writeAll` reifies its fold and re-raises through `Path.Op.Raise` once the handle is released, so the caller's row stays `PathWrite` and the runner reports the failure exactly as it would report a suspended operation.

**kyo-core** exposes `Stream.fromInputStream` publicly, which the file sinks need now that they are typed in terms of the capability rather than the concrete handle.

**kyo-tasty** routes the snapshot cache through runners: listing, `stat`, `remove` and `exists` each wrap in `Path.runReadOnly` or `Path.run` and recover on `FileSystemException` instead of the three separate types. `ClasspathOrchestrator`, `SnapshotReader`, `SnapshotWriter` and `DigestComputer` follow. A classpath fidelity test is pinned to scala-library so the count only moves on a deliberate stdlib bump.

**kyo-doctest** wraps one runner per operation rather than one per block, because each `Abort.recover` there carries a distinct operation label in its `IoError` message and a shared runner would lose which operation failed.

**kyo-browser** moves `Image.writeFileBinary` and `writeFileBase64` to `Abort[FileSystemException]` with the write discharged inside, so the public signature does not leak a capability the caller never asked for. `BrowserLauncher` and `ChromeDownloader` follow, and the download tests take their temp directories through the host runner.

**kyo-ai** collapses the Claude and Codex completion error unions from `FileStructureException | FileReadException | FileWriteException` to the single `FileSystemException`, and the MCP bridge config write is discharged where it happens so the config file stays an implementation detail rather than a capability the caller must grant.

**kyo-website** and **kyo-i18n** move their read and write paths to the umbrella type with runners at the boundary. `I18n.init` declares `PathRead` so a caller can load bundles from a substituted filesystem.

**kyo-aeron**, **kyo-jsonrpc**, **kyo-pod**, **kyo-net**, **kyo-ffi**, **kyo-http**, **kyo-ui**, **kyo-markdown**, **kyo-schema-ion**, **kyo-stats-machine**, **kyo-sql-mysql** and **kyo-sql-postgres** are callsite migrations: temp directories, socket cleanup, fixture reads and container binds move to the runner form with no behavior change. `PosixStructs` additionally stops reaching into `SystemPlatformSpecific` and reads the architecture through the public `kyo.System` unsafe boundary.

**build.sbt** promotes the `kyo-system` edge from test scope to compile scope for every module whose main sources now use `Path`, and consolidates the split `dependsOn` calls those modules had accumulated.

### Stack

Part of the path-capability stack: process fixes, filesystem vocabulary, abstraction, capability flip, channels, locks, watch, durability, confinement, in-memory, zip, then overlay (#1799). Merge bottom up; GitHub retargets the stack above each merge.

The lower branches (#1856, #1857, #1852, #1854, #1888, #1861, #1862, #1863) have landed, so this branch is now a single commit rebased onto `main`. The original incremental development history remains visible in #1799's earlier timeline and in the backup refs recorded in the working notes.

### DIFF
--- a/kyo-core/CONTRIBUTING.md
+++ b/kyo-core/CONTRIBUTING.md
@@ -4,25 +4,26 @@ This file documents the internal design contracts, invariants, and conventions
 specific to `kyo-core`. Read the root `CONTRIBUTING.md` first; everything there
 applies here, and this file extends it with module-local rules.
 
+File and OS-signal capabilities live in `kyo-system`; see `kyo-system/CONTRIBUTING.md` for those conventions.
+
 ---
 
 ## Architecture overview
 
 `kyo-core` is the primary effect module. It provides:
 
-- **Concurrency**: `Async`, `Fiber`, `Channel`, `Queue`, `Hub`, `Latch`, `Barrier`, `Meter`, `Gate`
-- **Time and scheduling**: `Clock`, `Duration`, `Deadline`
-- **Streams**: `Stream`, `Pipe`, `Sink` (in `kyo-prelude`; `StreamCoreExtensions.scala` adds core-specific combinators)
+- **Concurrency**: `Async`, `Fiber`, `Channel`, `Queue`, `Hub`, `Latch`, `Barrier`, `Meter`, `Gate`, `Exchange`
+- **Time and scheduling**: `Clock`, `Duration`, `Deadline`, `Retry`, `Timeout`
+- **Streams**: `StreamCoreExtensions` (core-specific combinators; `Stream`, `Pipe`, and `Sink` live in `kyo-prelude`)
 - **Atomic primitives**: `AtomicInt`, `AtomicLong`, `AtomicBoolean`, `AtomicRef`, `Adder`
-- **Environment**: `Console`, `Log`, `Signal`
+- **Environment and process**: `Console`, `Log`, `Signal` (reactive)
 - **Application entry**: `KyoApp`, `KyoAppInterrupts`
 - **Resource management**: `Scope`, `Sync`, `Sync.Unsafe`
-- **Data types**: `Chunk`, `Span` (in `kyo-data`; re-exported here)
 - **Observability**: `Stat`
 
 Every API in this module is cross-platform (JVM, Scala.js / Node, Scala Native)
-unless it is in a `jvm/`, `js/`, or `native/` source tree and explicitly
-documented as platform-specific.
+unless it is in a `jvm/`, `jvm-native/`, `native/`, or `js-wasm/` source tree and
+explicitly documented as platform-specific.
 
 ---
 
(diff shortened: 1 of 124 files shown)


kyo-core/CONTRIBUTING.md
```diff
@@ -4,25 +4,26 @@ This file documents the internal design contracts, invariants, and conventions
 specific to `kyo-core`. Read the root `CONTRIBUTING.md` first; everything there
 applies here, and this file extends it with module-local rules.
 
+File and OS-signal capabilities live in `kyo-system`; see `kyo-system/CONTRIBUTING.md` for those conventions.
```
hearnadam: Do we need this?

fwbrasil (approved):

PR #1866, opened 2026-08-08
[kyo-system] take an advisory lock on a path

DamianReeves:
### What this adds

Advisory file locks. `Path.Lock` is scope-managed: acquiring registers the release with the enclosing `Scope` before anything is claimed, so an interrupt landing between the claim and the finalizer cannot strand the lock. `LockMode` selects shared or exclusive compatibility, and `LockWait` selects whether a contended attempt fails immediately, waits, or waits until a deadline.

```scala
import kyo.*

val guarded = Scope.run {
    Path.runReadOnly {
        Path("state.bin").lock(Path.LockMode.Exclusive, Path.LockWait.Immediate).map(_.check)
    }
}
```

`tryLock` answers `Absent` for a conflicting holder and never waits for one. It still carries `Async`, for a different reason: a backend that merges same-process claims onto one platform lock has a span in which a compatible claim is being taken but is not yet shareable, and answering during that span would report a conflict that does not exist. Waiting it out is bounded and needs no holder to release anything.

There is no public manual release. A lock goes when its `Scope` closes, and `check` reports `FileLockOwnershipLostException` for the backends that can lose a claim underneath the handle.

### What changed

**`Path`** gains `LockMode`, `LockWait`, the public `Lock` trait, and the `lock` and `tryLock` extension methods. The private surface a platform implements is `RawLock` plus `LockAttempt`, which separates "no" from "not yet": only the unsettled answer is worth retrying on its own, since a real conflict clears when the holder releases and that is the waiting caller's concern.

**`FileSystem.Read`** gains `tryLock` and `lock`, so a read-only backend still vends locks. `FileSystem.awaitLock` holds the waiting policy in one place and paces its retries with `Async.sleep` rather than `Clock.sleep`, since the latter hands back the timer fiber instead of suspending on it and would turn the retry into a busy spin against the holder it is waiting on.

**`FileSystemException`** gains the `FileLockException` marker and the `FileLockUnavailableException`, `FileLockTimeoutException` and `FileLockOwnershipLostException` leaves. `FileSystemOperation` gains `Lock`, so an acquisition failure names what it was doing instead of borrowing `Read`.

**JVM and Native** back a lock with `FileChannel.tryLock` behind a process-wide `NioPathLockRegistry`. POSIX advisory locks do not exclude the owning process from reacquiring its own lock, so same-process exclusion is enforced explicitly rather than assumed from the platform call. The registry key is the resolved real path, so two names for one file through a symbolic link share one entry instead of opening two channels the JVM then refuses as overlapping.

**JS and Wasm** implement an `O_EXCL` lockfile protocol. A claim records host, pid and a random token, and an owner whose process is provably gone is reclaimed rather than left blocking the path for good.

**Tests.** `FileSystemLockTest` joins the read, write and channel suites as the fourth cross-backend contract, so every backend added later inherits the same laws. `HostPathLockTest` covers the interrupt window from both ends: an interrupt delivered from inside the claim node with the OS lock already held, and repeated interrupts landing before the claim is made. The repeated-interrupt stress case is currently marked ignored: it reliably exposes #1928, a core defect where an interrupted fiber can lose its `Scope.ensure` finalizers under load, which no backend-side code can compensate for. The claim path additionally records the holder through a tombstone, so a claim that completes after the finalizer has already drained releases the lock in place instead of stranding it.

### Stack

Part of the path-capability stack: process fixes, filesystem vocabulary, abstraction, capability flip, channels, locks, watch, durability, confinement, in-memory, zip, then overlay (#1799). Merge bottom up; GitHub retargets the stack above each merge.

The lower branches (#1856, #1857, #1852, #1854, #1888, #1861, #1862, #1863, #1864, #1865) have landed, so this branch is now a single commit rebased onto `main`. The original incremental development history remains visible in #1799's earlier timeline and in the backup refs recorded in the working notes.

### DIFF
--- a/kyo-system/shared/src/test/scala/kyo/FileSystemLockTest.scala
+++ b/kyo-system/shared/src/test/scala/kyo/FileSystemLockTest.scala
@@ -1,9 +1,7 @@
 package kyo
 
 /** Shared behavioral contract for [[FileSystem.Read]] lock implementations. */
-abstract class FileSystemLockTestSuite extends kyo.test.Test[Any]:
-
-    private given Frame = Frame.internal
+abstract class FileSystemLockTest extends kyo.test.Test[Any]:
 
     protected def withFileSystem(
         use: (FileSystem.Read[Sync], Path) => Unit < (Async & Sync & Scope & Abort[FileSystemException])
@@ -283,4 +281,4 @@ abstract class FileSystemLockTestSuite extends kyo.test.Test[Any]:
         }
     }
 
-end FileSystemLockTestSuite
+end FileSystemLockTest
--- a/kyo-system/shared/src/test/scala/kyo/FileSystemLockTest.scala
+++ b/kyo-system/shared/src/test/scala/kyo/FileSystemLockTest.scala
@@ -80,6 +80,18 @@ abstract class FileSystemLockTest extends kyo.test.Test[Any]:
     }
 
     "lifecycle" - {
+        "locking claims a sentinel sibling and leaves the data path untouched" in {
+            withFileSystem { (fileSystem, path) =>
+                Scope.run {
+                    fileSystem.lock(path, Path.LockMode.Exclusive, Path.LockWait.Immediate).map { _ =>
+                        fileSystem.exists(path).map { dataExists =>
+                            assert(!dataExists, s"locking created the data file $path")
+                        }
+                    }
+                }
+            }
+        }
+
         "scope closure releases the owning claim" in {
             withFileSystem { (fileSystem, path) =>
                 Scope.run(fileSystem.lock(path, Path.LockMode.Exclusive, Path.LockWait.Immediate).unit).andThen {
--- a/kyo-system/shared/src/test/scala/kyo/FileSystemLockTestSuite.scala
+++ b/kyo-system/shared/src/test/scala/kyo/FileSystemLockTestSuite.scala
@@ -1,0 +1,286 @@
+package kyo
+
+/** Shared behavioral contract for [[FileSystem.Read]] lock implementations. */
+abstract class FileSystemLockTestSuite extends kyo.test.Test[Any]:
+
+    private given Frame = Frame.internal
+
+    protected def withFileSystem(
+        use: (FileSystem.Read[Sync], Path) => Unit < (Async & Sync & Scope & Abort[FileSystemException])
+    )(using Frame): Unit < (Async & Sync & Scope & Abort[FileSystemException])
+
+    private def assertImmediate(first: Path.LockMode, second: Path.LockMode, expected: Boolean)(using
+        kyo.test.AssertScope
+    ): Unit < (Async & Sync & Scope & Abort[FileSystemException]) =
+        withFileSystem { (fileSystem, path) =>
+            fileSystem.lock(path, first, Path.LockWait.Immediate).map { held =>
+                fileSystem.tryLock(path, second).map { attempted =>
+                    assert(held.mode == first)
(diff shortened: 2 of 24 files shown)


DamianReeves:

|---|---|

```scala
}
```

### What this branch needs

The rename set above is the whole job here, and this branch is first in line for it. It adds no `FileSystem.Write` implementation of its own, so the new `temp` member arrives already satisfied through `HostFileSystem`.

The temp swap described above lands on this branch in `CommandTest` (five callsites) and `PathTest` (two). All seven call `remove` on the file themselves, so all seven are the unscoped kind and want `tempUnscoped`, not the new `temp`.

kyo-system/shared/src/test/scala/kyo/FileSystemLockTestSuite.scala
```diff
@@ -0,0 +1,286 @@
+package kyo
+
+/** Shared behavioral contract for [[FileSystem.Read]] lock implementations. */
+abstract class FileSystemLockTestSuite extends kyo.test.Test[Any]:
+
+    private given Frame = Frame.internal
```
fwbrasil: can you remove the `Suite` suffix? it makes this implicit unnecessary

kyo-system/shared/src/test/scala/kyo/FileSystemLockTest.scala
```diff
@@ -0,0 +1,286 @@
...
+        }
+        "exclusive with shared waits for release" in {
+            assertWaiting(Path.LockMode.Exclusive, Path.LockMode.Shared, compatible = false)
+        }
+        "exclusive with exclusive waits for release" in {
+            assertWaiting(Path.LockMode.Exclusive, Path.LockMode.Exclusive, compatible = false)
+        }
+    }
+
+    "lifecycle" - {
```
fwbrasil, line 82: If I'm not mistaken the lock marker in the inode gets cleared whenever the process closes a handle to a locked file. We need to either document the behavior or close the gap. Example problematic scenario:

```
        for
            lock <- Path("state.bin").lock(Path.LockMode.Exclusive) 
            data <- Path("state.bin").readBytes // will open and close, releasing locks at the OS level
            _    <- lock.check 
            _    <- Path("state.bin").writeBytes(data) 
        yield ()
```

kyo-system/shared/src/test/scala/kyo/FileSystemLockTestSuite.scala
DamianReeves: Done. Dropped the `Suite` suffix and the `Frame` given on all four conformance suites (read, write, channel, lock) and on the fixtures object; everything compiles on auto-derivation.

kyo-system/shared/src/test/scala/kyo/FileSystemLockTest.scala
DamianReeves: Confirmed with a cross-process probe on Linux (Temurin 21) and macOS: closing the second handle silently releases the lock and `FileLock.isValid` still reports true, so `check` couldn't catch it either. The JDK still takes plain `fcntl` locks, not OFD.

Closed rather than documented: JVM and Native now claim a sentinel sibling (`<path>.kyo-lock`, the same name the Node protocol already used), so data I/O never opens a handle to the claimed file and the scenario above is safe by construction; a test enacts it. The suffix is an optional parameter on `lock`/`tryLock` and is part of the lock's identity.

The trade is that the lock no longer interoperates with a foreign process fcntl-locking the data file directly, which is noted on `Path.Lock`.

PR #1867, opened 2026-08-08
[kyo-system] watch a path for filesystem changes

DamianReeves:
### What this adds

Filesystem watching as its own capability. `PathWatch` is separate from `PathRead`, discharged by `Path.runWatch` and `Path.runWatchWith`, so a program that only observes changes says so in its type. `FileSystem.Watch` supplies the backend; the host watcher polls through `PathWatch.polling` rather than binding a native notification API.

### Stack

### DIFF
--- a/kyo-system/shared/src/main/scala/kyo/FileSystem.scala
+++ b/kyo-system/shared/src/main/scala/kyo/FileSystem.scala
@@ -154,10 +154,10 @@ object FileSystem:
       * This tier is independent of [[Read]] and [[Write]]. A filesystem
       * advertises it only when it can provide the complete watch contract.
       */
-    trait Watch[S]:
+    trait Watch:
         def openWatcher(path: Path, options: WatchOptions)(using
             Frame
-        ): Path.Watcher < (S & Async & Scope & Abort[FileWatchException])
+        ): Path.Watcher < (Async & Scope & Abort[FileWatchException])
     end Watch
 
     abstract class Write[S] extends Read[S]:
@@ -238,8 +238,8 @@ object FileSystem:
         FileSystem.host.asInstanceOf[FileSystem.Read[Any]]
     )
 
-    private val watchLocal = Local.init[FileSystem.Watch[Any]](
-        FileSystem.host.asInstanceOf[FileSystem.Watch[Any]]
+    private val watchLocal = Local.init[FileSystem.Watch](
+        FileSystem.host
     )
 
     /** Runs `value` with `fileSystem` selected as the backend used by [[Path.run]] and
@@ -255,12 +255,12 @@ object FileSystem:
 
     /** Runs `value` with a coherent read, write, and watch backend selection. */
     @scala.annotation.targetName("letWatchable")
-    def let[A, S, FS](fileSystem: FileSystem.Write[FS] & FileSystem.Watch[FS])(value: A < S)(using
+    def let[A, S, FS](fileSystem: FileSystem.Write[FS] & FileSystem.Watch)(value: A < S)(using
         Frame
     ): A < (FS & S) =
         local.let(fileSystem.asInstanceOf[FileSystem.Write[Any]])(
             readLocal.let(fileSystem.asInstanceOf[FileSystem.Read[Any]])(
-                watchLocal.let(fileSystem.asInstanceOf[FileSystem.Watch[Any]])(value)
+                watchLocal.let(fileSystem)(value)
             )
         )
 
@@ -270,7 +270,7 @@ object FileSystem:
     private[kyo] def useReadErased[A, S](f: FileSystem.Read[Any] => A < S)(using Frame): A < S =
         readLocal.use(f)
 
-    private[kyo] def useWatchErased[A, S](f: FileSystem.Watch[Any] => A < S)(using Frame): A < S =
+    private[kyo] def useWatchErased[A, S](f: FileSystem.Watch => A < S)(using Frame): A < S =
         watchLocal.use(f)
 
     private[kyo] def letErased[A, S, FS](fileSystem: FileSystem.Write[FS])(value: A < S)(using Frame): A < S =
@@ -301,6 +301,6 @@ object FileSystem:
       * `Result[File*Exception, A]` into `Abort[FileSystemException]`, so it preserves current
       * `Path` behavior exactly.
       */
-    def host: FileSystem.Write[Sync] & FileSystem.Watch[Sync] = HostFileSystem()
+    def host: FileSystem.Write[Sync] & FileSystem.Watch = HostFileSystem()
 
 end FileSystem
--- a/kyo-system/shared/src/main/scala/kyo/FileSystem.scala
+++ b/kyo-system/shared/src/main/scala/kyo/FileSystem.scala
@@ -1,6 +1,7 @@
 package kyo
 
 import java.nio.charset.Charset
+import kyo.Path.WatchOptions
 
 /** Filesystem backend capabilities, effect-polymorphic in the backend effect `S`. [[Read]] exposes
   * inspection, content reads, channels, and locks. [[Write]] extends it with mutation and structure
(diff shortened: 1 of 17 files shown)


DamianReeves:

|---|---|

```scala
}
```

### What this branch needs

The rename set above, nothing more. No `FileSystem.Write` implementation of its own, so the new `temp` member arrives already satisfied through `HostFileSystem`.

Worth settling on this branch: a naming collision that neither side saw coming. This branch adds a `FileSystem.Watch[S]` tier and `PathWatch.scala` for filesystem change notification. Independently, `main` renamed the private tail-polling loop `Path.follow` to `Path.watch`, and renamed `Jsonl.follow` to `Jsonl.watch` with it. After the rebase `Path.watch` means "poll an open file for appended bytes" while `PathWatch` and `FileSystem.Watch` mean "subscribe to filesystem change events". Two unrelated concepts, one word, same module. This is the branch that introduces the second meaning, so it is the natural place to decide whether one of them gets a different name.

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
```diff
@@ -143,6 +144,22 @@ object FileSystem:
...
+      *
+      * Implementations register observation before returning a watcher,
+      * so mutations made immediately after acquisition remain visible.
+      * The returned watcher and its asynchronous stream are owned by the
+      * surrounding [[Scope]]. Closing that scope releases backend resources.
+      *
+      * This tier is independent of [[Read]] and [[Write]]. A filesystem
+      * advertises it only when it can provide the complete watch contract.
+      */
+    trait Watch[S]:
```
fwbrasil: I see it's a pre-existing pattern but can you elaborate on why we allow a free `S`? Isn't it always sync or async? If not, effects can still be added when handling the effect instead of tracking here

kyo-system/shared/src/main/scala/kyo/Path.scala
```diff
@@ -371,6 +407,27 @@ object Path extends PathPlatformSpecific:
             }
     end isolateWrite
 
+    /** Stateless isolation for watch operations. See [[isolateRead]]. */
+    given isolateWatch: Isolate[PathWatch, Async, PathWatch] with
+        type State        = FileSystem.Watch[Any]
```
fwbrasil: I think this isn't sound. We use the same pattern in `kyo-ai` with parametrized interfaces like `Tool[S]` and `Prompt[S]` but I've been wanting to review it as well. If the `S` effects in the interface aren't properly handled by the `Isolate`, which is the case here since it's erased, a forked computation can end up executing with a missing handler. If you use `Watcher[Var[Int]]` and fork a computation, it should fail on any use of the `Var` effect since it won't propagate with the fork. Earlier versions of `kyo-ai` avoided this unsoundness by requiring an isolate at the time of an enablement but the mechanism was lost at some point. We can try to make this code safe but I have trouble to see the value of tracking `S` in service interfaces here

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
DamianReeves: You are right. I carried `S` over from `Read`/`Write` for consistency, but watcher acquisition already exposes its complete `Async & Scope & Abort[FileWatchException]` contract and no watcher implementation needs another effect. I removed the type parameter from `Watch` and the backend row from `runWatchWith`; implementations now handle any internal effects before exposing the capability. The focused watcher suites pass on JVM, JS, Native, and Wasm.

kyo-system/shared/src/main/scala/kyo/Path.scala
DamianReeves: Agreed. That erased `S` path through `Isolate` is the decisive issue. Commit `72d1504f8` removes the backend effect parameter from `Watch`, removes `FS` from `runWatchWith`, and makes the isolate state concrete `FileSystem.Watch`, so a watcher cannot advertise an effect whose handler is then lost across a fork.

kyo-system/shared/src/main/scala/kyo/Path.scala
```diff
@@ -327,6 +332,37 @@ object Path extends PathPlatformSpecific:
...
+        )(
+            [C] =>
+                (op, cont) =>
+                    op match
+                        case WatchOp.Open(path, options) => fileSystem.openWatcher(path, options).map(cont)
+                        case WatchOp.Raise(error)        => Abort.error(error)
+        )
+
+    /** Runs `program`, discharging [[PathWatch]] against the Local-selected watch backend. */
+    def runWatch[A, S](program: A < (PathWatch & S))(using
```
fwbrasil, line 415: I wonder if we could track just `PathRead` for file watching? We've been very selective of what appears in the pending set so far.

DamianReeves: I considered that, but `PathRead` cannot preserve the current optional capability boundary. If `openWatcher` suspends only `PathRead`, every `PathRead` handler must interpret it. A program passed to `runReadOnlyWith(readBackend)` would then typecheck even when that backend cannot watch, leaving us to reject at runtime or silently use the Local/host watcher, both of which break the existing negative-capability and backend-coherence guarantees. The sound way to use only `PathRead` is to make the full watch contract mandatory on every `FileSystem.Read`, including read-only and virtual fixtures. Keeping `PathWatch` separate means it enters the pending set only at `openWatcher`, while backends that only provide reads remain valid. I would keep the separate effect unless we want watching to become mandatory for all read backends.

kyo-system/shared/src/main/scala/kyo/PathWatch.scala
```diff
@@ -0,0 +1,389 @@
...
+  *
+  * [[Immediate]] observes only direct children of the watched root.
+  * [[Recursive]] also observes descendants at every depth.
+  * The watched root itself is not passed through the glob filter.
+  * Its disappearance is instead reported as [[PathChange.Invalidated]].
+  *
+  * Depth is applied before glob matching, and paths emitted by either
+  * mode retain their complete backend path.
+  */
+enum WatchDepth derives CanEqual:
```
fwbrasil: Exposing members in the `kyo` package is an important decision given that we recommend `import kyo.*`. It should contain only the main API entrance points and, when multiple types are exposed like in `kyo-http`, we use a name prefix. For example, exposing `kyo.Request` as the http request could eventually conflict with `Request` in other contexts so we use `kyo.HttpRequest`. In this case, it seems we can nest these types in the companion?

DamianReeves: Agreed. Commit `93b916b74` keeps only the `PathWatch` effect at the package level alongside `PathRead` and `PathWrite`, and moves the supporting API to `Path.Change`, `Path.WatchOptions`, `Path.WatchDepth`, and `Path.MatchCase`. JVM, JS, Native, and Wasm watcher suites pass with the new names.

kyo-system/shared/src/main/scala/kyo/Path.scala
DamianReeves: @fwbrasil - Would you prefer that I fold watch into the Read capability? I can do that.

fwbrasil: Good point on the collapse implying a requirement for all impls! What implementations do you foresee we having? Any where it'd become problematic?

DamianReeves: The next branches in this stack already provide both cases. `InMemoryFileSystem` and `RootConfinedHostFileSystem` implement `Write & Watch`, so they can preserve observation. The zip backend deliberately returns only `FileSystem.Read[Sync]`: it is an archive snapshot and has no complete change-notification contract. Folding watch into `Read` would make that backend implement a fake watcher or reject at runtime. The same issue applies to user-defined read-only adapters and plausible remote/object-store readers that can list and read but cannot provide loss-aware, scope-managed change streams. Given the concrete zip backend, I would keep `Watch` optional and `PathWatch` separate.

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
ahoy-jon, line 54: not sure about that one to be honest.
```suggestion
    abstract class Read[-S]:
```

ahoy-jon (approved):

PR #1868, opened 2026-08-08
[kyo-system] make a write survive a crash

DamianReeves:
### Problem

Writing a file does not by itself guarantee that its contents and directory entry survive a crash. Atomic replacement must also preserve an existing file's access permissions.

### Solution

`syncDirectory` persists directory entries. `durableReplace` writes a create-new sibling temporary, syncs its content and metadata, closes it, atomically replaces the target, then syncs the parent directory. When it creates missing parent directories, it also syncs their entries through the first existing ancestor. A directory-sync failure is reported even if replacement has already happened.

Temporary names use a random identifier per execution, have bounded length, and skip the target itself. Exclusive creation handles name collisions. Invalid parent paths fail without collision retries. Cleanup ownership transfers synchronously with acquisition, before an effect continuation can be interrupted. Cleanup failures are logged and attached to the primary exception, and failures in the logger cannot replace that exception. Scoped sibling temporaries also clean up if acquisition finishes after their scope closes. Directory sync preserves precise missing-path and non-directory errors, including Windows Node errors for paths below regular files. Unsupported host platforms return a typed failure.

Host replacement preserves the existing regular file's owner, group, permission mode, and access ACLs. A native helper acquires the replacement with restrictive access, restores and verifies the captured controls before file synchronization, and fails before replacement when the controls cannot be preserved. The helper is shared by JVM, Node.js, Scala Native, and Wasm through `kyo-ffi`. Node.js and Wasm require `koffi` and the native library for the host platform. On Windows, restoration preserves legacy and automatically inherited security descriptors, including mandatory integrity labels, without adding inherited ACEs. Delete authorization is checked without backup privileges before temporary creation. Metadata changes by another process are not serialized.

Rebased onto `origin/main` (`d0847e051`, Scala 3.9.0), preserving the merged capability, channel, lock, watcher, temporary-file, and conformance-suite changes. The FFI loader tests move to the integration-test module to allow kyo-system to depend on kyo-ffi without creating a test dependency cycle.

#### Stack

The lower PRs through #1867 have merged into `main`. This PR is the next layer, followed by #1869, #1870, #1871, and #1799. Merge bottom up.

The earlier stack history remains available in the original PR timeline.

#### Validation

After rebasing onto `d0847e051`, `kyo-systemJVM/test` passed all 671 tests. A clean Native build passed all 53 focused durability and permission tests. `git range-diff` confirmed that the durability patch is unchanged.

Earlier validation:

After extracting the unrelated fixes, this PR passed its isolated JVM (671 tests) and Native (637 tests) suites directly on `main`. The FFI JVM suite passed 345 tests, and the relocated loader integration tests passed all nine cases. The rebuilt filesystem stack also passed all four platform suites and compiled all 41 README examples.

- `kyo-systemJVM/test`
- `kyo-systemJS/test`
- `kyo-systemNative/test`
- `kyo-systemWasm/test`
- `kyo-systemJVM/doctest`

Shared regression coverage includes exact mode and ACL preservation, inherited ACL protection, denied metadata/deletion access, native path validation, protocol ordering, new parent entries, concurrent replacement, forced filename collisions, invalid-parent retry behavior, cleanup and logger failures, acquisition panics, long and reserved filenames, interruption at acquisition and ownership transfer, and acquisition after scope closure. Host durability tests explicitly cover Windows' unsupported-directory-sync result. Local four-platform checks ran on macOS; a targeted Windows permission matrix passed all 26 checks, and the helper compiled and linked for Windows x64 and ARM64 at both ends of the stack.

#### Separate fixes

Merged into the current base: #1946, #1947, #1948, #1949, #1952, and #1953. The remaining unrelated fixes continue in their own PRs:

- [ci] retry Coursier download and setup failures: https://github.com/getkyo/kyo/pull/1946
- [build] align cross-platform checks with current toolchains: https://github.com/getkyo/kyo/pull/1947
- [core] keep inline macro JVM signatures resolvable: https://github.com/getkyo/kyo/pull/1948
- [stats] use the Native runtime weak-reference hierarchy: https://github.com/getkyo/kyo/pull/1949
- [scheduler] bound blocked-worker queue migration: https://github.com/getkyo/kyo/pull/1950
- [net] retain descriptors through poller operations: https://github.com/getkyo/kyo/pull/1951
- [test] preserve sampled workers in leak checks: https://github.com/getkyo/kyo/pull/1952


DamianReeves:

|---|---|

```scala
}
```

### What this branch needs

Beyond the rename set, one direct collision.

This branch declares `Path.TempFileHandle` for `siblingTemporary`:

```scala
/** Service-owned sibling temporary cleanup handle. Internal. */
abstract private[kyo] class TempFileHandle:
    def path: Path
    def remove()(using AllowUnsafe): Unit
end TempFileHandle
```

#1864 now declares `TempFileHandle` with the same two members, for the scoped `Path.temp`. The rebase will conflict on the declaration. Resolution is to drop this branch's copy and keep #1864's, then `siblingTemporary` and `tempFileHandle` compile against it unchanged. Worth widening the scaladoc at that point so it covers both uses rather than only the sibling one.

`siblingTemporary` itself is unaffected. It stays a distinct operation from `temp`: a create-new file next to a specific target for the durable-replace path, not a general temp factory.

kyo-ffi/plugin/src/main/scala/kyo/ffi/sbt/KyoFfiPlugin.scala
```diff
@@ -961,13 +961,13 @@ object KyoFfiPlugin extends AutoPlugin {
         // references to the system-lib symbols resolve against the `-l` flags that follow.
         ffiNativeLinkingOptions := {
             val platform = ffiTargetPlatform.value
+            val libs     = ffiLibrariesResolved.value
```
fwbrasil, line 964: I imagine these changes are to fix the sbt warnings?

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
```diff
@@ -231,6 +274,129 @@ object FileSystem:
...
+        Frame
+    ): Unit < (S & Sync & Abort[FileReadException | FileWriteException | FileStructureException]) =
+        Sync.defer {
+            val parent                                                          = target.parent.getOrElse(Path())
+            var acquired: Maybe[(Path.ChannelCloseHandle, Path.TempFileHandle)] = Absent
+            var channelClosed                                                   = false
+            var moved                                                           = false
+            def open(attempt: Int): (Path, Path.WriteChannel[S & Sync], () => Unit < (S & Sync)) <
+                (S & Sync & Abort[FileWriteException | FileStructureException]) =
+                val temporary = parent / s".kyo-durable-$attempt"
```
fwbrasil: if `parent` isn't a valid path, it seems this code can keep retrying up to the attempt limit

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
```diff
@@ -231,6 +274,129 @@ object FileSystem:
...
+                    }
+                end if
+            end loop
+            Sync.ensure {
+                // Unsafe: cleanup owns acquisition before any effect continuation can be interrupted.
+                Sync.Unsafe.defer {
+                    if !completed then
+                        acquired.foreach { (close, handle) =>
+                            try close.close()
+                            catch case _: Throwable => ()
```
fwbrasil: can you avoid swallowing exceptions?

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
```diff
@@ -231,6 +274,129 @@ object FileSystem:
...
+        fileSystem: FileSystem.Write[S & Sync],
+        target: Path,
+        onAcquire: Path.TempFileHandle => Unit
+    )(using Frame): Path.TempFileHandle < (S & Sync & Abort[FileWriteException | FileStructureException]) =
+        Sync.defer {
+            val parent                                                          = target.parent.getOrElse(Path())
+            var acquired: Maybe[(Path.ChannelCloseHandle, Path.TempFileHandle)] = Absent
+            var completed                                                       = false
+            def loop(attempt: Int): Path.TempFileHandle < (S & Sync & Abort[FileWriteException | FileStructureException]) =
+                val temporary = parent / s".kyo-temporary-$attempt"
```
fwbrasil: can't we have conflicts with different executions using the same file name?

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
```diff
@@ -231,6 +274,129 @@ object FileSystem:
...
+                            catch case _: Throwable => ()
+                            try handle.remove()
+                            catch case _: Throwable => ()
+                        }
+                }
+            }(Abort.run[FileWriteException | FileStructureException](loop(0))).map(Abort.get(_))
+        }
+    end siblingTemporary
+
+    private[kyo] def durableReplace[S](fileSystem: FileSystem.Write[S & Sync], target: Path, bytes: Span[Byte])(using
```
fwbrasil: can you check if this method keeps the same permissions as the original?

kyo-ffi/plugin/src/main/scala/kyo/ffi/sbt/KyoFfiPlugin.scala
DamianReeves: Yes. These remove sbt task-linter warnings about `.value` lookups inside an `if`. Regular tasks already evaluate those dependencies unconditionally, so moving `ffiLibrariesResolved.value` and `streams.value` above the condition makes that behavior explicit without changing dependency evaluation or the task results. This change is now in `main` and retained by the rebased stack.

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
DamianReeves: Implemented in [8fdda0d52](https://github.com/getkyo/kyo/commit/8fdda0d529155463f0667860dbd6d463dbb1806d), with the Windows restoration follow-up in [480c5f7e3](https://github.com/getkyo/kyo/commit/480c5f7e35f676a20c85b7bd699e2945edc8d32b). Existing regular files retain their owner, group, permission bits, and access ACLs. The native helper creates a restrictive temporary, restores and verifies the captured controls before synchronization and atomic replacement, and fails before replacing the target if preservation is unsupported. New files use their destination parent's normal creation permissions.

Windows now preserves both legacy and automatically inherited descriptors, including mandatory integrity labels, and checks ordinary delete authorization before creating the temporary. The Windows matrix passed all 26 checks. The [Windows ARM64 Scala run](https://github.com/getkyo/kyo/actions/runs/35139657243) passed all durability and permission tests on JVM and Node.

Shared regressions cover private modes, full ACLs, inherited ACLs, and denied metadata/deletion access. PR #1799 carries the same requirements through nested overlays and recovery, protects recovery logs with private directories, and retains destination-parent defaults for new files.

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
DamianReeves: Fixed in the bundled update ([c873f0260](https://github.com/getkyo/kyo/commit/c873f0260e2385edfa4398549dc3120deb8a6c1d), acquisition changes in [8fdda0d52](https://github.com/getkyo/kyo/commit/8fdda0d529155463f0667860dbd6d463dbb1806d)). Parent creation runs before exclusive candidate acquisition. A regular-file parent or ancestor fails immediately; only a genuine temporary-name collision triggers a retry. Shared regressions cover both invalid-parent cases and forced candidate collisions.

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
DamianReeves: Fixed in [c873f0260](https://github.com/getkyo/kyo/commit/c873f0260e2385edfa4398549dc3120deb8a6c1d). Cleanup errors are logged and attached to the original failure, while both close and removal are attempted. A throwing logger cannot replace that failure. The host removal handle also propagates a failed Result instead of discarding it. Regressions cover deletion failure, both cleanup actions failing, a failing logger, and preservation of the original panic.

kyo-system/shared/src/main/scala/kyo/FileSystem.scala
DamianReeves: Fixed in [c873f0260](https://github.com/getkyo/kyo/commit/c873f0260e2385edfa4398549dc3120deb8a6c1d). Each execution gets a random UUID, while retaining exclusive creation, collision retries, and the check against the target itself. The regression reproduces one execution targeting `.kyo-durable-0` while another uses that name as its temporary; the corrected implementation installs the requested bytes. A separate test forces a genuine candidate collision.

fwbrasil:
@DamianReeves the recent leak check improvements uncovered a new leak that I fixed directly in main: https://github.com/getkyo/kyo/commit/d0518a73604b45a5dba0285280e9db8e4504c988

There's only an option to rebase the PR so I'll let you handle it to avoid issues

PR #1871, opened 2026-08-08
[kyo-system] add the zip filesystems and the staged-changes contract

DamianReeves:
### What this adds

The zip backends and the staged-changes contract. `ZipReadOnlyFileSystem` serves an archive as a read-only backend; `ZipRewriteFileSystem` stages edits and rewrites the archive on commit. Both sit on `ZipArchive` and `ZipInflate`, which parse and emit the container without a host zip library, with a codec shared across all four platforms.

`FileSystem.StagedChanges`, `Conflict`, `Resolution` and `CommitConflict` arrive here rather than with the overlay because the zip rewrite backend is the first implementor of the contract. Zip has no live lower, so its commit rewrites the whole archive unconditionally and never raises a conflict.

Both ZIP backends follow the current lock-sentinel and typed-walk contracts. Rewrite acquisition forwards cleanup ownership synchronously. Directory sync validates the combined archive and staged view, including baseline-only directories and non-directory ancestors.

### Stack

The lower PRs through #1867 are in `main`. The remaining stack is #1868 → #1869 → #1870 → #1871 → #1799, rebased onto `origin/main` (`3b8679117`). Merge bottom up.

### Validation

The complete stack passes full `kyo-system` tests on JVM, JS, Native, and Wasm, plus JVM doctests. This layer also passes independently with `kyo-systemJVM/testOnly kyo.Zip* kyo.PathDurability*`. Local runtime checks ran on macOS.


DamianReeves:

|---|---|

```scala
}
```

### What this branch needs

Beyond the rename set, `ZipRewriteFileSystem` needs a `temp` member, since `FileSystem.Write` now declares one:

```scala
def temp(prefix: String, suffix: String)(using Frame): Path.TempFileHandle < (S & Abort[FileStructureException])
```

There is a design call here rather than a mechanical fill-in: whether a temp file is an entry in the staged rewrite or scratch space outside it. Scratch space is probably right, since a temp file is by definition not part of the archive being produced, but that should be a stated decision and it should match whatever `tempDir` already does on this branch.

PR #1876, opened 2026-08-14
feat(kyo-ui): drag and drop support and structurally valid reactive HTML ranges

### LINKED ISSUE #1875: Foreach/foreachKeyed wraps output in <span>, which is invalid (and gets foster-parented) directly inside <table>/<tbody>/<tr>
## Summary

`Signal[Chunk[A]].foreach` / `.foreachKeyed` / `.foreachKeyedIndexed` unconditionally wrap their
rendered batch in a single `<span data-kyo-reactive>` (or `<g>` in SVG context), regardless of
where the `Foreach` sits in the tree. This makes it unsafe to use any of these APIs to render
`<tr>` rows directly inside a `<table>`/`<tbody>`, or `<th>`/`<td>` cells directly inside a `<tr>`
— those HTML5 insertion modes only accept a restricted child set (`tr`/`tbody`/`colgroup`/etc., or
`td`/`th`, respectively), so a bare `<span>` is invalid there. Browsers apply the HTML5
tree-construction algorithm's foster-parenting step and move the invalid `<span>` out of the
table, producing duplicated/malformed markup instead of the intended table content.

## Repro

```scala
import kyo.*
import kyo.UI
import kyo.UI.*

def rowsUI(using Frame): UI < Async =
  for rows <- Signal.initRef(Chunk("A", "B", "C"))
  yield table(
    tr(th("Name")),
    rows.foreachKeyed(identity) { name => tr(td(name)): UI }
  )
```

Rendering this (either via `UI.runRenderPage`/server HTML, or via `UI.runMount` in the browser)
produces something like:

```html
<table>
  <tr><th>Name</th></tr>
</table>
<span data-kyo-reactive>
  <tr><td>A</td></tr>
  <tr><td>B</td></tr>
  <tr><td>C</td></tr>
</span>
```

The browser's parser foster-parents the `<span>...` block to just before the `<table>`, so the
rows never end up as descendants of the table at all. The same problem occurs for `<th>`/`<td>`
cells rendered via `Foreach` directly inside a `<tr>`.

## Why this is surprising

- `kyo-ui/shared/src/test/scala/kyo/TableTest.scala`'s `"foreach rows add remove"` test uses
  exactly this pattern (`UI.table(rows.foreach(r => UI.tr(UI.td(r))))`) and passes, but its
  assertions (`assertContains("A")`, etc.) only check that text is present somewhere on the page —
  they don't assert on DOM structure, so they don't catch the foster-parenting duplication. The
  test's "success" gave a false impression that this pattern is safe.
- `UI.Tbody`'s doc comment says wrapping rows in `UI.tbody(...)` avoids the problem ("rows patched
  into a live table programmatically... become direct `<table>` children, which renders fine")
  — but `<tbody>`'s insertion mode has the same restricted child set as `<table>`, so this doesn't
  actually resolve the foster-parenting issue for a `Foreach`'s `<span>` wrapper either.

## Impact

There's currently no documented, safe way to render a dynamically-keyed list of table rows or
cells using `foreach`/`foreachKeyed`/`foreachKeyedIndexed`. This forces application code to either
(a) avoid keyed rendering for table structure entirely (losing focus/DOM-identity stability across
reorders — the reason to reach for `foreachKeyed` in the first place), or (b) avoid semantic
`<table>` markup altogether in favor of a div/ARIA-role-based grid layout, which is what we ended
up doing.

## Possible fixes

- Have `Foreach`'s HTML rendering special-case table-context parents (similar to how `Tbody`
  already gets special treatment) and either omit the wrapper element or use a context-appropriate
  wrapper (e.g. no wrapper at all when the immediate children are all `tr`/`td`/`th`).
- At minimum, update the `Tbody` doc comment and the `TableTest` "foreach rows add remove" test to
  either demonstrate this is actually safe (with a structural assertion, e.g. asserting
  `table > tbody > tr` relationships) or document the limitation explicitly.

## Environment

- kyo version: v1.0.0-RC6
- Found while building a Scala.js kyo-ui application (a RACI matrix editor) that needed
  focus-preserving reorder of table rows/columns via `foreachKeyedIndexed`.

DamianReeves:
## Summary

This PR delivers complete drag and drop support for kyo-ui, two working proof applications, and the issue #1875 correction that makes reactive HTML structurally valid.

### Drag and drop

Full drag and drop for kyo-ui applications across every input mode:

- A typed drag value model with `Drag.MediaType` and `Drag.MediaTypePattern`, validated media boundaries and URI collision rejection, and a hardened wire protocol with bounded event paths.
- The pure `Sortable` engine plus a managed helper layer: `Sortable.moveBy` for typed collections, `Sortable.moveGroups` for multi-collection boards with anchor threading, `Sortable.expandSelection` for the standard multi-select rule, `locked` keys that can neither move nor anchor a move, `Drag.Source.sortable`/`Drag.Target.sortable` payload-free wiring, and `Drag.Decision.fromResult` to adapt reducer results into handler decisions.
- Accessible sensors in the local runtime: pointer drags with an activation distance, touch drags with a hold delay, and full keyboard sessions (Enter/Space lift, orientation-aware arrows, Home/End, Tab target traversal, confirm, Escape) with per-session live-region announcements, reserved `data-kyo-dragging`/`data-kyo-drop-valid`/`data-kyo-drop-position` presentation attributes, and focus restoration by stable key.
- An embedded server-push runtime with the same state machine and attributes, installed once per served page: sortable drops emit semantic `SortMove` wire events, decisions resolve over the WebSocket, and teardown is idempotent on socket close and page hide.
- Lazy dropped file and directory reads: `Drag.DroppedFile.from(item).bytes(chunkSize)`/`text(charset)` and `Drag.DroppedDirectory.from(item).entries(limits)` stream content over the session in bounded chunks with typed `Drag.FileError` failures, `ByteSize` quantities throughout, cancellation through `Scope`, and `Disconnected` failure of every pending read on session close.
- Lifecycle proof: one hundred iteration stress suites over sensor sessions, file reads, disconnects, and wire sessions pin sessions, tokens, live regions, frames, and presentation attributes at zero after close.

The README documents the whole surface: wiring, sortable helpers, files and directories, sensor behavior, local versus server-push data flow, and the cleanup guarantees.

### Proof applications

Two demos exercise the surface end to end, both proven by `DragScenarioItTest` on JVM, JS, Native, and Wasm. The scenario tests drive the real demo views through the real server dispatch path and assert the demo state plus the resolved decisions.

**Kanban** (`demo.KanbanDemo`): cards drag across lanes, a checkbox multi-selection moves together in board-visible order (including selections spanning lanes, which land contiguously at the destination anchor), and one total `applyMove` reducer built on `Sortable.move` serves pointer drags, keyboard moves, and the arrow-button fallback controls.

The board as served:

Kanban demo

A live interaction against the running server: two cards selected and armed for the next move, then both moved together in visible order by one action through the shared reducer (To Do drains to its empty state, In Progress counts 3, the selection stays ticked):

Kanban selection armed

Kanban selection moved together

A drag in flight against the running server: the sensor runtime's semi-transparent preview clone follows the pointer over In Progress, placed after its anchor card, while the live region announces the pickup:

Kanban drag in progress

**InventoryGrid** (`demo.InventoryGridDemo`): rows and columns reorder as two independent sortable collections over one state ref, cells follow the column order, the two collections never mix, and the SKU column is locked: it can neither move nor anchor a move.

InventoryGrid demo

Two rows selected against the running server, armed to move together on the next drag:

InventoryGrid selection armed

Screenshots show real interactions against the served demos; the mid-drag frame is the embedded sensor runtime's own in-page presentation.

### Issue #1875: logical comment ranges

Closes #1875. HTML `Reactive`, `Foreach`, and independently signal-bound elements now render paired logical comment ranges (`<!--kyo-rs:id-->` / `<!--kyo-re:id-->`) instead of invalid `span[data-kyo-reactive]` wrappers, so reactive content no longer breaks restricted HTML parents (tables, lists, selects). SVG keeps `g[data-kyo-reactive]` group hosts. Renderer recursion decides through closed `Namespace`, `BoundaryMode`, `ParentContext`, `TableContent`, and `RenderHost` ADTs. Direct table rows get a synthetic `tbody` range host; authored `tbody` sections stay single lifecycle roots with JavaScript properties, focus-auto, enter and leave, and animation hooks; bound text inputs morph in place preserving node identity, focus, and caret. `HtmlOp.ReplaceRange` carries validated range replacements; embedded `HtmlOp.Replace` is SVG-only.

### kyo-core and kyo-kernel left untouched

An earlier revision of this branch changed `IOPromise`, `IOTask`, `Scope`, and `Safepoint` to order interrupt completion after nested `Scope` finalizers. Per review, those changes are reverted and this PR no longer touches kyo-core or kyo-kernel: the known finalizers-on-interrupt issue stays open until `bracket` lands in kyo-kernel and covers every scenario, not only computations running on fibers. The `ReactiveUITeardownTest` cases that prove that ordering keep the same known-issue ignore main carries, including the two cases this branch adds; they become live with the bracket work.

The branch also merges current `main`, so the range transport composes with #1903's send-what-moved diffing: an HTML region replaces the content between its comment anchors with `ReplaceRange`, and an SVG region diffs against its previously rendered tree and sends path-addressed `Replace` ops only for the nodes that changed.

Manual testing of the demos surfaced two fixes: a drained Kanban lane kept a zero-height card list, so no pointer could land a drop in it (the list now keeps a minHeight, proven by a real-Chrome regression test against the board view), and the delegation teardown suite still expected the pre-sensors listener set of nineteen where the sensors commit had grown it to twenty-eight (stale counts reproduce at the pre-merge branch head; the suite now passes complete with zero ignored cases).

## Verification

Local, after the merge with `main` and the kyo-core revert:

- `kyo-uiJVM/test` full suite green: 2357 passed, 0 failed. The only ignored cases are the finalizer-ordering teardown tests, with the same known-issue reason `main` carries.
- `kyo-uiJS/test` full suite green, including the delegation teardown suite at its corrected listener counts.
- `kyo-schemaJVM/test` green (the `Schema[ByteSize]` instance and its round-trip tests).
- Manual runs of both demos against real Chrome: Kanban pointer drag across lanes, multi-select move spanning lanes (lands contiguously, selection stays ticked), add-card, drop into a drained lane, empty-state placeholder; InventoryGrid row reorder, column reorder with cells following, and the locked SKU column refusing to move or anchor.
- Native and Wasm run in CI; they were not re-run locally.

## Follow-ups

- Evaluate moving `Drag.MediaType` and `Drag.MediaTypePattern` to a higher-level module such as `kyo-data`, since media type modeling is not UI-specific.
- `KyoException.getMessage` appends a trailing space in production mode when no cause is provided (`Maybe("")` is `Present("")` in the detail fold). Cosmetic, pre-existing in kyo-data, surfaced while pinning range diagnostics; worth a one-line fix with a test in a kyo-data change.


DamianReeves:
Note on the red Native builds: I chased the kyo-browserNative crash to ground before concluding it is not this branch. It reproduces locally about every other run on the base commit as well, and the caught backtrace shows the abort inside Scala Native's stack-trace capture for an ordinary NumberFormatException thrown by CDP schema decoding, under concurrent worker threads. Full evidence and possible directions are in #1879.

While bisecting I did find and fix a real latent defect on the new deferred-interrupt path: the waiter-chain walk recursed per wrapper (quadratic time, unbounded stack). The last commit makes it a constant-stack loop with a 500k-deep regression test.

kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -38,6 +38,8 @@ private[kyo] class IOPromise[E, A](init: State[E, A]) extends Safepoint.Intercep
                     false
                 case l: Linked[E, A] @unchecked =>
                     doneLoop(l.p)
+                case _: Interrupting[E, A] @unchecked =>
```
fwbrasil: This is quite sensitive and hot code and I'd like to avoid these changes. The PR description mentions the known finalizers bug but it should remain open until we have bracket in `kyo-kernel`, which will provide proper guarantees for all scenarios. This fix would cover only one arm of the problem and computations that run outside of fibers will still face the issue.

DamianReeves: Reverted, the branch doesn't touch kyo-core or kyo-kernel anymore. I put the teardown tests that needed that fix back to ignored, same reason as the existing ones, so they can be re-enabled whenever the proper fix lands in the kernel. Also merged latest main while I was at it.

Issue #1881, opened 2026-08-15
[ci] kyo-pod podman tests fail on the new GitHub runner images: preinstalled podman 5.8.4 shadows the apt-installed 4.9.3


DamianReeves:
## Summary

Since roughly August 12 the linux CI jobs fail on main and on open PRs. Every failure is the same: the kyo-pod container suites fail all of their `[podman]` cases on JVM, JS, Native, and Wasm, while the docker cases pass. The trigger is a GitHub runner image update, not a change in this repository.

## Symptom

Every failing test reports the same error when it tries to exec into a container it just started:

```
Daemon returned HTTP 500: {"cause":"container state improper","message":"can only create exec sessions on running containers: container state improper","response":500}
```

Containers are created and started, then exit before the first exec reaches them. The logs show no start error; the death is only visible at exec time.

## Root cause

The new runner images ship a preinstalled podman 5.8.4 under `/usr/local`, which shadows the podman 4.9.3 that our setup action installs from the Ubuntu repositories into `/usr/bin`:

- Our setup step (`.github/actions/setup/action.yml`, "Install podman") runs `apt-get install podman`, which installs podman 4.9.3 plus its matching stack (conmon 2.1.10, passt from Feb 2024, slirp4netns).
- dpkg does not know about the binary in `/usr/local/bin`, so apt happily installs 4.9.3, but `/usr/local/bin` comes first on PATH, so `podman` still resolves to 5.8.4.
- The API service we start therefore runs podman 5.8.4 on top of the old Ubuntu 4.9.x support stack, together with our `containers.conf` that pins the OCI runtime to runc (a workaround for a broken crun on the old images).
- Under that mixed stack, rootless containers (rootless network command is now pasta) exit immediately after start, and every exec gets the HTTP 500 above.

## Evidence

| Run | Date | Runner image | podman | Result |
|-----|------|--------------|--------|--------|
| dependabot coursier 3.0.2 branch, linux-x64 | Aug 10 | ubuntu-24.04 20260720.247.2 | 4.9.3 | pass |
| main `[kyo-data] add Glob`, linux-x64 job | Aug 12 | ubuntu-24.04 20260720.247.2 | 4.9.3 | pass |
| main `[kyo-data] add Glob`, linux-arm64 job | Aug 12 | ubuntu-24.04-arm 20260810.90.1 | 5.8.4 | fail (kyo-pod) |
[...]

PR #1882, opened 2026-08-15
[ci] survive the new runner images: use the image podman, smoke-probe the stack at setup

DamianReeves:
Fixes the linux CI failures tracked in #1881.

## What broke

The 20260810.x runner images (rolled out gradually from Aug 12) ship podman 5.8.4 under `/usr/local` with matching helpers (conmon, netavark) in `/usr/local/lib/podman`. Our setup action still apt-installed Ubuntu's podman 4.9.3 on top of it. The package's `containers.conf` selects the journald log driver, the image's conmon is built without journald support, so conmon exits 1 on every container start. The container never leaves the `created` state and each exec fails with HTTP 500 `container state improper`, which took down all `[podman]` cases of the kyo-pod suites on JVM, JS, Native, and Wasm, on both x64 and arm64.

A probe matrix on the new images (see #1881) confirmed: the image's own podman stack passes a create + start + exec cycle with crun, runc, and the default runtime on both architectures, while any mix with the apt package fails with the exact CI error.

## The fix

- Use the podman the runner image ships when there is one, and apt-install podman only when the image provides none, so each stack stays coherent on its own.
- Add a health gate right after the podman API service comes up. A broken container stack now fails the setup step with a named cause instead of surfacing two hours later inside the kyo-pod suites.
- The runc pin stays where it was and stays unconditional: the probe verified the image podman works pinned to runc on both architectures, so there is no need to branch on where podman came from.

## Approach: logic in scripts, workflows stay declarative

The first version of this branch put both the install conditionals and the health checks inline in the setup action, which hurt the readability of the workflow. Both now live in scripts, and each script carries a flag that changes policy without restructuring the workflow.

**`scripts/apt-install.sh --if-missing`** expresses "install unless present". The callers read as plain one-liners:

```sh
bash scripts/apt-install.sh --if-missing podman uidmap
bash scripts/apt-install.sh --if-missing cmake go:golang-go liburing-dev
```

It skips a package when dpkg reports it installed, or when a binary with its name is already on PATH. The PATH check is the piece that matters for #1881: it is what keeps the image's `/usr/local` podman in place instead of layering the apt package on top of it. When the command name differs from the package name, a `cmd:pkg` token maps one to the other (`go:golang-go`: the runners ship go as a tarball, so neither dpkg nor a `golang-go` lookup would find it). When nothing is missing the script exits without touching apt, which also saves the `apt-get update` round-trip. The same flag replaces every inline `command -v X || apt-install` and `dpkg -s X || apt-install` guard that had accumulated across `release.yml`, `deploy-site.yml`, `readme.yml`, `release-setup/action.yml`, and the libaeron step of the setup action, so the repo now has one idiom for "install unless present" instead of three.

**`scripts/container-check.sh`** is the health gate. It verifies both runtimes the kyo-pod suites depend on: the full podman cycle (create, start, exec a real container, after printing the resolved OCI runtime and conmon so a failure names the component) and the docker daemon. Both checks always run, so one broken runtime does not hide the state of the other. The workflow calls it as a single line:

```sh
bash scripts/container-check.sh
```

By default a failed check fails the job. `--warn-only` runs the same checks and diagnostics but emits a GitHub warning annotation and exits 0: the switch for maintainers who later decide the gate should be loud instead of fatal, one flag in the workflow instead of a rewrite of the step.

### DIFF
--- a/.github/actions/release-setup/action.yml
+++ b/.github/actions/release-setup/action.yml
@@ -23,8 +23,7 @@ runs:
     - name: Prepare BoringSSL
       shell: bash
       run: |
-        command -v cmake >/dev/null 2>&1 || bash scripts/apt-install.sh cmake
-        command -v go >/dev/null 2>&1 || bash scripts/apt-install.sh golang-go
+        bash scripts/apt-install.sh --if-missing cmake go:golang-go
         bash kyo-net/build/boringssl/build-boringssl.sh linux-x86_64
 
     # The staged libaeron tree is gitignored, so anything that compiles kyo-aeron's C shim has to
@@ -32,10 +31,9 @@ runs:
     - name: Prepare libaeron
       shell: bash
       run: |
-        command -v cmake >/dev/null 2>&1 || bash scripts/apt-install.sh cmake
         # The Aeron driver links -luuid on Linux and the package providing the libuuid.so link
         # target is not preinstalled.
-        dpkg -s uuid-dev >/dev/null 2>&1 || bash scripts/apt-install.sh uuid-dev
+        bash scripts/apt-install.sh --if-missing cmake uuid-dev
         bash kyo-aeron/scripts/build-aeron.sh linux-x86_64
 
     - uses: coursier/setup-action@v3.0.0
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -107,7 +107,27 @@ runs:
       if: ${{ contains(inputs.os, 'linux') }}
       shell: bash
       run: |
-        bash scripts/apt-install.sh podman uidmap
+        # Use the podman the runner image ships when there is one. Since image 20260810.x the
+        # runners carry podman under /usr/local with matching helpers (conmon, netavark) in
+        # /usr/local/lib/podman, and installing Ubuntu's podman package ON TOP of that breaks
+        # every container start: the package's containers.conf selects the journald log driver,
+        # the image's conmon is built without journald support, conmon exits 1, the container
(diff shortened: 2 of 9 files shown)


DamianReeves:
The linux-x64 JVM job hit a transient NoClassDefFoundError in kyo-net (IoUringDriverAcceptTransientErrnoTest) on both runs of this branch. It is unrelated to the podman work here: the classfile exists, the suite passes in isolation (8 probe runs), fd and memory are healthy at the failure moment, and the same suite passed on main on the same runner image. Full evidence dossier and open directions are in #1884. I will rerun the failed job once the current run completes.

DamianReeves:
Run 31894626796 update: the kyo-pod suites now pass everywhere they run, so both podman 5.8.4 fixes hold (image podman selection and the PidsLimit omission). The two JVM job failures in this run are distinct pre-existing races that the recently added end-of-run checks surface: the kyo-net classload flake (#1884, linux-x64) and a kyo-sql leaked client connection (#1885, linux-arm64). Both are tracked with full evidence; neither is caused by this branch. I will rerun the failed jobs when the run completes.

DamianReeves:
CI is fully green: all eight build jobs pass, including the kyo-pod container suites on every platform and architecture. The two JVM job failures along the way were the pre-existing races tracked in #1884 and #1885, not this branch; #1885 is now precisely diagnosed (the pool's exit decision never runs when an externally interrupted fiber is parked on a socket read) with an on-demand reproduction kept on the ci/podman-runner-image-probe branch. Ready for review.

PR #1890, opened 2026-08-20
[kyo-sql] prepare the SQL modules for Scala 3.9

fwbrasil:
## Problem

Scala 3.9 is the next LTS release. Validating kyo against 3.9.0-RC6 surfaced three issues in the SQL modules, all addressed here on 3.8.4 so the eventual bump is a two-line `build.sbt` change.

Two are straightforward `-Werror` failures, self-evident in the diff.

The third is silent: whole-table aggregates stop folding at compile time, because 3.9 wraps the inlined construction in a cast the fold does not see through. `.runStatic` fails the compile, so we caught it, but `.run` would have degraded to runtime rendering unnoticed.

## Solution

Strip the cast in `resolveBindings`, beside the `substituteCo` arm already there for the same reason. Every `FromExpr` entry point runs `resolveBindings` first, so one arm covers all nesting levels, where patching the individual `peel` and `unwrap` sites would cover only the shapes we anticipated.

## Notes

The cast strip is not 3.9-only: it also fires on `asInstanceOf` nodes present in 3.8 trees, so it changes behaviour on the current compiler. The SQL suites pass on 3.8.4 and 3.9.0-RC6 is green across JVM, JS, Native and Wasm.

### DIFF
--- a/kyo-sql-mysql/shared/src/test/scala/kyo/internal/mysql/MysqlEncoderShortTest.scala
+++ b/kyo-sql-mysql/shared/src/test/scala/kyo/internal/mysql/MysqlEncoderShortTest.scala
@@ -201,7 +201,7 @@ class MysqlEncoderShortTest extends Test:
         case class Row(n: Short) derives CanEqual
 
         val schema = summon[SqlSchema[Row]]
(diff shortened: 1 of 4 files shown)


DamianReeves (approved):
LGTM

hearnadam (approved):

PR #1894, opened 2026-08-25
[kyo-schema-json, kyo-ai] let an agent use an MCP server's tools

fwbrasil:
### Problem

kyo ships an MCP server module and an LLM module, and they could not compose. MCP's premise is that a tool set drops into any host, and this was the one place in the stack where that visibly did not hold.

The blocker was a missing primitive, not a missing convenience. `Tool.init[In]` needs a static `Schema[In]` at compile time, while an MCP server publishes its tools' schemas over the wire, so a host has no Scala type to name. Because no `Tool` constructor took a runtime schema, a user could not write the bridge either. The only way through was hand-wrapping each tool while naming both its types, which works for a server you wrote yourself and never for a third-party one.

The same modules also reported things that were not true. Codex reported zero tokens on every turn, so the budget guardrail the README documents summed zero and could never fire. `turns` reported 1 for a generation that made three tool calls. `AI.stream[String]` is documented as incremental with no caveat, yet a 250-word answer arrives from the command harnesses as a single chunk, so a chat UI shows a blank screen for the whole generation with nothing in the API saying why.

### Solution

`Tool.fromMcp(client)` walks a connected server's published tools and builds one `Tool` per entry from its wire metadata. It takes no type parameters, so it works against a server whose Scala types are not on the classpath.

Underneath it, `Tool.initDynamic` is the primitive that was missing: a tool built from a JSON Schema held as a value, with arguments arriving as an open `Structure.Value`. The declared schema binds, checked before the body runs and on the tool-error channel, so a malformed call is repaired by the model rather than reaching a body that never agreed to receive it. Recovering the structural type that check needs is what `JsonSchema.toStructure` does, and it is why this spans two modules: the converter serves only the constructor and is dead code without it. Tools also now carry the schema they advertise instead of always deriving one, since re-deriving a third-party schema drops `pattern`, bounds and descriptions before the model sees them.

The reporting fixes are smaller. Codex's usage was a mapping gap, not a harness limit: the app-server does report it, on a notification absent from the route table, and an unrouted notification is dropped rather than delivered unhandled. `turns` is floored at tool calls plus one on both harness backends, which run their own tool loop inside a single kyo turn. Streaming incrementality is declared per backend instead of promised everywhere. `stream` gains the inline-input overloads `gen` has.

### Notes

The bridge's acceptance test is stricter than "types not on the classpath": the server's tool is built from a runtime `ToolMeta`, so its input type does not exist in Scala at all, and an `AI.gen` still calls it end to end. JVM, JS and Native.

Verified live against OpenAI, Codex and Claude Code, which caught an earlier form of the turn-count fix still reporting 1 for a real tool call.

Correction to the finding this came from: object streaming is not incremental on the harness backends either, so it cannot serve as evidence the streaming machinery is fine. Both modes are the same defect.

Does not point at a server launched as a subprocess; that needs a client-side stdio transport. The bridge works over `inMemory`, `unixDomain` and any custom wire, and needs no change when that lands.

### DIFF
--- a/kyo-ai/shared/src/main/scala/kyo/AI.scala
+++ b/kyo-ai/shared/src/main/scala/kyo/AI.scala
@@ -102,6 +102,43 @@ object AI:
     def stream[A: Schema](using Frame, Tag[Emit[Chunk[A]]]): Stream[A, LLM & Async & Scope & Abort[AIStreamException]] < LLM =
         init.map(ai => ai.stream[A])
 
+    /** [[stream]] seeded with an input, the streaming counterpart of `gen(input)`. The input is appended
+      * as a user message before the stream is projected, so `AI.stream[String]("...")` reads the way
+      * `AI.gen[Answer]("...")` does rather than requiring the caller to seed the context by hand.
+      */
+    def stream[A: Schema](using
+        Frame,
+        Tag[Emit[Chunk[A]]]
+    )[B: Schema](input: B): Stream[A, LLM & Async & Scope & Abort[AIStreamException]] < LLM =
+        init.map(ai => ai.stream[A](input))
+
+    def stream[A: Schema](using
+        Frame,
+        Tag[Emit[Chunk[A]]]
+    )[B: Schema, C: Schema](input1: B, input2: C): Stream[A, LLM & Async & Scope & Abort[AIStreamException]] < LLM =
+        stream[A]((input1, input2))
+
+    def stream[A: Schema](using
+        Frame,
+        Tag[Emit[Chunk[A]]]
+    )[B: Schema, C: Schema, D: Schema](
+        input1: B,
+        input2: C,
+        input3: D
+    ): Stream[A, LLM & Async & Scope & Abort[AIStreamException]] < LLM =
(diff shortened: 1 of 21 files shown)


kyo-ai/shared/src/main/scala/kyo/AI.scala
```diff
@@ -102,6 +102,43 @@ object AI:
     def stream[A: Schema](using Frame, Tag[Emit[Chunk[A]]]): Stream[A, LLM & Async & Scope & Abort[AIStreamException]] < LLM =
         init.map(ai => ai.stream[A])
 
+    /** [[stream]] seeded with an input, the streaming counterpart of `gen(input)`. The input is appended
+      * as a user message before the stream is projected, so `AI.stream[String]("...")` reads the way
+      * `AI.gen[Answer]("...")` does rather than requiring the caller to seed the context by hand.
+      */
+    def stream[A: Schema](using
```
DamianReeves, line 109: These will be useful

DamianReeves (approved):
LGTM

PR #1913, opened 2026-08-26
[kyo-schema] fix recurrent NoClassDefFoundError on incremental builds

fwbrasil:
### Problem

Incremental builds of kyo-schema fail with a false compile error:

```
java.lang.NoClassDefFoundError: kyo/internal/FocusMacro$
```

reported at every `derives Schema` site. Only a full clean clears it. Any edit to a macro source kyo-schema depends on triggers it, and so does reverting that edit, so it recurs through a working session and forces a multi-minute clean into loops that should take seconds. It is indistinguishable from a genuine breakage and gets misread as one; AI agents take it at face value and start fixing code that was never broken.

### Solution

`Schema.derived` splices a small object that delegates to `FocusMacro`. Move it into `FocusMacro.scala`.

The compiler suspends a unit whose macro implementation is being compiled in the same run, but decides that from the spliced symbol alone, one hop deep. Alone in its own file the delegate referenced too little for zinc to invalidate it, while `FocusMacro` is invalidated by nearly any kyo-data change, so the compiler saw no current-run dependency and expanded against a class file zinc had already deleted. Sharing a source file makes the two inseparable.

It stays a separate object because that is what gives it its own class file. Splicing `FocusMacro` directly fails a clean build with `Cyclic macro dependencies`: the deriving units are suspended, so the types named in `FocusMacro`'s signatures have no class files yet.

### Notes

An incremental build can also report green while sources are broken. That is a separate defect, not addressed here and not yet diagnosed.


DamianReeves (approved):
Interesting... this is a good find.
LGTM

PR #1915, opened 2026-08-27
[kyo-schema] recursive transform propagation, Protobuf collection handling, and a differential oracle (#1887, #1747)

### LINKED ISSUE #1747: [kyo-schema] Protobuf codec mishandles nested and empty collections

### LINKED ISSUE #1887: [BUG]: Decoding recursive case classes leads to Unknown variant errors
### Version

1.0.0-RC6

### Scala Version

3.8.4

### Expected Behavior

When decoding JSON to recursive case classes I assume that excess fields are ignored.

### Actual Behavior

For recursive case classes, excess fields seem to be picked up by and treated as unknown variants. This happens for enums as well as sealed traits.

### Steps to Reproduce

This scala-cli script shows the error nicely.

```scala
#!/usr/bin/env scala-cli
//> using scala 3.8.4
//> using dep "io.getkyo::kyo-schema-json:1.0.0-RC6"
//> using dep "io.getkyo::kyo-core::1.0.0-RC6"

import kyo.*
import kyo.schema.*

sealed trait Node
case class Parent(ref: String, children: Seq[Node]) extends Node
case class Child(ref: String) extends Node

final case class Container(items: Seq[Node]) derives Schema

given Schema[Node] = Schema[Node]
  .discriminator("type")
  .renameAllVariants(Schema.NameCase.KebabCase)

object BugDemo extends KyoApp:
  run {
    for
      // Issue 1: discriminator field treated as variant name
      json1 = """{"items":[{"type":"parent","ref":"p1","children":[{"type":"child","ref":"c1"}]}]}"""
      r1 = Json.decode[Container](json1, 32, Int.MaxValue)
      _ <- Console.printLine(s"Issue 1 (nested discriminator): ${r1 match
        case Result.Success(_) => "OK"
        case Result.Failure(e) => e.getMessage}")

      // Issue 2: extra fields treated as variant names
      json2 = """{"items":[{"type":"parent","ref":"p1","children":[{"extra":"val","type":"child","ref":"c1"}]}]}"""
      r2 = Json.decode[Container](json2, 32, Int.MaxValue)
      _ <- Console.printLine(s"Issue 2 (extra field in nested): ${r2 match
        case Result.Success(_) => "OK"
        case Result.Failure(e) => e.getMessage}")
    yield ()
  }
```

Save the code into a file an run with `scala-cli run <name>`

### Current Workaround

Either manually clean the raw JSON before parsing or switch to another library.

I created the sample above using AI. The issue occurred in handwritten code and puzzled me enough to let me let the agent investigate the problem.

DamianReeves:
Fixes two kyo-schema bug reports and adds the differential testing #1747 called for. Three commits, one per concern.

Closes #1887. Closes #1747.

## Behavior changes at a glance

| Change | Scope | Kind |
|---|---|---|
| Fluent config (discriminator, variant/field renames) now reaches recursive occurrences | all codecs | breaking for data persisted in the old mixed format |
| Empty collection / absent optional map values round-trip instead of corrupting or throwing | Protobuf | fix; wire unchanged |
| Nested collections get framed records instead of collapsing | Protobuf | wire change, only for shapes that previously corrupted |
| Non-string map entries share one field number; top-level maps work | Protobuf | wire change, only for shapes that previously corrupted |
| Absent scalar field decodes to its proto3 default instead of `MissingFieldException` | Protobuf reader only | behavior change; matches every conformant implementation |
| Dropped field decodes to typed empty value (`""`), not `null` | all codecs | behavior change, strictly safer |

Everything that round-tripped before is byte-identical: the pre-existing byte-exact conformance suites pass unchanged.

### The bug

The reporter's setup, reduced:

```scala
sealed trait Node
case class Parent(ref: String, children: Seq[Node]) extends Node
case class Child(ref: String) extends Node

final case class Container(items: Seq[Node]) derives Schema

given Schema[Node] = Schema[Node]
    .discriminator("type")
    .renameAllVariants(Schema.NameCase.KebabCase)
```

The configuration silently applied only at the top level. Encoding shows it directly:

```scala
Json.encode(Container(Seq(Parent("p1", Seq(Child("c1"))))))
// before: {"items":[{"type":"parent","ref":"p1","children":[{"Child":{"ref":"c1"}}]}]}
//                    outer level flat, as configured ^^^^^^^^  nested child in raw wrapper format
// after:  {"items":[{"type":"parent","ref":"p1","children":[{"type":"child","ref":"c1"}]}]}
```

Decoding the flat format the user actually wrote failed symmetrically, producing the reported errors, because the nested decoder expected `{"Child":{...}}` and treated the first flat field name as a variant name:

```
Issue 1: Unknown variant 'type'
Issue 2: Unknown variant 'extra'
```

### Root cause

The derivation macro resolves recursive references by summoning. Inside the right-hand side of a **top-level** given, the self-summon finds nothing eligible and falls back to a fresh, unconfigured `Schema.derived`; an **object-scoped** given ties the knot to the configured instance. Measured across a 2x2 matrix (top-level vs object-scoped, given before vs after the container): both top-level arrangements fail, both object-scoped arrangements work, declaration order is irrelevant. The same mechanism affects enums and recursive products with `renameAllFields` (nested fields kept their original names).

### The fix

Make every arrangement behave like the working one, at runtime: a schema with structural transforms registers itself on the Writer/Reader for the duration of its transformed write/read, and a transform-free Product or Sum schema reached deeper in the same operation delegates to the innermost registered override with the same nominal identity (name plus tag). Wrapper readers share the wrapped reader's stack; internally constructed replay readers/writers inherit it. When nothing is registered the raw path costs one empty-list check.

### Behavior change

Recursive sums configured through the fluent API previously produced the mixed wire format above, which round-tripped only through kyo itself. They now encode uniformly, so pre-existing persisted data in the mixed format does not decode unchanged. Object-scoped givens and annotation-configured schemas already produced the uniform format and are unaffected.

### Symptoms fixed

All four reported symptoms confirmed at HEAD by measurement.

**Empty collection map values corrupted the next entry** (symptom 1). proto3 encodes an empty `repeated` value as a key-only entry; the read loop discarded the one signal that reveals this and read a value anyway, consuming the following entry's bytes:

```
input:   Map(1 -> Chunk("a"), 2 -> Chunk.empty, 3 -> Chunk("c"))
decoded: Map(1 -> Chunk("a"), 2 -> Chunk("c"), 3 -> Chunk("c"))   // entry 2 stole entry 3's value
```

The value read is now gated on `hasNextField`; a key-only entry decodes to the value schema's `absentDefaultValue`:

```scala
val v =
    if reader.hasNextField() then
        discard(reader.field()) // "value"
        vSchema.serializeRead(reader)
    else
        vSchema.absentDefaultValue match
            case Maybe.Present(dv) => dv
            case _                 => throw MissingFieldException(Seq.empty, "value")(using reader.frame)
```

**Nested collections collapsed at encode** (symptom 2). Three distinct values, one encoding:

```
1 -> []        = 8aa9ea04020802
1 -> [[]]      = 8aa9ea04020802
1 -> [[],[]]   = 8aa9ea04020802
```

A nested collection is now one length-delimited record under the enclosing repeated field number with bare elements inside; an empty inner collection is a zero-length record, so all three values above get distinct bytes and round-trip. Verified to depth three, including messages and mixed packability.

**Maybe/Option-valued maps failed** (symptom 4): `maybeSchema` and `optionSchema` gained `absentDefaultValue`, and the reader's string-keyed entry arm presents a key-only entry through `isNil`.

(Symptom 3, the Dict decode failures, was partially fixed in #1744; the remainder is the finding below.)

### New finding: the map entry envelope leaked field numbers

Every entry after the first inherited the previous entry's value field id, and a top-level map's first entry had no field number at all:

```
Dict(1 -> 10, 2 -> 20) at top level, before:
12 04 08 04 10 28   02 02 02 14
^^ second entry, accidentally tagged with field 2 (the previous entry's "value" id)
                    ^^ first entry, written as a bare packed run with no entry framing at all
decoded: Dict(1 -> 1, 4 -> 8)
```

Top-level and multi-entry non-string maps could not round-trip; the token-based tests the issue flagged had routed around exactly this. Entries now share the map's field number (synthetic field 1 at top level), and top-level collection decode learns its field number from the first tag.

### Behavior and wire implications

- Depth-one repeated fields and every previously round-tripping shape are byte-identical; the existing conformance suites pass unchanged.
- The new framing applies only to shapes that previously produced corrupt bytes (nested collections, multi-entry and top-level non-string maps), so no working persisted data is affected.
- `validateCanonical`'s doc now states the measured rejection set: non-native map keys remain the only rejected shape; nested and empty repeated are lossless codec extensions, while `protoSchema` still rejects them in `.proto` text generation, where they genuinely have no declaration.
- The two bypassing token-based Dict/OrderedDict tests now go through a real codec, and a new `ProtobufCollectionsTest` matrix (22 cases: empty values in every position, nesting to depth three, optional values, non-string keys) guards the codec.

### Overview

The conformance tests validated kyo against kyo's own reading of proto3, so a misreading shared by writer and reader could not fail. `ProtobufDifferentialTest` (JVM-only) checks kyo against two independent implementations, chosen small to keep CI time flat: in-memory encodes only, two cached test-scope jars, no protoc, no codegen.

- **protobuf-java 4.35.0 (wire oracle)**: descriptors equivalent to `protoSchema`'s mapping are built programmatically; kyo bytes must parse to the same values through `DynamicMessage`, protobuf-java bytes must decode to the same values through kyo, and re-serialization is byte-identical for non-default values. Covers scalars (zigzag proof on negatives), packed repeated, nested and repeated messages, canonical map entries, and `Maybe` presence.
- **Proteus 0.6.0 (schema-mapping oracle)**: code-first Scala 3 library with kyo's constraint set, included because it had to answer the questions `.proto` cannot express. String shapes are byte-identical and cross-decode in both directions; `None` and empty-collection decisions agree; the integer mapping divergence is asserted as a recorded fact rather than skipped:

```scala
Protobuf.encode(KDNum(1)).toArray.toList        // List(0x08, 0x02)  kyo: sint32, zigzag
ProtobufCodec[PRNum].encode(PRNum(1)).toList    // List(0x08, 0x01)  Proteus: int32, plain varint
```

### What it caught on its first run

A canonical proto3 serializer omits default-valued scalar fields, so an all-defaults message is zero bytes on the wire, and kyo raised `MissingFieldException` on such input:

```scala
val oracleBytes = DynamicMessage.newBuilder(descriptor).build().toByteArray  // Array() : empty
Protobuf.decode[PDScalars](Span.from(oracleBytes))
// before: Failure(MissingFieldException)      canonical external output was undecodable
// after:  Success(PDScalars(0, 0L, 0.0, 0.0f, false, ""))
```

### Behavior changes

- Primitive schemas now declare their proto3 default as `absentDefaultValue`. The existing absent-default machinery turns that into absence-as-default **on the Protobuf reader only**; self-describing codecs (JSON, YAML, ...) still require field presence. This matches proto3 semantics and every conformant implementation, at the cost that a genuinely missing scalar field on Protobuf now reads as its default instead of failing.
- A dropped field's decode seed is now the typed empty value, so a dropped `String` reads back as `""` rather than `null`. This applies on every codec and is strictly safer; two test expectations documenting the old behavior were updated.

## Verification

- New regression suites: `SchemaRecursiveTransformTest` (8 tests, the failing top-level arrangement, sealed trait + enum + product renames), `ProtobufCollectionsTest` (22), `ProtobufDifferentialTest` (13).
- All seven codec modules green on JVM; kyo-schema, kyo-schema-json, and kyo-schema-tests green on JS and Native.
- Byte-exact Protobuf conformance tests pass unchanged, proving zero wire drift for previously working shapes.

### DIFF
--- a/kyo-schema-protobuf/shared/src/main/scala/kyo/internal/ProtobufReader.scala
+++ b/kyo-schema-protobuf/shared/src/main/scala/kyo/internal/ProtobufReader.scala
@@ -183,43 +183,51 @@ final class ProtobufReader(data: Array[Byte])(using _frame: Frame) extends Reade
 
     def hasNextElement(): Boolean =
         repeatedFrames match
+            case Nil => hasNextField()
             case frame :: _ =>
-                if frame.bare then
-                    if pos < frame.bareLimit then
-                        // Bare elements: length-prefixed blobs (messages, strings, deeper
-                        // collections) consume their own length, which is what LengthDelimited
-                        // signals; bare scalars never consult the wire type.
-                        currentWireType = LengthDelimited
-                        true
-                    else false
-                else
-                    if frame.packed && pos >= frame.packedLimit then
-                        frame.packed = false // packed run exhausted; a mixed producer may continue
-                    hasNextElementUnpacked(frame)
-            case Nil =>
-                hasNextField()
+                if frame.packed && pos >= frame.packedLimit then
+                    frame.packed = false // packed run exhausted; a mixed producer may continue
+                if frame.bare then nextBareElement(frame)
+                else if frame.packed then true // still inside the current packed run, no tag to peek
+                else if frame.firstPending then firstElement(frame)
+                else peekNextElementTag(frame)
+                end if
     end hasNextElement
 
-    private def hasNextElementUnpacked(frame: RepeatedFrame): Boolean =
-        if frame.packed then
-            true // still inside the current packed run, no tag to peek
-        else if frame.firstPending then
-            frame.firstPending = false
-            if frame.fieldNumber == 0 && limits.size == 1 then
-                // Top-level collection: no enclosing field consumed the first tag; read it
-                // now to learn the real field number (empty input is an empty collection).
-                if pos >= limits.head then false
-                else
-                    val tag = readVarint().toInt
-                    frame.fieldNumber = tag >>> 3
-                    currentFieldNumber = tag >>> 3
-                    currentWireType = tag & 0x7
-                    true
-            else true
-            end if
+    /** Bare elements of a nested-collection record are bounded by the record limit, with no tags
+      * to peek. Length-prefixed blobs (messages, strings, deeper collections) consume their own
+      * length, which is what LengthDelimited signals; bare scalars never consult the wire type.
+      */
+    private def nextBareElement(frame: RepeatedFrame): Boolean =
+        val more = pos < frame.bareLimit
+        if more then currentWireType = LengthDelimited
+        more
+    end nextBareElement
+
+    /** First hasNextElement of a frame. The enclosing field() consumed the first element's tag,
+      * except at top level, where no enclosing field exists: consume the tag now to learn the
+      * real field number. Empty input is an empty collection.
+      */
+    private def firstElement(frame: RepeatedFrame): Boolean =
+        frame.firstPending = false
+        val topLevel = frame.fieldNumber == 0 && limits.size == 1
+        if !topLevel then true
         else if pos >= limits.head then false
         else
-            // Peek the next tag: another element only if it repeats this field's number.
+            val tag = readVarint().toInt
+            frame.fieldNumber = tag >>> 3
+            currentFieldNumber = tag >>> 3
+            currentWireType = tag & 0x7
+            true
+        end if
+    end firstElement
+
+    /** Peek the next tag: another element only if it repeats this field's number; any other tag
+      * is un-read and left for the enclosing message loop.
+      */
+    private def peekNextElementTag(frame: RepeatedFrame): Boolean =
+        if pos >= limits.head then false
+        else
             val start = pos
             val tag   = readVarint().toInt
             if (tag >>> 3) == frame.fieldNumber then
@@ -227,10 +235,10 @@ final class ProtobufReader(data: Array[Byte])(using _frame: Frame) extends Reade
                 currentWireType = tag & 0x7
                 true
             else
-                pos = start // un-read; the tag belongs to the enclosing message loop
+                pos = start
                 false
             end if
-    end hasNextElementUnpacked
+    end peekNextElementTag
 
     // Establishes (or continues) a packed scalar run for the current repeated frame. Returns true
     // when the current element is a bare packed value. On first call of a packed run, consumes the
@@ -421,6 +429,7 @@ final class ProtobufReader(data: Array[Byte])(using _frame: Frame) extends Reade
 
     def hasNextEntry(): Boolean =
         mapFrames match
+            case Nil => hasNextField()
             case f :: _ =>
                 if f.entryLimitPushed then
                     // Finished the previous entry: drop its limit and return to the map level.
@@ -428,44 +437,53 @@ final class ProtobufReader(data: Array[Byte])(using _frame: Frame) extends Reade
                     limits = limits.tail
                     f.entryLimitPushed = false
                 end if
-                val more =
-                    if f.firstPending then
-                        f.firstPending = false
-                        if f.fieldNumber == 0 && limits.size == 1 then
-                            // Top-level map: no enclosing field consumed the first entry's tag;
-                            // read it now (empty input is an empty map).
-                            if pos >= limits.head then false
-                            else
-                                val tag = readVarint().toInt
-                                f.fieldNumber = tag >>> 3
-                                currentWireType = tag & 0x7
-                                true
-                        else true
-                        end if
-                    else if pos >= limits.head then false
-                    else
-                        val start = pos
-                        val tag   = readVarint().toInt
-                        if (tag >>> 3) == f.fieldNumber then
-                            currentWireType = tag & 0x7
-                            true
-                        else
-                            pos = start // un-read; the tag belongs to the enclosing message loop
-                            false
-                        end if
-                if more then
-                    // Enter the MapEntry message: consume its length prefix and push a limit.
-                    val len = readVarint().toInt
-                    if len < 0 || pos + len > data.length then
-                        throw TruncatedInputException(Protobuf(), s"map entry length $len exceeds remaining data")
-                    limits = (pos + len) :: limits
-                    f.entryLimitPushed = true
-                    true
-                else false
-                end if
-            case Nil =>
-                hasNextField()
+                val more = if f.firstPending then firstEntry(f) else peekNextEntryTag(f)
+                if more then enterMapEntry(f)
+                more
     end hasNextEntry
+
+    /** First hasNextEntry of a frame. The enclosing field() consumed the first entry's tag,
+      * except at top level, where no enclosing field exists: consume the tag now to learn the
+      * real field number. Empty input is an empty map.
+      */
+    private def firstEntry(f: MapFrame): Boolean =
+        f.firstPending = false
+        val topLevel = f.fieldNumber == 0 && limits.size == 1
+        if !topLevel then true
+        else if pos >= limits.head then false
+        else
+            val tag = readVarint().toInt
+            f.fieldNumber = tag >>> 3
+            currentWireType = tag & 0x7
+            true
+        end if
+    end firstEntry
+
+    /** Peek the next tag: another entry only if it repeats the map's field number; any other tag
+      * is un-read and left for the enclosing message loop.
+      */
+    private def peekNextEntryTag(f: MapFrame): Boolean =
+        if pos >= limits.head then false
+        else
+            val start = pos
+            val tag   = readVarint().toInt
+            if (tag >>> 3) == f.fieldNumber then
+                currentWireType = tag & 0x7
+                true
+            else
+                pos = start
+                false
+            end if
+    end peekNextEntryTag
+
+    /** Enter the MapEntry message: consume its length prefix and push a limit. */
+    private def enterMapEntry(f: MapFrame): Unit =
+        val len = readVarint().toInt
+        if len < 0 || pos + len > data.length then
+            throw TruncatedInputException(Protobuf(), s"map entry length $len exceeds remaining data")
+        limits = (pos + len) :: limits
+        f.entryLimitPushed = true
+    end enterMapEntry
 
     def bytes(): Span[Byte] =
         val len = readVarint().toInt
--- a/kyo-schema-protobuf/shared/src/main/scala/kyo/internal/ProtobufReader.scala
+++ b/kyo-schema-protobuf/shared/src/main/scala/kyo/internal/ProtobufReader.scala
@@ -46,7 +46,23 @@ final class ProtobufReader(data: Array[Byte])(using _frame: Frame) extends Reade
     // consumed entry tag is wire type 2 (LengthDelimited) reads the length once, sets packedLimit,
     // flips packed=true, then reads bare values until packedLimit. A mixed producer (some packed,
     // some unpacked under the same field) is handled by resetting packed when the run is exhausted.
-    final private class RepeatedFrame(val fieldNumber: Int, var firstPending: Boolean, var packed: Boolean, var packedLimit: Int)
+    // bare/bareLimit describe a NESTED collection (a collection directly inside another
+    // collection), which the writer emits as one length-delimited record of bare elements:
+    // scalars as raw varint/fixed bytes, strings/bytes/messages/deeper collections as
+    // length-prefixed blobs with no tag. limitDepth records limits.size at frame creation; a
+    // second arrayStart at the same depth (no message boundary in between) is by construction a
+    // nested collection. fieldNumber is a var because a TOP-LEVEL collection's first tag is
+    // consumed by hasNextElement (nothing enclosing consumed it), which is when the real field
+    // number becomes known.
+    final private class RepeatedFrame(
+        var fieldNumber: Int,
+        var firstPending: Boolean,
+        var packed: Boolean,
+        var packedLimit: Int,
+        val bare: Boolean,
+        val bareLimit: Int,
+        val limitDepth: Int
+    )
     private var repeatedFrames: List[RepeatedFrame] = Nil
 
     // Per-map decode frame. A map is a repeated MapEntry message under the map field number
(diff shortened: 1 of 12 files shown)


kyo-schema-protobuf/shared/src/main/scala/kyo/internal/ProtobufReader.scala
```diff
@@ -154,38 +184,65 @@ final class ProtobufReader(data: Array[Byte])(using _frame: Frame) extends Reade
...
+                    else false
+                else
+                    if frame.packed && pos >= frame.packedLimit then
+                        frame.packed = false // packed run exhausted; a mixed producer may continue
+                    hasNextElementUnpacked(frame)
             case Nil =>
                 hasNextField()
     end hasNextElement
 
+    private def hasNextElementUnpacked(frame: RepeatedFrame): Boolean =
```
DamianReeves: There is quite a bit of nested ifs here. Can we maybe clean this up using pattern matches or some better (zero cost) use of functional programming principles?

DamianReeves: Flattened in 4278f9d73. Each case is now its own small named method (nextBareElement, firstElement, peekNextElementTag), so hasNextElement reads as a flat decision list, and I gave hasNextEntry the same treatment since it had grown the same shape. All private methods on a final class, so no allocation or dispatch cost, and the suites pass unchanged.

fwbrasil (approved):

PR #1916, opened 2026-08-27
[kyo-data] refuse Tag derivations an opaque type's scope makes ambiguous

### LINKED ISSUE #1367: [BUG]: type inference within scope of opaque types breaks effect handling
### Version

0.19.0

### Scala Version

3.7.1

### Expected Behavior

Adding extension methods that involve effect handling with an effect type should work.

```scala
import kyo.*


opaque type X = Int
object X:
  def apply(i: Int): X = i
  extension (x: X)
    def emit: Unit < Emit[X] = Emit.value(x)

object Test:
  import X.*
  val x = X(0)
  @main def main = println(Emit.run(x.emit).eval)
```

### Actual Behavior

`def emit: Unit < Emit[X] = Emit.value(x)` infers `Emit.value[Int]` instead of `Emit.value[X]`, causing an effect to register with the `Tag[Emit[scala.Int]]`, instead of `Tag[Emit[X]]`
```
kyo.bug$KyoBugException: BUG Unexpected pending effect while handling scala.Any: Kyo(kyo.Emit[scala.Int], Input(0), main.scala:13:36, @main def main = println(Emit.run) Please open an issue 🥹 https://github.com/getkyo/kyo/issues
	at kyo.bug$KyoBugException$.apply(data.scala:14)
	at kyo.bug$.apply(data.scala:26)
	at kyo.kernel$package$.failTag(kernel.scala:36)
	at kyo.kernel.Pending$package$$less$.inline$failTag$i1(Pending.scala:18)
	at  @main def main = println(Emit.run(x.emit).eval @ Test$.main(main.scala:13)
	at Test$.main(main.scala:13)
	at main.main(main.scala:13)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at sbt.Run.invokeMain(Run.scala:144)
	at sbt.Run.execute$1(Run.scala:94)
	at sbt.Run.$anonfun$runWithLoader$5(Run.scala:121)
	at sbt.Run$.executeSuccess(Run.scala:187)
	at sbt.Run.runWithLoader(Run.scala:121)
	at sbt.Run.run(Run.scala:128)
	at com.olegych.scastie.sbtscastie.SbtScastiePlugin$$anon$1.$anonfun$run$1(SbtScastiePlugin.scala:38)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at sbt.util.InterfaceUtil$$anon$1.get(InterfaceUtil.scala:32)
	at sbt.ScastieTrapExit$App.run(ScastieTrapExit.scala:258)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

### Steps to Reproduce

https://scastie.scala-lang.org/wjcx4Tj5TrypdwC2jKOpNA

### Current Workaround

Manual type annotation or moving effect handlers outside the opaque object.

fwbrasil:
Fixes #1367

### Problem

Inside an opaque type's own template and companion, the compiler replaces the opaque type with its underlying wherever it infers, before any macro runs. A `Tag` derived there names the underlying type while one derived outside names the opaque type, so a value crossing that boundary is looked up under a different key than it was stored under (#1367).

The ticket's repro (`Emit.value(x)` in an extension method) no longer fails on main under Scala 3.8.4: the compiler stopped rewriting that query. The same substitution still fails through other APIs:

```scala
opaque type X = Int
object X:
    def apply(i: Int): X = i
    def getInside: X < Env[X] = Env.get[X]

@main def repro() = println(Env.run(X(5))(X.getInside).eval)
```

```
RuntimeException: fatal: kyo.TypeMap of contents
[TypeMap((X >: Nothing <: Any) -> 5)] missing value of type: scala.Int
```

`Env.run` stores the value under `Tag[X]`; `Env.get[X]` inside the companion was derived as `Tag[Int]`. `TypeMap(x)` built inside and read outside fails the mirror way, and `Tag[X]` summoned inside equals `Tag[Int]`, not the outside `Tag[X]`.

Separately, parameterized opaque types dropped their arguments, making `Tag[Maybe[Int]]` equal to `Tag[Maybe[String]]`.

### Solution

At the derivation site an `Int` may have been written `Int` or `X`, and no declaration can settle it: a `given Tag[X]` there is also a `Tag[Int]` and answers the query before the macro runs. So the macro walks the owner chain for the opaque types transparent at the site and refuses any derivation whose surface carries one of their underlying types. The repro above now fails to compile on the `Env.get[X]` line:

```
[Tag.opaque.collapsed] Cannot derive Tag[scala.Int] here.

This code is inside the scope of opaque type X, its declaring template or
companion object. There the compiler treats X and scala.Int as the same
type and substitutes scala.Int for X before Tag can see it. So what reaches
Tag is scala.Int, and nothing says whether it was written as X or scala.Int.
The two need different tags: a tag derived as scala.Int here would not
match the one every use outside derives as X, and a value stored under one
would not be found under the other.

Fixes:
  - Pass the tag explicitly, naming the opaque type: `(using Tag.derive[X])`.
    Tag.derive[X] survives the substitution and derives the same tag as outside.
  - Or move this code out of the scope, into a sibling object or another file.
Do not declare `given Tag[X]` inside the scope: there it is also a
Tag[scala.Int] and would silently answer every such derivation with X's tag.
```

`Tag.derive[X]` survives the substitution because it names the opaque type explicitly, so in-scope code passes it. Declaring a `given Tag[X]` in `X`'s scope is refused with `[Tag.opaque.given]`, and an underlying that cannot be recognized once substituted (a match type or refinement) refuses every derivation in that scope with `[Tag.opaque.unwalkable]`.

Arguments of parameterized opaque types come from the applied type, with the declared bounds instantiated by them. Union and intersection members sort by their dealiased form, since a member's spelling differs by site while `=:=` compares positionally.

### Notes

`Tag.hash` and `Tag.show` change for every parameterized opaque type, reaching aeron stream ids and persisted kyo-flow schemas.

`Memo`, `Memory.Arena` and `JsonRpcId` now pass the tag explicitly, and `JsonRpcId`'s discriminator no longer stores the underlying union's tag.

An inline body or `inline given` written inside the scope resolves where nothing is transparent, out of the macro's sight; CONTRIBUTING records it.


DamianReeves (approved):
LGTM

kyo-jsonrpc/shared/src/main/scala/kyo/JsonRpcId.scala
```diff
@@ -81,8 +81,11 @@ object JsonRpcId:
                 throw TypeMismatchException(Seq.empty, "number or string", "null")(using reader.frame)
             else
                 try JsonRpcId(reader.long())
-                catch case _: TypeMismatchException => JsonRpcId(reader.string()),
-        structure = Structure.Type.Open(Tag[JsonRpcId].asInstanceOf[Tag[Any]])
+                catch
+                    case _: TypeMismatchException => JsonRpcId(reader.string()),
+        // JsonRpcId is String | Long inside its own companion, so a summoned Tag[JsonRpcId] would
+        // arrive as the union's; deriving it by name keeps the discriminator naming JsonRpcId.
+        structure = Structure.Type.Open(Tag.derive[JsonRpcId].asInstanceOf[Tag[Any]])
```
hearnadam:
```suggestion
        structure = Structure.Type.Open(Tag.derive[JsonRpcId].erased
```

hearnadam (approved):

PR #1920, opened 2026-08-29
[kyo-schema, kyo-schema-json] preserve arbitrary-precision Structure numbers

DamianReeves:
`Structure.Value.BigNum` is the exact numeric case for self-describing structure trees, but JSON could neither read nor write it as an exact JSON number.

## Problem

The untyped `Structure.Value` JSON path had three different numeric behaviors:

- integral JSON numbers were parsed with `String.toLong`, so a valid number outside the signed 64-bit range failed instead of becoming `BigNum`
- decimal JSON numbers were parsed through `Double`, silently narrowing values with more precision than a `Double` can retain
- `Structure.Value.BigNum` reused the typed `BigDecimal` writer, whose existing compatibility contract encodes the value as a JSON string

That made the structure representation asymmetric:

```scala
val value = Structure.Value.BigNum(BigDecimal("18446744073709551615"))

Json.encode(value)
// before: "\"18446744073709551615\""
// after:  "18446744073709551615"

Json.decode[Structure.Value]("18446744073709551615")
// before: failure from Long overflow
// after:  Result.succeed(value)
```

This surfaced in a downstream JSON-RPC protocol whose document version is an unsigned 64-bit number. The maximum value, `18446744073709551615`, is valid protocol data but could not cross Kyo's self-describing JSON boundary without precision loss or a second JSON stack.

## Fix

`Codec.Writer` gains a `bigNumber` hook for an arbitrary-precision numeric scalar. Its default delegates to `bigDecimal`, so existing codecs retain their current wire representation. `SchemaSerializer` uses that hook only for `Structure.Value.BigNum`.

The JSON codec then supplies the self-describing behavior:

- `JsonWriter.bigNumber` emits the decimal text as a bare JSON number
- `JsonReader.readStructure` parses the token exactly as `BigDecimal`
- integral values that fit in `Long` remain `Structure.Value.Integer`; larger values become `BigNum`
- decimal values that round-trip exactly through a finite `Double` remain `Structure.Value.Decimal`; higher-precision values become `BigNum`

The reader therefore keeps the existing compact cases when they are lossless and promotes only values that need arbitrary precision.

## Compatibility

- Typed `BigInt` and `BigDecimal` schemas still use their existing JSON string representation.
- Existing `Long`-range integers and exactly representable decimals retain their current `Structure.Value` cases.
- Other codecs inherit `bigNumber = bigDecimal`, so their wire behavior is unchanged.
- The behavior change is limited to the previously non-round-tripping `Structure.Value.BigNum` JSON path and JSON numbers that previously overflowed or narrowed.

## Verification

- `sbt 'kyo-schema-jsonJVM/test'`
- `sbt 'kyo-schema-jsonJS/test'` (706 passed)
- `sbt 'kyo-schema-jsonNative/test'` (706 passed)
- release-probe passed
- downstream `morphir-scala` proof passed all 59 MEP tests plus its native-image smoke target, including bare-number round trips for `0` and `18446744073709551615` and invalid-params rejection for negative, fractional, and above-u64 document versions

### DIFF
--- a/kyo-schema-json/shared/src/main/scala/kyo/internal/JsonReader.scala
+++ b/kyo-schema-json/shared/src/main/scala/kyo/internal/JsonReader.scala
@@ -666,10 +666,14 @@ final class JsonReader private (private var input: Span[Byte], private var _fram
                 Structure.Value.Null
             case _ =>
                 val numStr = readNumber()
-                if numStr.indexOf('.') >= 0 || numStr.indexOf('e') >= 0 || numStr.indexOf('E') >= 0 then
-                    Structure.Value.Decimal(numStr.toDouble)
+                val exact  = BigDecimal(numStr)
+                if numStr.indexOf('.') < 0 && numStr.indexOf('e') < 0 && numStr.indexOf('E') < 0 then
+                    if exact.isValidLong then Structure.Value.Integer(exact.toLong)
+                    else Structure.Value.BigNum(exact)
                 else
-                    Structure.Value.Integer(numStr.toLong)
+                    val decimal = numStr.toDouble
+                    if decimal.isFinite && BigDecimal(decimal.toString) == exact then Structure.Value.Decimal(decimal)
+                    else Structure.Value.BigNum(exact)
                 end if
         end match
     end readStructure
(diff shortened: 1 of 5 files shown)


DamianReeves:
Downstream integration proof: morphir-scala decodes MEP requests as raw JSON -> Structure.Value -> a u64 domain schema. Against commit a5badd7, 19 focused tests pass, including numeric 18446744073709551615 without precision loss and invalid-params rejection for negative, fractional, and above-u64 values. The same tests fail on 1.0.0-RC6 at the JSON/Structure boundary.

PR #1926, opened 2026-09-01
kyo-http: let a path capture's codec select the route and document the parameter

Salim-belkhir:
Two commits that, together, let a path capture's codec both **select** the route and **document** the parameter. They are separable; the first stands alone, the second is the one with a behaviour change worth scrutiny.

## Motivation

Mirroring an existing Tapir service, whose block endpoints are declared like this:

```scala
def txsByHeight = prefix.get
  .in(path[BlockHeightInput].name("height").onDecodeFailureNextEndpoint)
  .in("txs")

def txsByHash = prefix.get
  .in(path[BlockHashInput].name("hash"))
  .in("txs")

object BlockHashInput:
  private val validator = Validator.pattern("(0x)?[a-fA-F0-9]{64}")
  given PlainCodec[BlockHashInput] = Codec.string.mapValidate(validator)(BlockHashInput.apply)(_.value)
```

Neither half was expressible in kyo-http. The two routes collapsed onto one, and the accepted shape could not reach the generated document. What we ended up doing instead was a single `Capture[String]("ref")` plus a hand-written parser in the handler, with the constraint described in prose — which is how the two drift apart.

## 1. `let a codec publish its own parameter schema` (07c7028)

`OpenApiGenerator` derived every parameter schema by probing the codec with sample values, which recovers a JSON *type* and nothing else. For `HttpCodec[String]` that is `type: string`, so a parameter accepting only a subset of strings had no way to say so. Path captures were worse: `extractPathParams` hardcoded `description = None`.

`HttpCodec` gains `def schema: Maybe[HttpOpenApi.SchemaObject] = Absent`, and the generator prefers it over inference. Built-in codecs are untouched.

The intended constructor is `HttpCodec.pattern`, which applies one regular expression twice — rejecting a non-matching value in `decode`, and publishing the same expression as the parameter's `pattern`:

```scala
given HttpCodec[BlockHash] =
    HttpCodec.pattern("^0x[0-9a-fA-F]{64}$")(_.value, BlockHash(_), "32-byte block hash")
```

`SchemaObject` gains `pattern`, `minLength`, `maxLength`, `minimum`, `maximum`, `description`, all defaulting to `None`, so the existing positional constructions are unaffected.

The metadata sits on the codec rather than on `HttpPath.Capture` for two reasons: adding a field to the enum case would break every `case HttpPath.Capture(fn, wn, _)` in the test suite, and the codec is shared by path, query, header and cookie parameters, which all reach the same `inferCodecJson`.

## 2. `try the next route when a path capture refuses to decode` (623eca4)

**This one fixes a silent bug.** A capture matches any segment, so two routes differing only in what their captures accept reach the same trie node. `insert` wrote them into the same per-method slot with `+`, so the second registration replaced the first and one route became permanently unreachable, with no diagnostic.

`MutableNode.endpoints` now holds a list per method in registration order, and `serialize` links the candidates through a new `nextCandidates` span indexed by endpoint. The dispatch takes the head as before; when the captures fail to decode it follows the link and tries the next.

Three properties I tried to preserve:

- **Only a *path* decode failure falls through.** A bad query parameter, header or body is a genuine client error on a route that did match, and must not silently select a different one. The retry checks `HttpFieldDecodeException.fieldType == "path"`.
- **No new API and no opt-in flag.** The only configuration whose behaviour changes is one that was already broken. A single route has `-1` in `nextCandidates` and reports its own decode failure exactly as before.
- **Nothing added to the success path.** The trie walk is not redone — every candidate sits at the same node, so the captured segment indices stay valid. The extra array read happens only on a request that was going to be rejected.

## Testing

`kyo-httpJVM/test` passes in full.

- `OpenApiGeneratorTest`, six new cases: the published pattern, the description reaching the path parameter, the same pattern rejecting a value, presence in the serialized document, the inference fallback for built-in codecs, and `withSchema`.
- `HttpServerTest`, five new end-to-end cases, each in plain and TLS: a numeric segment reaching the height route, a hash segment falling through, registration order not deciding the outcome, a segment neither route accepts giving 400, and a single route still reporting its own decode failure.

## Notes for reviewers

CONTRIBUTING asks that new API be discussed first, and this does add some (`HttpCodec.schema`, `HttpCodec.pattern`, `withSchema`, the `SchemaObject` fields). Happy to open an issue or move to Discord if you would rather shape it there — or to split the two commits into separate PRs, since the router change is the one that deserves independent review.

I have not run the fork CI workflow; local `kyo-httpJVM/test` is the only signal so far. I can run it and link the result if useful.

Built and tested on JDK 25 per `.jvmopts`.


kyo-http/shared/src/main/scala/kyo/internal/server/UnsafeServerDispatch.scala
```diff
@@ -644,29 +644,57 @@ private[kyo] object UnsafeServerDispatch:
...
+                                current.serveBuffered(currentCaptures, queryParam, headers, bodyBytes, path, method)
+
+                        serveResult match
+                            case Result.Failure(error) =>
+                                // Only a *path* decode failure can mean "wrong
+                                // candidate". A bad query param, header or body is
+                                // a genuine client error on a route that did match,
+                                // and must not silently fall through to another.
+                                val pathMismatch = error match
+                                    case e: HttpFieldDecodeException => e.fieldType == "path"
```
fwbrasil, line 677: Can you introduce a new exception type so we can avoid the string comparison?

Salim-belkhir:
Done in 9459ae5.

`HttpPathDecodeException` is a new sibling under the sealed `HttpDecodeException`. RouteUtil's path site raises it, the query/header/cookie site keeps `HttpFieldDecodeException`, and the dispatch matches on the type.

Two notes on the choices, in case you would have made them differently:

**Sibling, not subclass.** `HttpFieldDecodeException` is a `case class`, so it cannot be extended. That means a path decode failure is no longer an `HttpFieldDecodeException`. Nothing in the repo or its tests pattern-matches that type, and the only match on `HttpDecodeException` is on the supertype, so no call site changed. It is still an observable change for anyone downstream catching the field type to handle path errors. If you would rather not move it, the alternative is to keep one class and replace `fieldType: String` with an enum, which removes the string comparison without moving anything between types. Say the word and I will switch.

**Message unchanged.** The new class formats `Failed to decode path '<name>'`, which is byte-identical to what the old one produced with `fieldType = "path"`, so error output does not move.

**Name.** `HttpPathDecodeException` seemed the obvious fit next to `HttpFieldDecodeException` and `HttpMissingFieldException`, but it is your naming scheme, so rename freely.

I also added the test this type exists to protect, which was missing from the original six: a query parameter that fails to decode must **not** fall through. Both routes match `/block/7` on the path, so two candidates are live by the time the query is read, and answering 200 from the sibling route would hide a real client error behind whichever route happens to decode.

`kyo-httpJVM/test` passes in full.

kyo-http/shared/src/main/scala/kyo/internal/server/UnsafeServerDispatch.scala
Salim-belkhir: Fixed in 9459ae5: `HttpPathDecodeException`, matched on the type

kyo-http/shared/src/main/scala/kyo/HttpCodec.scala
```diff
@@ -34,6 +51,63 @@ object HttpCodec:
...
     end apply
 
+    @nowarn("msg=anonymous")
+    inline def apply[A](inline enc: A => String, inline dec: String => A, published: HttpOpenApi.SchemaObject): HttpCodec[A] =
+        val e = enc; val d = dec; val s = published
+        new HttpCodec[A]:
+            def encode(value: A)                          = e(value)
+            def decode(raw: String): Result[Throwable, A] = Result.catching[Throwable](d(raw))
+            override def schema                           = Present(s)
+        end new
```
ahoy-jon, line 61: no shortnames

kyo-http/shared/src/main/scala/kyo/HttpCodec.scala
```diff
@@ -34,6 +51,63 @@ object HttpCodec:
...
+      *     HttpCodec.pattern("^0x[0-9a-fA-F]{64}\$")(_.value, BlockHash(_), "32-byte block hash")
+      * }}}
+      */
+    @nowarn("msg=anonymous")
+    inline def pattern[A](regex: String)(
+        inline enc: A => String,
+        inline dec: String => A,
+        description: String = ""
+    ): HttpCodec[A] =
+        val e        = enc; val d = dec; val r = regex; val desc = description
```
ahoy-jon, line 86: same

ahoy-jon (approved):

PR #1934, opened 2026-09-04
fix: forward JsonSchema constraints to OpenAPI SchemaObject

vbergeron-ledger:
Fixes a silent data loss in the OpenAPI generator: constraint information
recorded in `Json.JsonSchema` (via `Schema.Constraint` and
`JsonSchemaEnricher`) was always dropped when translating to
`HttpOpenApi.SchemaObject`, so generated specs never contained
`minLength`, `maxLength`, `pattern`, `minimum`, `maximum`,
`exclusiveMinimum`, `exclusiveMaximum`, `minItems`, `maxItems`,
`uniqueItems`, or `description` on any schema.

## Root cause

`HttpOpenApi.SchemaObject` was missing all constraint fields. The
translation function `jsonSchemaToHttpOpenApi` in `OpenApiGenerator`
discarded every `Json.JsonSchema` variant down to the bare primitive
helper (`SchemaObject.string`, `SchemaObject.number`, etc.) regardless
of what constraints were present on the source node.

## Changes

**`HttpOpenApi.SchemaObject`** (`kyo-http/shared/.../HttpOpenApi.scala`)

Adds the 11 missing constraint fields with `= None` defaults:
`minimum`, `exclusiveMinimum`, `maximum`, `exclusiveMaximum`,
`minLength`, `maxLength`, `pattern`, `minItems`, `maxItems`,
`uniqueItems`, `description`. All existing positional and named
call sites continue to compile unchanged.

**`OpenApiGenerator.jsonSchemaToHttpOpenApi`** (`kyo-http/shared/.../OpenApiGenerator.scala`)

Forwards all constraint fields from each `JsonSchema` variant:
- `Str`: `minLength`, `maxLength`, `pattern`, `format`, `description`
- `Num`: `minimum`, `exclusiveMinimum`, `maximum`, `exclusiveMaximum`, `description`
- `Integer`: same numeric constraints widened to `Double`, `description`
- `Arr`: `minItems`, `maxItems`, `uniqueItems`, `description`
- `Obj`: `description`
- `Bool`: `description`

Also widens the method from `private` to `private[kyo]` to allow
direct unit testing.

**`OpenApiGeneratorTest`** (`kyo-http/shared/.../OpenApiGeneratorTest.scala`)

11 new tests in a `constraint forwarding` group: one per variant and
constraint combination, plus round-trip JSON serialization checks
verifying fields are absent when empty and present when set.

### DIFF
--- a/kyo-http/shared/src/main/scala/kyo/HttpOpenApi.scala
+++ b/kyo-http/shared/src/main/scala/kyo/HttpOpenApi.scala
@@ -53,7 +53,18 @@ object HttpOpenApi:
         additionalProperties: Option[SchemaObject],
         oneOf: Option[List[SchemaObject]],
         `enum`: Option[List[String]],
-        `$ref`: Option[String]
+        `$ref`: Option[String],
+        minimum: Option[Double] = None,
+        exclusiveMinimum: Option[Double] = None,
+        maximum: Option[Double] = None,
+        exclusiveMaximum: Option[Double] = None,
+        minLength: Option[Int] = None,
+        maxLength: Option[Int] = None,
+        pattern: Option[String] = None,
+        minItems: Option[Int] = None,
+        maxItems: Option[Int] = None,
+        uniqueItems: Option[Boolean] = None,
+        description: Option[String] = None
     ) derives Schema, CanEqual
 
     case class Info(
--- a/kyo-http/shared/src/main/scala/kyo/internal/codec/OpenApiGenerator.scala
+++ b/kyo-http/shared/src/main/scala/kyo/internal/codec/OpenApiGenerator.scala
@@ -226,9 +226,9 @@ private[kyo] object OpenApiGenerator:
       *   - Nullable(inner) → recurse on inner (optionality is captured in the parent's required list).
       *   - OneOf(variants) → if all variants map to an empty Obj, emit enum with variant names; otherwise emit oneOf.
       */
-    private def jsonSchemaToHttpOpenApi(js: JsonSchema): HttpOpenApi.SchemaObject =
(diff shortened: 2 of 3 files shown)


kyo-http/shared/src/main/scala/kyo/HttpOpenApi.scala
```diff
@@ -53,7 +53,18 @@ object HttpOpenApi:
         additionalProperties: Option[SchemaObject],
         oneOf: Option[List[SchemaObject]],
         `enum`: Option[List[String]],
-        `$ref`: Option[String]
+        `$ref`: Option[String],
+        minimum: Option[Double] = None,
```
DamianReeves, line 57: I'm debating whether or not a Double is the correct type here. Floating points can be lossy and do we actually want something arbitrary precision like a BigDecimal here.

cc/ @fwbrasil @hearnadam

vbergeron-ledger: The JSON spec specify that it represents an arbitrary decimal number so I would go for BigDecimal as well. However, since JSON means JavaScript Object Notation and that JS numers are only safe on 53 bits of precision, the case for Double still holds. Do we want to be technically correct or pragmatically correct ?

ahoy-jon: hard one
> `minimum: Option[String]` ?

sadly BigDecimal is impractical to use.

ahoy-jon: @DamianReeves if you are ok, let's keep `Double` for now, and change it later if needed 💪

ahoy-jon (approved):

PR #1944, opened 2026-09-15
[kyo-ui] remove the drag runtime's pagehide listener on teardown

fwbrasil:
## Problem

The server-push client installs a fresh drag runtime each time a socket opens after a drop. Every install registers a `pagehide` listener on `window`, and the runtime's `cleanup` removed its document listeners but not that one. Each reconnect therefore left a listener behind, holding the torn-down runtime's closure for the life of the tab:

- one per drop in normal operation;
- one every few seconds while a server accepts connections and closes them at once, as during a restart.

## Solution

`cleanup` in `DragClientJs` now removes its own `pagehide` listener along with the document capture listeners, so only the current runtime is listening. The comments that described that listener as never removed (`HtmlRenderer`) and the runtime as installed once per page (`DragClientJs`) are corrected.

## Notes

- Behavior was unaffected: a stale listener firing on `pagehide` did nothing, since each runtime's `cleanup` runs only once. The cost was retained memory.
- `DragClientJsTest` reproduces it: it drops the socket twice through `window.__kyoWs`, proves each reconnected session live with a click, and counts the `pagehide` listeners still registered. It fails on main with 2 and passes with 1.


DamianReeves (approved):
LGTM

PR #1945, opened 2026-09-17
[kyo-sql] add the SQLite backend, and make engine neutrality checkable

fwbrasil:
## Problem

kyo-sql only talked to servers. PostgreSQL and MySQL are both network drivers on kyo-net, so every program needed a database process running somewhere: no file-backed or in-memory database, nothing that runs in a browser or ships inside a single binary, and a container for any test that wanted real SQL.

Adding an embedded engine was not just a third driver. PostgreSQL and MySQL are alike in every way the abstraction cares about, so core had been shaped around their similarities without anyone deciding to, and the conformance battery could not tell real agreement between engines from two engines that merely resemble each other.

## Solution

**`kyo-sql-sqlite`** is an embedded SQLite driver over the C library through kyo-ffi. A database is a file path or `:memory:`, with no server. It cross-builds for JVM, JS, Native and Wasm; in a browser it runs on the WebAssembly build.

**`kyo-sql-sqlite-driver`** is that driver's connection, pool plumbing, codecs, row reader, dialect and URL parsing, plus the binding trait as declarations only: no C, no FFI plugin, no engine. The split exists because on Scala Native an FFI module's C is compiled INTO the binary, so any module depending on an engine module links that engine in. A second engine speaking the same `sqlite3_*` API therefore cannot depend on `kyo-sql-sqlite` without carrying the vendored `sqlite3.c` too, at which point the calls bind to whichever linked first. Depending on declarations carries no C.

**kyo-ffi gained a second JS transport.** A generated JS impl asks `NativeFacade` for its dispatch table: koffi over a real shared library on Node, or a registered WebAssembly provider where there is neither, which is every browser. The choice is made by asking whether koffi can be loaded, not by sniffing for a browser, and `-Dkyo.ffi.js.transport=wasm` forces the WASM path so it runs under the same driver suites as the native one.

### What this changed in core

SQLite is the first engine that is not a server, and four things in kyo-sql had assumed one:

- `SqlConfig.Address` was a host and a port. It is now sealed over `Network` and `Local`, so neither kind of backend can be handed the other's coordinate.
- Core parsed the URL and then looked up the backend, so it defined what a database URL looks like. The scheme is now resolved first, and `Backend.parseUrl` hands each backend the raw URL to parse itself.
- A defaulted column was rendered by substituting a cell, which needs a `DEFAULT` keyword. `Idiom.overrideFor` is now `private[kyo]`, so a flavor without one can drop the column from the insert list instead.
- `supportsGroupingSets` implied rollup. MySQL has rollup and no grouping sets, SQLite has neither, so `supportsRollup` is its own flag.

`SqlJsonArray` also moves from `kyo-sql-mysql` into core. MySQL and SQLite both lack an array type and both map the three array kinds onto `JSON`, so one wire means a value written by either is read by either.

### Conformance

Every suite now runs against three engines. Telling a real divergence from an accident needed the descriptors to say what an engine can do, so they gained seventeen members: isolation levels honoured and defaulted to, advisory locks, concurrent write transactions, JSON key order, case folding past ASCII, and others. A leaf reads a capability, never an engine name, and capabilities are used in complementary pairs, so an engine that lacks a feature asserts what it does instead of being skipped. An engine that cannot deliver an isolation level has to refuse it rather than quietly substitute a weaker one, which is the lost-update bug this suite exists to catch.

`SqlBackendNeutralitySourceTest` enforces that mechanically: it scans the shared conformance source and fails on an engine name in a body, with the banned list derived from the registered descriptors, so a new backend joins the check by existing.

## Notes

**This is the first of two.** A Dolt backend, its embedded DoltLite counterpart, and a `FileSystem` stored in a version-controlled database follow separately. DoltLite is the second engine the driver split above exists for: it keeps the `sqlite3_*` API over different storage, so it reuses this driver's connection, codecs and row reader with its own library. Nothing here depends on that landing.

**SQLite is compiled from a pinned C source** rather than linked against a host `libsqlite3`, since several conformance answers depend on the version: `RETURNING` since 3.35.0, the `RIGHT`/`FULL JOIN` floor at 3.39.0. A current macOS ships 3.43.2 against the 3.53.4 pinned here. The source is fetched and checksummed by `kyo-sql-sqlite/scripts/build-sqlite.sh`, and the staged tree is a gitignored build artifact. Nothing is staged per os-arch: it is one architecture-independent `.c`, and the FFI plugin already compiles per platform.

**`SqliteVfs` is a `SqlConfig` extension** rather than a config field, because it means nothing to any other engine. It is how a browser gets persistence: the WebAssembly build has no filesystem, so the default VFS keeps a database in memory and loses it at reload.

**What this engine refuses.** It has no advisory locks and no configurable isolation, and both are reported as typed refusals naming what was asked rather than silently downgraded. That is the shape engine-specific API takes here: what SQLite has over the others is an absence of things, so `SqliteClient` adds no methods.

**Verification.** PostgreSQL and MySQL run in containers, SQLite embedded. The module cross-builds for JVM, JS, Native and Wasm.


NPCRUS:
god bless you for adding sqlite! that's exactly what i needed

PR #1949, opened 2026-09-17
[stats] use the Native runtime weak-reference hierarchy

DamianReeves:
### Problem

The JavaScript WeakReference shim was also compiled for Scala Native, shadowing its runtime implementation. That broke reference casts and inherited thread-local copying, including the runtime stack-unwinding buffers.

### Solution

Restrict the shim to JS and Wasm. Add regressions for clearing references, replacing cleared registry entries, and child-local inheritance. Enable Native machine-metric readback tests and remove the scheduler tests' exception suppression.

### Notes

Validated on this standalone branch: 282 registry tests across JVM, Native, JS, and Wasm; three machine-metric tests each on JVM and Native; both README doctests; and serial `scalafmtCheckAll`. The three Native weak-reference regressions reproduced the original defects before the fix.

Extracted from #1868 so the filesystem stack can be reviewed independently.

Rebased onto `6d50eecae` on September 17. Revalidated all 282 registry tests across JVM, Native, JS, and Wasm, plus six machine-metric tests across JVM and Native. The previous Linux JVM CI failures are addressed separately in #1953 and #1954.


kyo-scheduler/jvm-native/src/test/scala/kyo/scheduler/TestExecutors.scala
```diff
@@ -10,35 +10,20 @@ import kyo.scheduler.util.Threads
   * same process. The returned executors ignore shutdown calls to prevent accidental closure by test cleanup code.
   */
 object TestExecutors {
```
hearnadam, line 12: Why is this changing?

DamianReeves: This removes the workaround for the Native bug fixed here. Our JS `WeakReference` shim was being compiled into Native without extending `Reference`, so Native’s inherited-thread-local copy skipped `childValue`. Its stack-trace implementation uses that hook to allocate separate unwind buffers for each child thread; skipping it left threads sharing those buffers and could crash in libunwind.

Using Native’s real weak-reference hierarchy restores that copy. `StatsRegistryWeakReferenceTest` reproduces the broken hierarchy and missing `childValue` call before the fix. With the suppression removed, the scheduler suites pass (215 JVM tests and 213 Native tests). Keeping the handlers would continue swallowing uncaught test exceptions, including globally on JVM.

DamianReeves:
Rebased onto `main` at `6d50eecae`. #1950 and #1951 were restacked in the same atomic push, with their patches unchanged.

The two previous Linux JVM failures led to separate fixes:

- #1953 corrects the UI test's synchronization. Its render counter advances before the observer registers its next waiter; a gated reproduction consistently observed zero where the old check expected one.
- #1954 fixes a resource leak in `Scope.acquireRelease`, which also backs `Fiber.init`. Interrupting the final acquisition step could leave one acquired resource with zero releases. This independently reproduces a defect consistent with the SQL cancellation failure; the CI log alone cannot establish the exact interleaving.

Both fixes pass their focused tests on JVM, JS, Native, and Wasm. The core fix also passes all 1,756 active JVM core tests, 681 JVM kernel tests, and 21 SQL cancellation tests.

On this PR's rebased head, all 282 registry tests and six machine-metric tests passed locally. GitHub checks and the release probe have passed; the [build matrix](https://github.com/getkyo/kyo/actions/runs/35294696374) is still running.

PR #1950, opened 2026-09-17
[scheduler] bound blocked-worker queue migration

DamianReeves:
### Problem

Task placement called `Worker.checkAvailability` while choosing a destination. That method also drains a blocked worker's queue, which calls task placement again. Inspecting another blocked destination could therefore start another drain. When all workers were blocked, one maintenance pass repeatedly moved the same tasks between queues.

This has a controlled regression, independent of worker-thread timing. It queues 24 tasks across six workers, marks five or all six blocked, and explicitly drives maintenance with worker execution and timers held back. The old placement behavior fails the all-blocked case at 145 insertions against a 144-operation bound. Restoring the equality guard before CAS does not fix that failure.

### Solution

Use `isAvailable` during placement to check candidates without draining their queues. Keep migration in the maintenance path through `checkAvailability`. Preserve the equality guard before CAS so non-running workers avoid an unnecessary atomic read-modify-write.

The regression also verifies all 24 tasks are conserved and each executes exactly once after the workers can run. Test fixtures close their owned schedulers and clocks.

### Notes

The Native weak-reference fix is the parent because origin/main alone reproduced the separate runtime stack-unwinding crash. This PR contains only scheduler changes relative to its parent.

Extracted from #1868 so the filesystem stack can be reviewed independently. Prerequisite: #1949.

Previous full scheduler validation passed 215 JVM tests and 213 Native tests, with one platform cancellation. The focused Linux ARM64 JVM CI run passed: https://github.com/getkyo/kyo/actions/runs/35183612550.

After restoring the CAS guard in `419d3b103`, both full suites passed again: JVM 215 and Native 213, with one platform cancellation. Reinstating the old placement calls with the guard present reproduced the bounded-migration failure before this validation.


kyo-scheduler/jvm-native/src/main/scala/kyo/scheduler/Worker.scala
```diff
@@ -228,10 +229,18 @@ abstract private class Worker(
...
+
+    /** Checks availability and migrates work away from an unavailable worker.
+      *
+      * Used by the maintenance cycle. Placement uses isAvailable so draining a queue cannot recursively drain destination queues.
+      */
+    def checkAvailability(nowMs: Long): Boolean = {
+        val available = isAvailable(nowMs)
         if (!available) {
-            if ((st eq State.Running) && state.compareAndSet(State.Running, State.Stalled))
+            if (state.compareAndSet(State.Running, State.Stalled))
```
fwbrasil: is there a motivation for this change? the equality guard is a performance optimization to avoid the CAS

DamianReeves: The equality guard removal was incidental to splitting the availability check; it is not needed for the fix. Restored it in 419d3b103 so non-running workers avoid the CAS.

The underlying issue is that placement used `checkAvailability`, which can drain a blocked candidate's queue and recursively call placement again. With all workers blocked, the same tasks keep being migrated during a single maintenance pass. The regression queues 24 tasks across six workers and drives maintenance with worker execution and timers held back. I reran the old placement behavior with the equality guard restored: it still fails at 145 insertions against the 144-operation bound. With placement using `isAvailable`, both the five-blocked and six-blocked cases pass and verify every task executes exactly once.

Both full scheduler suites pass with the guard restored: 215 JVM tests and 213 Native tests. I also expanded the PR description with this reproduction.

PR #1954, opened 2026-09-18
[core] retain resource cleanup across acquisition interrupts

DamianReeves:
`Scope.acquireRelease` could produce a resource and then lose its cleanup when the fiber was interrupted before the registration continuation. `Fiber.init` uses this path, so a child that had already started could survive its parent scope's shutdown.

Register the release in the same step that receives the acquired value, using a private kernel completion observer. Suspended acquisition stays interruptible. If the scope already closed, run the rejected release exactly once on a detached fiber and retain the `Closed` error. Public effect rows and `Scope.ensure` rejection behavior remain unchanged.

Investigated after [#1949](https://github.com/getkyo/kyo/pull/1949)'s [Linux ARM64 JVM job](https://github.com/getkyo/kyo/actions/runs/35183452786/job/105080376944) reported no reclaim events after interrupting a SQL lease's parent. A deterministic core regression interrupts inside the final acquisition step and observes scope teardown: acquisition ran once, release ran zero times. It fails directly in 16 ms before the fix. This proves the core defect independently; the CI log does not establish its exact interleaving.

Shared regressions cover interruption during acquisition, a scope closing before acquisition completes, rejected asynchronous cleanup failure, acquisition failure/panic, exactly-once release, and the original scope creation frame in `Closed`. The existing closed-scope acquisition regression is enabled.

Validation:

- Full JVM core suite: 1,756 passed across 42 suites.
- Full JVM kernel suite: 681 passed; SQL cancellation suite: 21 passed.
- ScopeTest: 53 passed on JVM, JS, Native, and Wasm.
- FiberTest: JVM 113, JS 109, Native 111, Wasm 109 passed.
- PendingTest and RuntimeEffectTest: 61 passed on each of JS, Native, and Wasm.
- Serial `scalafmtCheckAll` and `git diff --check` passed.


fwbrasil:
I'm preparing to open a PR with the new arrow-based kernel and it has major changes in how this code works. I'll close this PR since the new kernel fixes this issue as well.

PR #1980, opened 2026-09-21
[kyo-core][kyo-website][kyo-browser] fix the idle docs-tab leak (#1923)

### LINKED ISSUE #1923: [BUG]: https://getkyo.io memory leak
### Version

0

### Scala Version

0

### Expected Behavior

eats around 200-300mb max?

### Actual Behavior

when docs website is opened over substantial amount of time (>8 hours) - i find out the tab eats around 5gb of memory, immediately released after i closed it.

### Steps to Reproduce

_No response_

### Current Workaround

_No response_

fwbrasil:
Fixes #1923

## Problem

An open getkyo.io docs tab grows its memory without bound, reaching around 5 GB over a long session per the report.

Measured against the real generated site in a real Chrome, forcing a collection before every sample, with nothing on the page touched:

| elapsed | retained heap | DOM nodes |
| ------- | ------------- | --------- |
| 0 s     | 49.9 MB       | 3652      |
| 120 s   | 67.3 MB       | 3652      |
| 240 s   | 84.4 MB       | 3652      |

Linear at ~8.8 MB/min with the DOM flat, so this is retained JS state, not DOM growth and not collection lag.

`IOPromise.Pending` is append-only, and the one removal it offered (`removeInterrupt`) dropped an interrupt link and nothing else. Two registrations, both made by every fiber, outlived what they were registered for:

- A task joining a promise registers a resume callback on it, and that callback holds the task's continuation. Interrupting the parked task left it registered. When the awaited promise is masked it refuses the interrupt and never completes, so nothing ever cleared it.
- A forked task registers an interrupt link on its parent. The parent kept that link after the child had completed.

Neither matters on a short-lived promise; both are unbounded on a long-lived one. `Signal.observe` drives both once per second, because it parks on a repair race that abandons one waiter and forks two fibers per tick, and a Signal's `next` promise is masked. A docs page carries 68 reactive regions.

## Solution

`Pending` gets a single removal, keyed by what is able to register:

```scala
final def remove(key: IOPromise[?, ?] | Function1[?, ?]): Boolean
```

It replaces `removeInterrupt`, so a registration of any kind can be dropped by whoever made it: a link is keyed by the promise it targets, a callback by itself. It stops at the first match and returns the same chain when nothing matched, which is how the promise-level `remove` reports whether the registration was there. The drop shares the CAS that completion uses, so `true` means the promise will never fire that callback.

The join link now carries the task's resume callback. No new state on `IOTask`:

- When the awaited promise accepts the interrupt, its completion fires the callback, as before.
- When it refuses (masked), the link removes the callback from it and fires it with the interrupt, only if `remove` returned `true`. The parked task is woken exactly once and nothing stays registered.

Links to completed promises are dropped as new links are registered, so a long-lived parent holds links only to children that are still running.

`merge`'s removal did not reach registrations merged in by `become`. It now searches both chains.

## Notes

`SiteAppHeapTest` is the regression guard for the reported symptom: the real generated site, the real `fullLinkJS` bundle, a real Chrome, and the tab's retained heap sampled across a 20 s idle window against a 1 MB budget. It depends on the wall clock deliberately, since the page's own 1 s timer inside the browser drives the leak. Measured growth after the baseline sample:

| window | with the fix | without it |
| ------ | ------------ | ---------- |
| 10 s   | 54 KB        | 1.5 MB     |
| 20 s   | 59 KB        | 3.0 MB     |
| 30 s   | 63 KB        | 4.5 MB     |
| 60 s   | 127 KB       | 9.0 MB     |
| 120 s  | 160 KB       | 17.9 MB    |

### Manual check: is the residual growth a slower leak?

With the fix the heap still drifts up by tens of KB per minute, so the idle tab was profiled by hand to see what that is. This is not part of the test suite.

Method: the same served site and Chrome the test uses, left idle for 10 minutes. Once a minute the heap was read after a forced collection (`HeapProfiler.collectGarbage`, then `Runtime.getHeapUsage`). Two full heap snapshots were taken 5 minutes apart with `HeapProfiler.takeHeapSnapshot`, from a second CDP client attached to the page over Chrome's remote-debugging port, and diffed by object type and name.

| minute | used after forced collection |
| ------ | ---------------------------- |
| 1      | 49,711 KB                    |
| 2      | 49,764 KB                    |
| 3      | 49,773 KB                    |
| 4      | 21,464 KB                    |
| 5      | 21,477 KB                    |
| 6      | 21,475 KB                    |
| 7      | 21,487 KB                    |
| 8      | 21,478 KB                    |
| 9      | 21,521 KB                    |
| 10     | 21,519 KB                    |

The 28 MB drop at minute 4 is V8 flushing aged code on its own; a forced collection does not reclaim it earlier. After it the heap stays inside a 60 KB band with no trend.

Snapshot diff, minute 5 to minute 10, total self size 67,010 KB to 67,047 KB (+37 KB):

| grew     | objects | what                          |
| -------- | ------- | ----------------------------- |
| +15.7 KB | +21     | `InstructionStream` (machine code) |
| +9.1 KB  | +8      | `WeakArrayList` (V8 internal) |
| +6.1 KB  | +51     | `TrustedByteArray` (bytecode and metadata) |
| +2.4 KB  | +15     | `ProtectedFixedArray` (code metadata) |
| +1.5 KB  | +21     | `Code`                        |

Every byte of growth is V8's own compiled code and its metadata: the page's once-per-second timers keep code warm, V8 keeps tiering functions up, and compiled code counts in `usedSize`. Closures, arrays, strings and Scala.js class instances: zero change. Nothing the application allocates is retained.


kyo-core/shared/src/main/scala/kyo/scheduler/IOPromise.scala
```diff
@@ -310,47 +325,78 @@ private[kyo] object IOPromise:
...
-        def removeInterrupt(other: IOPromise[?, ?]): Pending[E, A]
+
+        /** Costs a frame per node walked, so it stops at the first match; the registration being dropped is normally
+          * the newest, at the head. Returns `this`, not a copy, when nothing matched.
+          */
+        def remove(key: IOPromise[?, ?] | Function1[?, ?]): Pending[E, A]
         def run(v: Result[E, A]): Pending[E, A]
 
+        /** The rest of the chain if this node is a link to a completed promise, else this node. */
+        def dropIfDead: Pending[E, A] = this
```
hearnadam, line 336: This comment is a bit confusing. If 'this' is a link to a completed promise, wouldn't the rest of the chain be non-pending?

